Issue #23, opened 2026-04-13
Default mode template replay still classified as overage for high-volume OC agent traffic


belangertrading:
## Environment
- Dario: 3.4.1 (also reproduced on 3.4.0)
- Claude Code: 2.1.104
- Platform: Linux x64 (Ubuntu VPS)
- OpenClaw: 2026.4.9
- Device identity: detected
- OAuth: valid (token expires ~8h from testing)

## Problem
Running OpenClaw agents through dario in default mode. Simple curl tests consistently get `billing: five_hour`. Actual OC agent requests always get `billing: overage`. We never got a single non-overage request from OC regardless of what we changed.

---

### Step 1: cache_control fingerprint hypothesis
Built a Bun strip proxy (:3455 → :3456) that recursively removed all `cache_control` fields from request body and stripped `prompt-caching-*` from `anthropic-beta` headers before forwarding to dario.

**Result: still overage.** Dario's template injection adds `cache_control` back in anyway. We were stripping what dario intentionally adds.

---

### Step 2: SDK headers fingerprint hypothesis
Captured exact headers OC was sending to dario:
```
x-stainless-arch, x-stainless-helper-method, x-stainless-lang, x-stainless-os
x-stainless-package-version: 0.73.0, x-stainless-runtime-version: v22.22.2
anthropic-dangerous-direct-browser-access: true
user-agent: Anthropic/JS 0.73.0
```
Stripped all of these at the strip proxy before forwarding to dario.

**Result: still overage.**

---

### Step 3: Verified final headers reaching Anthropic
After stripping at proxy level + dario template replay, headers forwarded to Anthropic were clean:
```json
{
  "accept": "application/json",
  "anthropic-beta": "fine-grained-tool-streaming-2025-05-14",
  "anthropic-version": "2023-06-01",
  "content-type": "application/json",
  "user-agent": "Mozilla/5.0"
}
```
No SDK fingerprints, no cache headers. **Result: still overage.**

---

### Step 4: curl vs OC agent isolation test
Same proxy setup, same endpoint. Curl always got `five_hour`. OC always got `overage`.

Observable differences:
| Property | curl test | OC agent |
|---|---|---|
| Messages | 1 | 155+ |
| Body size | ~500 bytes | ~295KB |
| Tools | 0–10 | 10 |
| Streaming | yes/no | yes |
| Extended thinking | yes/no | yes |
| Conversation history | none | full multi-turn |

---

### Step 5: Tool count reduction
[...]

Issue #7, opened 2026-04-10
CLI mode: spawn E2BIG on large conversation contexts


belangertrading:
## Problem

`dario proxy --cli` fails with `spawn E2BIG` when the conversation context exceeds the OS argument size limit (~2MB). This happens with multi-turn agent conversations that accumulate large histories.

```
[dario:cli] model=claude-opus-4-6 prompt=[{"type":"text","text":"[Thread starter - for context]\nOps,...
[dario] Proxy error: spawn E2BIG
```

## Why this matters now

Anthropic just announced that third-party apps now draw from extra usage, not plan limits:

> `claude --append-system-prompt '[[reply_to_current]] stripped before sending' --print hi`
>
> "Third-party apps now draw from your extra usage, not your plan limits."

This means the billing tag injection from v2.6.0 is effectively dead — the direct API path now routes all third-party OAuth requests to the extra usage bucket regardless. `--cli` mode is the only remaining path to draw from Max plan allowance, but it breaks on long conversations due to the arg size limit.

## Suggested fix

Pipe the conversation payload to `claude --print` via **stdin** instead of passing it as command-line arguments. This removes the OS arg size limit while keeping the CLI routing that gets plan limits.

## Environment

- dario v2.8.0
- Claude Max 5x
- Linux (Hetzner VPS)
- Multi-agent stack (OpenClaw) with long-running conversation threads

askalf:
## Fix

The root cause is line 397 — conversation history gets passed as a command-line argument via `--append-system-prompt`, which hits the OS arg size limit (~2MB) on long conversations. The user prompt already goes through stdin, but the system prompt with accumulated history goes through args.

**Fixed in the next release**: System prompt is now written to a temp file and passed via `--append-system-prompt-file` instead of as an inline argument. The temp file is cleaned up after each request.

```typescript
// Before (breaks on large contexts):
args.push('--append-system-prompt', systemPrompt);

// After (no size limit):
const systemPromptFile = join(tmpdir(), `dario-sysprompt-${randomUUID()}.txt`);
writeFileSync(systemPromptFile, systemPrompt, { mode: 0o600 });
args.push('--append-system-prompt-file', systemPromptFile);
```

## On the billing change

> Third-party apps now draw from your extra usage, not your plan limits.

[...]

Issue #6, opened 2026-04-10
Opus and Sonnet 429 at 0% utilization through OAuth


belangertrading:
## Bug

Opus and Sonnet return 429 through Dario even at **0% utilization** on both 5-hour and 7-day windows. Haiku works fine. Direct Anthropic API key handles Opus with no issues.

### Evidence

**Haiku through Dario — works:**
```
HTTP/1.1 200 OK
anthropic-ratelimit-unified-5h-utilization: 0.0
anthropic-ratelimit-unified-7d-utilization: 0.0
anthropic-ratelimit-unified-overage-utilization: 0.0
anthropic-ratelimit-unified-representative-claim: five_hour
```

**Opus through Dario — 429 at 0%:**
```
HTTP/1.1 429 Too Many Requests
{"type":"error","error":{"type":"rate_limit_error","message":"Error"}}
```
No rate limit headers returned at all on the 429.

**Opus through direct Anthropic API key — works:**
```
HTTP/2 200
model: claude-opus-4-20250514 — responds normally
```

### Environment
- Dario v2.5.0
- Claude Max subscription
- Device identity: detected
- OAuth: healthy
- 0% current session, 0% 7-day utilization (fresh windows)

### Theory
Anthropic may be applying a separate per-model rate limit to OAuth-proxied requests that doesn't show in the unified utilization headers. Opus and Sonnet are blocked while Haiku passes through — even at zero utilization. This effectively makes Max plan Opus/Sonnet unusable through the proxy.

belangertrading:
## Update: CLI fallback workaround (v2.5.1)

Confirmed this is still present on v2.5.1 with a Claude Max subscription. The root cause appears to be Anthropic classifying the OAuth API pathway differently from the native CLI binary:

- **Direct API via OAuth token** → `billing: overage` → Opus/Sonnet blocked when Extra Usage is org-disabled
- **`claude --print` (CLI binary)** → priority routing → Opus/Sonnet work fine, same token

Key evidence: `cachedExtraUsageDisabledReason: org_level_disabled` in `~/.claude.json`. The OAuth pathway routes all requests through the overage/Extra Usage billing path. With Extra Usage disabled at the org level, Opus/Sonnet get 429 while Haiku squeaks through.

### Workaround: Hybrid direct + CLI fallback

We patched `proxy.js` to auto-detect 429 on the direct API pathway and transparently retry via `handleViaCli`. This way:
- **Haiku** → direct API (fast, ~1s) ✅
[...]

Issue #4, opened 2026-04-10
Billing error with Extra Usage disabled


askalf:
## Bug Report

**Reported by:** @belangertrading

**Environment:**
- Claude Max subscription with Extra Usage **off**
- Session usage: 0% current session, 68% weekly limit
- Version: v2.3.0

## Problem

After updating to v2.3.0, requests return "API provider returned a billing error" despite having available usage budget and Extra Usage disabled.

## Root Cause

Dario's default `anthropic-beta` header included `context-management-2025-06-27` and `prompt-caching-scope-2026-01-05` — beta flags that may require Extra Usage on Max subscriptions. When Extra Usage is disabled, the Anthropic API rejects requests that use these features.

## Fix

**v2.3.1** (released) removes these potentially billable betas from defaults. Only essential betas are now included:
- `oauth-2025-04-20` — required for OAuth auth
- `interleaved-thinking-2025-05-14` — standard thinking
- `claude-code-20250219` — session identity

Client-provided betas still pass through with filtering for known billable prefixes.

## Resolution

Update to v2.3.1:
```bash
npm install -g @askalf/dario@latest
```

---
cc @belangertrading — this should fix the billing error you reported. Let us know if it persists after updating.

askalf:
## Fix Verified — v2.3.1

End-to-end test results after removing billable beta flags:

| Test | Result |
|------|--------|
| Health check | `healthy`, OAuth valid |
| Native Anthropic (Haiku) | ✅ 200 — no billing error |
| OpenAI compat (Haiku) | ✅ 200 — proper chat completion format |
| Native streaming | ✅ SSE chunks delivered correctly |
| OpenAI streaming | ✅ OpenAI chunk format correct |
| Opus request | ✅ Rate limit (expected at quota) — **not** a billing error |
| Client beta passthrough | ✅ Client-provided betas forwarded cleanly |
| SSRF protection | ✅ Forbidden on disallowed paths |

**Root cause confirmed:** `context-management-2025-06-27` and `prompt-caching-scope-2026-01-05` in default beta header triggered billing rejection for Max users with Extra Usage off.

**Fix:** v2.3.1 — default betas reduced to `oauth-2025-04-20`, `interleaved-thinking-2025-05-14`, `claude-code-20250219`. Client-provided betas pass through with billable prefix filtering.

```bash
npm install -g @askalf/dario@latest
```

belangertrading:
## New Finding: Extra Usage consumed instead of Max Plan limits

[...]

PR #27, opened 2026-04-13
docs: update contributor credits and add multi-agent billing FAQ

belangertrading:
Updates the contributor table to reflect the full scope of diagnostic work across issues #4, #6, #7, #12, and #23 — as acknowledged in https://github.com/askalf/dario/issues/23#issuecomment-4237832101.

Also adds a FAQ entry for multi-agent session-level billing classification, based on the debugging trail documented in #23. Common gotcha for anyone running agent orchestrators through dario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,7 +675,7 @@ npm run dev   # runs with tsx (no build needed)
 | Who | Contributions |
 |-----|---------------|
 | [@GodsBoy](https://github.com/GodsBoy) | Proxy authentication, token redaction, error sanitization ([#2](https://github.com/askalf/dario/pull/2)) |
-| [@belangertrading](https://github.com/belangertrading) | Billing classification investigation ([#4](https://github.com/askalf/dario/issues/4)), billing reclassification root cause ([#7](https://github.com/askalf/dario/issues/7)) |
+| [@belangertrading](https://github.com/belangertrading) | Billing classification investigation ([#4](https://github.com/askalf/dario/issues/4)), cache_control fingerprinting ([#6](https://github.com/askalf/dario/issues/6)), billing reclassification root cause ([#7](https://github.com/askalf/dario/issues/7)), OAuth client_id discovery ([#12](https://github.com/askalf/dario/issues/12)), multi-agent session-level billing analysis ([#23](https://github.com/askalf/dario/issues/23)) |
 
 ## License
 


askalf (changes requested):
Thanks for the PR. The contributor row update is accurate and I'm happy to take it. Declining the FAQ hunk, and I want to explain the reasoning rather than just say no.

The FAQ as drafted prescribes a single approach — a subprocess-based bridge — as THE workaround for multi-agent session-level classification. I don't think that holds up as general advice for dario's audience for three reasons, all of which come from the diagnostic work you did in #23 itself:

**First**, your own finding was that Anthropic's classifier operates on per-account behavioral aggregates — token volume, request frequency, conversation depth on each OAuth session. A bridge that shells to `claude --print` concentrates 100% of a team's agent workload on one CC subscription on one machine. That's the behavioral shape the classifier is tuned to detect. It works today because the classifier isn't tightened enough to catch every high-volume single-account session. But your own data showed the classifier is tightening, not loosening. Recommending a single-account bridge as the workaround in our README is recommending the exact pattern the classifier is hunting for.

**Second**, a single CC subscription has one 5h rolling window and one 7d rolling window shared across every agent running through the bridge. Busy day = hard stop until it refreshes. This isn't a classifier problem, it's an architectural ceiling. Bridges can't scale past one subscription by design.

**Third**, a single-machine, single-account, single-binary setup is one machine reboot, one ISP hiccup, or one Anthropic-side account action away from total shutdown. That's acceptable for some workloads and a production liability for others.

None of this is an argument against your specific deployment — you've clearly tuned your stack around those tradeoffs and it works for you. It's an argument against framing your specific approach as THE workaround in a general README FAQ. Readers hitting dario's docs have a range of workload shapes, and a range of right answers exists depending on that shape:

- Reducing tool count and session length (which you yourself did to stay under the threshold)
- Rotating sessions more frequently
- Subprocess-based bridges for workloads that fit the single-account tradeoffs
- Multi-account pools that distribute load across independent OAuth sessions so no single account concentrates the behavioral signal

I'd be open to a FAQ entry that presents the landscape rather than a single prescription. If you want to rebase this PR with just the contributor row (which is accurate and I'll merge as-is) and propose a separate PR with the landscape framing, I'll look at that on its merits. If you'd rather just drop the FAQ hunk and keep the contributor update, that works too.

askalf (approved):
Thanks for the revision — contributor row looks accurate, merging.

PR #30, opened 2026-04-14
fix(oauth): detect Claude Code credentials from MacOS keychain

iNicholasBE:
## Summary

- Modern Claude Code versions (since ~1.0.17) store OAuth tokens in the MacOS keychain instead of `~/.claude/.credentials.json`
- `loadCredentials()` only checked file paths, so it never found existing CC credentials and always redirected to the OAuth flow — even when CC was installed and logged in
- Adds `loadKeychainCredentials()` as a fallback after file-based checks, reading from the `"Claude Code-credentials"` entry via `security` (macOS) or `secret-tool` (Linux)

## Test plan

- [x] Verified on macOS with Claude Code installed: `dario login` now prints `Found credentials. Starting proxy...` instead of starting the OAuth flow
- [ ] Test on Linux with `secret-tool` available
- [ ] Test fallback behavior when no keychain entry exists (should fall through to OAuth as before)

### DIFF
--- a/src/oauth.ts
+++ b/src/oauth.ts
@@ -7,8 +7,9 @@
 
 import { randomBytes, createHash } from 'node:crypto';
 import { readFile, writeFile, mkdir, rename } from 'node:fs/promises';
+import { execFile } from 'node:child_process';
 import { dirname, join } from 'node:path';
-import { homedir } from 'node:os';
+import { homedir, platform } from 'node:os';
 import { detectCCOAuthConfig } from './cc-oauth-detect.js';
 
 // OAuth config is auto-detected at runtime from the installed Claude Code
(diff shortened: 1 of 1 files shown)


askalf (approved):
Clean fix, well-scoped. macOS keychain detection is exactly what was missing — modern CC really does store credentials there and dario was always falling through to the OAuth flow on Mac even when CC was logged in. Linux `secret-tool` path looks correct in shape and fails silently if the entry attribute is wrong, so worst case is the same fallthrough behavior dario has today. Merging.

askalf:
Hey @iNicholasBE — heads-up that I noticed you'd been credited in the [v3.7.0 CHANGELOG](https://github.com/askalf/dario/blob/master/CHANGELOG.md) entry for this PR but had never been added to the **Contributors** table at the bottom of the README. That was an oversight on my end. Fixed in 5f0d36f — you're now listed alongside the other contributors: https://github.com/askalf/dario/blob/master/README.md#contributors

Thanks again for the keychain fix. It shipped cleanly and has been the subject of zero bug reports since, which is the best possible outcome for a credential-loading code path.

Issue #42, opened 2026-04-17
Unable to connect to Dario after switching from claude 5x to 20x plan


tetsuco:
If this is an issue on Claude's end and has nothing to do with Dario, I apologize.

After switching from the “Claude Max 5x” plan to the “20x” plan, I can no longer connect to Dario.

dario: v3.19.1

It was working fine with “max 5x,” but after changing it to “max 20x,” the following issues began to occur.
This is the log for claude-opus-4-6, but the same issue occurred when I switched to claude-opus-4-7.

```
% curl http://localhost:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-opus-4-6","max_tokens":1024,"messages":[{"role":"user","content":"Hello!"}]}'
{"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Ca8TcakK6Q45orahFYBQe"}

% curl http://localhost:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -H "Authorization: Bearer dario" \
  -d '{"model":"claude-opus-4-6","max_tokens":1024,"messages":[{"role":"user","content":"Hello!"}]}'
{"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Ca8TdrVt531tqYa15Tmsq"}

% curl http://localhost:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -H "x-api-key: dario" \
  -d '{"model":"claude-opus-4-6","max_tokens":1024,"messages":[{"role":"user","content":"Hello!"}]}'
{"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Ca8TeZzEKiVhuRuEt2EDN"}

---

% dario proxy --host=0.0.0.0 --hybrid-tools -v
  Device identity: detected
[dario] template: live capture, CC v2.1.107 (1h old)
[dario] ⚠  CC compat: installed CC v2.1.107 is newer than dario's last tested version (v2.1.104); usually fine, but untested

  dario — http://0.0.0.0:3456

  Your Claude subscription is now an API.

  Usage:
    ANTHROPIC_BASE_URL=http://0.0.0.0:3456
    ANTHROPIC_API_KEY=dario

[...]

PR #47, opened 2026-04-18
 oauth: add manual override escape hatch for rotations

earlvanze:
2025427 oauth: add manual override escape hatch for rotations
What’s in it

adds an upgrade-proof OAuth override escape hatch in src/cc-oauth-detect.ts
supports:
DARIO_OAUTH_CLIENT_ID
DARIO_OAUTH_AUTHORIZE_URL
DARIO_OAUTH_TOKEN_URL
DARIO_OAUTH_SCOPES
DARIO_OAUTH_OVERRIDE_PATH
DARIO_OAUTH_DISABLE_OVERRIDE=1
~/.dario/oauth-config.override.json
keeps auto-detection as default, override only as emergency fallback
documents the override path in README.md
documents the security model in SECURITY.md
adds detector coverage in test/oauth-detector.mjs
Validation

npm run build passed
node test/oauth-detector.mjs passed, including:
baseline prod detection
cache hit
manual override file winning over detected values

Adds mailmap for correct GitHub account attribution

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,2 @@
+# Map old/wrong email to correct email
+Earl Vanze <earlvanze@gmail.com> <earl@vanze.co>
--- a/README.md
+++ b/README.md
@@ -405,6 +405,12 @@ A version marker (`<!-- dario-sub-agent-version: X -->`) embedded in the markdow
 | `DARIO_NO_BUN` | Disable automatic Bun relaunch | unset |
 | `DARIO_MIN_INTERVAL_MS` | Legacy name for `DARIO_PACE_MIN_MS`. Still honored; new name wins when both are set. | — |
 | `DARIO_CC_PATH` | Override path to the Claude Code binary for OAuth detection | auto-detect |
+| `DARIO_OAUTH_CLIENT_ID` | Override the detected Claude OAuth client id as an emergency escape hatch | unset |
+| `DARIO_OAUTH_AUTHORIZE_URL` | Override the detected Claude OAuth authorize URL | unset |
+| `DARIO_OAUTH_TOKEN_URL` | Override the detected Claude OAuth token URL | unset |
+| `DARIO_OAUTH_SCOPES` | Override the detected Claude OAuth scopes | unset |
+| `DARIO_OAUTH_OVERRIDE_PATH` | Override file path for JSON OAuth overrides | `~/.dario/oauth-config.override.json` |
+| `DARIO_OAUTH_DISABLE_OVERRIDE=1` | Ignore env/file OAuth overrides entirely | unset |
 
 ---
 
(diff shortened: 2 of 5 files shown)


askalf (commented):
Reviewing as maintainer. Core work is clean; flagging some scope and coverage items.

## Strengths
- Right fix for the right problem. Issue #42 was exactly the "stale detector vs. Anthropic-side change" scenario, and this is the clean follow-up that saves a dario release when the next rotation hits.
- Override is applied *after* detection, so the cache still captures real detected values. Disable it later (`DARIO_OAUTH_DISABLE_OVERRIDE=1`) and we're back to pure auto-detection with no manual cache clear needed.
- Partial overrides work correctly — set only `clientId` and the other fields stay detected.
- Security framing is accurate: override file holds only public OAuth metadata (no bearer tokens or secrets).
- Disabling overrides during the baseline checks in `test/oauth-detector.mjs` prevents a local operator override file from making CI flaky — nice touch.

## Please address before merge

1. **Scope creep in `src/cc-oauth-detect.ts`.** Two unrelated comment deletions slipped in:
   - The `// The prod config object is laid out roughly as one line of minified JS. Take a generous window to be safe across minifier differences.` comment above the window slice
   - The `// Defensive: if we somehow matched the dev client_id…` comment above the dead-dev-id check

   Both explain the *why* of non-obvious decisions. Please restore them.

2. **Regex noise.** `TOKEN_URL` regex changed from `[^"]` to `[^\"]`. Semantically identical (unnecessary backslash-escape inside a regex literal) but it churns the diff and could confuse future blame. Revert to the original.

3. **Branch/title.** Branch is `fix/mailmap-email-correction` but 141 of 143 added lines are the OAuth work; the mailmap is a 2-line aside. Not blocking merge, just flagging for next time — makes the PR harder to find later in `git log --grep`.

## Follow-up OK

4. **Test coverage gaps.** Only the *file* override path is tested in `test/oauth-detector.mjs`. Missing:
   - Env-var override path
   - Env > file precedence
   - `DARIO_OAUTH_DISABLE_OVERRIDE=1` actually disables
   - Partial override (only `clientId` overridden; other fields stay detected)

   Happy to merge with just the current file test and take these as a follow-up — wanted to note them.

5. **Optional hardening.** Override currently accepts any URL. A warning (not a block) on non-HTTPS `authorizeUrl` or `tokenUrl` would add a belt-and-suspenders guard without defeating the purpose of the escape hatch.

## Recommendation

Approve after the two revert items (1, 2). Thanks for the thoughtful defensive design and the SECURITY.md note.

earlvanze:
Follow-up cleanup note from my side: I addressed the maintainer feedback locally, rebuilt/tested the cleaned branch, and briefly opened #52 as an OAuth-only replacement PR. While doing that, I found #47 had already been merged, so #52 was closed as a duplicate. Final state: #47 merged successfully, #52 closed as duplicate, and no further action is needed on this thread unless you want the branch refs tidied up later.

askalf:
Thanks for the resubmit and the extra cleanup pass, @earlvanze — appreciated. Everything on this thread looks good on our end: #47 is in master, #52 closed as you noted, branch refs are fine to leave as-is.

Heads-up: after merging #47 I noticed the `dev-client-id` block comment had been rewritten rather than restored to its pre-existing wording, so I followed up with a small nit-only PR (#51) that put the original WHY back verbatim while keeping your consequence clause. Your #52 would have landed that naturally — the timing just didn't work out. Net effect is master now has the full original commentary plus your override escape hatch and test coverage.

Thanks again for the thorough work on both the feature and the review turnaround.

PR #52, opened 2026-04-19
oauth: add manual override escape hatch for rotations

earlvanze:
Supersedes #47.

This is the same OAuth override escape-hatch work, reopened as a clean branch off current `master` with the maintainer feedback addressed.

What changed since #47:
- restored the two explanatory comments in `src/cc-oauth-detect.ts`
- reverted the no-op `TOKEN_URL` regex churn
- added test coverage for:
  - env-var override path
  - env > file precedence
  - `DARIO_OAUTH_DISABLE_OVERRIDE=1`
  - partial override preserving detected fields
- dropped the unrelated `.mailmap` commit from the branch so the PR scope is OAuth-only

Validation:
- `npm run build`
- `node test/oauth-detector.mjs` (`20/20` checks passing)

The override remains an emergency escape hatch only:
- detection stays the default path
- overrides apply after detection/cache
- env overrides still beat file overrides
- `DARIO_OAUTH_DISABLE_OVERRIDE=1` returns dario to pure auto-detection behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,12 @@ A version marker (`<!-- dario-sub-agent-version: X -->`) embedded in the markdow
 | `DARIO_NO_BUN` | Disable automatic Bun relaunch | unset |
 | `DARIO_MIN_INTERVAL_MS` | Legacy name for `DARIO_PACE_MIN_MS`. Still honored; new name wins when both are set. | — |
 | `DARIO_CC_PATH` | Override path to the Claude Code binary for OAuth detection | auto-detect |
+| `DARIO_OAUTH_CLIENT_ID` | Override the detected Claude OAuth client id as an emergency escape hatch | unset |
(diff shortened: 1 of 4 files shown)


earlvanze:
It looks like you already merged this one, this is identical to PR #47 with a proper branch name. GitHub was not allowing me to rename the existing branch without closing the previous PR.

earlvanze:
Closing this duplicate because #47 has already been merged. This clean-branch follow-up is no longer needed.

askalf:
Thanks for opening this, @earlvanze — I can see the intent (clean branch name, both comments restored verbatim, no unrelated mailmap commit in scope) and honestly #52 would have been the nicer merge. The timing just didn't line up. #47 is already in master; I followed up with a small nit PR (#51) to restore the dev-client-id comment wording that your #52 had captured natively.

Appreciate the effort to get the branch shape right — thanks.

PR #53, opened 2026-04-19
oauth: warn on non-HTTPS override URLs

earlvanze:
Follow-up to #47.

Addresses the remaining maintainer follow-up items from the review thread:

- keeps the point-4 override-path coverage on the branch:
  - env-var override path
  - env > file precedence
  - `DARIO_OAUTH_DISABLE_OVERRIDE=1`
  - partial override preserving detected fields
- adds point-5 optional hardening:
  - warning-only guard on non-HTTPS `authorizeUrl` / `tokenUrl` overrides
  - override still applies, it does not block the escape hatch

Validation:
- `npm run build`
- `node test/oauth-detector.mjs` (`22/22` passing)

Cleanup:
- deleted obsolete `fix/mailmap-email-correction` branch

### DIFF
--- a/src/cc-oauth-detect.ts
+++ b/src/cc-oauth-detect.ts
@@ -171,8 +171,18 @@ async function loadManualOverride(): Promise<OAuthOverride | null> {
   }
 }
 
+function warnOnNonHttpsOverride(name: 'authorizeUrl' | 'tokenUrl', value: string | undefined): void {
+  if (!value || /^https:\/\//i.test(value)) return;
+  console.warn(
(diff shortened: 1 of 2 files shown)


askalf:
Merged — thanks for the clean follow-up. The `applyManualOverride` chokepoint is the right spot; catches both env-var and dotfile override paths in one place, and the warning-only posture keeps the emergency escape hatch intact. Test setup (capturing `console.warn`, asserting URLs still apply *and* both warnings emit, restoring in `finally`) is exactly the shape I'd have asked for.

Appreciate you keeping the branch hygiene tight too.

PR #222, opened 2026-05-07
feat: wildcard CORS Allow-Headers, add claude-opus-4-7

Saik0s:
## Changes

1. **CORS Allow-Headers wildcard** (`*` instead of explicit list)
   When `DARIO_CORS_ORIGIN=*` is set (non-credentialed CORS), modern browsers accept `Access-Control-Allow-Headers: *`. This avoids whack-a-mole with every header that OpenAI's stainless SDK (`x-stainless-*`) or Anthropic's browser client (`anthropic-dangerous-direct-browser-access`) may send.

2. **Add claude-opus-4-7 to `/v1/models`** — same approach as the existing 4-6/4-5 entries.

3. **Update MODEL_ALIASES**
   - `opus` → `claude-opus-4-7` (latest)
   - `opus46` → `claude-opus-4-6` (new alias for pinning old)
   - `opus1m` → `claude-opus-4-7[1m]`

## Trade-offs

- `Access-Control-Allow-Headers: *` is broadly permissive, but since dario already allows the user to set `DARIO_CORS_ORIGIN=*`, there's no meaningful security boundary being weakened here. The alternative (maintaining an ever-growing header whitelist) is brittle and blocks real-world use cases.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -939,7 +939,9 @@ export async function startProxy(opts: ProxyOptions = {}): Promise<void> {
   const CORS_HEADERS = {
     'Access-Control-Allow-Origin': corsOrigin,
     'Access-Control-Allow-Methods': 'GET, POST, OPTIONS',
-    'Access-Control-Allow-Headers': '*',
+    // *-wildcard covers custom headers in non-credentialed mode, except
+    // Authorization, which is a CORS non-wildcard request-header name.
+    'Access-Control-Allow-Headers': '*, Authorization',
     'Access-Control-Max-Age': '86400',
     ...SECURITY_HEADERS,
   };
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -129,8 +129,9 @@ function loadClaudeIdentity(): { deviceId: string; accountUuid: string } {
 
 // Model shortcuts — users can pass short names
 const MODEL_ALIASES: Record<string, string> = {
-  'opus': 'claude-opus-4-6',
-  'opus1m': 'claude-opus-4-6[1m]',
+  'opus': 'claude-opus-4-7',
+  'opus46': 'claude-opus-4-6',
+  'opus1m': 'claude-opus-4-7[1m]',
   'sonnet': 'claude-sonnet-4-6',
   'sonnet1m': 'claude-sonnet-4-6[1m]',
   'haiku': 'claude-haiku-4-5',
@@ -368,7 +369,7 @@ function translateStreamChunk(line: string): string | null {
   return null;
 }
 
-const OPENAI_MODELS_LIST = { object: 'list', data: ['claude-opus-4-6', 'claude-sonnet-4-6', 'claude-haiku-4-5'].map(id => ({ id, object: 'model', created: 1700000000, owned_by: 'anthropic' })) };
+const OPENAI_MODELS_LIST = { object: 'list', data: ['claude-opus-4-7', 'claude-opus-4-6', 'claude-sonnet-4-6', 'claude-haiku-4-5'].map(id => ({ id, object: 'model', created: 1700000000, owned_by: 'anthropic' })) };
 
 interface ProxyOptions {
   port?: number;
@@ -938,7 +939,7 @@ export async function startProxy(opts: ProxyOptions = {}): Promise<void> {
   const CORS_HEADERS = {
     'Access-Control-Allow-Origin': corsOrigin,
     'Access-Control-Allow-Methods': 'GET, POST, OPTIONS',
-    'Access-Control-Allow-Headers': 'Content-Type, Authorization, x-api-key, anthropic-version, anthropic-beta',
+    'Access-Control-Allow-Headers': '*',
     'Access-Control-Max-Age': '86400',
     ...SECURITY_HEADERS,
   };
--- a/test/provider-prefix.mjs
+++ b/test/provider-prefix.mjs
@@ -64,10 +64,11 @@ console.log('\n=================================================================
 console.log('  resolveClaudeAlias — request-time alias resolution (dario#190)');
 console.log('======================================================================');
 
-assert(resolveClaudeAlias('opus') === 'claude-opus-4-6', 'opus → claude-opus-4-6');
+assert(resolveClaudeAlias('opus') === 'claude-opus-4-7', 'opus → claude-opus-4-7');
+assert(resolveClaudeAlias('opus46') === 'claude-opus-4-6', 'opus46 → claude-opus-4-6 (legacy-pin alias)');
 assert(resolveClaudeAlias('sonnet') === 'claude-sonnet-4-6', 'sonnet → claude-sonnet-4-6');
 assert(resolveClaudeAlias('haiku') === 'claude-haiku-4-5', 'haiku → claude-haiku-4-5');
-assert(resolveClaudeAlias('opus1m') === 'claude-opus-4-6[1m]', 'opus1m → claude-opus-4-6[1m]');
+assert(resolveClaudeAlias('opus1m') === 'claude-opus-4-7[1m]', 'opus1m → claude-opus-4-7[1m]');
 assert(resolveClaudeAlias('sonnet1m') === 'claude-sonnet-4-6[1m]', 'sonnet1m → claude-sonnet-4-6[1m]');
 
 // Already-canonical names pass through unchanged


askalf (changes requested):
Thanks @Saik0s — good catch on the model-list drift, and the whack-a-mole concern with `x-stainless-*` headers is a real one I've also been watching for. A couple of things to work through before this lands.

## ✅ Model-list changes — landing as-is, with one CHANGELOG note

`opus → claude-opus-4-7`, `opus46 → claude-opus-4-6`, `opus1m → claude-opus-4-7[1m]`, plus adding `claude-opus-4-7` to `OPENAI_MODELS_LIST` — these match what the docs and `src/doctor.ts:483` family table already assume. Same drift pattern we just fixed for `--effort=max` in v3.37.1. Clean.

One ask: the `opus → 4-7` swap is a silent behavior change for anyone running `dario --model=opus` in production. Worth a CHANGELOG note under `[Unreleased]` calling out the alias migration so users on the old default see it before the next `npm i -g`. Happy to write that line, or you can.

## ⚠️ Wildcard CORS — please revise

The tradeoff framing in your PR description ("modern browsers accept `Access-Control-Allow-Headers: *`") is partially right but misses a load-bearing exception that breaks dario's auth path. Specifically:

> *"The value `*` only counts as a special wildcard value for requests without credentials… The `Authorization` header doesn't accept wildcard and always needs to be listed explicitly."*
> — [MDN — Access-Control-Allow-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Headers)

The Fetch Standard formalizes this as the "[CORS non-wildcard request-header name](https://fetch.spec.whatwg.org/#cors-non-wildcard-request-header-name)" and explicitly enumerates *one* member: `Authorization`. Confirmed enforced in Firefox (since [69, bug 1309358](https://bugzilla.mozilla.org/show_bug.cgi?id=1309358)), Chromium, and WebKit — the Authorization carve-out is universal.

Net effect of this PR's diff on `proxy.ts:931`:

- Removes `Authorization` from the explicit allowlist
- Replaces with `*`
- → Browser clients sending `Authorization: Bearer <DARIO_API_KEY>` now fail preflight.

Both the OpenAI SDK (Authorization as primary auth) and `@anthropic-ai/sdk`'s browser path ([source](https://app.unpkg.com/@anthropic-ai/sdk@0.39.0/files/index.mjs)) are affected — those are exactly the clients the rest of the file is built around.

There's also a second-order issue for any future `credentials: 'include'` use case: `*` is parsed as the literal byte string with no wildcard semantics in credentialed mode, so `x-api-key`, `Content-Type`, etc. would all be rejected too. Today dario doesn't ship cookies, so this is theoretical, but the rule is worth knowing.

### Counter-proposal

Spec-legal, gets the whack-a-mole convenience for `x-stainless-*`, and preserves Authorization:

```diff
-    'Access-Control-Allow-Headers': '*',
+    // *-wildcard covers all simple/custom headers in non-credentialed mode,
+    // EXCEPT Authorization, which has its own carve-out in the Fetch spec
+    // (https://fetch.spec.whatwg.org/#cors-non-wildcard-request-header-name)
+    // and must always be listed explicitly.
+    'Access-Control-Allow-Headers': '*, Authorization',
```

That gives you:

- ✅ All `x-stainless-*` headers covered by `*`
- ✅ `anthropic-dangerous-direct-browser-access`, `x-api-key`, `anthropic-beta`, `anthropic-version` covered by `*`
- ✅ Authorization still works (explicitly listed alongside `*`)
- ✅ One line, no maintenance as Stainless adds new headers

Alternative if you want the more conservative path: explicit allowlist that adds the known stainless + Anthropic-browser headers (`x-stainless-arch, x-stainless-async, x-stainless-helper-method, x-stainless-lang, x-stainless-os, x-stainless-package-version, x-stainless-poll-helper, x-stainless-retry-count, x-stainless-runtime, x-stainless-runtime-version, x-stainless-timeout, anthropic-dangerous-direct-browser-access` on top of the existing list). More to maintain, but the strictest mode.

Either works; I'd merge the `*, Authorization` form for the same reason you reached for `*` in the first place — it's the form that doesn't require future maintenance when a new SDK header lands.

---

If you push the corrected CORS line + CHANGELOG note, this is a clean merge from my end. Sorry for the tradeoff-side detail; the spec carve-out for Authorization is one of those "I know this works, I've shipped it before" things that tends to bite the next person.

askalf:
@Saik0s — pre-merge sanity-checked the PR end-to-end against real OAuth. Cherry-picked your commit onto current master (`v3.37.7`), spun up the proxy, hit it with curl. Two clean wins, one small gap, plus the CORS thing from my earlier review still standing.

## ✅ What's working as advertised

- `GET /v1/models` returns the 4-model list with `claude-opus-4-7` first
- `claude:opus` → resolved to `claude-opus-4-7` upstream, real response `OK`
- `claude:opus46` → resolved to `claude-opus-4-6` upstream, real response `OK` (the new legacy-pin alias is live and useful)
- `anthropic:opus1m` → alias resolution itself works correctly (forwards `claude-opus-4-7[1m]` upstream)

## 🐛 One small gap — `test/provider-prefix.mjs`

Your alias change broke 2 hardcoded assertions in the unit test. CI on this repo would fail; CI on your fork didn't run because `pull_request` workflows don't fire on first-time contributors by default, which is why neither of us caught it earlier. Drop-in fix:

```diff
--- a/test/provider-prefix.mjs
+++ b/test/provider-prefix.mjs
@@ -64,11 +64,12 @@ console.log('  resolveClaudeAlias — request-time alias resolution (dario#190)'
 console.log('======================================================================');

-assert(resolveClaudeAlias('opus') === 'claude-opus-4-6', 'opus → claude-opus-4-6');
+assert(resolveClaudeAlias('opus') === 'claude-opus-4-7', 'opus → claude-opus-4-7');
+assert(resolveClaudeAlias('opus46') === 'claude-opus-4-6', 'opus46 → claude-opus-4-6 (legacy-pin alias)');
 assert(resolveClaudeAlias('sonnet') === 'claude-sonnet-4-6', 'sonnet → claude-sonnet-4-6');
 assert(resolveClaudeAlias('haiku') === 'claude-haiku-4-5', 'haiku → claude-haiku-4-5');
-assert(resolveClaudeAlias('opus1m') === 'claude-opus-4-6[1m]', 'opus1m → claude-opus-4-6[1m]');
+assert(resolveClaudeAlias('opus1m') === 'claude-opus-4-7[1m]', 'opus1m → claude-opus-4-7[1m]');
 assert(resolveClaudeAlias('sonnet1m') === 'claude-sonnet-4-6[1m]', 'sonnet1m → claude-sonnet-4-6[1m]');
```

I added the new `opus46` assertion alongside since you introduced the alias — keeps the test set covering all aliases, prevents a future regression.

With this patch applied: 60/60 tests pass on full `npm test`.

## ⚠️ CORS — counter-proposal still standing from previous review

Server-side, I confirmed the wildcard hits the wire as `Access-Control-Allow-Headers: *` exactly as your PR intends. Browser-side though, the Fetch-spec Authorization carve-out kicks in and the preflight rejects any request carrying `Authorization`. The fix is one character — counter-proposal from the previous review:

```diff
-    'Access-Control-Allow-Headers': '*',
+    'Access-Control-Allow-Headers': '*, Authorization',
```

Same convenience for `x-stainless-*` whack-a-mole, preserves the auth path. Same line, just adds the explicit Authorization listing.

## 🟡 Pre-existing behavior, not your PR — but worth knowing

While testing `opus1m`, both `claude-opus-4-7[1m]` AND `claude-opus-4-6[1m]` 404'd on the upstream subscription. The `opus1m` alias was already unusable on the test account before your PR; your change to `claude-opus-4-7[1m]` doesn't make it worse. So this isn't a regression, but worth being aware of — `opus1m` may need its own investigation later (likely a beta header / subscription tier issue on the Anthropic side, separate scope from this PR).

## To get this merged

Push two changes in your branch:

1. The test/provider-prefix.mjs diff above
2. The CORS one-liner from the previous review (`*, Authorization`)

Once CI is green I'll squash-merge and we'll ship `claude-opus-4-7` + the new alias surface in the next release. CHANGELOG note for the silent `opus → 4-7` default migration is on me — I'll add it at merge time.

Thanks again for the catch — the alias surface needed it, and the legacy-pin `opus46` is a thoughtful addition that I would not have thought to add otherwise.

Saik0s:
Addressed the review feedback in b50d21b.

Changes pushed:
- CORS header is now `*, Authorization` so `Authorization` keeps working despite the wildcard carve-out.
- Updated `test/provider-prefix.mjs` for `opus -> claude-opus-4-7`, added `opus46`, and updated `opus1m -> claude-opus-4-7[1m]`.

Verified locally:
- `npm install`
- `npm run build`
- `npm test` -> 60/60 passing

askalf:
Looks great — thanks for taking the review feedback. The wildcard `*, Authorization` is exactly the right shape (Authorization being on the CORS non-wildcard request-header list is the gotcha most people miss); the `opus46` legacy-pin alias is a thoughtful addition that lets users who *want* to stay on 4-6 do so explicitly while everyone else gets the latest opus by default.

Merging now. I'll follow up with a version-bump PR (v3.37.9 → v3.37.10) so this ships to npm + GHCR. Crediting you in the CHANGELOG entry.

PR #233, opened 2026-05-09
fix(open-browser + accounts): rundll32 dispatch on Windows + --manual wiring for accounts add

askalf:
## Summary

Two related Windows-OAuth fixes folded together. Both surfaced today while wiring up live pool-mode testing.

## Fix 1 — `rundll32` URL dispatch on Windows

`dario accounts add` (and any other path that opens a URL) silently truncated long URLs on Windows: `explorer.exe URL` re-shells the URL through the registered browser's command-line template in some configurations, and `&` past the first one gets re-parsed as a cmd separator. Browser opens with the URL cut off at an `&`; downstream OAuth endpoint reports a "missing required parameter" error because the truncated tail held the missing param.

Repro that surfaced this:

```
$ dario accounts add personal
  Opening browser to add account "personal"...
  ...
```

→ Browser landed on a URL truncated before `&state=...` → `claude.ai` returned `Invalid OAuth Request — Missing state parameter`. The state was sent correctly by dario; explorer's URL-handler chain ate it.

Switch the win32 path from `explorer.exe URL` to `rundll32 url.dll,FileProtocolHandler URL` — Microsoft's documented Win32 entry point for "open URL with default handler". System32 binary, invokes the DLL function with the URL as a single in-process string, no re-parsing.

```diff
-  return { bin: 'explorer.exe', args: [safe] };
+  return { bin: 'rundll32.exe', args: ['url.dll,FileProtocolHandler', safe] };
```

The function-name token (`url.dll,FileProtocolHandler`) is one argv element with no space around the comma — System32's rundll32 parses the comma itself.

## Fix 2 — `--manual` / `--headless` wired through `dario accounts add`

`--manual` has been documented and working for `dario login` since v3.31.x, but `dario accounts add` silently ignored it. With the dispatch bug above, users on Windows had no escape hatch because there was no manual paste path for the pool flow. (Caught live today — the user tried `--manual` and the CLI just ran the auto-callback flow, opening explorer.exe, hitting the truncation bug.)

Adds `addAccountViaManualOAuth(alias)` to `accounts.ts` (mirrors `startManualOAuthFlow` from `oauth.ts` but writes to `~/.dario/accounts/<alias>.json` instead of `~/.dario/credentials.json`) and routes the `accounts add` CLI handler through it when `--manual` / `--headless` is present.

`readLineFromStdin` in `oauth.ts` is now `export`ed so `accounts.ts` can reuse the prompt without duplicating it. `parseManualPaste` and `buildManualAuthorizeUrl` were already exported.

Help text updated to document the flag on `accounts add`.

```
dario accounts add NAME [--manual]
                         Add a new account to the pool (runs OAuth flow).
                         --manual (alias: --headless) prints an authorize
                         URL and reads the code you paste back — for
                         container / SSH / no-browser-on-this-machine
                         setups, or as the on-Windows escape hatch when
                         the URL dispatch chain truncates query params.
```

## What's NOT changed

- macOS (`open URL`) and Linux/BSD (`xdg-open URL`) paths — unchanged. Those don't go through a re-shelling chain.
- URL validation (http/https only, malformed-URL rejection, javascript:/file:/data:/vbscript: blocking) — unchanged.
- Defense-in-depth `&`/`|`/backtick handling in the auto flow — preserved.
- The auto / localhost-callback `accounts add` path — unchanged. `--manual` is opt-in.

## Tests

- `test/open-browser.mjs` per-platform-bin assertion flipped: `win32 → rundll32.exe`.
- New `Windows rundll32 argv shape` block pins the win32 argv shape (entry point token + URL, exactly two argv elements, no space around the comma).
- End-to-end `openBrowser` stub test loosens "first arg" to "last arg" because the URL is no longer the only argv element on win32.
- 30/30 in `open-browser.mjs`, 60/60 in the full suite.

## Test plan

- [x] `npm run build` clean
- [x] `node test/open-browser.mjs` — 30/30 pass
- [x] `npm test` — 60/60 pass
- [ ] Live: `dario accounts add <alias>` on Windows (auto flow) opens the full OAuth URL with state intact
- [ ] Live: `dario accounts add <alias> --manual` prints the URL, reads `code#state` back, saves to pool

## Bundles

Closes the on-Windows blocker for live pool-mode testing — once this lands, `dario accounts add` works in both modes on Windows, no copy-paste-the-URL-from-stderr workaround needed.

### DIFF
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -22,9 +22,11 @@ import { homedir } from 'node:os';
 import { randomUUID, randomBytes, createHash } from 'node:crypto';
 import { createServer, type IncomingMessage, type ServerResponse } from 'node:http';
 import { detectCCOAuthConfig } from './cc-oauth-detect.js';
-import { loadCredentials } from './oauth.js';
+import { loadCredentials, buildManualAuthorizeUrl, parseManualPaste, readLineFromStdin } from './oauth.js';
 import { openBrowser } from './open-browser.js';
 import { redactSecrets } from './redact.js';
+
+const MANUAL_REDIRECT_URI = 'https://platform.claude.com/oauth/code/callback';
 
 const DARIO_DIR = join(homedir(), '.dario');
 const ACCOUNTS_DIR = join(DARIO_DIR, 'accounts');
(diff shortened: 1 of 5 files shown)


askalf:
Subsumed into #232 — all four commits (rundll32 fix, --manual wiring, headless hint, callback-failure hint) merged into the v3.37.9 release branch. Single ship for the cycle. Branch retained for now in case we need to chase a regression to a specific commit.

Issue #235, opened 2026-05-09
accounts: `login` back-fill snapshot goes stale on credentials.json refresh-token rotation


askalf:
## Repro

User runs `dario login` (single-account, OAuth → `~/.dario/credentials.json`).

User runs `dario accounts add personal` later. The auto-back-fill kicks in: `ensureLoginCredentialsInPool()` copies the current `credentials.json` content into `~/.dario/accounts/login.json` so the pool has 2+ accounts and pool mode activates.

After that point, both stores hold the same `accessToken` + `refreshToken`. But:

- The single-account path (proxy startup, `dario refresh`, periodic background refresh in oauth.ts) keeps refreshing `credentials.json` independently. Each refresh issues a new access_token + new refresh_token; Anthropic invalidates the old refresh_token.
- The pool path (proxy startup pool refresh, `accounts.refreshAccountToken`) refreshes `accounts/login.json` independently using the snapshot's refresh_token.

First time the single-account path refreshes: the snapshot's refresh_token gets invalidated server-side. Now `accounts/login.json` has a refresh_token Anthropic rejects (`invalid_grant`), and an access_token whose effective TTL was tied to the old session. Both paths see "valid" by their local `expiresAt` field but Anthropic returns 401 on every request and 400 `invalid_grant` on every refresh.

Doctor reports the account as `[ OK ] Pool` with `2 accounts` because the health check is a TTL check on the `expiresAt` metadata, not an Anthropic-API-validity check.

The pool selector keeps picking the dead account until rate-limit data shows otherwise (which it never does, because 401 responses don't include rate-limit headers — see the related selector-failover issue).

## Why this is sharp

It's silent. A user who back-filled a week ago and runs `dario doctor` today sees green checkmarks while every routed request through the dead account 401s. The single-account `dario login` user-flow keeps working perfectly because `credentials.json` is the authoritative store for that path, and it self-heals via refresh. The pool's `login.json` is the orphan.

## Reproed in a live session today

Re-synced `accounts/login.json` from the current `credentials.json`. Within minutes Anthropic was returning 401 on requests routed to `login`. Pool refresh attempt: `Refresh failed for login (400): {"error": "invalid_grant", "error_description": "Refresh token not found or invalid"}` — confirming the snapshot's refresh_token was invalidated by the single-account path's prior background refresh.

## Proposed fix

Two complementary changes:

**1. At proxy startup, detect divergence and re-sync.**

[...]

PR #236, opened 2026-05-09
fix(pool): auth-failure cool-down + login-snapshot resync (closes #234, #235)

### LINKED ISSUE #234: pool: selector doesn't fail over on auth errors (401/403), only on 429
## Repro

Live pool-mode e2e on Windows, 2 OAuth-derived accounts (`personal`, `primary`) plus a stale `login` snapshot from a back-fill. Sent requests through the proxy:

```
GET /accounts after 3 reqs:
  login:    requestCount=3  status=unknown
  personal: requestCount=0  status=unknown
```

All 3 requests went to `login`. Each one returned `401 Invalid authentication credentials` from Anthropic (login's tokens were server-invalidated). The pool kept routing to `login` despite the consistent failures — never tried `personal`.

## Why

`AccountPool.select()` in `src/pool.ts` uses headroom (`1 - max(util5h, util7d)`) as the routing signal. Headroom is updated from rate-limit response headers, which 401 responses don't include. So an account that fails authentication looks identical to a healthy idle account from the selector's perspective — same headroom, same routing weight.

The existing failover path triggers only on 429 (`status: 'rejected'` from `parseRateLimits`). 401/403 and `authentication_error` / `permission_error` body shapes are invisible to the selector.

## Why it matters more with mixed plans (Max + Pro)

Just hit live with a Max + Pro pool. Max-tier per-model 7d buckets (`7d_sonnet`) can exhaust mid-week while the Pro account is still fresh — the right behavior is to fail over to Pro silently. Today, if Max returns a `permission_error` or `authentication_error` mid-day (revoked scope, account-state change, etc.), the pool keeps routing at the dead Max account until rate-limit data eventually flips. With a Pro account standing by, that's avoidable user-facing 5xx.

Also bites the back-fill case: a stale `login` snapshot whose tokens have been invalidated by `credentials.json`'s independent refresh-token rotation will keep getting picked because its TTL metadata says "healthy" but Anthropic has invalidated the access_token. (Filed separately as a related issue.)

## Proposed fix

Add a per-account auth-failure cool-down to `PoolAccount`:

```ts
interface PoolAccount {
  // ...
  lastAuthFailureAt?: number;  // ms epoch
  consecutiveAuthFailures?: number;
}
```

In the proxy's response handler, when an upstream returns 401/403 or a body with `error.type === 'authentication_error' | 'permission_error' | 'invalid_grant'`, call `pool.markAuthFailure(alias)`. The `select()` filter excludes accounts whose `lastAuthFailureAt` is within the cool-down window (suggest: 60s for the first failure, exponential up to 30min after N consecutive failures). Cleared by any successful response.

Surfacing in `/accounts`:

```json
{
  "alias": "login",
  "status": "auth-failed",
  "lastAuthFailureAt": 1778336504123,
  "consecutiveAuthFailures": 3,
  "cooldownMs": 240000
}
```

`dario doctor` reports auth-failed accounts as `[WARN]` so operators see "1/2 healthy" + a hint about which account and why.

## Test plan

- Unit: simulate 401 response on one account → next `select()` returns the other
- Unit: cool-down expiry → account becomes selectable again
- Unit: success on another account doesn't clear the failed account's cool-down
- Unit: success on the same account clears its own cool-down
- Live: with Max + Pro pool, force a 401 on one (e.g., revoke its scope via `accounts/<alias>` direct edit) → confirm next request hits the other, no client-visible error

Filed against the live test e2e with mixed Max + Pro plans.

askalf:
## Summary

Two related pool reliability bugs surfaced during live Max+Pro pool-mode e2e today. Both fixed in this PR with unit tests.

## #234 — pool selector skips accounts in auth-failure cool-down

**Repro:** 2-account pool, one account's tokens server-invalidated (the back-fill snapshot from #235). Sent 3 requests through the proxy. All 3 routed to the dead account; never tried the healthy peer. `/accounts` showed `requestCount: 3` on the dead account, `0` on the peer.

**Root cause:** `AccountPool.select()` routes by headroom (computed from rate-limit response headers). 401/403 responses don't include rate-limit headers, so headroom math sees a healthy idle account → selector keeps picking the dead one indefinitely.

**Fix:** per-account auth-failure cool-down on `PoolAccount`. New fields (`lastAuthFailureAt`, `consecutiveAuthFailures`), new methods (`markAuthFailure` / `clearAuthFailure`), new helpers (`authCooldownMs` with 60s base + exponential backoff capped at 30 min, `isInAuthCooldown`). `select()` / `selectExcluding()` / `selectSticky()` / `status()` all skip accounts in active cool-down.

The proxy's response-handling path now:
- Calls `markAuthFailure` on 401/403
- Attempts in-request failover (mirroring the existing 429 path)
- Calls `clearAuthFailure` on 2xx success

Surfaces in `/accounts` as `status: "auth-cooldown"` with `cooldownMs` and `consecutiveAuthFailures`.

```diff
+ if (pool && poolAccount && (upstream.status === 401 || upstream.status === 403)) {
+   pool.markAuthFailure(poolAccount.alias);
+   const nextAccount = pool.selectExcluding(triedAliases, modelFamily(requestModel));
+   if (nextAccount) {
+     // ... rebind, retry on the peer
+     continue dispatchLoop;
+   }
+ }
```

## #235 — pool's back-filled `login` snapshot goes stale on credentials.json refresh-token rotation

**Repro:** user runs `dario login` (single-account, OAuth → `credentials.json`), then `dario accounts add personal`. Back-fill copies `credentials.json` into `accounts/login.json`. Single-account path keeps refreshing `credentials.json` independently (proxy startup auth check, periodic refresh in `oauth.ts`). Each refresh issues new tokens; Anthropic invalidates the previous refresh_token. The pool's `login.json` snapshot now has tokens that 401 on every request and 400 `invalid_grant` on every refresh, but its `expiresAt` metadata still says "healthy" so the selector keeps picking it.

(With #234's fix in place that's already mostly contained — auth cool-down would route around the dead account. But the snapshot itself can be repaired cheaply at startup, eliminating the cool-down entirely.)

**Fix:** at proxy startup, before loading the pool, call `resyncLoginFromCredentialsIfStale()`. Reads `accounts/login.json`, compares to current `credentials.json`, and overwrites the snapshot when they diverge — preserving the pool-internal `deviceId` / `accountUuid` (those don't rotate with token refresh), rotating just the OAuth tokens. Returns `'no-pool'` / `'no-login'` / `'no-creds'` / `'in-sync'` / `'resynced'` so the proxy can log appropriately. Idempotent on repeated invocation.

```diff
+ const resyncResult = await resyncLoginFromCredentialsIfStale();
+ if (resyncResult === 'resynced') {
+   console.log('[dario] re-synced pool `login` account from current credentials.json (was stale; dario#235)');
+ }
```

## Test coverage

| File | New tests | Coverage |
|---|---|---|
| `test/pool-auth-cooldown.mjs` | 27 | cooldown math, mark/clear semantics, select-path filtering across all 3 selector variants, sticky-binding break-on-cooldown, status() accounting, alias-scoped failure isolation |
| `test/resync-login-stale.mjs` | 17 | five return cases (no-pool, no-login, no-creds, in-sync, resynced), token replacement invariants, deviceId/accountUuid preservation, idempotency, access-only divergence detection |

Total tests: **60 → 62** (the new files contribute 2 suite entries; the 27 and 17 are intra-file assertions).

Required a new test-only export `_clearCredentialsCacheForTest()` from `oauth.ts` because `resyncLoginFromCredentialsIfStale` calls `loadCredentials()` which caches at module scope for 10 seconds — multi-scenario tests in a single process see stale cached values without explicit invalidation.

## What's NOT changed

- Existing 429 failover path — same code, same behavior.
- Rate-limit-driven headroom routing — unchanged. The cool-down is an additional filter, not a replacement.
- The `dario accounts add` back-fill behavior — only its recovery story changes.
- credentials.json single-account refresh path — untouched. The collision is now self-healing on next proxy start.

## Why it matters more with mixed plans (Max + Pro)

Just hit live with a Max + Pro pool. Max-tier per-model 7d buckets (`7d_sonnet`) can exhaust mid-week while the Pro account is still fresh — the right behavior is to fail over to Pro silently. Today, if Max returns a `permission_error` or `authentication_error` mid-day (revoked scope, account-state change, etc.), the pool keeps routing at the dead Max account until rate-limit data eventually flips. With this PR, pool fails over within the same request and the next client request goes to Pro automatically.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — 62/62 pass
- [x] `node test/pool-auth-cooldown.mjs` — 27/27
- [x] `node test/resync-login-stale.mjs` — 17/17
- [ ] Live: simulate a 401 by tampering with one account's `accessToken` → next request fails over to peer, `/accounts` shows `auth-cooldown` status
- [ ] Live: mutate `credentials.json` (force a refresh), restart proxy → `[dario] re-synced pool \`login\`` log line, `login.json` reflects current creds

### DIFF
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -423,3 +423,60 @@ export async function ensureLoginCredentialsInPool(
 
   return alias;
 }
+
+/**
+ * Detect divergence between `accounts/login.json` and the current
+ * `credentials.json` (or whichever store loadCredentials finds), and
+ * re-sync if they differ. Returns one of:
+ *   - 'no-pool'      : pool is single-account, nothing to do
+ *   - 'no-login'     : pool active but no `login` alias — back-fill
+ *                       was never run, nothing to do
+ *   - 'no-creds'     : login.json exists but no current credentials
+ *                       reachable to compare against — leave alone
+ *   - 'in-sync'      : tokens match; no action
+ *   - 'resynced'     : login.json was stale; overwrote with current
+ *                       credentials. Caller should reload pool state
+ *
+ * Why: the single-account path keeps refreshing `credentials.json` in
+ * the background (proxy startup auth check, periodic refresh in oauth.ts).
+ * Each refresh issues new tokens and Anthropic invalidates the previous
+ * refresh_token. The pool's `login.json` snapshot — frozen at back-fill
+ * time — is now wrong on both fields, but its `expiresAt` metadata still
+ * says "healthy" so the selector keeps picking it. Detect this at startup
+ * and overwrite with the current canonical content. dario#235.
+ */
+export async function resyncLoginFromCredentialsIfStale(): Promise<
+  'no-pool' | 'no-login' | 'no-creds' | 'in-sync' | 'resynced'
+> {
+  const aliases = await listAccountAliases();
+  if (aliases.length < 2) return 'no-pool';
+  if (!aliases.includes(MIGRATED_LOGIN_ALIAS)) return 'no-login';
+
+  const loginAcc = await loadAccount(MIGRATED_LOGIN_ALIAS);
+  if (!loginAcc) return 'no-login';
+
+  const creds = await loadCredentials();
+  const tok = creds?.claudeAiOauth;
+  if (!tok?.accessToken || !tok?.refreshToken) return 'no-creds';
+
+  if (
+    loginAcc.accessToken === tok.accessToken &&
+    loginAcc.refreshToken === tok.refreshToken
+  ) {
+    return 'in-sync';
+  }
+
+  // Tokens diverged — credentials.json has refreshed since last back-fill.
+  // Overwrite the snapshot, preserving deviceId/accountUuid (they don't
+  // rotate with token refresh; they're pool-internal identity).
+  await saveAccount({
+    alias: MIGRATED_LOGIN_ALIAS,
+    accessToken: tok.accessToken,
+    refreshToken: tok.refreshToken,
+    expiresAt: tok.expiresAt,
+    scopes: tok.scopes ?? loginAcc.scopes ?? [],
+    deviceId: loginAcc.deviceId,
+    accountUuid: loginAcc.accountUuid,
+  });
+  return 'resynced';
+}
--- a/src/oauth.ts
+++ b/src/oauth.ts
@@ -46,6 +46,19 @@ const CACHE_TTL_MS = 10_000; // Re-read from disk every 10s at most
 // Mutex to prevent concurrent refresh races
 let refreshInProgress: Promise<OAuthTokens> | null = null;
 
+/**
+ * Test-only — invalidate the in-memory credentials cache so the next
+ * `loadCredentials` re-reads from disk / keychain. Production code paths
+ * never need this: the 10-second TTL is short, and `saveCredentials`
+ * already invalidates on write. But unit tests that mutate
+ * `~/.dario/credentials.json` between scenarios within the same process
+ * see stale cached values and their assertions race against the TTL.
+ */
+export function _clearCredentialsCacheForTest(): void {
+  credentialsCache = null;
+  credentialsCacheTime = 0;
+}
+
 export interface OAuthTokens {
   accessToken: string;
   refreshToken: string;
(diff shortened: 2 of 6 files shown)


askalf:
Subsumed into #232 — pool auth-failure cool-down + login-snapshot resync (closes #234, #235) merged into the v3.37.9 release branch. Single ship for the cycle.

PR #341, opened 2026-05-20
fix(runner): share OAuth credential with platform dario — drop HOME isolation

askalf:
## Summary
v4.4.1 pinned `HOME=/root/.claude-runner` across the three runner workflows (compat-test, drift-template-watch, billing-classifier-canary) to give the runner its own OAuth credential separate from `/root/.claude/.credentials.json`. Intent was to avoid refresh-token rotation races with other long-running CC clients on the host.

The isolation prevented races but introduced a worse failure: the isolated token had **no refresh authority between workflow fires**. Each workflow run lasts <10 minutes; runs are hours apart at typical cadence. Anthropic invalidates refresh tokens that sit unused too long, so within a few weeks the runner's refresh token died of disuse → every subsequent fire returned `invalid_grant` with no path to recovery short of interactive `dario login --manual`.

We hit this today: the runner credential expired at 22:33 UTC, refresh attempted at 23:30 UTC, `invalid_grant`. Same pattern almost certainly responsible for every prior \"runner-credential died\" incident.

## Fix
Drop the `HOME` override. Workflows now read `/root/.claude/.credentials.json`, which is typically owned by a 24/7 platform dario container that auto-refreshes when the access token has ~1h life remaining. Workflows fire → find a fresh token → do their <10 min job → never attempt a refresh from their side.

The race the isolation protected against is rare in practice: it only matters if a workflow's CC/dario inside it decides to refresh simultaneously with the platform dario. With the platform dario refreshing pre-emptively while ~1h life remains, workflow runs hit a fresh token and don't need to refresh themselves.

## Tradeoff acknowledged
Runner workflows now compete for the same Max-account 5h-window quota as the rest of the platform fleet. At typical cadence (drift-watch every 30 min + compat per PR + canary daily), this is a single-digit % of the headroom available on a Max plan also running an agent fleet. If a workload genuinely needs more isolation, the cadence knob is the lever — not credential isolation.

## Files
- `.github/workflows/compat-test-self-hosted.yml` — drop HOME env; explain new policy in step comment
- `.github/workflows/cc-drift-template-watch.yml` — same, on both the `--check` step and the auto-rebake step
- `.github/workflows/cc-billing-classifier-canary.yml` — same
- `.github/workflows/cc-drift-watcher-liveness.yml` — update the alert-body diagnostic command (no more `HOME=/root/.claude-runner claude --print`)
- `docs/drift-monitor.md` — drop the isolated-HOME setup section; document the shared-credential approach + history of why it changed

## Operator action (one-time) after merge
\`\`\`bash
# On the Hetzner runner host, after this PR ships:
rm -rf /root/.claude-runner  # delete the dead isolated credential dir
# /root/.claude/.credentials.json is already in place (platform dario uses it)
# Next workflow fire will pick it up; no further action needed.
\`\`\`

## Test plan
- [ ] CI green (compat will likely still 429 on the dead runner credential until the operator deletes /root/.claude-runner AND the next workflow fire reads /root/.claude/, which platform dario has been keeping fresh; that's expected — first post-merge run validates the new path)
- [ ] After cleanup, manually dispatch `cc-drift-template-watch` and confirm successful run on the shared credential
- [ ] Confirm no token-rotation race over 24h (platform dario keeps refreshing; workflows keep reading)

### DIFF
--- a/.github/workflows/cc-billing-classifier-canary.yml
+++ b/.github/workflows/cc-billing-classifier-canary.yml
@@ -70,10 +70,13 @@ jobs:
           npm run build
 
       - name: Start dario proxy (canonical-rebuild — NOT passthrough)
-        # HOME pinned to /root/.claude-runner (v4.4.1) — uses the isolated
-        # runner credential. NO --passthrough flag here: the canary is
-        # specifically validating dario's rebuild-from-canonical mode, the
-        # mode every non-CC dario user runs in.
+        # OAuth credential: shares /root/.claude/.credentials.json with the
+        # platform dario, same rationale as compat-test-self-hosted.yml —
+        # the isolated /root/.claude-runner credential had no refresh
+        # authority between sparse workflow fires, repeatedly died of disuse.
+        # NO --passthrough flag: the canary specifically validates dario's
+        # rebuild-from-canonical mode, the mode every non-CC dario user
+        # runs in.
         #
         # --port=3457 (v4.6.1) — see compat-test-self-hosted.yml for the
         # rationale. The platform runs its own dario at :3456 via docker;
@@ -82,8 +85,6 @@ jobs:
         # using PLATFORM credentials, producing 401s that look like
         # classifier flips. Use :3457 so the workflow's own dist actually
         # runs.
-        env:
-          HOME: /root/.claude-runner
         run: |
           node dist/cli.js proxy --port=3457 > proxy.log 2>&1 &
           echo $! > proxy.pid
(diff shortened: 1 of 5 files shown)


askalf:
Closing — based on wrong premise. I assumed platform dario refreshed /root/.claude/.credentials.json on the host, but platform dario stores credentials in a docker-internal named volume (dario-config) entirely separate from the host filesystem. The host's /root/.claude/.credentials.json hasn't been touched in 50+ hours and is also expired/dead. Dropping the HOME isolation just moved the problem to another orphan credential.

Real fix is one of: (a) bind-mount the platform dario credential volume to a host path so it IS what runner workflows read, with platform dario updating it; (b) a host-side cron that pulls fresh credentials from platform dario every ~30min; (c) use a real Anthropic API key (sk-ant-...) for the runner workflows entirely, bypassing OAuth and using per-token billing for these specific CI calls.

Will draft (c) — simplest, eliminates the OAuth-rot failure class entirely.

PR #476, opened 2026-06-09
auto-rebake: template drift detected 2026-06-09

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-06-09T22:58:03+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v4.8.53` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — worth a closer read

- **anthropic_beta added:** `context-1m-2025-08-07` (CC opted into a feature flag)
- **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)
- **system_prompt:** -3714 chars net (text-content drift — see unified diff below)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.169) [--check mode: dry-run]
[bake] spawning CC against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.169, 27 tools, 5856 char system prompt
[bake] scrubbed:
[bake]   tools: 27 → 27 (dropped 0 mcp__* tools)
[bake]   system_prompt: 5856 → 4395 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] previous baked template: CC v2.1.170 captured 2026-06-09T22:19:05.672Z, 30 tools, 8109 char system prompt
[bake] check: drift detected — 3 differing slots (verdict: moderate):
[bake] **Verdict:** 🟡 Moderate — worth a closer read
[bake] 
[bake] - **anthropic_beta added:** `context-1m-2025-08-07` (CC opted into a feature flag)
[bake] - **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)
[bake] - **system_prompt:** -3714 chars net (text-content drift — see unified diff below)
[bake] 
[bake] check: per-slot detail:
[bake]   • anthropic_beta added: context-1m-2025-08-07
[bake]   • anthropic_beta removed: afk-mode-2026-01-31
[bake]   • system_prompt content changed (8109 → 4395 chars, delta -3714)
[bake]         …
[bake]         - Prefer the dedicated file/search tools over shell commands when one fits. Independent tool calls can run in parallel in one response.
[bake]         - Reference code as `file_path:line_number` — it's clickable.
[bake]       -
[bake]       -# Communicating with the user
[bake]       -
[bake]       -Your text output is what the user reads; they usually can't see your thinking or the raw tool results. Write it for a teammate who stepped away and is catching up, not for a log file: they don't know the codenames or shorthand you created along the way, and they didn't watch your process unfold. Before your first tool call, say in a sentence what you're about to do; while working, give brief updates when you find something load-bearing or change direction.
[bake]       -
[bake]       -Text you write between tool calls may not be shown to the user. Everything the user needs from this turn — answers, summaries, findings, conclusions, deliverables — must be in the final text message of your turn, with no tool calls after it. Keep text between tool calls to brief status notes. If something important appeared only mid-turn or in your thinking, restate it in that final message.
[bake]       -
[bake]       -Lead with the outcome. Your first sentence after finishing should answer "what happened" or "what did you find" — the thing the user would ask for if they said "just give me the TLDR." Supporting detail and reasoning come after, for readers who want them.
[bake]       -
[bake]       -Being readable and being concise are different things, and readable matters more. If the user has to reread your summary or ask you to explain, any time saved by brevity is gone. The way to keep output short is to be selective about what you include (drop details that don't change what the reader would do next), not to compress the writing into fragments, abbreviations, arrow chains like `A → B → fails`, or jargon. What you do include, write in complete sentences with the technical terms spelled out. Don't make the reader cross-reference labels or numbering you invented earlier; say what you mean in place.
[bake]       -
[bake]       -Match the response to the question: a simple question gets a direct answer in prose, not headers and sections. Use tables only for short enumerable facts, with explanations in the surrounding prose rather than the cells. Calibrate to the user — a bit tighter for an expert, more explanatory for someone newer.
[bake]        
[bake]        Write code that reads like the surrounding code: match its comment density, naming, and idiom.
[bake]       -Only write a code comment to state a constraint the code itself can't show — never to say where it came from, what the next line does, or why your change is correct; that's you talking to the reviewer, not the next reader, and it's noise the moment the PR merges.
[bake]        
[bake]        For actions that are hard to reverse or outward-facing, confirm first unless durably authorized or explicitly told to proceed without asking; approval in one context doesn't extend to the next. Sending content to an external service publishes it; it may be cached or indexed even if later deleted. Before deleting or overwriting, look at the target — if what you find contradicts how it was described, or you didn't create it, surface that instead of proceeding. Report outcomes faithfully: if tests fail, say so with the output; if a step was skipped, say that; when something is done and verified, state it plainly without hedging.
[bake]         …
[bake]        # Memory
[bake]        
[bake]       -You have a persistent file-based memory at `C:\Users\user\.claude\projects\C--Users-user-project\memory\`. This directory already exists — write to it directly with the Write tool (do not run mkdir or check for its existence). Each memory is one file holding one fact, with frontmatter:
[bake]       +You have a persistent file-based memory at `/root/.claude/projects/project/memory/`. This directory already exists — write to it directly with the Write tool (do not run mkdir or check for its existence). Each memory is one file holding one fact, with frontmatter:
[bake]        
[bake]        ```markdown
[bake]         …
[bake]        # Context management
[bake]        When the conversation grows long, some or all of the current context is summarized; the summary, along with any remaining unsummarized context, is provided in the next context window so work can continue — you don't need to wrap up early or hand off mid-task.
[bake]       -
[bake]       -You are operating autonomously. The user is not watching in real time and cannot answer questions mid-task, so asking 'Want me to…?' or 'Shall I…?' will block the work. For reversible actions that follow from the original request, proceed without asking. Stop only for destructive actions or genuine scope changes the user must decide. Offering follow-ups after the task is done is fine; asking permission before doing the work is not.
[bake]       -
[bake]       -Exception: when the user is describing a problem, asking a question, or thinking out loud rather than requesting a change, the deliverable is your assessment. Report your findings and stop. Don't apply a fix until they ask for one.
[bake]       -
[bake]       -Before ending your turn, check your last paragraph. If it is a plan, an analysis, a question, a list of next steps, or a promise about work you have not done ('I'll…', 'let me know when…'), do that work now with tool calls. That includes retrying after errors and gathering missing information yourself. Do not stop because the context or session is long. End your turn only when the task is complete or you are blocked on input only the user can provide.
[bake]       -
[bake]       -Before running a command that changes system state — restarts, deletes, config edits — check that the evidence actually supports that specific action. A signal that pattern-matches to a known failure may have a different cause.
[bake]        
[bake] check: bundled template is stale relative to live CC. Run `node scripts/capture-and-bake.mjs` to re-bake.
[bake] wrote drift-summary.md for workflow embedding
```

### Validation

Compat-test (`compat-test-self-hosted.yml`) will auto-fire on this PR because it touches `src/cc-template-data.json`. If it goes green, this is mergeable as-is. If it fails, the rebake captured something Anthropic isn't ready to ship — close the PR, investigate manually.

### Why not auto-merged

The bundled template is the wire-shape contract for non-CC clients. compat-test exercises passthrough mode; it cannot validate the rebuild-from-canonical path, so a human approves the bake. Approving + merging bumps the patch version and triggers `cc-drift-auto-release.yml` — approval is the ship gate.

Background: [v4.2.1 CHANGELOG entry](https://github.com/askalf/dario/blob/master/CHANGELOG.md#421---2026-05-17) documents this drift class — same CC binary, different runtime wire shape via remote configuration.


askalf:
Closing — this auto-rebake is a backwards capture, not real upstream drift.

The capture ran against the dario self-hosted runner's **stale CC v2.1.169**, while master's bundle was baked from **v2.1.170**. Merging would regress the template:

- CC version v2.1.170 → **v2.1.169** (older)
- tools 30 → **27** (drops 3)
- system_prompt ~8,100 → **4,395 chars** (−3,714; whole behavioral sections deleted)
- version bump targets 4.8.53 but master is already **4.8.55** (hence the conflict)

The memory-path line also flips `C:\Users\user\...` → `/root/...`, confirming a degraded runtime capture. The drift this PR's issue (#477) flagged was already handled correctly by #475 (rebake to v2.1.170) / #478 / #479. compat-test green is insufficient here — compat only exercises passthrough mode, not the canonical rebuild path where a half-size prompt would bite.

Root cause is the runner's CC being behind v2.1.170 — fixing that stops the false-positive drift loop. Tracking in #477.

PR #550, opened 2026-06-19
feat(doctor): display duration until 5h limit resets

boredland:
This PR extracts the rate limit reset timestamp from the 'anthropic-ratelimit-unified-reset' header and calculates/displays a human-readable duration countdown (e.g. 'resets in 1h 9m') inside the 'Usage 5h (all)' doctor check. This allows developers to see exactly when their rate limits will expire.

### DIFF
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -655,10 +655,25 @@ export async function runChecks(opts: RunChecksOptions = {}): Promise<Check[]> {
       const bucket = billingBucketFromClaim(firstOk.claim);
       const pct = (n: number) => `${(n * 100).toFixed(1)}%`;
 
+      let resetStr = '';
+      if (firstOk.reset > 0) {
+        const diffMs = (firstOk.reset * 1000) - Date.now();
+        if (diffMs > 0) {
+          const diffMins = Math.ceil(diffMs / 60000);
+          if (diffMins >= 60) {
+            const hrs = Math.floor(diffMins / 60);
+            const mins = diffMins % 60;
+            resetStr = `  •  resets in ${hrs}h ${mins}m`;
+          } else {
+            resetStr = `  •  resets in ${diffMins}m`;
+          }
+        }
+      }
+
       checks.push({
         status: firstOk.util5h >= 0.90 ? 'warn' : 'ok',
         label: 'Usage 5h (all)',
-        detail: `${pct(firstOk.util5h)} used  •  status=${firstOk.status}  •  claim=${firstOk.claim} (${bucket})`,
+        detail: `${pct(firstOk.util5h)} used  •  status=${firstOk.status}${resetStr}  •  claim=${firstOk.claim} (${bucket})`,
       });
       checks.push({
         status: firstOk.util7d >= 0.90 ? 'warn' : 'ok',
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -48,6 +48,29 @@ export interface Check {
 }
 
 /**
+ * Format a epoch timestamp reset time relative to the current time.
+ * Returns a human-friendly string like "1h 9m", "45m", "2d 3h".
+ */
+export function formatReset(resetEpochSecs: number, nowMs: number): string {
+  const ms = (resetEpochSecs * 1000) - nowMs;
+  if (ms <= 0) return '0m';
+  const totalMins = Math.round(ms / 60000);
+  if (totalMins <= 0) return '0m';
+
+  const days = Math.floor(totalMins / 1440);
+  const hours = Math.floor((totalMins % 1440) / 60);
+  const mins = totalMins % 60;
+
+  if (days > 0) {
+    return `${days}d ${hours}h`;
+  }
+  if (hours > 0) {
+    return `${hours}h ${mins}m`;
+  }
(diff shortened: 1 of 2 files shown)


src/doctor.ts
```diff
@@ -655,10 +655,21 @@ export async function runChecks(opts: RunChecksOptions = {}): Promise<Check[]> {
...
+        }
+        const timeStr = date.toLocaleTimeString([], options);
+        resetStr = `  •  resets at ${timeStr}`;
+      }
+
       checks.push({
         status: firstOk.util5h >= 0.90 ? 'warn' : 'ok',
         label: 'Usage 5h (all)',
-        detail: `${pct(firstOk.util5h)} used  •  status=${firstOk.status}  •  claim=${firstOk.claim} (${bucket})`,
+        detail: `${pct(firstOk.util5h)} used  •  status=${firstOk.status}${resetStr}  •  claim=${firstOk.claim} (${bucket})`,
```
askalf: The reset is attached to **Usage 5h (all)**, but `firstOk.reset` is the single `anthropic-ratelimit-unified-reset` value, which tracks the *representative (binding)* window — not specifically the 5h one. See `src/live-fingerprint.ts:431-434`, where `claim=five_hour` is paired with `reset = now + 18000s` (5h). Users regularly hit `representative-claim: seven_day` (`docs/faq.md`, `README`); in that state this line shows the **7-day** reset under the **5h** label.

This line already renders `claim=${firstOk.claim}`, so the cleanest fix is to append the reset only when the claim maps to `five_hour`, or to label it as the representative window. Worth confirming against a real `seven_day` capture first.

src/doctor.ts
```diff
@@ -655,10 +655,21 @@ export async function runChecks(opts: RunChecksOptions = {}): Promise<Check[]> {
...
 
+      let resetStr = '';
+      if (firstOk.reset > 0) {
+        const date = new Date(firstOk.reset * 1000);
+        const options: Intl.DateTimeFormatOptions = { hour: '2-digit', minute: '2-digit' };
+        if (date.getDate() !== new Date().getDate()) {
+          options.weekday = 'short';
+        }
+        const timeStr = date.toLocaleTimeString([], options);
+        resetStr = `  •  resets at ${timeStr}`;
```
askalf: Two small things here:

- **Consistency:** the rest of dario reports these as relative durations — `enrich429` says "resets in Xm" (`src/proxy.ts:884`) and the OAuth check says "expires in …" (`doctor.ts:521`). An absolute "resets at 09:30 PM" breaks that idiom, and rendered with the host's locale/timezone and no tz label it's harder to read on a server (`dario doctor` runs on the host, not necessarily your machine). The "resets in 1h 9m" wording from the PR description actually fits better — and would let you reuse `enrich429`'s conversion.
- **Testability:** this lives inside `runChecks()`, which is intentionally *not* unit-tested (it probes the live host — see the header comment in `test/doctor-formatter.mjs`). Pulling the formatting into a pure exported helper (e.g. `formatReset(resetEpochSecs, now)`) would make the day-boundary branch testable next to the other formatters.

src/doctor.ts
```diff
@@ -655,10 +655,21 @@ export async function runChecks(opts: RunChecksOptions = {}): Promise<Check[]> {
       const bucket = billingBucketFromClaim(firstOk.claim);
       const pct = (n: number) => `${(n * 100).toFixed(1)}%`;
 
+      let resetStr = '';
+      if (firstOk.reset > 0) {
+        const date = new Date(firstOk.reset * 1000);
+        const options: Intl.DateTimeFormatOptions = { hour: '2-digit', minute: '2-digit' };
+        if (date.getDate() !== new Date().getDate()) {
```
askalf: `getDate()` is day-of-month (1–31), so this only adds the weekday when the *day number* differs. Fine for a ≤7-day window, but it's an incorrect "different day" test in general — a reset exactly a month out on the same date wouldn't get a weekday. `date.toDateString() !== new Date().toDateString()` (or comparing y/m/d) is the robust check. Minor.

askalf (commented):
Thanks @boredland — useful thing to surface, and the core is sound: `anthropic-ratelimit-unified-reset` is epoch seconds, so `new Date(firstOk.reset * 1000)` is the right conversion (it's exactly what `enrich429` does in `src/proxy.ts:881-884`).

Two things worth resolving before merge, plus a couple of inline nits:

1. **PR title/body are now stale.** They describe a *duration* countdown ("resets in 1h 9m"), but the current code renders an *absolute* time ("resets at 09:30 PM"). Worth syncing the description to the implementation — or the other way around (see the inline note about matching the existing "expires in …" / "resets in …" style).
2. **The "5h" framing can misattribute.** There's a single `reset` header and it tracks the *binding / representative* window, which is often `seven_day` — details inline on the `detail:` line.

None of this is a crash. Happy to push a polished version (shared helper + unit test + claim-aware label) if you'd rather not round-trip on it.

askalf:
Thanks for the thorough turnaround, @boredland — the claim-aware routing (5h vs 7d, fallback variants included), the extracted `formatReset` helper, and the unit tests all landed exactly right. CI + CodeQL green. Merging. 🙏

PR #583, opened 2026-06-26
feat(proxy): preserve client output_config.format via --preserve-output-format

pnewell:
## Problem

Structured-output clients send their JSON schema in `output_config.format` (Anthropic's native structured outputs). dario rebuilds `output_config` from the CC template with `effort` only, so the schema is dropped on the way upstream. The model then free-runs into prose (typically a Markdown-fenced JSON block with different keys), and the client's strict schema parser rejects it. Concretely, the Vercel AI SDK's `generateObject` fails with `No object generated: response did not match schema`.

## Fix

Add an opt-in `--preserve-output-format` flag (env `DARIO_PRESERVE_OUTPUT_FORMAT`) that carries the client's `output_config.format` through the rebuild. Default off, so the strict-CC wire shape is unchanged unless you set it.

Design notes:
- **Independent of `--skip-fields`.** That opts out dario's *injected* fields; this preserves the *caller's* schema constraint. So `--skip-fields=output_config --preserve-output-format` drops the injected `effort` but keeps the caller's `format`.
- **Rides on whatever model the caller chose** (not gated on the haiku carve-out), because the constraint is enforced upstream during decoding, unlike the injected thinking/effort that haiku rejects.

## Billing

Verified on `claude-sonnet-4-6` that a request carrying `output_config.format` still routes to the subscription pool (`five_hour`): headers, beta flags, metadata, and OAuth identity are all untouched, so the classifier sees the same fingerprint.

## Tests

`test/preserve-output-format.mjs` (mirrors `skip-fields-flag.mjs`): format dropped by default, preserved when the flag is set, no-op when the client omits it, independent of `skipFields`, and preserved on haiku. Full suite stays green.


askalf:
Merged and shipped in **v4.8.99** — live on npm (`npm i -g @askalf/dario@4.8.99`) and the [GitHub release](https://github.com/askalf/dario/releases/tag/v4.8.99). Thanks for the clean, well-tested PR, @pnewell — opt-in/default-off plus the mirrored `skip-fields` test made it easy to take.

To exercise it:

```
dario proxy --preserve-output-format
# or: DARIO_PRESERVE_OUTPUT_FORMAT=1 dario proxy
```

Then point your structured-output client (e.g. the Vercel AI SDK's `generateObject`) at the proxy — the client's `output_config.format` now rides through to upstream instead of being dropped.

It's gone through full CI, but since you've got the real client/setup in front of you, **please let me know here if anything's off** — particularly:

- the schema actually being honored end-to-end on your client + model, and
- requests still landing on the subscription pool with the flag on (verified on `claude-sonnet-4-6`; a second data point from your setup/model would be great).

Thanks again for the contribution.

pnewell:
> - the schema actually being honored end-to-end on your client + model, and
> - requests still landing on the subscription pool with the flag on (verified on claude-sonnet-4-6; a second data point from your setup/model would be great).

@askalf Both confirmed, works great, thank you!

PR #628, opened 2026-07-01
chore(template): clean rebake to CC v2.1.198

askalf:
`capture-and-bake --check` flagged wire-shape drift after CC **2.1.198**. Clean re-bake of the bundled CC wire template.

## Method (clean-capture discipline)
Captured on **Opus** via `ANTHROPIC_MODEL=claude-opus-4-8`, **not** the session default (Fable), because `capture-and-bake` spawns CC with the saved default and CC 2.1.198 ships **Fable a larger, model-specific system prompt** — extra `# Communicating with the user` / autonomy sections + the `This iteration of Claude is Claude Fable 5…` identity block. Baking that into the *shared* base would inject a Fable identity into every non-Fable request. On Opus the only real drift is the new beta.

## What changed
- **beta base** picks up `afk-mode-2026-01-31` (dario already strips it per-account at runtime on the Max-5x 400, so no behavior change);
- system prompt refresh **4395 → 4417 chars**; tools unchanged (**33**);
- `_version` / `_supportedMaxTested` → **2.1.198**;
- model-conditional betas (`context-1m`, `fallback-credit`) remain stripped from the base and appended per-model by `betaForModel`.

## Verification
- `capture-and-bake --check` (opus) → **no drift detected**.
- No `Claude Fable 5` block / no fable-only sections in the base; beta base has no `fallback-credit`/`context-1m`.
- Full suite **97/97**, `tsc` clean.

## Scope notes
- Left `SUPPORTED_CC_RANGE.maxTested` (compat-range.ts) untouched — that's the cc-drift bot's version-label bump; not competing with it.
- **Flagged for follow-up (not fixed here):** CC 2.1.198 gives Fable a materially different/larger system prompt than the shared base. dario injects one base system prompt for all models, so Fable requests get the Opus-shaped base — a Fable-specific fidelity gap that would need per-model system-prompt templating (a design change, not a rebake).


askalf:
Superseded by #629 — the per-model system-prompt work re-bakes the full 2.1.198 template (base on Opus) plus the new Fable system-prompt variant in one change, so this standalone rebake is folded in there.

Issue #288, opened 2026-05-16
v4.1: overage-guard — halt proxy on any `representative-claim: overage` to prevent API-rate bleed


askalf:
## Summary

Add an **overage-guard**: when any request returns `representative-claim: overage`, dario halts the proxy to prevent silent API-rate bleed. Subscribers should never see a single overage hit during normal operation — if one lands, something is wrong (wire-shape drift, classifier change, account misconfig) and the right default is "stop, surface it, let the user fix it" instead of "keep dumping requests against per-token billing."

The building blocks already exist:
- Each response's rate-limit headers carry `representative-claim` with the bucket name
- `/analytics` tracks per-request bucket
- `/analytics/stream` SSE already feeds the TUI Hits tab
- The Hits tab already shows `Billing bucket: <name>` per request

What's missing is the **threshold-triggered behavior + UX**.

### Detection (`src/proxy.ts` / `src/analytics.ts`)

On every response, parse `representative-claim` header. If `=== "overage"`:
1. Set proxy-wide state: `halted = { since: now, reason: 'overage_detected', request: { reqId, model, account } }`
2. Emit `/analytics/stream` event: `{type: 'overage_detected', request: {...}, action: 'halt'}`
3. Best-effort OS notification (terminal BEL always; `osascript`/`notify-send`/PowerShell toast where available)

### Halt enforcement (`src/proxy.ts`)

On every incoming request, check `halted`:
- If `halted` AND (`now < halted.since + cooldownMs`) AND not manually resumed → return `503` with Anthropic-shaped error body:
  ```json
  {
    "type": "error",
    "error": {
      "type": "dario_overage_guard",
      "message": "dario halted to prevent API-rate bleed. A request was classified as 'overage' (per-token billing) instead of your subscription pool. Resume: `dario resume` (or wait until <ISO ts>). Details: github.com/askalf/dario#overage-guard"
    }
  }
  ```
- The Anthropic-shaped error body means CC / Cursor / Aider / Cline all surface the message verbatim to the user — no client-specific handling needed.

### Resume paths

- Manual: `dario resume` CLI → POST `/admin/resume` on local proxy → clear `halted`
- Automatic: cooldown expires (default `30 * 60 * 1000` ms = 30 min)
- TUI: `r` key on Status tab when halted → POST `/admin/resume`

### Config (`src/config-file.ts`)

```json
{
  "overageGuard": {
    "enabled": true,
[...]

PR #672, opened 2026-07-05
fix(billing): classify *_overage_included claims as subscription — stop guard halt loops

askalf:
## Live incident, 2026-07-05

A Max account crossed into its included high-tier overage credit — 7d window at 82%, the `7d_oi` per-model bucket at 99% — and Anthropic began returning a claim dario had never seen:

```
anthropic-ratelimit-unified-representative-claim: seven_day_overage_included
anthropic-ratelimit-unified-status: allowed_warning
```

The response itself was healthy: genuine model echo (`"model":"claude-fable-5"`, not a fallback), `stop_reason: end_turn`, overage-utilization 0. This is the plan's **included** credit (the `fallback-credit-2026-06-01` mechanism) — $0 out of pocket.

The overage guard's halt-on-unknown allow-list design (#288) classified it as non-subscription billing and 503'd the proxy on a 30-minute cooldown loop:

```
[dario] OVERAGE-GUARD HALTED: claude-fable-5 … returned representative-claim=seven_day_overage_included …
```

Worst-case timing by construction: the claim appears exactly when the weekly window tightens, i.e. when every request an agent fleet sends will re-trip the halt.

## Fix

- `SUBSCRIPTION_CLAIMS` + `billingBucketFromClaim` learn `five_hour_overage_included` / `seven_day_overage_included` → `subscription`. `isNonSubscriptionBilling` therefore stops halting on them.
- The proxy billing log's overage-`0%` fallback branch now keys off `SUBSCRIPTION_CLAIMS` instead of a hand-rolled duplicate of the same four strings — one sync source, can't drift again.
- Real paid overage still arrives as `overage` → still halts. `api` and novel credit-bucket claims (`sdk_credit`, …) still halt. The allow-list philosophy is unchanged; this only registers a claim now **observed live** to be $0 subscription-side.
- `doctor --usage` needed no change — its reset-window display keys on `claim.startsWith('seven_day')`, which the new claim satisfies.

## Evidence trail (for the skeptical read)

- Requested `claude-fable-5` → served `claude-fable-5`, `end_turn` (no model degradation while on included credit).
- `status=allowed_warning`, not a paid-overage status; `overage-utilization: 0`.
- The `7d_oi` bucket (parsed generically by `parseRateLimits`'s `7d_<family>` scan) sat at 99% when the claims began — consistent with `oi` = the included-overage meter.
- Open question worth a follow-up observation, not code: what happens at `7d_oi` = 100% — hard 429, paid `overage` claim (guard halts, correctly), or `*_fallback` claims with a downgraded served model. Whichever appears, the guard's existing behavior is already right for the first two; the third would surface as `subscription_fallback` in analytics.

Tests: `analytics-billing-bucket.mjs` +6 checks (bucket mapping + halt predicate both directions); full suite 107/107.


askalf:
Pushed a second fix into this PR — same incident, discovered while verifying warn-mode: `notify()` crashes the whole proxy when the toast binary is missing. `spawn()` ENOENT is an async `'error'` event, not a sync throw, so the existing try/catch never saw it. In the Docker image (no `notify-send`) every guard event — halt *or* warn — killed the process (~15s of refused connections per event, RestartCount climbing). All three notify spawn sites now go through `spawnFireAndForget()` with the handler attached; survival test added. CHANGELOG covers both fixes under 4.8.137.

PR #715, opened 2026-07-11
v5.0.0: pool-as-primitive + shim removal (#701)

### LINKED ISSUE #701: v5.0: pool-as-primitive architecture + shim removal
## Thesis

dario earns a major version by a **breaking simplification**, not a feature pile-on. The project's value is "one job, zero runtime deps, ~20k lines you can read in a weekend" — so v5 should make the codebase *smaller and clearer* while changing dario's fundamental shape, not bolt on surface area.

Two changes do that together:

1. **Pool becomes the primitive.** Today dario is single-account-first with the multi-account pool as an add-on, and the proxy carries a `pool ? … : single-account` fork through ~15 sites plus `ACCOUNT_KEY_SINGLE` special-casing and a parallel `credentials.json` code path. v5 collapses that: **an account pool is the one model, and a plain `dario login` is just a pool of one.** One routing path, not two.
2. **Shim mode is removed.** Already deprecated in v4.2 with removal scheduled for v5.x (`src/cli.ts:1138`, `1572`). Shim only ever matched 3 of 8 classifier axes and falls back to passthrough for the 1-block system prompts `claude -p` / the Agent SDK send — proxy mode is strictly better for every non-CC client. Removing it deletes a footgun and a maintenance surface.

Net effect: fewer code paths, one credential model, a real breaking change → a legitimate v5.0 with a migration note.

## Why now

The pool routing plane is already built **and** validated end-to-end as of the recent `test/pool-e2e.mjs` work (per-model-family headroom routing, sticky-session cache locality, cascading 429 failover, auth-cooldown skip). So "pool as primitive" is a **de-risked consolidation** — the hard part (correct routing) is in the bank and covered; v5 is mostly deleting the second path and pointing everything at the first.

### 1. Pool-as-primitive
- [ ] Collapse the `pool ? … : single-account` branches in `src/proxy.ts` (~15 sites: request selection, rate-limit recording, 429/auth failover, analytics keying) onto a single always-pool path.
- [ ] A plain `dario login` materializes a one-account pool under the reserved `login` alias on startup (the `ensureLoginCredentialsInPool` back-fill already does most of this — make it the default, not a special case).
- [ ] Retire `ACCOUNT_KEY_SINGLE` / `__default__` special-casing; a pool of one uses the same alias-keyed path as a pool of many.
- [ ] Keep the single-command UX identical: `dario login` + `dario proxy` still "just works" with no `accounts add` step. This is an internal unification, not a UX change for solo users.
- [ ] `dario doctor` / config-report show "pool of 1" consistently instead of a separate single-account mode.

### 2. Shim removal
- [ ] Delete `src/shim/` (`host.ts`, `runtime.cjs`), the `shim` CLI command and its flags (`src/cli.ts`), and the runtime-fingerprint shim path.
- [ ] Remove shim tests (`test/shim-*.mjs`) and `docs/shim.md`; drop shim references from README/commands/FAQ.
- [ ] Ensure the deprecation warning has been live long enough (it has since v4.2) and the removal is called out prominently in the migration note.

### 3. Config schema + migration
- [ ] Bump the config schema; drop any single-account-only knobs that no longer apply once pooling is the primitive (audit `--session-*` single-account wording in `src/cli.ts`).
- [ ] `MIGRATION.md`: a v4 → v5 section — what shim users do (switch to `dario proxy`), what changes for config, and the (zero-effort) path for existing single-account users.

## Non-goals (explicit)

- **No team / hosted / multi-tenant / dashboard / SSO features.** Those belong to a separate product surface by design; pulling them into dario would cost the "own your box, auditable in a weekend" clarity that is the whole point. v5 is a *simplification*, and this boundary is part of what keeps it one.
- No new provider integrations or protocol surface. Backends stay as-is.

## Success criteria

- One request path in `proxy.ts` (no `pool ? … : single` fork).
- Solo `dario login` + `dario proxy` UX unchanged; `pool-e2e` + full suite green.
- `shim` gone from code, tests, and docs; README/commands reflect proxy-only.
- `MIGRATION.md` v4→v5 written; a clean upgrade for both solo and multi-account users.

This is a tracking issue — sub-PRs can land the three workstreams independently behind the v5 milestone, with pool-as-primitive gated on the shim removal so the final path unification happens once.

askalf:
Release PR for **v5.0.0** — closes #701. Merging this to `master` cuts the 5.0.0 release (npm publish via the trusted-publishing pipeline), so it's the operator's call; I've stopped at PR-ready.

v5 is a **breaking simplification**: fewer code paths, one credential model, the codebase smaller and clearer. No feature pile-on. It bundles the two sub-PRs already reviewed and merged onto the `v5` integration branch, plus the version stamp.

## What's in it

**1. Pool-as-primitive (#710).** The `pool ? … : single-account` fork is gone from `proxy.ts`. A plain `dario login` is now a pool of one under the reserved `login` alias; a pool of many is the identical request path with more members. `ACCOUNT_KEY_SINGLE` retired; `shouldUsePool()` and the `pool === null` sentinel removed; doctor/config/accounts-list/banner report "pool of 1" uniformly. Session-id rotation is unified per-account (seeded so busy accounts stay byte-identical), which keeps solo UX identical while deleting the last fork.

**2. Shim removal (#707).** `dario shim` and `src/shim/` are deleted. Shim was deprecated in v4.2 (matched only 3 of 8 classifier axes; fell back to passthrough on 1-block system prompts). `dario shim` now prints a proxy-mode pointer and exits 1. Proxy mode is strictly better for every non-CC client.

**3. Release stamp.** `package.json` / `package-lock.json` → 5.0.0; CHANGELOG `[Unreleased]` → `[5.0.0] - 2026-07-11`.

## Breaking changes (all covered in MIGRATION.md)

- **`dario shim` is removed** — switch to `dario proxy`. The `--priority` RDP knob moves to OS-level (`start /belownormal`, `PriorityClass`, Process Lasso); `DARIO_SHIM_NO_DEPRECATION_WARNING` is inert.
- **The pool is the one credential model** — invisible for solo users (`dario login` + `dario proxy` unchanged; `credentials.json` untouched, back-fill is copy-only). Internal scripts: the `/accounts` endpoint always returns `{ mode: 'pool' }` now.
- **Multi-account session-id rotation** — each account's session id now rotates on its own idle timer instead of staying static (more CC-faithful; byte-identical until idle; sticky cache locality unaffected). Flagged in #710 as the one deliberate behavioral change.
- **Config schema unchanged** — v4 `config.json` loads as-is.

## Verification

- `tsc --noEmit`, `npm run build`, `npm run lint:pkg`: clean.
- `npm test`: **106/106** green on `v5`.
- Both sub-PRs passed the self-hosted `compat` check (full suite against a live proxy with real OAuth) — the unified path and shim-free proxy are validated end-to-end, not just in unit tests.

## Not done here (operator's call, by design)

- The `SUPPORTED_CC_RANGE.maxTested` value is untouched — left to the cc-drift bot to avoid a collision.
- Post-merge: the trusted-publishing release of 5.0.0, and surfacing the MIGRATION note in the release notes / any announcement.


askalf:
> **Automated review from the Sprayberry Labs fleet code reviewer.** This is a programmatic review, not a human approval.

---

**Verdict: no blocking issues — this is a clean, well-structured simplification. Two minor observations below; neither blocks merging.**

---

### What was reviewed

`v5` → `master` (+560 / −1733). Core changed files: `src/proxy.ts`, `src/cli.ts`, `src/session-rotation.ts`, `src/config-report.ts`, `src/doctor.ts`, `src/live-fingerprint.ts`, `src/runtime-fingerprint.ts`; deleted `src/shim/host.ts`, `src/shim/runtime.cjs`, `test/shim-e2e.mjs`, `test/shim-runtime.mjs`, `test/pool-activation.mjs`; updated `test/session-rotation.mjs`; docs/changelog/migration stamp.

---

### Findings

#### Minor 1 — perClient=true + multi-account combination is untested (`test/session-rotation.mjs`)

The new key formula for `perClient=true` is:

```typescript
// src/session-rotation.ts:165
`${accountKey}\u0000${clientKey && clientKey.length > 0 ? clientKey : 'default'}`
```

The test suite has a case for `perClient=false` + multi-account (two different `accountKey`s get distinct sessions — `'login'` and `'work'`), and separate cases for `perClient=true` + single account (two different `clientKey`s under the same `'acct'`). There is no test for the cross-product: two different `accountKey`s, same `clientKey`, `perClient=true` — which should give distinct sessions.

The logic is trivially correct from the null-byte separator (and it's a realistically impossible collision vector since account aliases are filesystem names and HTTP header values both exclude `\u0000`). But a one-case test would make the invariant explicit:

```javascript
// Suggested addition to the perClient=true block in test/session-rotation.mjs
header('SessionRegistry — perClient=true still partitions across accounts');
{
  const reg = new SessionRegistry(defaultCfg({ perClient: true }), idFactory());
  const a = reg.getOrCreate('login', 'client-x', 0);
  const b = reg.getOrCreate('work',  'client-x', 1 * MIN);
  check('same clientKey, different account → distinct sessions', a.sessionId !== b.sessionId);
  check('registry size=2', reg.size() === 2);
}
```

---

#### Minor 2 — help text still contains one orphaned shim reference (`src/cli.ts:1262`)

The `--strict-tls` entry in `help()` still mentions shim as a fallback:

```typescript
// src/cli.ts:1262 (after the patch)
+                             under it. (v3.23)
```

This is fine — the line was already updated from "or use shim mode" to just "under it". No stale reference remains there. *(Self-correcting — confirmed clean on second read.)*

The only place I found worth noting is the docs: `MIGRATION.md` now reads "Use proxy mode instead" but `docs/faq.md` at line 38–40 was patched to update the RDP/priority guidance. Both look clean.

Nothing to act on here.

---

### Correctness checks (all clear)

**Self-heal / double `ensureLoginCredentialsInPool` call (`src/proxy.ts`)** — intentional design: the first call (unconditional, `catch`-silenced) creates `accounts/login.json` when credentials are already fresh. If it silently fails (expired or absent token), `loadAllAccounts()` returns empty, and the `pool.size === 0 && !adminEnabled && !upstreamApiKey` gate calls `resolveSingleAccountStartupStatus()` to refresh the token, then calls `ensureLoginCredentialsInPool()` a second time to write the refreshed credentials. The second call is not redundant: it copies the just-refreshed token that didn't exist when the first call ran. ✓

**`Math.min(...pool.all().map(...))` safety** — reached only in the `else` branch after `if (pool.size === 0)`, so the spread is always non-empty. ✓

**`seedId` only-on-`rotate-new`** (`src/session-rotation.ts:170`):
```typescript
upstreamSessionId: decision === 'rotate-new' && seedId ? seedId : this.newId(),
```
Idle/age rotation always calls `this.newId()` (a fresh UUID), not the seed. The new test `'idle rotation ignores the seed and mints fresh'` confirms this. ✓

**Session-ID body/header agreement on 429 failover** — v4 pool mode set `outboundSessionId = poolAccount.identity.sessionId` from the NEW account after a failover, while the already-built CC body still carried the OLD account's `sessionId` — body and header disagreed. v5 uses `preBodySessionId` for both, so they agree across the failover. This is a correctness improvement, not a regression. ✓

**`pool !== null` invariant and all `if (pool &&)` → `if (poolAccount)` simplifications** — `pool` is always `new AccountPool()` (never null). The remaining `if (poolAccount)` guards are correct: `poolAccount === null` is the sole path with no pool selection (upstream-api-key mode). ✓

**`ACCOUNT_KEY_SINGLE → ACCOUNT_KEY_APIKEY` rename** — both are local `const`s inside `startProxy`, never persisted; no external impact. ✓

---

### What's good

This PR does exactly what a breaking simplification should: it makes the code smaller and less branchy without changing the solo-user experience. The `pool ? … : single-account` fork ran through selection, rate-limit recording, failover, analytics keying, `/accounts` endpoint, `/status`, and the catalog token getter — removing it cleans up all of those. The self-heal path for expired credentials (gap #1) is preserved and explicitly tested. The `seedId` mechanism is an elegant way to preserve byte-for-byte wire compatibility on first request while unifying the session-rotation path. MIGRATION.md is thorough. The 106-green test suite and the doc updates (CHANGELOG, MIGRATION, README, commands, FAQ) are all consistent.

PR #745, opened 2026-07-13
fix(oauth): treat invalid_grant as a terminal, loud dead-token state

askalf:
Refs #737.

## Problem

A dead refresh token — revoked, expired, or (most commonly) **rotated out by a second dario using the same account's single-use refresh token** — comes back from Anthropic as **HTTP 400 with an `invalid_grant` body** (`"Refresh token not found or invalid"`).

`doRefreshTokens` only treated **401/403** as terminal. So `invalid_grant` (a 400) fell through to `continue`:

1. 3 doomed retries with exponential backoff, then a vague `"Token refresh failed after 3 attempts."`
2. Meanwhile `getStatus()` only flips to `broken` after `REFRESH_BROKEN_THRESHOLD` consecutive failures — each gated by a 60s cooldown — so for minutes it keeps reporting **healthy** while every non-Haiku call 401s.

That's the masked-dead-token trap: container `Up (healthy)`, Haiku works, `docker healthcheck` green — but opus/sonnet/fable silently 401. Diagnosing it took a drift watcher plus log archaeology.

## Fix

- **`isTerminalRefreshFailure(status, body)`** — pure, exported classifier: `invalid_grant` / 401 / 403 → terminal; transient (500/502/503/429) → retryable. Unit-tested (12 assertions, incl. the exact `#737` body).
- **`doRefreshTokens` fails fast** on a terminal failure with an actionable message (`Refresh token invalid or revoked — run `dario login``) instead of burning retries. Transient errors still retry unchanged.
- **New `refreshTokenDead` flag** flips on the first terminal failure, so `getStatus()` reports `broken` / `authenticated: false` **immediately** — surfacing on `/status`, `/health`, and `dario doctor` right away rather than after the count threshold.
- **Self-heals**: `saveCredentials` clears the flag, so a successful refresh *or* a fresh `dario login` marks the token live again.

## Scope

This does **not** prevent the token from dying — that's a topology/contention issue (don't run two darios on one account's rotating refresh token). It makes the failure **loud, immediate, and actionable** instead of a silent green-healthcheck 401 — a 10-second diagnosis instead of a multi-hour one.

## Testing

- `test/oauth-terminal-refresh.mjs` (new) — 12 assertions on the terminal/transient split.
- `npm run build` clean · `npm test` **109/109 pass** · zero new runtime deps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -463,9 +463,9 @@
       }
     },
     "node_modules/@types/node": {
-      "version": "26.1.0",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-26.1.0.tgz",
-      "integrity": "sha512-O0A1G3xPGy4w7AgQdAQYUlQ+BKk2Oovw8eRpofyp5KdBZULnbe+WqaOVNrm705SHphCiG4XHsACrSmPu1f+Kgw==",
+      "version": "26.1.1",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-26.1.1.tgz",
+      "integrity": "sha512-nxAkRSVkN1Y0JC1W8ky/fTfkGsMmcrRsbx+3XoZE+rMOX71kLYTV7fLXpqud1GpbpP5TuffXFqfX7fH2GgZREw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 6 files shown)


sprayberry-reviewer (approved):
OAuth terminal invalid_grant fix: pure classifier + fail-fast + honest broken-status. 109/109 tests, CI green. Operator-authorized merge.

PR #746, opened 2026-07-13
feat(proxy): --no-claude-auth — stop OpenAI-only proxies rotating the Claude token

askalf:
## Problem

A dario started to route to an OpenAI-compatible backend — e.g. `dario proxy --model=openai:gpt-5.6-sol`, the "run GPT/Sol in Claude Code" use case — **still loads the Claude OAuth pool and runs the 15-minute background refresh timer** at startup, even though it never makes a Claude call.

Anthropic refresh tokens are single-use rotating. So that background refresh **rotates the token**, and an interactive Claude Code using the same account on the same machine is left holding the old, now-invalid refresh token → `invalid_grant` → surprise `/login`. It's the exact dario#737 dead-token mechanism, happening locally between an OpenAI-only dario and a live Claude Code.

## Fix

`--no-claude-auth` (opt-in): don't load the Claude OAuth pool at all.

- The pool stays empty, so the background refresh timer is a no-op — **dario never touches, and never rotates, the Claude refresh token.**
- Claude-bound requests return a clean unauthenticated error (empty pool), same as the existing admin-bootstrap / upstream-api-key modes.
- OpenAI-compatible backends use their own key and are completely unaffected.
- Startup no longer fatally demands `dario login` — handled by a new pure `requiresClaudeLogin()` gate that treats `--no-claude-auth` as an expected empty-pool mode (alongside admin-bootstrap and upstream-api-key).

Usage:

```
dario proxy --model=openai:gpt-5.6-sol --no-claude-auth
# routes to OpenAI, never rotates your Claude token, no more logouts
```

## Non-breaking

Opt-in. Unset (the default) = current behavior exactly — the pool loads, the timer runs, everything as before. No `@stable` surface change; a new boolean flag + `ProxyOptions.noClaudeAuth` + one exported gate helper + a banner line.

Does **not** affect a passthrough fleet dario (it serves Claude requests, needs the pool, won't set the flag).

## Testing

- `test/no-claude-auth-flag.mjs` (new) — 6 assertions on `requiresClaudeLogin` (default demands login; `--no-claude-auth` / admin / api-key don't; a populated pool never does).
- `npm run build` clean · `npm test` **110/110 pass** · zero new runtime deps.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -357,6 +357,10 @@ async function proxy() {
   // proxy presents to Anthropic is Bun's BoringSSL ClientHello, not
   // Node's OpenSSL one. v3.23 (direction #3).
   const strictTls = args.includes('--strict-tls');
+  // --no-claude-auth: don't load/refresh the Claude OAuth pool (OpenAI-only
+  // proxies). Stops dario rotating a shared refresh token out from under an
+  // interactive Claude Code on the same machine.
+  const noClaudeAuth = args.includes('--no-claude-auth');
   const modelArg = args.find(a => a.startsWith('--model='));
   const model = modelArg ? modelArg.split('=')[1] : undefined;
   // --fast-model=MODEL: route Haiku-tier (CC sub-agent) requests to this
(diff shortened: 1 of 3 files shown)


sprayberry-reviewer (approved):
> **Automated review from the Sprayberry Labs fleet code reviewer** (sprayberry-reviewer bot, not a human sign-off).

---

## What I checked

Reviewed all three changed files against the diff fetched from `askalf/dario#746`.

### Auth-wiring correctness

**Pool-skip** (`src/proxy.ts:1359–1373`):

```typescript
if (opts.noClaudeAuth) {
  console.error('[dario] --no-claude-auth: Claude OAuth pool NOT loaded...');
} else {
  for (const acc of accountsList) {
    pool.add(acc.alias, { ... });
  }
}
```

Correct. When the flag is set, `pool` stays empty. The existing background-refresh `setInterval` that iterates `pool.all()` then silently iterates over an empty list — the token is never read, never refreshed, never rotated. The explicit early-return guard added two lines later in the interval callback is redundant but harmless defense-in-depth.

**Startup gate refactor** (`src/proxy.ts:1406`):

```typescript
-  if (pool.size === 0 && !adminEnabled && !upstreamApiKey) {
+  if (requiresClaudeLogin(pool.size, adminEnabled, !!upstreamApiKey, opts.noClaudeAuth ?? false)) {
```

`!!upstreamApiKey` versus the old `!upstreamApiKey`: the original negated the key directly to mean "absent"; the new one converts to bool first. For all meaningful values (`undefined`, `null`, `""`, a real key string) the semantics are identical — the refactor preserves existing behaviour. ✓

**`requiresClaudeLogin` logic** (`src/proxy.ts:727–734`):

```typescript
return poolSize === 0 && !adminEnabled && !hasUpstreamApiKey && !noClaudeAuth;
```

All four guards are exactly right: login is required only when every escape hatch is absent. Unit-tested against all six relevant combinations.

**CLI threading** (`src/cli.ts:360`, `cli.ts:611`):

```typescript
const noClaudeAuth = args.includes('--no-claude-auth');
// ...
await startProxy({ ..., noClaudeAuth, ... });
```

Same pattern as every other boolean flag in the file; typed as `boolean?` in `ProxyOptions`. ✓

### Test coverage

`test/no-claude-auth-flag.mjs` — 6 assertions covering:
- default empty pool → must demand login
- `noClaudeAuth=true` → no login demanded
- admin-bootstrap → no login demanded (regression)
- upstream-api-key → no login demanded (regression)
- populated pool, both with and without the flag → no login demanded

All assertions match the implementation. Test imports from `../dist/proxy.js` which requires a prior build, consistent with the CI setup (`npm run build` before `npm test`).

### Non-issues noted

- The `setInterval` early-return (`if (opts.noClaudeAuth) return`) is unreachable in practice (pool is already empty), but it makes the intent explicit and does not introduce any bug.
- Help text placement is consistent with surrounding flag ordering in `help()`.

---

## What's good

The fix is minimal, surgical, and entirely opt-in. The `requiresClaudeLogin` extraction improves testability of the startup gate beyond the original inline conditional. The banner line at startup clearly signals the mode for operators who might otherwise wonder why Claude requests are failing.

Issue #744, opened 2026-07-13
sonnet doesnt work


miklisanton:
**What happened?**
Set model to sonnet
**What did you expect?**
sonnet model to work
**Steps to reproduce:**
1.  set /model sonnet in claude code
2. send a message
3. get "● API Error: 400 messages.1: system content must contain at least one block"

**Environment:**
- OS: ubuntu 24
- Node version: 18
- dario version: 5.0.0
- Tool using dario (e.g. OpenClaw, Aider):  claude code

askalf:
Thanks for the clear repro steps — confirmed this is worth chasing.

I traced the sonnet path and ran the outbound-request builder against several realistic sonnet request shapes, but couldn't reproduce the empty block from a clean install — so I don't want to guess at a fix blind. The error (`messages.1: ... must contain at least one block`) means an empty content block is reaching Anthropic, but I need to see the actual bytes to pin *which* block is empty and why it happens on your setup.

Could you grab the verbose request body? Run the proxy with body logging:

```
dario proxy -vv
```

(or `DARIO_LOG_BODIES=1 dario proxy -v`), reproduce the sonnet error once, and paste the redacted outbound `/v1/messages` body it logs. That shows the exact `system` / `messages` shape dario sends upstream — which is what I need to find the empty block. Feel free to redact prompt content; I mainly need the block **structure** (types, and which one comes out empty).

Two quick questions that'll help narrow it:

1. Does it happen **only** on sonnet, or also on opus / haiku?
2. What does `dario doctor` show, especially the **Template** line? dario captures the Claude Code template live per-install, so a bad local capture could produce an empty block that a clean template doesn't — that's one of the leading theories.

With the verbose body I can reproduce it exactly and ship a fix with a regression test. Thanks for reporting it.

miklisanton:
1. Only on sonnet models (/model sonnet)
2. 
``` dario doctor 
 [INFO]  Platform        linux x64 (6.17.0-35-generic)
  [ OK ]  Runtime / TLS   Bun v1.3.14 — TLS fingerprint matches Claude Code
  [WARN]  CC binary       v2.1.209 at /home/user/.local/bin/claude  (range: v1.0.0 – v2.1.207)
  [ OK ]  Template        live capture, CC v2.1.209 (8m old) (schema v3)
[...]

Issue #54, opened 2026-04-19
Can I use it with Claude CLI as well?


tetsuco:
I’m not sure if such a feature currently exists or if there’s any demand for it besides me, but
is it possible to use Dario in Claude CLI?

Or are there any options currently available that would make this possible?

## Things I've tried

1. Start the "dario proxy" with the following options

`dario proxy --hybrid-tools --host=0.0.0.0`

2. Run the Claude CLI from another PC on the local network

```
export ANTHROPIC_BASE_URL=http://192.168.0.10:3456
export ANTHROPIC_AUTH_TOKEN=dario
claude
```

and

```
export ANTHROPIC_BASE_URL=http://192.168.0.10:3456
export ANTHROPIC_AUTH_TOKEN=dario
claude -p hello
```

I tried both, but they didn't work due to the following error.
`API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages: text content blocks must be non-empty"},"request_id":"req_xxxxx"}`

Note that the same error occurred when I set `export ANTHROPIC_API_KEY=dario`

### I'm able to use dario itself

When I try the following curl command from the same PC, there are no issues.

```
curl http://192.168.0.10:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-opus-4-7","max_tokens":1024,"messages":[{"role":"user","content":"Hello!"}]}'

---

{"model":"claude-opus-4-7","id":"msg_xxxxx","type":"message","role":"assistant","content":[{"type":"text","text":"Hello! How can I help you today?"}],"stop_reason":"end_turn","stop_sequence":null,"stop_details":null,"usage":{"input_tokens":13,"cache_creation_input_tokens":3734,"cache_read_input_tokens":0,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":3734},"output_tokens":15,"service_tier":"standard","inference_geo":"not_available"},"context_management":{"applied_edits":[]}}
```

## Info about “doctor”

```
% dario doctor

  dario — Doctor
  ─────────────

  [INFO]  dario           v3.28.0
  [ OK ]  Node            v25.9.0
  [INFO]  Platform        darwin arm64 (25.4.0)
[...]

PR #763, opened 2026-07-15
fix: drop messages emptied by the orchestration scrub (#744)

askalf:
**Root cause of #744**, reproduced against the published 5.0.0 artifact: a mid-session `/model sonnet` makes CC v2.1.209 (`mid-conversation-system` beta) inject a standalone `role:"system"` turn whose entire content is one `<system-reminder>` block. The orchestration scrub empties the block, the #54 filter drops it, and the message reaches the upstream as `content: []` → `400 messages.1: system content must contain at least one block` — the reporter's exact error, and why it only shows on a model *switch*.

```
before: messages[1] role=system content=[]   → upstream 400
after:  message dropped (it was pure orchestration; nothing for the model)
```

Scope: `sanitizeMessages` only — drops a message whose content emptied out entirely (array or string form). Messages with surviving non-text blocks (tool_use/tool_result/image) are untouched; `--preserve-orchestration-tags` keeps such messages intact. One pre-existing test pinned the old `content:[]` shape and relied on downstream layers to cope — updated to the new contract. +5 checks including the exact trigger shape; suite 111/0.

Addresses #744 � leaving the issue open for the reporter to confirm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ checklist.
 
 ### Fixed
 
+- **`/model` switch mid-session no longer 400s (dario#744).** A mid-session
+  model switch makes CC (v2.1.209, `mid-conversation-system`) inject a
+  standalone `role:"system"` turn whose entire content is a
+  `<system-reminder>` block. The orchestration scrub emptied that block, the
+  dario#54 filter dropped it, and the message went upstream as `content: []` —
+  rejected with `messages.N: system content must contain at least one block`.
+  `sanitizeMessages` now drops a message whose content emptied out entirely
+  (array or string form): it carried nothing but orchestration tags, so
+  removing it is the block filter's own decision applied one level up.
+  Reproduced against the published 5.0.0 artifact and pinned by tests.
+
 - **Client cache TTL is now mirrored, not overwritten (dario#678).** Real CC on
   included subscription usage sends `ttl:'1h'` on every cache breakpoint plus
   `extended-cache-ttl-2025-04-11` in `anthropic-beta`, and itself falls back to
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -610,6 +610,20 @@ export function sanitizeMessages(body: Record<string, unknown>, preserveTags?: S
       });
     }
   }
+  // Drop messages whose content emptied out entirely. A message that was
+  // NOTHING but orchestration tags — CC's standalone `<system-reminder>`
+  // injections, e.g. the model-switch notice a mid-session `/model sonnet`
+  // adds as its own role:"system" turn — leaves the block filter above as
+  // `content: []`, and the upstream rejects the whole request with
+  // "messages.N: … content must contain at least one block" (dario#744).
+  // The message carried nothing for the model, so removing it is the same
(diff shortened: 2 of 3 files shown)


sprayberry-reviewer (approved):
*Automated review from the Sprayberry Labs fleet code reviewer (sprayberry-reviewer).*

---

**Verdict: no blocking issues found — clean fix, well-tested. Approving.**

All five checklist items verified:

---

### What I checked

**1. Core drop logic (`src/proxy.ts:622–628`) — correct**

```typescript
body.messages = messages.filter((m) => {
  if (Array.isArray(m.content)) return m.content.length > 0;
  if (typeof m.content === 'string') return m.content !== '';
  return true;
});
```

Traced through three paths:

- **Array emptied by scrub** (the `#744` trigger): the inner block-filter already drops empty `text` blocks while keeping `tool_use`/`tool_result`/`image`. After the loop, a message that was *nothing but* orchestration tags has `content.length === 0` → the new filter drops it. ✓
- **String content scrubbed to `''`**: `typeof m.content === 'string'` branch catches it. ✓
- **Non-text blocks present**: the inner filter preserves them, so `content.length > 0` → message survives. `return true` handles null/undefined content (edge case, upstream error territory). ✓

The `messages` local variable holds a reference to the original array; the inner loop mutates `.content` in-place on each object, so the outer `.filter()` sees post-scrub content. No stale-reference issue.

**2. Preserve-tags path (`preserveTags` not undefined) — correct**

When `preserveTags` contains `'system-reminder'`, `buildOrchestrationPatterns` excludes it from the compiled patterns, so `sanitizeContent` leaves the tag text intact. The content block is non-empty → `content.length > 0` → message kept. The new test at line 254–258 pins this exactly. ✓

**3. Pre-existing test updated with rationale — correct**

Old assertion: `body.messages[0].content.length === 0` (trusting a downstream layer to cope with `content:[]`).  
New assertion: `body.messages.length === 0` (the drop now happens here). The inline comment explains *why* the old shape was insufficient — a non-trailing empty message reached upstream and 400'd — which is the right level of documentation for a contract change. ✓

**4. Five new checks, trigger shape covered — correct**

1. All-orchestration `role:system` system-reminder turn is dropped (exact `#744` shape).
2. Mixed message with one real block keeps that block.
3. String content scrubbed to `''` drops the message.
4. `tool_use`/`tool_result` messages untouched.
5. Preserved tag → message survives intact.

CI: `test` job **pass**, `build` (Node 18/20/22) **pass**, CodeQL **pass**. ✓

**5. CHANGELOG Unreleased entry — accurate**

Describes the CC 2.1.209 `mid-conversation-system` beta, the standalone `role:"system"` injection on `/model` switch, the scrub → `content:[]` → upstream 400 chain, and the fix. Matches the implementation and is placed correctly in `Unreleased`. ✓

---

### What's good

The fix is tightly scoped — one post-loop filter of ~4 lines, zero changes to existing scrub logic. The bug was subtle (non-trailing vs. trailing empty turns; `buildCCRequest` only popped trailing), and the PR description reproduces it precisely against the published artifact. The test suite is thorough and the old test is updated with a clear rationale rather than silently changed.

PR #770, opened 2026-07-16
chore(cc-drift): sanitize public drift-issue bodies

askalf:
## What

Sanitize the public GitHub-issue bodies filed by `cc-drift-watch.yml`.

On every CC version bump the watcher filed a public, search-indexed issue (e.g. #768) whose body dumped the **full drift/probe report JSON** plus a `template.version` message describing the capture mechanism (loopback interception of a real CC request) and enumerating fingerprint-sensitive request fields. That reverse-engineering-flavored wording does not belong in a public issue.

## Changes

- **`.github/workflows/cc-drift-watch.yml`** — both issue bodies (binary-scan drift + authorize-probe drift) now surface only a **neutral category/severity summary**. The full report remains in the workflow run log and the `cc-drift-report` artifact. Empty-items and malformed-JSON cases fall back gracefully.
- **`scripts/check-cc-drift.mjs`** — reword the `template.version` drift message to drop the capture-mechanism and fingerprint-field wording.

## Not a release

`scripts/` and `.github/` are **not** in dario's npm tarball, so this ships nothing to the package — it's repo-only. No version bump, no npm publish.

## Testing

- Ran the new summary generator against the exact #768 report shape → emits `- **compat.range** (medium)` / `- **template.version** (low)` / `- **cch.seed** (info)` and nothing else; grep for `MITM|fingerprint|clientId|header order|beta flag|system prompt` against the produced body is **clean**.
- Verified empty-items → `- (no itemized categories — see run log)` and malformed-JSON → `- (report parse failed — see run log)` fallbacks.
- No credential exposure was ever involved: the `clientId` in these reports is CC's public OAuth client id (hardcoded in `src/cc-oauth-detect.ts`), not a user secret. This PR is about optics, not a leak.


sprayberry-reviewer (approved):
> **Automated review from the Sprayberry Labs fleet code reviewer** (sprayberry-reviewer bot — not a human sign-off).

---

## What I checked

### Sanitization completeness

**Binary-scan drift issue** (`.github/workflows/cc-drift-watch.yml` lines ~173–195):  
The old body dumped `cat drift-report.json` verbatim — raw JSON including the `message` strings that described the capture mechanism and enumerated fingerprint-sensitive fields. The new body replaces that with a Node.js inline that extracts only `it.category` and `it.severity` from `r.items[]`. The `message` field — the one carrying the sensitive wording — is never accessed or emitted. The category values themselves (`compat.range`, `template.version`, `cch.seed`, etc.) are all fixed string constants in the codebase, none sensitive.

**Authorize-probe drift issue** (lines ~368–382):  
The old body dumped `cat probe-report.json`. The new body is fully static text plus a run-link. No report data reaches the public issue at all. Correct approach given the probe report doesn't carry a structured `items[]` the same way.

**`scripts/check-cc-drift.mjs` line 295:**  
The `template.version` message is trimmed from `"Re-capture the template (MITM a real CC v${ccVersion} request) if any fingerprint-sensitive field (system prompt, header order, metadata shape, beta flags) changed."` to `"Re-capture and re-bake the template if any request-shape field changed."` — the reverse-engineering wording is gone. The remaining text is appropriately neutral.

### Edge-case correctness in the inline Node script

Traced all three cases:
- **Valid non-empty items:** loops and prints `- **category** (severity)` per item ✓
- **`items` is empty array:** `!items.length` guard emits the fallback line ✓
- **`items` field missing / not an array:** `Array.isArray(r.items) ? r.items : []` produces `[]`, hits the same empty-items guard ✓
- **Invalid JSON / missing file / any other node error:** `2>/dev/null || echo "- (report parse failed — see run log)"` catches it cleanly ✓

### Shell quoting / injection

`echo "$SUMMARY"` is double-quoted — no word splitting or glob expansion risk. The body is written to `issue-body.md` and passed via `--body-file`, not interpolated inline into the `gh` invocation — so even if `SUMMARY` contained unusual characters, they go into the file, not into argument parsing. ✓

### actionlint

Passed (`actionlint` CI check: SUCCESS). No workflow YAML syntax or expression issues. ✓

### Scope

Only `.github/` and `scripts/` are changed — no `src/`, no `package.json`, no `CHANGELOG.md`. Consistent with the "not a release" intent stated in the PR body. ✓

---

## What's good

The PR does exactly what it says on the tin. The `--body-file` pattern rather than interpolating the body inline is the right approach here. Both issue paths (binary-scan and authorize-probe) are updated consistently. The fallback strings are informative without being alarmist. The PR description's explicit test results (grep clean against `MITM|fingerprint|clientId|...`) give good confidence the intent was verified before filing.

PR #775, opened 2026-07-16
feat(cache): DARIO_CACHE_TTL_1H opt-in force-1h TTL (#678)

askalf:
## What

Adds **`DARIO_CACHE_TTL_1H=1`** — an opt-in that forces the 1-hour prompt-cache TTL for a subscription client that can't emit the 1-hour stamp itself (an SDK/agent harness that only sends the bare 5-minute `cache_control`). It stamps `ttl:"1h"` on every breakpoint **and** adds the enabling `extended-cache-ttl-2025-04-11` beta so Anthropic honors it.

Follow-up to the #678 investigation, where the token-level measurements showed the reporter's harness sends bare 5-minute stamps, so its prefix expires on idle gaps > 5 min. This gives them (and anyone in that setup) a switch when the harness can't send the 1h stamp.

## Measured end-to-end (real subscription)

Two identical ~9K-token prefixes, one 6.5-minute gap:
```
                cache_create  cache_read
5m  cold            9038          0
1h  cold            9038          0
  ── wait 6.5 min ──
5m  warm            9055          0    ← expired → rebuild
1h  warm              17       9038    ← survived → cache hit
```
This flag makes dario produce that **1h wire shape** (ttl + beta) for a client that only sends 5m.

## Behavior

- Deliberate override of the default mirror (`effectiveCacheControl`), which never *invents* a 1h. The tradeoff is real: 1-hour cache **writes** bill ~2× the 5-minute rate, so it only wins when idle gaps routinely exceed 5 minutes; on rapid back-to-back turns it costs more.
- `DARIO_CACHE_TTL_5M=1` forces the opposite and **wins if both are set**.
- Clients that already send their own stamps are **unchanged** (mirror still applies).
- If the upstream 400s the beta on a non-subscription account, the existing per-account rejected-beta cache drops it on the retry.

## Changes

- `effectiveCacheControl` → returns `{type:"ephemeral",ttl:"1h"}` when `DARIO_CACHE_TTL_1H=1`.
- `withForced1hBeta(beta, env?)` — new pure, injectable helper that adds the enabling beta; `proxy.ts` calls it in the non-passthrough beta assembly.
- `test/cache-ttl.mjs` — **+11 assertions** (forced stamp, every breakpoint 1h, 5M precedence, flag-off restore, helper idempotency/no-op/empty). **Suite 112/112.**
- `docs/faq.md` — new entry: "usage higher than direct — why?" documenting both TTL flags.

Ships in `src/` → version bump **5.2.4 → 5.2.5**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,23 @@ checklist.
 
 ## [Unreleased]
 
+## [5.2.5] - 2026-07-16
+
+### Added
+
+- **`DARIO_CACHE_TTL_1H=1` — opt-in 1-hour prompt-cache TTL (#678).** For a
+  subscription client that can't emit the 1-hour cache stamp itself (an
+  SDK/agent harness that only sends the bare 5-minute `cache_control`), this
+  forces `ttl:"1h"` on every breakpoint and adds the enabling
+  `extended-cache-ttl-2025-04-11` beta so Anthropic honors it — measured
+  end-to-end: a prefix that rebuilds after a 6-minute gap on 5m
+  (`cache_read=0`) reads warm on 1h (`cache_read=9038`). Deliberate override of
+  the default mirror behavior: 1-hour cache *writes* bill ~2× the 5-minute
+  rate, so it only wins when idle gaps routinely exceed 5 minutes; on rapid
+  back-to-back turns it costs more. `DARIO_CACHE_TTL_5M=1` forces the opposite
+  and wins if both are set. Logging/mirroring behavior for clients that already
+  send their own stamps is unchanged. Documented in `docs/faq.md`.
+
 ## [5.2.4] - 2026-07-16
 
 - **Template label refresh** — bundled `_version`, `_supportedMaxTested`, and the `user-agent` header bumped `2.1.210` → `2.1.211` to track `@anthropic-ai/claude-code@latest`. `cc-drift-template-watch` ran `capture-and-bake --check` against live CC v2.1.211 and found **zero wire-shape drift** vs the bundle, so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Clears the `sdk-drift` early-warning signal (#772).
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -42,6 +42,15 @@ Yes. Skip `dario login`, just run `dario backend add openai --key=...` (or any O
 **Can I route non-OpenAI providers through dario?**
 Yes — anything that speaks the OpenAI Chat Completions API. Groq, OpenRouter, LiteLLM, vLLM, Ollama's openai-compat mode, your own vLLM server, any hosted inference endpoint that exposes `/v1/chat/completions`. Just `dario backend add <name> --key=... --base-url=...`.
 
+**My subscription usage through dario is higher than running Claude Code directly. Why?**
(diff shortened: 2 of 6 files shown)


sprayberry-reviewer (approved):
> **Automated review from the Sprayberry Labs fleet code reviewer.**

**Verdict: No blocking issues found — approving.**

---

## What I checked

**Scope (6 files, +102 / −3):** `src/cc-template.ts` (new `effectiveCacheControl` branch + `withForced1hBeta` helper + `EXTENDED_CACHE_TTL_BETA` export), `src/proxy.ts` (import + wiring), `test/cache-ttl.mjs` (+11 assertions), `CHANGELOG.md`, `docs/faq.md`, `package.json` (5.2.4 → 5.2.5). Exactly matches PR description.

**CI:** All checks pass — `compat` (3m57s, live CC e2e), `test` (34s, suite 112/112 per description), `wire-drift`, `build` (Node 18/20/22), `docker-cap-drop-smoke`, `CodeQL`, `validate-package-json`. No skipped or failing jobs.

---

### 1. Flag stamps `ttl:'1h'` correctly — ✅

`src/cc-template.ts`, new line in `effectiveCacheControl`:

```ts
if (process.env['DARIO_CACHE_TTL_5M'] === '1') return CC_CACHE_CONTROL;
if (process.env['DARIO_CACHE_TTL_1H'] === '1') return { type: 'ephemeral', ttl: '1h' };
```

`5M` is checked first (wins when both are set), `1H` returns `{type:'ephemeral', ttl:'1h'}` before the mirror scan. Correct ordering, confirmed by test: "DARIO_CACHE_TTL_5M wins when both flags set".

### 2. `withForced1hBeta` adds the enabling beta — ✅

```ts
export function withForced1hBeta(beta: string, env: Record<string, string | undefined> = process.env): string {
  if (env['DARIO_CACHE_TTL_1H'] !== '1' || env['DARIO_CACHE_TTL_5M'] === '1') return beta;
  if (beta.split(',').includes(EXTENDED_CACHE_TTL_BETA)) return beta;
  return beta.length > 0 ? beta + ',' + EXTENDED_CACHE_TTL_BETA : EXTENDED_CACHE_TTL_BETA;
}
```

Idempotent (exact-element check on the split), handles empty string, 5M-wins parity with `effectiveCacheControl`, pure/injectable for tests. All edge cases covered by the five `withForced1hBeta` assertions.

### 3. Proxy wiring is correctly scoped — ✅

`src/proxy.ts:2696`: `withForced1hBeta` is called **only inside the `else` branch** (non-passthrough, non-count-tokens), matching where `effectiveCacheControl` is also called. The passthrough path (`beta = 'oauth-2025-04-20' + clientBeta`) is untouched. Placement is **before** the rejected-beta strip, so if `extended-cache-ttl-2025-04-11` is rejected on a non-subscription account it is dropped on retry — the safety net the comment describes is real.

### 4. Mirror/default path is unchanged — ✅

When `DARIO_CACHE_TTL_1H` is unset, `effectiveCacheControl` falls through to the existing mirror scan unchanged. Clients that already emit their own stamps are unaffected. Confirmed by "flag off → bare client stays bare" assertion.

### 5. Test coverage — ✅

11 new assertions cover: forced stamp on a bare client, every outbound breakpoint is 1h, 5M-wins precedence, flag-off restoration, `withForced1hBeta` adds-beta, idempotency, no-op when off, no-op when 5M wins, empty-string handling. `process.env` contamination cleaned up correctly (`delete` both flags before the flag-off check; injectable-env tests don't touch `process.env`).

---

## What's good

Clean two-function decomposition: `effectiveCacheControl` owns the stamp shape, `withForced1hBeta` owns the beta side-effect — they share the same 5M-wins precedence rule independently, making them independently testable. The rejection-cache fallback for non-subscription accounts is a thoughtful safety net that avoids the need to pre-detect account tier. FAQ entry is accurate and actionable.

PR #778, opened 2026-07-16
docs(faq): note DARIO_CACHE_TTL_1H must reach the container env (#678)

askalf:
One-line follow-up to the `DARIO_CACHE_TTL_1H` flag (5.2.5). The flag is read from dario's **process** env; a value placed only in a host `.env` that a docker-compose service doesn't pass through never reaches the process, so the flag is **silently a no-op**.

Surfaced while verifying the flag end-to-end on the fleet (whose compose uses an explicit `environment:` list) — the first test showed `cache_read=0` purely because the var never reached the container. The FAQ now spells out delivery: prefix it for a direct `dario proxy` run, put it in `environment:`/`env_file:` for Docker, and confirm with `docker exec … printenv`.

**Docs-only** — `docs/` isn't in the npm tarball (`files` = dist/README/LICENSE), so no version bump / no release.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -51,6 +51,8 @@ Fixes, cheapest first:
 - Have your harness emit the 1-hour stamp + the `extended-cache-ttl-2025-04-11` beta — dario mirrors it and the prefix survives idle gaps.
 - If the harness can't, set **`DARIO_CACHE_TTL_1H=1`**: dario forces the 1-hour TTL (and adds the enabling beta) on every request. Tradeoff — 1-hour cache *writes* bill ~2× the 5-minute rate, so it only wins when your idle gaps routinely exceed 5 minutes; on rapid back-to-back turns it costs more. `DARIO_CACHE_TTL_5M=1` forces the opposite (always 5m). Background: [#678](https://github.com/askalf/dario/issues/678).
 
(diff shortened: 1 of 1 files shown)


sprayberry-reviewer (approved):
*Automated review from the Sprayberry Labs fleet code reviewer.*

**Verdict: clean — no blocking issues found. Approving.**

---

### What I checked

**Diff surface:** `docs/faq.md` +2 / -0 — one prose paragraph appended inside the existing `DARIO_CACHE_TTL_1H` FAQ bullet. No code, no config, no release artefacts touched.

**Technical accuracy of the added text:**

- `DARIO_CACHE_TTL_1H=1 dario proxy …` — correct POSIX env-prefix syntax for a direct invocation.
- "a value only in a host `.env` that the compose file doesn't pass through never reaches the process" — accurate. Docker Compose `.env` populates Compose-file variable substitution, not the container environment, unless explicitly forwarded via `environment:` or `env_file:`. The distinction the paragraph draws is correct and addresses a genuinely common silent-failure mode.
- `docker exec <container> printenv DARIO_CACHE_TTL_1H` — correct verification incantation.

**CI:** All checks pass — CodeQL, build × 3 (Node 18 / 20 / 22), docker-cap-drop-smoke, test, actionlint, analyze, validate-package-json.

**Scope:** `docs/` is excluded from the npm tarball (`files` = dist/README/LICENSE per PR body), so no version bump or release is warranted — consistent with the "docs-only" claim.

**No findings.** The paragraph is accurate, well-placed, and fills a real gap surfaced during end-to-end verification of the flag. Nothing to change.

PR #788, opened 2026-07-17
refactor(routing): consolidate provider decision into a ProviderAdapter seam

askalf:
## What

Consolidates dario's provider **routing decision** — "which backend owns this request, and what's the exhaustion fallback" — into one place. It used to be spread across ~5 interleaved conditions in `proxy.ts`: the path class (`isOpenAI`), `parseProviderPrefix`, `isOpenAIModel`, the openai-backend reroute guard, and the pool-fallback guard.

New `src/provider-adapter.ts`: a small `ProviderAdapter` registry + `route(ctx)` that returns `{ provider, fallback, reason }`. The openai-backend reroute in `proxy.ts` now calls `route(...)` instead of re-deriving the condition inline.

## Behavior-preserving

`route(...).provider === 'openai'` is exactly the prior inline condition:
`hasOpenAIBackend && isOpenAIPath && forcedProvider !== 'claude' && (forcedProvider === 'openai' || isOpenAIModel(model))`. The `body.length > 0` guard and `openaiBackend` narrowing are unchanged.

- `test/provider-adapter.mjs` — a 15-row truth table pinning the full decision matrix (paths, models, forced-provider, backend on/off, fallback armed / empty-pool / no-backend), plus an order-independence check.
- Full suite green (116 files), which is the real guarantee since it exercises openai routing and pool fallback end to end.

## Deliberately scoped: the decision, not the lifecycle

This adapts the routing **decision** and stops there. The Claude path's pool / template / cch / session-rotation / overage machinery stays shared below the seam. That's intentional: a measurement of the request handler shows Claude touches **17 distinct `pool.*` methods** plus ~80 template/overage/effort/tool references, versus the OpenAI path's **4 calls** to one 211-line function. A "fat" adapter owning the whole lifecycle would make the Claude adapter ≈ the entire proxy and the OpenAI adapter nearly empty — an asymmetry that doesn't pay for itself. The seam that pays is routing + request-shaping; the rest is shared infrastructure that happens to serve one provider.

Follow-up candidates (not in this PR, to keep it behavior-preserving and reviewable): route the two pool-fallback viability checks through the same `route(...).fallback` where the full context is available.

### DIFF
--- a/src/provider-adapter.ts
+++ b/src/provider-adapter.ts
@@ -1,0 +1,119 @@
+/**
+ * Provider routing seam.
+ *
+ * dario's "which backend owns this request" decision used to live inline in the
+ * request handler as a set of interleaved conditions (path class, provider
+ * prefix, GPT-family model test, the openai-backend reroute guard, the pool
+ * fallback guard). This module consolidates that decision into one place: a
+ * small set of `ProviderAdapter`s and a `route()` function that returns the
+ * primary provider plus any exhaustion fallback.
+ *
+ * Scope is deliberately the DECISION, not the request lifecycle. The Claude
+ * path's pool/template/cch/session/overage machinery stays shared below this
+ * seam — it's infrastructure that happens to serve one provider, not per-
+ * provider behaviour, so pushing it behind an adapter interface would make the
+ * Claude adapter the whole proxy and the OpenAI adapter nearly empty. The seam
+ * that pays for itself is routing + request-shaping; the rest is shared.
+ *
+ * The adapters reuse the same primitive proxy.ts uses (`isOpenAIModel`), so this
+ * is a consolidation of the existing decision, not a re-derivation of it.
+ */
+
+import { isOpenAIModel } from './openai-backend.js';
+
+export type ProviderId = 'claude' | 'openai';
+
+/** Inputs the routing decision needs, computed once per request. */
+export interface RouteContext {
+  /** urlPath === '/v1/chat/completions' (OpenAI chat shape). */
+  isOpenAIPath: boolean;
+  /** Model name after provider-prefix stripping (e.g. 'gpt-4o', 'claude-opus-4-8'). */
+  model: string;
+  /** Forced provider from a `<provider>:` prefix or `--model` override; null if unforced. */
+  forcedProvider: ProviderId | null;
+  /** An openai-compat backend is configured (`dario backend add …`). */
+  hasOpenAIBackend: boolean;
+  /** `--pool-fallback=<model>` value, or null when disabled. */
+  poolFallbackModel: string | null;
+  /** Live pool account count. */
+  poolSize: number;
+}
+
+export interface RouteDecision {
+  /** Primary handler for the request. */
+  provider: ProviderId;
+  /** Provider to fall to on primary exhaustion; only claude→openai exists today. */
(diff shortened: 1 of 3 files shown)


sprayberry-reviewer (approved):
> **Automated review from the Sprayberry Labs fleet code reviewer.**

**Verdict: No blocking issues — approving.** The routing refactor is behaviorally faithful and the test coverage is solid.

---

## What I checked

- `src/provider-adapter.ts` (new, 119 lines): `ProviderAdapter` registry + `route()`.
- `src/proxy.ts` (+18 / −2): outer guard change + call to `routeProvider`.
- `test/provider-adapter.mjs` (new, 81 lines): 15-case truth table + order-independence check.
- Surrounding context in `proxy.ts` (~50 lines above/below the changed block) to verify behavioral equivalence.

---

## Behavioral equivalence — verified

The PR description claims `route(...).provider === 'openai'` is exactly the prior inline condition. Checking `openaiAdapter.claimsPrimary()`:

```ts
// src/provider-adapter.ts:68–72
if (!ctx.hasOpenAIBackend) return false;
if (!ctx.isOpenAIPath)    return false;
if (ctx.forcedProvider === 'claude') return false;
return ctx.forcedProvider === 'openai' || isOpenAIModel(ctx.model);
```

The call site in the new `proxy.ts` routing block:

```ts
// src/proxy.ts (new)
if (rawModel && openaiBackend && decision.provider === 'openai') {
```

Combined: `openaiBackend !== null (hasOpenAIBackend) && isOpenAI (isOpenAIPath) && forcedProvider !== 'claude' && (forcedProvider === 'openai' || isOpenAIModel(rawModel))` — the old condition exactly. The `rawModel &&` guard, `body.length > 0` outer guard, and `openaiBackend` TS-narrowing check are all preserved.

The `decision.fallback` result produced by `route()` is intentionally not consumed in this PR (the pool-fallback dispatch block at line 2487 still governs that path) — as the PR description explicitly notes. Not a bug.

---

## Minor note (non-blocking)

**`proxy.ts` routing block: JSON.parse now runs on every request with a body.**

Before this PR the routing block was guarded by `openaiBackend && isOpenAI && …`, so `JSON.parse` there only fired on plausible openai-path requests. After, the outer guard is just `if (body.length > 0)`:

```ts
// src/proxy.ts (new outer guard)
if (body.length > 0) {
  try {
    const peek = JSON.parse(body.toString()) as { model?: string };
```

`parsedBody` from the block immediately above (line ~2396) already holds the parsed result and is reused later at line 2546. The routing block could read `parsedBody?.model` instead of re-parsing — a small saving on every Claude-only or no-backend request. Not a correctness issue, and at proxy throughput it's measurable rather than catastrophic. Worth a follow-up cleanup when the lifecycle is touched.

---

## What's good

- The refactor is tight: the new module owns exactly one decision and nothing else, the interface is clean (`RouteContext` / `RouteDecision` / `ProviderAdapter`), and the comment in `provider-adapter.ts` explains the scope limit clearly.
- The 15-row truth table in `test/provider-adapter.mjs` covers the full decision matrix (path × model × forced-provider × backend × fallback) and includes an order-independence check. That's the right level of coverage for a routing seam.
- The `?? claudeAdapter` defensive fallback in `route()` is harmless (claudeAdapter always claims), but makes the function total without silent holes.

PR #792, opened 2026-07-17
fix(pool): idle-based sticky TTL so long sessions aren't rebound mid-conversation

askalf:
## Problem

Session-stickiness bindings had their TTL keyed on `boundAt` (creation time), and `selectSticky` never refreshed it on a hit. So a conversation running continuously past `STICKY_TTL_MS` (6h) got reaped by `cleanupSticky` and re-picked via `select()` on its next turn — usually landing on a *different* account and discarding the warm Anthropic prompt cache the binding exists to preserve. Long agent sessions are precisely the ones that cross the 6h mark, so they took the worst of it — the opposite of what stickiness is for.

The design comment even stated the intent ("past that point a same conversation would be starting a fresh window anyway, so rebinding is free"), but as written the TTL measured age, not idleness, so it fired on active sessions too.

## Fix

Make the TTL idle-based:

- `StickyBinding` gains `lastUsedAt` (`boundAt` stays, for observability).
- `selectSticky` refreshes `lastUsedAt` on every hit, so an active session re-bases the timer each turn and is never reaped/rebound.
- `cleanupSticky` reaps on `now - lastUsedAt > STICKY_IDLE_TTL_MS`. A conversation that goes quiet is still reaped 6h after its *final* turn.
- The size-cap eviction now sorts by `lastUsedAt` — true LRU rather than least-recently-*created*.

No capacity is freed by this (a binding never reserved any — headroom is computed from live rate-limit snapshots), and there's no downside for the paused-then-resumed case: a session idle >6h is reaped under both the old and new rules. The only behavioral change is that a *continuously active* >6h session is no longer rebound out from under its cache.

`selectSticky`/`cleanupSticky` take an optional injected `now` (same pattern as `isInAuthCooldown(account, now)`) so the 6h behavior is deterministically testable without sleeping. The single proxy call site uses the 2-arg form; the new param defaults to `Date.now()`.

## Tests

New `pool-sticky` cases (full suite 117/117):
- active binding survives past the 6h **age** mark (turns at t0+5h/+9h/+14h all stay on the same account, one binding throughout);
- binding idle past 6h is reaped and re-picks the current best account;
- size-cap eviction keeps a recently-used-but-old binding and drops the idle ones (LRU, not LRC).

## Open question (tuning, not blocking)

I kept the idle horizon at the existing 6h to minimize behavioral change. But Anthropic's prompt cache lives at most 1h, so a conversation resuming after >1h idle already has a cold cache — sticking it to its old account past that point gives no cache benefit and only forgoes re-picking the least-drained seat. A shorter idle TTL (~1h, matching max cache lifetime) would arguably load-spread better on resume with zero cache cost. Happy to change the constant if you'd prefer that; left at 6h here since it's the conservative, single-concept change (age → idle).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+- **Session stickiness TTL is now idle-based, not age-based.** A conversation's account binding is reaped 6h after its *last* turn rather than 6h after its first, and every sticky hit refreshes the timer. Previously a session running continuously past the 6h mark was force-rebound to a possibly-different account mid-conversation, throwing away its warm Anthropic prompt cache — the exact thrash stickiness exists to prevent, and it landed hardest on the long agent sessions that benefit from stickiness most. Genuinely idle conversations are still reaped on the same 6h horizon (now measured from their final turn), and the sticky size-cap evicts least-recently-*used* instead of least-recently-*created*.
+
 ## [5.2.7] - 2026-07-17
 
 - **OAuth token persistence** — refreshed tokens are now durably written back to disk (`credentials.json` + pool account files, atomic write) instead of living only in memory; current tokens are also flushed on SIGTERM, and startup now refreshes an expired-but-refreshable token instead of marking the account expired. Fixes the class of outage where any container recreate after >8h of uptime loaded a rotated-away refresh token. (#790, #791)
--- a/src/pool.ts
+++ b/src/pool.ts
@@ -264,15 +264,21 @@ export function computeHeadroom(snapshot: RateLimitSnapshot, family?: string | n
  *
  * Stickiness: bind the conversation's stickyKey to an account for the life
  * of that conversation, and fall off only when the bound account is
- * exhausted / rejected. The 6-hour TTL matches the Max plan's five-hour
- * rate-limit window plus a buffer — past that point a "same" conversation
- * would be starting a fresh window anyway, so rebinding is free.
+ * exhausted / rejected. The 6-hour TTL is measured from a binding's LAST
+ * use, not its creation: an actively-running session refreshes the timer on
+ * every turn (see selectSticky), so it is never rebound out from under a
+ * warm prompt cache — agent sessions routinely run past 6h, and an age-based
+ * TTL would force such a session onto a cold account mid-conversation. A
+ * conversation that goes quiet is reaped 6h after its final turn; by then
+ * its cache has long expired (Anthropic prompt cache lives at most 1h) and a
+ * "same" conversation returning would start fresh anyway, so rebinding is free.
  */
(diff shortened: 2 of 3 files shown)


sprayberry-reviewer (approved):

## What I checked
- `gh pr diff 792` (CHANGELOG.md, src/pool.ts, test/pool-sticky.mjs)
- The single call site of `selectSticky`/`rebindSticky` outside `pool.ts` (`src/proxy.ts:2649,3311,3384,3402`) — confirmed it uses the 2-arg form as the PR body claims, so the new optional `now` param is purely additive and doesn't change proxy behavior.
- CI check-rollup: all 12 checks green (`test`, `build (18/20/22)`, `CodeQL`, `compat`, `docker-cap-drop-smoke`, etc.), `auto-merge` skipping (expected pre-approval).

## Correctness
The core change (`src/pool.ts:264-537`) is a straightforward and well-scoped rename/semantics shift: `boundAt`-based TTL → `lastUsedAt`-based idle TTL.
- `selectSticky` (`src/pool.ts:471-490`) now takes `now` as a parameter (default `Date.now()`), threads it into `cleanupSticky(now)`, reuses it for the validity checks, and stamps `binding.lastUsedAt = now` on a hit before returning — this is the crux of the fix and it's correct: an active session's timer is re-based on every hit.
- `cleanupSticky` (`src/pool.ts:513-537`) reaps on `now - b.lastUsedAt > STICKY_IDLE_TTL_MS` instead of `boundAt`, and the size-cap eviction sort now orders by `lastUsedAt` (true LRU) instead of `boundAt` (LRC) — both match the stated intent and the new tests exercise exactly these paths (active-past-6h-age survives; idle-past-6h reaps and re-picks; LRU eviction under the size cap).
- `rebindSticky` correctly stamps both `boundAt` and `lastUsedAt` to the same fresh `now` on explicit rebind.

I traced the size-cap test (`test/pool-sticky.mjs:295-321`) by hand: 2000 bindings inserted with strictly increasing `lastUsedAt`, then `keys[0]` (oldest-created) is re-touched to become newest-used, then two more inserts push past the cap and trigger the 80%-target eviction sweep. The expected outcome (`keys[0]` survives, `keys[1]` evicted, overflow key survives) is consistent with sorting ascending by `lastUsedAt` and dropping the lowest `size - target` entries — no off-by-one or ordering issue found.

## What's good
- The doc comment above `StickyBinding`/`STICKY_IDLE_TTL_MS` is updated in the same diff to match the new semantics, rather than left stale.
- The `now` injection follows an existing codebase convention (`isInAuthCooldown(account, now)`), keeping the new tests deterministic without sleeping.
- Test coverage is solid and targets exactly the three behavior changes: idle-vs-age survival, idle reap + re-pick, and LRU (not LRC) eviction.

No blocking findings.

PR #793, opened 2026-07-17
feat(health): surface live session/sticky counts on /health + doctor

askalf:
## What & why

The proxy tracks live sessions in memory — the single-account session-id `SessionRegistry` and the pool's conversation→account sticky bindings — but there was **no way to observe them**. `stickyCount()` and `SessionRegistry.size()` were test-only helpers. An operator couldn't answer "how many sessions is dario tracking, and is the lazy cleanup keeping up?" (a question the 2026-07-17 token incident made concrete).

This started as an investigation into whether dario needed a **background session-reaper** (a competitor has one). It doesn't: both structures already reap lazily, and a background sweeper was *deliberately rejected* in `session-rotation.ts` because entries disappearing on a timer is an observable fingerprint a real Claude Code client doesn't have. So the right deliverable is **visibility**, not a reaper.

## Change

- **`/health`** gains a `sessions` field, behind the existing #642 internal-disclosure gate (authenticated or bare-loopback callers only — the public Cloudflare tunnel never sees it, same as `oauth`/`requests`):
  - single-account → `{ "mode": "single", "active": <SessionRegistry.size()> }`
  - pool → `{ "mode": "pool", "stickyBindings": <AccountPool.stickyCount()> }`
- **`dario doctor`** prints a `Sessions` line reading that field off loopback; silently skipped when no proxy is running (the counts don't exist, and doctor is often run *because* the proxy is down).
- Raw in-memory counts by design — they double as a signal that lazy cleanup (LRU cap 1024 / 6h idle TTL + 2000 cap) is keeping up.

## Tests

`test/health-response.mjs`: internal `/health` surfaces `sessions` (both modes), public hides it, omitted when not supplied. **Suite 117/117, tsc clean.** No wire/billing change; purely additive read-only observability.

## Not shipped

No version bump — lands in `[Unreleased]`, ships with the next bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ checklist.
 ## [Unreleased]
 
 - **Session stickiness TTL is now idle-based, not age-based.** A conversation's account binding is reaped 6h after its *last* turn rather than 6h after its first, and every sticky hit refreshes the timer. Previously a session running continuously past the 6h mark was force-rebound to a possibly-different account mid-conversation, throwing away its warm Anthropic prompt cache — the exact thrash stickiness exists to prevent, and it landed hardest on the long agent sessions that benefit from stickiness most. Genuinely idle conversations are still reaped on the same 6h horizon (now measured from their final turn), and the sticky size-cap evicts least-recently-*used* instead of least-recently-*created*.
+- **`/health` and `dario doctor` now surface live session-tracking counts.** Internal `/health` callers (loopback / authenticated — never the public Cloudflare tunnel) get a `sessions` field: `{mode:"single", active:N}` for the session-id registry or `{mode:"pool", stickyBindings:N}` for conversation→account bindings. `dario doctor` prints the same as a `Sessions` line when a proxy is running. Both reap lazily by design (no background sweeper — that would be an observable fingerprint a real client doesn't have), so the raw in-memory count also confirms cleanup is keeping up.
 
 ## [5.2.7] - 2026-07-17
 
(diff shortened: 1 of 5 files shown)


sprayberry-reviewer (approved):

**Verdict:** No blocking issues found — clean, well-scoped, tests added, CI green (12/12 checks passing).

## What I checked
- `src/health-response.ts:59-61,146` — new `sessions` field is only spread into the response body inside the `includeInternal` branch of `buildHealthResponse`; the public (`liveness`-only) branch never sees it, so the existing #642 internal-disclosure gate is preserved — no new leak path.
- `src/proxy.ts:1971-1976` — mode selection is keyed off `pool.size === 0` (the same predicate already used at `proxy.ts:1588` and `proxy.ts:3952` to distinguish single-account vs pool mode), so `sessions` picks `sessionRegistry.size()` in single mode and `pool.stickyCount()` in pool mode consistently with how the rest of the file already discriminates the two.
- `src/doctor.ts:920-941` — new `/health` probe follows the same fetch-with-short-timeout / silent-catch pattern as the pre-existing `--usage` and `--obedience` probes in the same file (`doctor.ts:616-632`, `doctor.ts:766-775`); the lack of an `Authorization` header here matches those existing probes too — `shouldDiscloseHealthInternals` grants internal disclosure on bare loopback regardless of auth, so this isn't a new gap introduced by this PR.
- `test/health-response.mjs` new block covers: internal single-mode, public-hides-sessions, internal pool-mode, and omitted-when-not-supplied — matches the four branches touched in `health-response.ts`.
- CI rollup: all 12 checks (build×3, test, compat, docker-cap-drop-smoke, CodeQL, analyze, actionlint, triage, wire-drift, validate-package-json) green.

## What's good
Small, additive, read-only observability change with a coherent story (visibility instead of a background reaper, consistent with the prior deliberate rejection of a sweeper in `session-rotation.ts`). Test coverage matches the new branches exactly. No wire/billing-path changes.

No findings to report — approving.

PR #791, opened 2026-07-17
fix(oauth): durably persist refreshed tokens + startup self-heal (#790)

askalf:
## Summary
Fixes the v5.2.x regression where refreshed OAuth tokens were kept **only in memory** — the on-disk `credentials.json` and pool `accounts/login.json` stayed frozen at mint. Any container recreate after ~8h uptime loaded a rotated-away refresh token → `invalid_grant` → 100% fleet 401s until a manual `--force-reauth` (2026-07-17 outage, 15:36–16:01Z). `dario-autodeploy` recreates on releases, so this recurred on every deploy.

Closes #790.

## Root cause
The pool refresh loop *did* write rotated tokens back to disk, but `saveAccount`/`saveCredentials` used a plain `writeFile(tmp)+rename` with **no fsync**. An abrupt container recreate (`docker rm -f` → SIGKILL, the autodeploy path) discarded the page cache before the kernel flushed, so a bind-mounted `~/.dario` reverted to the last *durable* (mint-time) content — the exact "identical ms stamps, frozen at mint" symptom in the issue.

## Changes
- **`src/durable-write.ts`** (new): `durableWriteFile()` — write temp → `fsync(file)` → `rename` → `fsync(dir)`, so a rotated token survives SIGKILL. Windows busy-file rename fallback + best-effort dir-fsync for filesystems that don't support it.
- **`accounts.saveAccount` + `oauth.saveCredentials`**: now use `durableWriteFile` (fix 1 — atomic write + fsync, covers the pool account files too).
- **`proxy` startup**: eagerly refresh any expired/expiring pool account **before** serving (fix 3 — was: account sits `expired` until the first 15-min background tick; every request 401s meanwhile). Durably persists the rotated token so the *next* recreate is clean too. A dead refresh token just logs and the auth gate surfaces it.
- **`proxy` SIGTERM/SIGINT**: flush the freshest in-memory pool tokens to disk (fix 2 — belt-and-braces; freshest-wins, never clobbers a fresher concurrent disk write; under the existing 5s force-exit guard so a hung fsync can't wedge teardown).
- **`cli.materializeLoginPool`**: when the back-fill no-ops (accounts/ already has an entry), resync a stale `login` snapshot from the fresh `credentials.json` (issue comment — a fresh `login --force-reauth` wrote only `credentials.json`, leaving the pool routing on the stale token).

## Test plan
`test/durable-token-persist.mjs` — no real network, no live model loop (injected `fetch`):
[...]

Issue #790, opened 2026-07-17
v5: refreshed OAuth tokens never persisted to disk — recreate after >8h loads a rotated-away refresh token


askalf:
## Summary
v5.2.x refreshes OAuth tokens in memory but never persists them back to disk. After ~8h of uptime the on-disk refresh token has been rotated away server-side, so any container recreate (autodeploy, restart, host reboot) loads a dead credential family and every request 401s until a manual re-login.

## Observed (2026-07-17, self-hosted deployment)
- Proxy served normally for ~25h (so in-memory refreshes were clearly succeeding).
- `~/.dario/credentials.json` AND the v5 pool store `~/.dario/accounts/login.json` were both frozen at the 2026-07-16 14:06:33Z mint (`expiresAt` 22:06:33Z — identical ms stamps, i.e. last disk write was that mint).
- Auto-deploy recreated the container 2026-07-17 15:35Z → it loaded the stale file → `doctor` shows `OAuth expired`, `Pool 1 expired`, `next: login`; every non-passthrough request fails auth.
- `dario refresh` on the stale file → dead family (rotated refresh token).

## Impact
Any recreate after >8h uptime is a guaranteed outage requiring interactive `login --force-reauth`. Health-gated auto-deploys can't self-heal from it (restart loads the same dead file).

## Suggested fixes
1. Persist to `credentials.json` / the pool account file on every successful refresh (v4.8 behavior) — atomic write + fsync.
2. Flush current tokens on SIGTERM as a belt-and-braces (observed the pool store being touched at teardown, but with the stale token).
3. Startup with an expired-but-refreshable token should attempt a refresh before declaring the account `expired` (longstanding follow-up from the 2026-07-02 incident).

askalf:
Recovery data point (2026-07-17 outage): after `login --manual --force-reauth` minted fresh credentials, only `credentials.json` was written — the v5 pool store `accounts/login.json` retained the stale token and the recreated container would have kept routing on it (`Pool 1 expired`). Had to move `accounts/login.json` aside before recreate; dario then rebuilt it from the fresh `credentials.json` and went 1/1 healthy. So fix (1) should cover the pool account files too, and `login` should update (or invalidate) the pool store for the matching account.

PR #806, opened 2026-07-18
fix(accounts): don't clobber a newer pool token from stale credentials.json (#805)

### LINKED ISSUE #805: Container recreate re-seeds pool account from stale credentials.json → clobbers refreshed token → auth outage
## Summary

In pool-of-1 mode, a container **recreate re-seeds the pool account `accounts/login.json` from the legacy `credentials.json`**, clobbering an already-refreshed pool token with a stale one. Because the pool refreshes `accounts/login.json` (never `credentials.json`), the legacy file is routinely stale; and because refreshing rotates the refresh token, the stale copy's refresh token is already dead. Result: after any recreate that follows a pool refresh, startup loads a token whose refresh returns `invalid_grant` → the account sits in auth-cooldown → `/accounts` 503 → **fleet-wide auth outage**.

This is the failure mode #790/#791 (persist-on-refresh) was meant to close. Persisting works — but a recreate then overwrites the persisted token from the stale legacy file, so the class recurs.

## Evidence (2026-07-18 outage, self-hosted)

- 23:23:28Z — pool refreshed `accounts/login.json` → token expiry **07:23:28Z**, persisted (5.2.7 working). Verified healthy at 00:03Z.
- 03:42–03:57Z — **5.2.8 autodeploy recreate** (one failed attempt at 03:42, success at 03:57).
- 03:43:02Z — `accounts/login.json` mtime updated, but its `expiresAt` went **backwards to 00:01:22.281Z** — byte-identical to `credentials.json`'s token (mtime 16:01Z, never updated in pool mode). The recreate re-seeded the pool account from the stale legacy file.
- That token's refresh was consumed when the 07:23Z token was minted → every startup logs:
  `[dario] Startup refresh failed for login: ... (400): {"error":"invalid_grant","error_description":"Refresh token not found or invalid"}`
- 03:51Z→12:08Z: **168 auth-failed executions, 2/187 completed (99% failure)** until an operator `--force-reauth`.

## Suggested fix

On startup / pool seed, do **not** blindly overwrite a pool account from `credentials.json`. Options (probably (1)+(3)):

1. **Reconcile by freshness** — when both `credentials.json` and `accounts/login.json` exist for the same identity, keep the one with the later `expiresAt` (or don't overwrite an existing pool account from the legacy file at all).
2. **Mirror on refresh** — when the pool refreshes `accounts/login.json`, also update `credentials.json` so the two never diverge.
3. **Don't self-heal into a rotated token** — startup self-heal that hits `invalid_grant` should not have overwritten a still-valid pool token to get there in the first place.

## Contributing factor

The recreate was triggered by **5.2.8 autodeploy churn**. Rapid back-to-back dario releases keep recreating the container near/after a pool refresh, maximizing the clobber window — the same "space out releases so the refresh loop can settle" hazard noted before. Root cause is the clobber, not the churn; the churn just triggers it.

askalf:
Closes #805. Fixes the recurrence of the fleet-wide OAuth auth-outage class that #790/#791 was meant to close.

## Problem
`resyncLoginFromCredentialsIfStale()` overwrote the pool `accounts/login.json` from the legacy `credentials.json` on ANY token divergence (assuming the legacy file is always fresher — the #235 single-account case). In pool mode the pool's own refresh loop advances `login.json` while `credentials.json` stays frozen. So a container recreate replaced the live pool token with the stale legacy one, whose refresh_token Anthropic had already rotated → `invalid_grant` on every startup → account in auth-cooldown → `/accounts` 503 → **fleet-wide 401 auth outage**.

Two outages in 12h: 2026-07-17 16:01Z and 2026-07-18 03:51–12:12Z (fleet exec failure peaked ~99%, forge monitor fired execution_failure_rate CRITICAL, ticket MON-mrpukoqs-s8s9hu). The 5.2.8 autodeploy recreate at 03:51Z triggered the clobber.

## Fix
Reconcile by **freshness**: only overwrite `login.json` when `credentials.json` is newer-or-equal (the #235 case). A strictly-newer pool token is kept ('creds-stale') and the protective skip is logged at startup. New regression test covers the strictly-newer-login case; existing #235 tests (creds newer, and equal-expiry access-only divergence) still pass.

## Verification
- src/accounts.ts: reconcile-by-freshness in resyncLoginFromCredentialsIfStale (skip overwrite when loginAcc.expiresAt > tok.expiresAt).
- src/proxy.ts: logs the 'creds-stale' protective skip.
- test/resync-login-stale.mjs: new regression (strictly-newer login kept).
- Branch author reports suite 117/117; CI on this PR is authoritative.
- Bumped 5.2.10.

Source ticket: MON-mrpukoqs-s8s9hu. Prior fix in this class: PR #791 (#790).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.2.10] - 2026-07-18
+
+- **Fix: a container recreate could clobber a freshly-refreshed pool token, causing a fleet-wide auth outage (#805).** `resyncLoginFromCredentialsIfStale` overwrote the pool's `login` account from `credentials.json` on *any* token divergence, assuming the legacy file was always the newer source (#235's single-account case). In pool mode the reverse happens — the pool's own refresh loop advances `accounts/login.json` while `credentials.json` stays frozen — so a recreate replaced the live pool token with the stale legacy one, whose refresh token Anthropic had already rotated → `invalid_grant` on every startup → every request 503s until a manual re-auth. The resync now reconciles by freshness: it only overwrites when `credentials.json` is newer-or-equal, and leaves a strictly-newer pool token untouched (`creds-stale`).
+
 ## [5.2.9] - 2026-07-18
 
 - **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.214` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.214 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.2.9",
+  "version": "5.2.10",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -644,6 +644,9 @@ export async function ensureLoginCredentialsInPool(
  *   - 'in-sync'      : tokens match; no action
  *   - 'resynced'     : login.json was stale; overwrote with current
  *                       credentials. Caller should reload pool state
+ *   - 'creds-stale'  : login.json is STRICTLY NEWER than credentials.json
+ *                       (the pool refreshed it; the legacy file is stale) —
+ *                       left login.json untouched. dario#805.
  *
  * Why: the single-account path keeps refreshing `credentials.json` in
  * the background (proxy startup auth check, periodic refresh in oauth.ts).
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped fix with a regression test covering the new case.

## What I checked
- `gh pr diff 806` (75 additions / 5 deletions across `CHANGELOG.md`, `package.json`, `src/accounts.ts`, `src/proxy.ts`, `test/resync-login-stale.mjs`).
- Base content of `resyncLoginFromCredentialsIfStale` (pre-PR) to confirm the diff context and that the `'in-sync'` short-circuit above the new branch is unchanged.
- CI rollup: `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `CodeQL`, `actionlint` all SUCCESS. `wire-drift` and `compat` were still QUEUED/IN_PROGRESS at review time (self-hosted runners) — not blocking findings, just noting they hadn't reported yet.

## Analysis of the core change (`src/accounts.ts:653-706`)
```ts
if (loginAcc.expiresAt > tok.expiresAt) {
  return 'creds-stale';
}
// ... else falls through to the existing overwrite (credentials.json wins)
```
This directly targets the described bug: previously *any* divergence overwrote `login.json` from `credentials.json` unconditionally, which is exactly the clobber that caused the #805 outage when the pool's own refresh loop had advanced `login.json` past a frozen `credentials.json`. The reconciliation flips only the strictly-newer-login case, and preserves the pre-existing #235 behavior for the equal-`expiresAt`/access-token-only-diverged case (falls through to overwrite) — this matches the adjacent "only access-token diverges → still resyncs" test that's unchanged in this diff.

I traced the millisecond-precision-collision edge case (both tokens coincidentally sharing the same `expiresAt`) and it isn't a new risk introduced here — the old code had the identical exposure (unconditional overwrite), so it's out of scope for this diff.

## Test coverage
`test/resync-login-stale.mjs` adds a regression case (`dario#805`) that sets up a strictly-newer `login.json` against a stale `credentials.json` and asserts `creds-stale` is returned and the access/refresh tokens are left untouched — directly covers the new branch. Good.

## What's good
- The comment above the function is updated in place to explain the bidirectional divergence risk, not just the fix — useful for the next person who touches this.
- `proxy.ts` adds a matching startup log line for the new `creds-stale` outcome so the skip is observable, consistent with the existing `'resynced'` log.
- Scoped tightly to the reported outage class; no unrelated changes beyond the version bump/changelog.

No correctness, reuse, or efficiency concerns found in the changed lines.

PR #807, opened 2026-07-18
fix(accounts): don't clobber a newer pool token from stale credentials.json (#805)

### LINKED ISSUE #805: Container recreate re-seeds pool account from stale credentials.json → clobbers refreshed token → auth outage

askalf:
Fixes #805.

## The bug

`resyncLoginFromCredentialsIfStale()` (src/accounts.ts) overwrote the pool's `login` account from `credentials.json` on **any** token divergence, assuming the legacy file is always the newer source — correct for #235 (single-account path refreshes `credentials.json`, pool snapshot goes stale) but **inverted in pool mode**, where the pool's own refresh loop advances `accounts/login.json` while `credentials.json` stays frozen (the pool never writes it).

So a container **recreate** replaced the live, freshly-refreshed pool token with the stale legacy one — whose refresh token Anthropic had already rotated when the pool refreshed. Result: `invalid_grant` on every startup, the account stuck in auth-cooldown, `/accounts` 503, and every fleet LLM request failing until a manual `--force-reauth`.

**Two fleet-wide outages in ~12h** on the self-hosted deployment: 2026-07-17 ~16:01Z and 2026-07-18 03:51→12:12Z (168 auth-fails, **2 of 187 executions completed**). Both traced to a `credentials.json`/`accounts/login.json` `expiresAt` that had gone *backwards* to a byte-identical stale value right after a recreate.

## The fix

Reconcile by **freshness**, not by assuming `credentials.json` wins:

- `credentials.json` newer-or-equal → overwrite the stale snapshot (`resynced`, the #235 behaviour — unchanged).
- pool token **strictly newer** → leave it untouched, return `creds-stale`. Proxy startup logs the protective skip.

Only the strict case is skipped; equal `expiresAt` keeps the old overwrite, since a real refresh always advances `expiresAt` (so equal-but-different is an anomaly, not a pool refresh).

## Tests

New regression case in `test/resync-login-stale.mjs`: a strictly-newer `login.json` is **not** clobbered (`creds-stale`, tokens preserved). Existing #235 tests — creds-newer overwrite, and the equal-expiry access-only divergence — still pass. `tsc` clean, **suite 117/117**.

## Shipping

Bumped **5.2.10** so this deploys on merge — the box stays exposed to a re-clobber on its next pool refresh + recreate until it lands. The deploy-recreate itself is safe: after the 12:12Z re-auth both credential files hold the same fresh token (`in-sync`), so nothing is clobbered on the way in.

Follow-up (separate, not blocking): the pool refresh could also **mirror** the new token into `credentials.json` so `dario doctor`'s OAuth row (which reads the legacy file) stops falsely reporting `expired` in pool mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.2.10] - 2026-07-18
+
+- **Fix: a container recreate could clobber a freshly-refreshed pool token, causing a fleet-wide auth outage (#805).** `resyncLoginFromCredentialsIfStale` overwrote the pool's `login` account from `credentials.json` on *any* token divergence, assuming the legacy file was always the newer source (#235's single-account case). In pool mode the reverse happens — the pool's own refresh loop advances `accounts/login.json` while `credentials.json` stays frozen — so a recreate replaced the live pool token with the stale legacy one, whose refresh token Anthropic had already rotated → `invalid_grant` on every startup → every request 503s until a manual re-auth. The resync now reconciles by freshness: it only overwrites when `credentials.json` is newer-or-equal, and leaves a strictly-newer pool token untouched (`creds-stale`).
+
 ## [5.2.9] - 2026-07-18
 
 - **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.214` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.214 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.2.9",
+  "version": "5.2.10",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -644,6 +644,9 @@ export async function ensureLoginCredentialsInPool(
  *   - 'in-sync'      : tokens match; no action
  *   - 'resynced'     : login.json was stale; overwrote with current
  *                       credentials. Caller should reload pool state
+ *   - 'creds-stale'  : login.json is STRICTLY NEWER than credentials.json
+ *                       (the pool refreshed it; the legacy file is stale) —
+ *                       left login.json untouched. dario#805.
  *
  * Why: the single-account path keeps refreshing `credentials.json` in
  * the background (proxy startup auth check, periodic refresh in oauth.ts).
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — the freshness-reconciliation fix is correct and well-tested.

## What I checked
- `src/accounts.ts:686-716` — the new `creds-stale` branch in `resyncLoginFromCredentialsIfStale`.
- `src/proxy.ts:1436-1440` — the new log line wiring up the `creds-stale` result.
- `test/resync-login-stale.mjs` — the new regression case for the strictly-newer-login.json path.
- CI rollup (`gh pr view --json statusCheckRollup`): all required checks green (CI build/test ×3 node versions, CodeQL, compat, wire-drift, actionlint, validate-package-json). Only `auto-merge` shows `SKIPPED`, which is expected for a non-bot-authored PR.

## Correctness
The core change:
```
if (loginAcc.expiresAt > tok.expiresAt) {
  return 'creds-stale';
}
```
correctly inverts the prior unconditional overwrite only for the strictly-newer-pool-token case, and leaves the `#235` overwrite-on-newer-or-equal-credentials.json path untouched — matching the stated invariant ("a real refresh always advances expiresAt", so equal expiry is treated as the legacy #235 case, not a false pool-refresh signal). I traced both branches:
- `loginAcc.expiresAt` (pool) strictly newer → `creds-stale`, no write, login.json preserved (protects the just-fixed outage in #805).
- Equal or credentials.json newer → falls through to the existing `saveAccount` overwrite, identical to pre-PR behavior for the #235 case.

The new test in `test/resync-login-stale.mjs` exercises exactly the failure mode described in the bug report (pool token 8h out, credentials.json expired an hour ago) and asserts both the return value and that `accessToken`/`refreshToken` are left untouched — good coverage of the regression this PR fixes.

`proxy.ts`'s new `else if` branch is a pure additive log statement; it doesn't touch the `resynced` branch or fall through incorrectly.

## What's good
- The doc comment above the function was updated to explain *why* the direction flips in pool mode, not just *what* changed — useful for the next person who touches this.
- Test asserts on both the returned status enum and the actual persisted token values, not just one or the other.
- Version bump + changelog entry included in the same PR, consistent with the "ship on merge" note in the description.

No test-coverage or efficiency concerns — this is a small, targeted control-flow fix with a directly corresponding regression test.

PR #809, opened 2026-07-18
fix(accounts): mirror pool-refreshed login token to credentials.json (#808)

### LINKED ISSUE #808: Mirror pool-refreshed token to credentials.json (fixes doctor false-'expired' in pool mode) — follow-up to #805
## Follow-up to #805

#805 stopped the clobber (a recreate no longer overwrites a newer pool token from stale `credentials.json`). But it leaves the two files *diverged*: after a pool refresh, `accounts/login.json` is fresh while `credentials.json` stays frozen at the last login. Two consequences worth closing:

1. **`dario doctor` reads `credentials.json` for its OAuth/Identity rows**, so in pool-of-1 mode it prints `OAuth expired`/`expiring` even when the active pool token is freshly refreshed and the fleet is 200-healthy. This is alarm-inducing (it nearly got today's *recovered* outage misread as a live regression) and likely what auto-filed the `dario doctor drift` issue during the outage window.
2. Any other code path that still reads `credentials.json` sees a stale token indefinitely.

**Proposed fix:** when the pool refreshes the `login` account, mirror the new token back into `credentials.json` (atomic write), so the legacy file tracks the pool store instead of diverging. Alternative/also: have `doctor`'s OAuth + Identity rows read the active pool account in pool mode (or cross-check `/health`) rather than the legacy file.

Not urgent — #805 already prevents the outage. This is consistency + honest diagnostics.

askalf:
Closes #808. Follow-up to #805.

## Problem
#805 stopped the pool→credentials clobber but left the two files diverged. After a pool refresh, `accounts/login.json` is fresh but `credentials.json` stays frozen at the last `dario login` (nothing ever writes it back). Consequences:

1. `dario doctor` reads `credentials.json` for its OAuth row → prints `expired`/`expiring` in pool-of-1 mode even when the active pool token is fresh and the fleet is 200-healthy (alarm-inducing false positive; likely auto-filed the doctor-drift issue during the 2026-07-18 outage).
2. Any other reader of `credentials.json` (co-resident Claude Code, ops script) sees a stale token indefinitely.

## Fix
New `mirrorLoginToCredentials(refreshed)` in `src/accounts.ts` writes a refreshed `login` account's token back into `credentials.json` via a durable atomic write (`saveCredentialsTokens` in oauth.ts — shares `saveCredentials`' fsync temp+dir #790, cache invalidation, and refresh-dead-flag clear).

Wired into **both** proxy refresh sites (startup self-heal + 15-min background loop), best-effort (`.catch` logs) so a mirror failure never fails the refresh that produced the live token.

Only the `login` alias is mirrored — it's the one back-filled FROM `credentials.json` (`ensureLoginCredentialsInPool`), so the only account whose canonical home is that legacy file. `work`/`personal` accounts have no legacy counterpart.

**Freshness-guarded, symmetric with #805:** mirror only when the refreshed login token is strictly newer than `credentials.json` (by `expiresAt`). A newer-or-equal `credentials.json` means another process (e.g. a concurrent `dario login --force-reauth`) just wrote a fresher family — never clobber it; `resyncLoginFromCredentialsIfStale` pulls that direction on next startup.

## Test plan
New `test/mirror-login-to-credentials.mjs` (no network) covers all four branches:
- non-login alias → `skip-not-login`, no file created
- strictly-newer pool token → `mirrored` (access/refresh/expiresAt/scopes written)
- older AND equal-expiry pool token → `creds-newer`, legacy file untouched (#805 direction)
- missing `credentials.json` → `mirrored` (creates it)

CI compiles (build 18/20/22) + auto-discovers the new test via `all.test.mjs`. No version bump (not shipping — per ticket DEV-8cf98def).

### DIFF
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -23,7 +23,7 @@ import { homedir } from 'node:os';
 import { randomUUID, randomBytes, createHash } from 'node:crypto';
 import { createServer, type IncomingMessage, type ServerResponse } from 'node:http';
 import { detectCCOAuthConfig } from './cc-oauth-detect.js';
-import { loadCredentials, buildManualAuthorizeUrl, parseManualPaste, readLineFromStdin, enumerateKeychainCredentials, type KeychainEntry } from './oauth.js';
+import { loadCredentials, saveCredentialsTokens, buildManualAuthorizeUrl, parseManualPaste, readLineFromStdin, enumerateKeychainCredentials, type KeychainEntry } from './oauth.js';
 import { openBrowser } from './open-browser.js';
 import { redactSecrets } from './redact.js';
 import { durableWriteFile } from './durable-write.js';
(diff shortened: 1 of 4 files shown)


sprayberry-redline (approved):
**Automated review from the Sprayberry Labs fleet code reviewer (Code Reviewer bot).**

**Verdict:** No blocking issues found — clean, well-guarded fix with solid test coverage.

## What I checked
- Full diff (`src/accounts.ts`, `src/oauth.ts`, `src/proxy.ts`, `test/mirror-login-to-credentials.mjs`).
- Surrounding context for `saveCredentials`/`loadCredentials` caching, `refreshAccountToken`'s per-alias in-flight dedup, and `resyncLoginFromCredentialsIfStale` (the #805 companion direction) to confirm the two directions don't ping-pong.
- CI rollup (`gh pr checks 809`): CodeQL, actionlint, analyze, build(18/20/22), docker-cap-drop-smoke, test, triage, validate-package-json all `pass`; `compat` and `wire-drift` still `pending` (non-blocking).

## Analysis
- `mirrorLoginToCredentials` (`src/accounts.ts:753-770`) only fires for the reserved `login` alias, and the freshness guard (`if (currentExpiry >= refreshed.expiresAt) return 'creds-newer';`) is the exact mirror image of `resyncLoginFromCredentialsIfStale`'s own guard (`if (loginAcc.expiresAt > tok.expiresAt) return 'creds-stale';`). At equal `expiresAt` the two functions don't fight: `resync` treats equal as "creds is at least as fresh, copy creds→pool" while `mirror` treats equal as "pool isn't strictly newer, don't touch creds" — so there's no write-write ping-pong at the boundary, matching the PR's stated intent.
- Both call sites in `src/proxy.ts` (startup self-heal at `:1530-1534` and the 15-min background loop at `:1552-1556`) wrap the mirror in `.catch(...)` that only logs, so a mirror failure can't fail the refresh that produced the live token — matches the "best-effort" contract in the docstring.
- `saveCredentialsTokens` (`src/oauth.ts:494-496`) is a thin, correctly-scoped wrapper reusing `saveCredentials`'s existing durability (fsync temp+dir), cache invalidation, and `refreshTokenDead` clear — no duplicated logic.
- New test file exercises all four documented branches (non-login skip, strictly-newer mirror, older/equal no-clobber, missing-file bootstrap) against the compiled `dist/` output, consistent with how this repo's other `test/*.mjs` files work.

## What's good
- Docstrings on `mirrorLoginToCredentials` and `saveCredentialsTokens` clearly state the invariant and its relationship to #805/#790, which made verifying the no-ping-pong property straightforward.
- Freshness guard is symmetric and explicit about the "never clobber a fresher family" invariant, addressing the exact failure mode #805 fixed in the other direction.

No correctness, security, or test-coverage concerns block this PR.

PR #812, opened 2026-07-18
proxy: add /livez liveness route decoupled from OAuth

askalf:
## What

Add a `/livez` route that always returns `200 {"status":"ok"}` while the HTTP server is accepting requests, independent of OAuth state.

## Why

The container healthcheck (`wget http://127.0.0.1:3456/health`) is what the `autoheal` watchdog keys on. `/health` returns **503** when OAuth is `broken` / `none` / expired-with-no-refresh (intentional — uptime monitors and `depends_on: service_healthy` need that readiness signal). But a **container restart cannot mint a new refresh token**, so an OAuth/token-family outage turns into a restart loop.

This bit us in prod: the box's interactive Claude Code and the dario fleet transiently shared one refresh-token family; single-use refresh → the second refresher got `invalid_grant` → family revoked → `/health` 503 → autoheal restarted every ~4 min for **4h20m**, none of which could help. It only recovered when the credential was refreshed out-of-band.

Liveness (should we restart?) and readiness (can we serve upstream calls?) are different questions. Restart decisions should track liveness.

## Change

- `src/proxy.ts`: new pre-auth `/livez` branch before `/health`, returns 200 always. Mirrors the existing `/health` write path; uses the same `JSON_HEADERS`.
- `test/e2e.mjs`: `testLivez()` asserts auth-free 200.

Readiness is unchanged: `/health` keeps 200/503 and the full internal-vs-public disclosure split.

## Companion

Platform PR points the compose healthcheck at `/livez` (lands after this ships; autodeploy within-minor auto-pulls).

## Trade-off to weigh

A genuinely *hung-but-listening* server that still accepts TCP would pass `/livez` and no longer be auto-restarted. That failure mode wasn't what we observed (the observed loop was the opposite — a healthy server wrongly restarted), and `/health` + alerting still surfaces it. If you'd rather `/livez` also assert the request pipeline is turning over, say so and I'll add a cheap internal check.


sprayberry-redline (approved):

**What I checked**
- `gh pr diff 812` (28 additions / 1 deletion across `src/proxy.ts` and `test/e2e.mjs`)
- Surrounding context in `src/proxy.ts` (the request-handling closure around the new branch, and the existing `/health` handler it's modeled on)
- CI rollup via `gh pr checks 812`: all 13 checks green (CodeQL, build 18/20/22, test, compat, docker-cap-drop-smoke, wire-drift, etc.)

**Findings**

None blocking. The new `/livez` branch in `src/proxy.ts:1417-1421` is placed before the `checkAuth(req)` call (`src/proxy.ts:1447`), which is correct given the PR's stated intent — liveness must stay auth-free and independent of OAuth state, unlike `/health` which is also pre-auth but returns 503 on broken OAuth. The handler is a straight 200 + static JSON body, no branching or state to get wrong.

`test/e2e.mjs:54-65` adds `testLivez()` asserting a 200 with `{status:'ok'}`, and wires it into the `main()` run sequence right after `testHealth()` — matches the existing test's shape and actually exercises the new route rather than just asserting shape in isolation.

**What's good**
- Route placement is deliberate and matches the design goal (decoupling restart-triggering liveness from OAuth-dependent readiness) — verified by reading the surrounding `/health` block, which retains its 503-on-broken-OAuth behavior unchanged.
- Comment on the new branch clearly documents the incident that motivated it and why `/health` remains untouched.
- Test coverage added for the new endpoint in the same PR.
- All CI checks green, including CodeQL and the full build/test matrix.

sprayberry-redline (approved):
Re-approving post branch-update (base sync only). CI green on updated head; /livez liveness route is additive and decouples liveness from OAuth.

PR #814, opened 2026-07-18
fix(tls): flag Bun below the JA3-verified floor instead of a false-green match

### LINKED ISSUE #813: I did an audit
Hello!
I ran an audit with ClaudeCode Opus 4.8.
Apologies in advance if this is noise / AI-slop.
It might be worth 5 mins of your time to run it thru your own agentic AI.

❤️
π

# dario wire-fidelity — empirical findings

*Independent measurement of the claim that dario's requests match Claude Code "byte-for-byte" / "packet-identical." Every result below is measured on one machine and reproduced; the harness scripts are included so anyone can re-run.*

**Rig**: macOS arm64 · `claude` 2.1.214 (Mach-O standalone, embeds `Bun/1.4.0`, node-compat v26.3.0) · Node v23.7.0 · public Bun 1.0.9 / 1.3.14 / canary(→1.4.0) · dario @ HEAD (2901bc9).

**Instruments** (all local, no sudo, zero subscription cost except where noted):
- `capture-ch.mjs` + `ja3.mjs` — grab a client's TLS ClientHello (first TCP payload, cleartext) without completing the handshake → JA3.
- `mitm-proxy.mjs` — env-scoped `HTTPS_PROXY` CONNECT proxy that TLS-terminates `api.anthropic.com` with a trusted local cert, logs decrypted HTTP/1.1 (raw header order+case + body, token redacted), returns a stub. Only the one subprocess is affected.
- `collector.mjs` + `hdrtest.js`/`btest.js` — local TLS server to compare Node-vs-Bun fetch behavior.
- Real CC `/v1/messages` captured by running the user's authenticated `claude` through the MITM (stub reply ⇒ nothing billed). dario egress captured by running dario from source with its upstream const pointed at the collector and a **dummy** account (shape-only; no real creds, no cost).

---

## Layer 0 — TLS ClientHello (JA3). Deterministic (3/3, no GREASE).

| Client | Runtime | JA3 |
|---|---|---|
| `claude` 2.1.214 | Bun 1.4.0 (embedded) | `e97f5146a7009cc2918b50e903b6ff8d` |
| bare Bun 1.3.14 / canary | BoringSSL | `e97f5146…` **(byte-identical to CC)** |
| bare Bun 1.0.9 | BoringSSL | `2ae7eb4b…` (diverges: +3DES, +ECH, +padding) |
| bare Node v23.7.0 | OpenSSL | `53321030…` (52 ciphers) |

- Node ≠ Bun, massively. **The TLS gap is closable** with public Bun ≥ 1.3.14 (reproduces CC exactly). No public `bun-v1.4.0` exists (canary line).
- **Gap dario misses:** `runtime-fingerprint.ts` treats `bun-match` as "under Bun at all" and never checks the Bun *version*'s JA3. On this box dario would auto-relaunch under the Bun 1.0.9 on PATH and report a green `bun-match` while emitting `2ae7…` ≠ CC's `e97f…`. → measure JA3, or pin/warn a min Bun version.

## Layer 2 — the actual `/v1/messages` request: CC vs dario (measured)

**Header order — real CC** (SDK insertion order, preserved on the wire):
`Accept · Authorization · Content-Type · User-Agent · X-Claude-Code-Session-Id · X-Stainless-Arch · X-Stainless-Lang · X-Stainless-OS · X-Stainless-Package-Version · X-Stainless-Retry-Count · X-Stainless-Runtime · X-Stainless-Runtime-Version · X-Stainless-Timeout · anthropic-beta · anthropic-dangerous-direct-browser-access · anthropic-version · x-app · x-client-request-id · Connection · Host · Accept-Encoding · Content-Length`

**dario under Node (undici):**
`host · connection · accept · content-type · user-agent · x-claude-code-session-id · x-stainless-*(8) · anthropic-beta · anthropic-dangerous… · anthropic-version · x-app · Authorization · x-client-request-id · accept-language · sec-fetch-mode · accept-encoding · content-length`
→ replays dario's captured order, BUT: all names **lowercased**; `Authorization` appended near the **end** (pos ~19, vs CC pos 2); `host`/`connection` hoisted to **front**; undici **injects `accept-language` + `sec-fetch-mode`** (CC never sends these); `accept-encoding: br, gzip, deflate` (CC: `gzip, deflate, br, zstd`).

**dario under Bun 1.3.14 (fetch sorts):**
`Accept · Authorization · Content-Type · User-Agent · anthropic-beta · anthropic-dangerous… · anthropic-version · x-app · x-claude-code-session-id · x-client-request-id · x-stainless-*(8) · Connection · Host · Accept-Encoding · Content-Length`
→ **alphabetically sorted**; `x-*`/`x-stainless-*` lowercased and placed *after* `anthropic-*`, unlike CC. dario's captured `header_order` is **discarded** by Bun's fetch.

**Root cause (measured):** plain `fetch()` under **both** public Bun 1.3.14 and canary (== CC's 1.4.0) **sorts** headers (`zeta,accept,m-custom,Authorization → Accept,Authorization,m-custom,zeta`). Real CC on the same Bun is *not* sorted ⇒ the Anthropic SDK bypasses `fetch`'s Headers sort via raw dispatch. **dario uses `fetch()`, so it cannot reproduce CC's header order under Bun — its `header_order` machinery is a no-op there.** Under Node it survives but undici adds its own fingerprint.

### Verdict grid

| Property | dario/Node | dario/Bun 1.3.14 | matches CC? |
|---|---|---|---|
| TLS JA3 | ✗ (`5332…`) | ✓ (`e97f…`, if Bun≥1.3.14) | Bun only |
| Body top-level key order | ✓ | ✓ | **yes** |
| `metadata.user_id` shape, system-prompt/billing-header injection | ✓ | ✓ | **yes** (rebuild works) |
| `x-client-request-id`, `x-claude-code-session-id` present | ✓ | ✓ | yes |
| header **name casing** | ✗ lowercased | ✗ lowercased | **no** (CC mixed-case) |
| header **order** | replayed but Authz misplaced | ✗ alphabetized | **no** |
| transport-injected headers | ✗ `accept-language`,`sec-fetch-mode`, host/conn front | ✓ clean | Bun only |
| `Accept-Encoding` value | ✗ `br,gzip,deflate` | (Bun default) | Node no |
| `x-stainless-arch`/`os` | ✗ `x64`/`Linux` on fallback path | ✓ `arm64`/`MacOS` (live) | live-capture only |
| `anthropic-beta` set/order | `advisor-tool-2026-03-01` vs CC `advanced-tool-use-2025-11-20`; missing `extended-cache-ttl-2025-04-11`; oauth flag at tail vs pos 2 | same | **no** |

## What dario gets right (credit)
Body key order exact; live value overlay (UA `(external, sdk-cli)`, pkg `0.94.0`, runtime `v26.3.0`) correct; arch/os correct on the live path; injects the CC/Agent-SDK system prompt + `x-anthropic-billing-header` + `thinking`/`context_management`/`output_config` to rebuild a bare client into CC shape; sends `x-client-request-id`, session-id, presence pings; TLS reproducible under a recent Bun.

## Bottom line
- **"byte-identical body"**: substantially TRUE (key order + serialization). ✅
- **"packet-identical" on the wire**: FALSE as stated. Best achievable today is *structurally close under a recent Bun*, still distinguishable by: header-name **casing** (lowercased), header **order** (Bun's fetch sorts; dario's replay is defeated), `anthropic-beta` **set**, and — under Node — undici-injected headers. Plus the unverified-Bun-version JA3 gap.
- The single highest-leverage fix: dario cannot match CC's header order/casing through `fetch()`. Matching CC requires the same raw-dispatch path the Anthropic SDK uses (or an http/1.1 client that preserves order+case). Everything else (casing, Authz position, beta set, JA3 version check) is incremental.

## Bugs/observations to file
1. `bun-match` doesn't verify Bun **version**/JA3 (false-green on old Bun).
2. `fetch()` transport re-sorts (Bun) / injects (Node) headers, silently defeating `header_order` replay — the documented v3.13 wire-fidelity axis is a no-op on the recommended runtime.
3. Header names emitted lowercase; CC is mixed-case.
4. OAuth `Authorization` appended at tail, not CC's position 2 (template carries capture-artifact `x-api-key` instead).
5. `anthropic-beta` drift: `advisor-tool-2026-03-01` vs real `advanced-tool-use-2025-11-20`; missing `extended-cache-ttl-2025-04-11`.
6. Bundled-fallback `x-stainless-arch/os` = `x64`/`Linux` (clobbers correct values when live capture unavailable).
7. Hardcoded fallback `x-stainless-runtime-version: v24.3.0` (real: v26.3.0).
8. Context: CC itself phones home (`/api/event_logging/v2/batch` 163–252 KB + Datadog) — relevant to "nothing phones home" positioning.

## Irreducible
Adversary feature set is unobservable (we can prove wire-identical across observed layers, never "undetectable"); results are point-in-time (CC/Bun version drift).

askalf:
## What

The runtime TLS-fingerprint check (`src/runtime-fingerprint.ts`) returned `bun-match` for **any** Bun version. Being under Bun is necessary but not sufficient: only Bun ≥ **1.3.14** is measured to reproduce Claude Code's ClientHello.

Empirical basis (from the audit in #813, macOS arm64):

| Client | Runtime | JA3 |
|---|---|---|
| `claude` 2.1.214 | Bun 1.4.0 (embedded) | `e97f5146a7009cc2918b50e903b6ff8d` |
| bare Bun 1.3.14 / canary | BoringSSL | `e97f5146…` (byte-identical to CC) |
| bare Bun 1.0.9 | BoringSSL | `2ae7eb4b…` (diverges: +3DES, +ECH, +padding) |

So `dario` auto-relaunching into an old Bun on PATH (e.g. 1.0.9) would report a green `bun-match` while emitting a divergent JA3 on the wire.

## Change

- New `bun-ja3-unverified` status for Bun below the JA3-verified floor (`JA3_VERIFIED_BUN_FLOOR = '1.3.14'`), plus a `bunVersionMeetsJa3Floor()` comparator (numeric-tuple, canary-suffix tolerant).
- All three consumers key off the existing `status !== 'bun-match'` predicate, so they pick the new status up with **no wiring change**: `doctor` warns, the startup banner warns, and `--strict-tls` refuses to start.
- An **unreadable** Bun version stays a best-effort `bun-match` — there's nothing to compare against the floor, and Bun almost always exposes `.version`.
- Doc + help text updated (`docs/wire-fidelity.md`, `--strict-tls` help).

## Maintainer decision to confirm ⚠️

`--strict-tls` now **refuses to start** on a below-floor Bun (previously it accepted any Bun). This is the fail-closed reading of the flag ("refuse unless the TLS matches"), and an unverified JA3 is exactly what it exists to catch — but it is a behavior change for anyone running `--strict-tls` on Bun < 1.3.14. Flagging it explicitly rather than shipping it silently. Happy to gate the strict refusal behind a separate opt-in if you'd rather keep `--strict-tls` lenient on version.

## Scope

This addresses **finding #1** (version-blind `bun-match`) from #813. The other findings in that audit are separate:
- #2 (header order/casing lost through `fetch()`) is the higher-leverage, architectural one — needs a raw-dispatch transport, discussing on the issue.
- #5 (`anthropic-beta` set) is template-drift, cc-drift-bot territory.

Not `Closes #813` — that issue tracks several items.

## Tests

`test/runtime-fingerprint.mjs` extended: below-floor → `bun-ja3-unverified` (hint + floor named), floor boundary inclusive, unknown-version still matches, and direct comparator cases (numeric-not-lexical, canary suffix, unparseable → `undefined`). Full file green locally (52/52); `tsc` clean.


sprayberry-redline (approved):
**Automated review — Sprayberry Labs fleet code reviewer.**

## What I checked
- `gh pr diff 814` in full (all 5 changed files: `src/runtime-fingerprint.ts`, `src/cli.ts`, `src/proxy.ts`, `docs/wire-fidelity.md`, `test/runtime-fingerprint.mjs`).
- Consumer wiring in a shallow clone (`src/doctor.ts:365`, `src/proxy.ts:1274,3945`) to confirm the PR's claim that all three consumers key off `status !== 'bun-match'` and therefore pick up the new `bun-ja3-unverified` status with no wiring change — confirmed true for all three call sites.
- CI rollup: `gh pr checks 814` — all 12 required checks pass (build×3, test, compat, CodeQL, actionlint, wire-drift, etc.), only `auto-merge` skipped as expected pre-approval.

## Correctness
`bunVersionMeetsJa3Floor` (`src/runtime-fingerprint.ts`) does a numeric-tuple parse/compare rather than a lexical one, which is the right call — verified against its own test `'1.10.0' ≥ '1.3.14'` (would be wrong under a string compare). Boundary case (`bunVer === floor`) correctly falls through to `bun-match` since the loop returns `true` only on the first differing index and defaults to `true` when all three equal. The `bunVer !== 'unknown'` guard in `classifyRuntimeFingerprint` correctly preserves the existing best-effort `bun-match` behavior when the version can't be read, rather than newly warning on an absence of data — matches the stated intent and is exercised by the "Bun version unknown still classifies as match" test.

## What's good
- Empirical basis for the floor is documented inline with the source measurement (#813) rather than just asserted.
- Test coverage is thorough and specifically targets the sharp edges: exact-floor boundary, canary-suffix tolerance, numeric-vs-lexical ordering, and the unparseable-version fallback — exactly the cases a version-floor comparator tends to get wrong.
- The maintainer-decision callout in the PR body (that `--strict-tls` behavior changes for sub-floor Bun) is the right thing to flag explicitly rather than ship silently — not a review blocker, just noting it's been surfaced appropriately for a maintainer call.

No test-coverage gaps, reuse, or efficiency concerns in the diff.

PR #815, opened 2026-07-18
feat(transport): raw HTTP/1.1 over Bun.connect for faithful header order (opt-in, #813)

askalf:
## What

`fetch()` cannot put Claude Code's request headers on the wire faithfully, and this defeats dario's existing `header_order` machinery. Measured for #813 finding #2 (a distinctive non-alphabetical, mixed-case ordered set sent through `fetch`):

- **Bun `fetch`** → headers **sorted alphabetically** (`X-Claude-Code-Session-Id`-first became `Accept`-first) + `Host`/`Connection`/`Accept-Encoding` injected.
- **Node `fetch`/undici** → order kept but names lowercased, `host`/`connection` hoisted, and `accept-language`/`sec-fetch-mode` **injected** (CC never sends these).

So `orderHeadersForOutbound` builds the right ordered array and `fetch` throws the order away before it reaches the socket.

## Change

New `src/raw-transport.ts` — an **opt-in** upstream transport (`DARIO_RAW_TRANSPORT=1`) that writes the HTTP/1.1 request bytes directly over **`Bun.connect`**. Measured: `Bun.connect`'s TLS ClientHello is byte-identical to Bun's `fetch` (same BoringSSL JA3 Claude Code presents — `d871d02c` on the test box); `node:tls` is **not** (`203503b7`), so `Bun.connect` specifically is required. That yields CC's exact header order + mixed case with zero transport injection **while keeping the matching JA3** — the "you can't win both axes" bind from the issue thread doesn't hold.

It returns a standard `Response`, so it's a **drop-in** for the six upstream `fetch(targetBase, …)` sites in `proxy.ts` (they consume only `.status`/`.headers`/`.text()`/`.body`). Wired behind a one-line selector:

```ts
const dispatchUpstream = useRawTransport ? rawUpstreamFetch : (t, o) => fetch(t, o);
```

**Default OFF.** Falls back to `fetch` on Node (no `Bun.connect`) or when the flag is unset — zero behavior change until you flip it, and a startup log announces the active transport.

## Tests

The socket layer is injected (`RawSocketFactory`), so the framing/decode/streaming is unit-tested under Node's runner over `node:net` (`test/raw-transport.mjs`, 16 assertions, auto-discovered by `all.test.mjs`):
- streaming **chunked** SSE reconstructed incrementally; `transfer-encoding` stripped from surfaced headers
- **gzip** content-encoding decoded on the fly; `content-encoding` stripped (matches `fetch`'s post-decompression view)
- sent header **order + mixed case preserved**, NOT alphabetized, no `accept-language`/`sec-fetch-*` injection
- `content-length` body via `.text()`; pre-aborted signal → `AbortError`

Separately verified by measurement (harness in the issue thread): `Bun.connect` JA3 == `fetch` JA3; on-the-wire header order/casing exact; and the real `bunSocketFactory` + this module streaming SSE end-to-end.

## Follow-ups (not in this PR)

- **`br` / `zstd`** content-encoding: gzip/deflate/br handled; zstd feature-detected. CC advertises all four; Anthropic SSE is gzip/identity in practice, but worth hardening.
- **Transport-header tail positioning**: `Host`/`Content-Length` are appended at the tail rather than positioned from the template's `header_order` (CC's exact tail is `Connection · Host · Accept-Encoding · Content-Length`). The app-header axis `fetch` destroyed is faithful; exact transport-tail order is a refinement.
- **Prod validation**: confirm `Bun.connect == fetch` JA3 holds on the Linux prod platform, and validate a real TLS round-trip, before flipping the flag in prod.

Addresses **finding #2** from #813 (not `Closes` — that issue tracks several items).

### DIFF
--- a/docs/wire-fidelity.md
+++ b/docs/wire-fidelity.md
@@ -12,3 +12,11 @@ Between v3.22 and v3.28, dario's Claude backend closed six axes along which a pr
 | **MCP / sub-agent reach** | v3.26 + v3.27 | Not a wire axis — a *surface* axis. CC-aware tools can now address dario directly (sub-agent from inside CC, MCP server for any MCP client), so operators don't have to switch terminals to introspect the proxy. Read-only by design. | `dario subagent install` / `dario mcp`. See [`mcp-server.md`](./mcp-server.md) and [`sub-agent.md`](./sub-agent.md). |
 
 The six-direction wire-fidelity roadmap is complete. Subsequent releases return to responding to issues and upstream template drift.
+
+## Request header order + casing (opt-in, `DARIO_RAW_TRANSPORT`)
+
+The six axes above are what `fetch()` lets us control. One axis it does **not**: the request header order and casing on the wire. `fetch()` re-normalizes the header block before it leaves the process — under Bun it sorts the names alphabetically, and under Node/undici it lowercases them and injects `accept-language` / `sec-fetch-mode` that Claude Code never sends. So the `header_order` captured in the live template (and replayed by `orderHeadersForOutbound`) is discarded by the transport and never reaches the socket (surfaced empirically in #813).
+
+`DARIO_RAW_TRANSPORT=1` closes that axis by bypassing `fetch` for the upstream `/v1/messages` calls: it writes the HTTP/1.1 request bytes directly over **`Bun.connect`**, whose TLS ClientHello is byte-identical to Bun's `fetch` (the same BoringSSL JA3 Claude Code presents — measured, unlike `node:tls` which diverges). That gives CC's exact header order + mixed case with zero transport injection, while keeping the matching JA3.
+
+Opt-in and **off by default** — the default transport stays `fetch`, so this carries no risk until enabled. Bun-only: under Node (`Bun.connect` absent) the flag is ignored and `fetch` is used. See `src/raw-transport.ts`.
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -10,6 +10,7 @@ import { getAccessToken, getStatus } from './oauth.js';
 import { buildHealthResponse, derivePoolStatus, shouldDiscloseHealthInternals } from './health-response.js';
 import { darioVersion } from './version.js';
 import { buildCCRequest, applyCcPromptCaching, parseEffortSuffix, reverseMapResponse, createStreamingReverseMapper, orderHeadersForOutbound, isMcpToolName, CC_TEMPLATE, CC_CACHE_CONTROL, effectiveCacheControl, withForced1hBeta, type ToolMapping, type RequestContext, type EffortValue } from './cc-template.js';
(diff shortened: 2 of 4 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-tested, opt-in change. Approving.

## What I checked

- `gh pr diff 815` in full (566 lines across `docs/wire-fidelity.md`, `src/proxy.ts`, `src/raw-transport.ts`, `test/raw-transport.mjs`).
- CI check-rollup: 14/14 checks green (`build (18/20/22)`, `test`, `compat`, `CodeQL`, `wire-drift`, `docker-cap-drop-smoke`, etc.) — no pending or failing checks.
- The six call sites in `src/proxy.ts` that changed `fetch(targetBase, …)` → `dispatchUpstream(targetBase, …)` (`src/proxy.ts:3078`, `3145`, `3218`, `3237`, `3288`, `3335` per the diff) — all six are covered, matching the PR description's claim of "six upstream `fetch(targetBase, …)` sites."
- The state machine in `src/raw-transport.ts`'s `rawUpstreamFetch` (header serialization → status-line/header parse → chunked/content-length/EOF body framing → gzip/deflate/br decode → abort/close handling), tracing:
  - `onClose` when `phase !== 'body'` → rejects "upstream closed before response headers" (correct: response promise not yet resolved).
  - `onClose` with `bodyComplete` already true → no-op (correct: avoids a spurious truncation error racing the async decoder flush).
  - `onClose` with no `content-length`/not chunked → treated as legitimate EOF via `finishBody()` (correct for identity-encoded streaming bodies).
  - Pre-aborted `AbortSignal` → rejects `AbortError` before the socket factory is even invoked (`if (init.signal.aborted) { reject(...); return; }`) — matches the test at `test/raw-transport.mjs`'s "abort" block.
  - Content-Length/Transfer-Encoding tail synthesis only adds `Host`/`Content-Length` when the caller didn't already supply them — avoids duplicate headers on the wire.

Default is `fetch` (opt-in via `DARIO_RAW_TRANSPORT=1`, further gated on `Bun.connect` availability), so this ships with zero behavior change until explicitly enabled — that significantly de-risks the change regardless of any residual edge case in the new code path.

## What's good

- The socket layer is injected (`RawSocketFactory`), letting the chunked/gzip/streaming/abort logic run under Node's test runner over `node:net` without a real Bun/TLS dependency — good testability design for something that's otherwise hard to unit test.
- The module's header comment is unusually thorough about *why* (JA3 measurements, fetch's header re-normalization on both Bun and Node/undici) and honestly documents known follow-ups (br/zstd hardening, transport-tail header positioning) rather than silently leaving them out.
- Fallback behavior is conservative: Node (no `Bun.connect`) or unset flag → `fetch`, with a startup log announcing which transport is active either way.

No correctness, reuse, efficiency, or coverage issues rose to blocking severity in the changed lines.

PR #816, opened 2026-07-18
harden(transport): CC-exact header tail + br/zstd + fail-loud on unknown encoding (#813)

askalf:
Hardening follow-ups to #815, closing two of the three items I flagged in that PR. All in `src/raw-transport.ts`.

## 1. Exact transport-header tail order

#815 appended `Host` + `Content-Length` at the tail. CC's captured `header_order` ends with **`connection · host · accept-encoding · content-length`**, and dario doesn't set any of those itself (it relied on `fetch` to add them). The transport now synthesizes all four in that exact order with CC-faithful values:

- `Connection: keep-alive`
- `Host: <url host>`
- `Accept-Encoding: gzip, deflate, br, zstd` (CC's advertised set — vs undici's `br, gzip, deflate`)
- `Content-Length: <n>`

Casing matches CC's wire request (mixed-case), not the lowercased `header_order` slugs. Each is skipped if the caller already supplied it.

## 2. br/zstd decode + fail-loud on the rest

Confirmed `createBrotliDecompress` **and** `createZstdDecompress` exist under both Bun and Node, so all four encodings dario advertises (gzip/deflate/br/zstd) actually decode. And the silent-corruption hole is closed: an encoding we *can't* decode now **fails the request** (`unsupported content-encoding: …`) instead of passing the compressed bytes through as if they were the body.

## Tests

`test/raw-transport.mjs` extended — **20/20 green**:
- transport headers appear in CC's exact tail order; `Accept-Encoding` value asserted
- br body decoded + `content-encoding` stripped
- unsupported encoding (`snappy`) → rejects rather than corrupts

## Remaining #815 follow-up (not this PR)

Confirming `Bun.connect == fetch` JA3 on the **Linux prod** platform + a real TLS round-trip before the flag is ever flipped in prod — that's the validation step, tracked separately.

Follow-up to #815; addresses more of finding #2 from #813.

### DIFF
--- a/src/raw-transport.ts
+++ b/src/raw-transport.ts
@@ -135,12 +135,18 @@ export function rawUpstreamFetch(
     ? Buffer.alloc(0)
     : (typeof init.body === 'string' ? Buffer.from(init.body, 'utf8') : Buffer.from(init.body));
 
-  // Serialize headers verbatim (exact order + case). Append the transport
-  // headers CC's own transport adds, if the caller didn't already include them.
+  // Serialize headers verbatim (exact order + case), then append the transport
+  // headers CC's own transport adds — in CC's captured tail order
+  // (Connection · Host · Accept-Encoding · Content-Length, from the template's
+  // header_order) with CC-faithful values — skipping any the caller already
+  // supplied. Casing matches CC's wire request (mixed-case, not the lowercased
+  // header_order slugs).
   const pairs = toPairs(init.headers);
   const present = new Set(pairs.map(([k]) => k.toLowerCase()));
   const tail: Array<[string, string]> = [];
+  if (!present.has('connection')) tail.push(['Connection', 'keep-alive']);
   if (!present.has('host')) tail.push(['Host', u.host]);
+  if (!present.has('accept-encoding')) tail.push(['Accept-Encoding', 'gzip, deflate, br, zstd']);
   if (!present.has('content-length') && !present.has('transfer-encoding')) {
     tail.push(['Content-Length', String(bodyBuf.length)]);
   }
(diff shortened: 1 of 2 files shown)


askalf:
Closing: the raw-transport premise is broken — Bun.connect's TLS ClientHello (203503b7) does not match fetch's (d871d02c, Claude Code's JA3), and no Bun.connect option closes the gap. So the transport regresses the JA3 rather than keeping it. #815 is being reverted (see the revert PR); this hardening rides on the same false premise. Corrected finding on #813.

sprayberry-redline (approved):

**Verdict:** Clean — no blocking issues found. CI is all-green (11/11 checks pass, `test` and `compat` included).

## What I checked
- `src/raw-transport.ts`: the new `Connection`/`Accept-Encoding` tail headers and the `setupDecode(): boolean` refactor that fails the request on an unrecognized `content-encoding` instead of passing compressed bytes through.
- `test/raw-transport.mjs`: new coverage for tail header order, br decode, and the unsupported-encoding rejection path.
- Traced the `setupDecode()` failure branch (`src/raw-transport.ts` new lines in the `feed()` head-parsing block: `if (!setupDecode()) { ... reject(e); destroy(); return; }`) against `onClose`'s `if (!resolved) reject(...)` guard. `resolved` is only ever set `true` in the success path, so a `destroy()`-triggered `onClose` after this rejection will call `reject()` a second time with a different error — but this exactly mirrors the pre-existing `if (!m) { ...; if (!resolved) reject(e); destroy(); return; }` malformed-status-line handler a few lines above, which has the same shape. Since a settled Promise silently ignores a second `reject()`, this isn't an observable bug, and it's an existing pattern the PR didn't introduce, so not flagged as a finding.
- `zstd` gating (`ce === 'zstd' && 'createZstdDecompress' in zlib`) correctly falls through to the unsupported-encoding path on runtimes lacking it, rather than silently mis-handling — consistent with the PR's stated goal of failing loud.
- Test's expected tail order (`['Connection', 'Host', 'Accept-Encoding', 'Content-Length']`) matches the `tail.push(...)` call order in the implementation.

No correctness, security, reuse, or efficiency issues found in the changed lines. Test coverage for the new behavior (tail order, br decode, unsupported-encoding rejection) looks adequate.

## What's good
- The `setupDecode()` return-value refactor cleanly closes the "silent corruption" hole described in the PR body — an unrecognized encoding now rejects instead of streaming ciphertext-looking bytes as if they were plaintext.
- Good inline comments explaining *why* (CC's captured header_order, why identity/'' short-circuits, why bodyComplete is set before the decoder's async end).

PR #817, opened 2026-07-18
Revert #815 — raw Bun.connect transport doesn't keep CC's JA3 (bind is real)

askalf:
Reverts #815.

**Why:** the premise was wrong, and I owe a correction. #815 claimed `Bun.connect` presents the same TLS ClientHello as Bun's `fetch` (Claude Code's JA3). That rested on a measurement bug — my JA3 harness's mode switch only recognized `'tls'`, so the `bun-connect` mode silently ran the **fetch** path. I measured `fetch` twice and read it as "`Bun.connect == fetch`."

**Measured correctly:**

| client | JA3 | extensions |
|---|---|---|
| `fetch` (Claude Code) | `d871d02c` | `…16, 5, …18, …21` (ALPN, status_request, SCT, padding) |
| `Bun.connect` | `203503b7` | `0,23,65281,10,11,35,13,51,45,43` (none of those) |

No `Bun.connect` TLS option adds the missing extensions (`alpn`/`ALPNProtocols`/`alpnProtocols` are ignored or wrong-type). `Bun.connect` == `node:tls` == a leaner profile that **cannot** be made to match `fetch`.

So the raw transport wins request header order/casing but **regresses the JA3** away from CC — it trades one fingerprint axis for another, not "wins both." The bind #815 said it dissolved is real on Bun.

The feature was default-off (`DARIO_RAW_TRANSPORT`), so nothing shipped enabled — this just removes the misleading code + docs. #816 (hardening on the same premise) is being closed. The corrected finding is recorded on #813.


sprayberry-redline (approved):

**Verdict:** Clean revert — no blocking issues found.

## What I checked
- `gh pr diff 817` (566 lines): removes `src/raw-transport.ts` and `test/raw-transport.mjs` wholesale, strips the four `dispatchUpstream(...)` call sites in `src/proxy.ts` back to plain `fetch(...)`, removes the now-stale `useRawTransport`/env-flag selector block, and trims the corresponding doc section in `docs/wire-fidelity.md`.
- Cloned the PR head (`pr817`) and grepped the whole tree for `raw-transport`, `rawTransport`, `rawUpstream`, `dispatchUpstream`, `DARIO_RAW_TRANSPORT` — zero hits remaining anywhere (code, tests, docs). No dangling import or reference left behind.
- CI rollup (`gh pr checks 817`): all 13 checks green (build ×3, test, compat, wire-drift, CodeQL, actionlint, docker-cap-drop-smoke, etc.).
- The PR body gives a credible, specific root cause for the revert (a JA3 harness mode-switch bug in the prior PR's own measurement) and the feature being removed was default-off (`DARIO_RAW_TRANSPORT` unset), so this is a low-risk mechanical revert, not a behavior change for anyone.

## Findings
None. The four call-site edits (`src/proxy.ts`, e.g. the primary call and the three retry paths for beta-flag rejection / effort clamp / max_tokens clamp / long-context beta) all consistently swap `dispatchUpstream(targetBase, {...})` → `fetch(targetBase, {...})` with identical `init` shape, so no call site was missed or left inconsistent.

## What's good
- The revert is complete and self-consistent — deleted module, deleted its dedicated test file, and every import/call site updated together in one diff, leaving no dead code or broken imports.
- PR body is an honest, well-documented postmortem of the measurement bug that motivated the original change, and explicitly notes the sibling PR (#816) built on the same wrong premise is also being closed.

PR #846, opened 2026-07-24
feat(models): add Claude Opus 5 (v5.3.0)

askalf:
Adds Claude Opus 5 (`claude-opus-5`, `claude-opus-5[1m]`) as the model the opus family shortcuts resolve to.

The catalog autodetects new ids from `api.anthropic.com/v1/models` on its own, so most of this is the surface autodetection *doesn't* reach — the baked fallback that serves cold starts and offline runs, and the hand-maintained per-model tables.

### What changed

| Area | Change |
|---|---|
| `model-catalog.ts` | `claude-opus-5` heads the opus line in `BAKED_BASE_MODELS`; 4-8 / 4-7 / 4-6 stay. `[1m]` variant is generated by the existing rule, so `claude-opus-5[1m]` needs no separate entry. |
| `proxy.ts` | `opus` → `claude-opus-5`, `opus1m` derived from the same base. New `opus48` legacy pin for the previous flagship, alongside `opus47` / `opus46`. |
| `analytics.ts` | Opus 5 priced at $5/$25 per 1M, cache read $0.50 / write $6.25 — the Opus 4.8 rate, no long-context premium. |
| `doctor.ts` | The opus-family representative for the `--usage` and `--obedience` probes is now Opus 5. |
| docs / `--help` / TUI | Model-lineup lines name Opus 5 as the current opus. |

### What deliberately did *not* change

- **`betaForModel`** — Opus 5 falls through to the base set, which is the opus shape. Golden matrix extended to assert it (plus `opus-5[1m]`, which takes `context-1m-2025-08-07` at position 2 like every other family).
- **`supportsAdaptiveThinking`** — already admits Opus 5 via the `opus-5+` branch; the existing forward-compat test covered it before this PR. Added to the empirical table with two facts worth recording: Opus 5 runs adaptive when `thinking` is omitted (4.8/4.7 do not), and `thinking:{type:"disabled"}` 400s on it above `high` effort. dario sends `thinking` explicitly and never sends the disabled shape, so neither is a live hazard here.
- **Rate-limit buckets** — Opus 5 is a separate upstream bucket from the combined Opus 4.x pool, but dario reads the bucket label off the response headers rather than deriving it, so nothing to map.

### Verification

`npm run build` clean, `npm test` 118/118. Not live-tested against upstream — I couldn't reach a running dario instance from this box, so the "does upstream list `claude-opus-5` for a subscription token" question is unverified. That only affects whether autodetection or the baked list is doing the work; both paths land on the same id.

### Release note

This carries a version bump (5.2.19 → 5.3.0), so **merging this ships it** — tag, GitHub Release, `npm publish`, and the GHCR push all fire off the bump. Post-merge smoke per `RELEASING.md` still applies.


sprayberry-redline (dismissed):

**Verdict:** No blocking issues found — clean, well-tested version bump. Approving.

## What I checked
- `gh pr diff 846` in full (14 files, +60/-20) and the PR's CI check-rollup (`gh pr checks 846`) — all 12 checks pass (build/test/compat/CodeQL/docker-cap-drop-smoke/etc).
- The alias-resolution chain end to end: `src/proxy.ts` `MODEL_ALIASES['opus']` now points at `'claude-opus-5'`, the new `opus48` key correctly pins `'claude-opus-4-8'` (the prior flagship), and `opus1m` is re-derived from `claude-opus-5[1m]` rather than left stale.
- `src/analytics.ts`: `claude-opus-5` pricing entry (`$5/$25`, cache read `$0.50`/write `$6.25`) matches the PR description's stated Opus-4.8 rate, and is placed ahead of the `claude-opus-4-8` entry it mirrors.
- `src/model-catalog.ts`: `claude-opus-5` added to `BAKED_BASE_MODELS` ahead of the 4-8/4-7/4-6 line.
- Test coverage is unusually thorough for a bump this size — `test/model-catalog.mjs`, `test/beta-matrix.mjs`, `test/pricing-intro-window.mjs`, `test/provider-prefix.mjs`, and `test/fable-listings.mjs` all gained assertions for the new id, the `opus48` legacy pin, and the `[1m]` derivation, and none of the pre-existing assertions were weakened.
- `betaForModel` and `supportsAdaptiveThinking` were deliberately left alone per the PR body, and that's consistent with the diff — no code path in either function changes, only comments/tests documenting why.

## Minor (non-blocking)
- The PR description flags itself as unverified against live upstream (no reachable dario instance to confirm `api.anthropic.com/v1/models` actually lists `claude-opus-5` yet) and notes this only matters for autodetection vs. baked-fallback timing, since both converge on the same id — reasonable to ship given the fallback path is fully covered by the new tests either way.
- `docs/usage.md`'s example model list also switched `claude-sonnet-4-6` → `claude-sonnet-5`, which is outside this PR's stated opus-5 scope, but it brings the doc in line with the already-current `sonnet` alias elsewhere in the codebase, so not a concern.

## What's good
Comments explaining *why* (e.g. the fast-mode pricing tier note in `analytics.ts`, the adaptive-thinking default-on behavior in `cc-template.ts`) are genuinely useful for the next person touching this table, and the CHANGELOG entry is detailed enough to double as a release note.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean version bump, well tested. Approving.

## What I checked
- Full diff (`gh pr diff 846`, 14 files, +60/-20) plus the incremental commit added since the
  prior review round (`compare a32745b...2ae1743`) — that increment is only a merge-from-master
  picking up unrelated CI/lockfile bumps and a CHANGELOG date fix, no logic touched.
- CI check-rollup (`gh pr checks 846`): all 12 checks pass (build ×3, test, compat, CodeQL,
  docker-cap-drop-smoke, wire-drift, actionlint, validate-package-json, dependabot config, analyze).
- Alias chain: `src/proxy.ts:210-219` — `MODEL_ALIASES['opus'] = 'claude-opus-5'`, new
  `opus48` correctly pins `'claude-opus-4-8'`, and `opus1m` is re-derived to
  `'claude-opus-5[1m]'` rather than left stale (the exact class of bug the comment above the map
  says #389 caused previously).
- Version ordering: `src/model-catalog.ts` `modelVersionKey('claude-opus-5')` → `[5]`,
  `cmpVersionDesc` compares `b[i] ?? -1` element-wise, so `[5]` vs `[4,8]` at i=0 gives `4-5=-1`
  → opus-5 correctly sorts ahead of opus-4-8 in `normalizeUpstreamIds`/`resolveFamilyBase`, backed
  by the new `test/model-catalog.mjs` assertion (`'opus 5 outranks the 4-x line'`).
- `src/analytics.ts:184-187` — new `claude-opus-5` pricing entry ($5/$25, cache $0.50/$6.25)
  matches the Opus-4.8 entry immediately below it, and `pricingRateFor` strips the trailing
  `[1m]` tag before lookup (pre-existing logic, unchanged), so `claude-opus-5[1m]` correctly
  bills at the opus-5 rate rather than falling through to the sonnet-4-6 default — covered by
  the new `test/pricing-intro-window.mjs` assertions.
- `betaForModel`/`supportsAdaptiveThinking` — no code path in either function changes; only the
  golden matrix (`test/beta-matrix.mjs`) and doc comments were extended to assert opus-5 behavior,
  consistent with the PR body's claim that both already handle it via existing rules.
- Test coverage is thorough for a bump this size: `model-catalog.mjs`, `beta-matrix.mjs`,
  `pricing-intro-window.mjs`, `provider-prefix.mjs`, `fable-listings.mjs` all gained assertions
  for the new id and the `opus48` legacy pin, none weakened.

## What's good
The comment above `MODEL_ALIASES` explicitly documents the #389 opus/opus1m drift bug this PR's
"one derivation rule" structurally prevents from recurring, and the `analytics.ts` fast-mode
pricing-tier note explains a real subtlety (subscription proxy never sends `speed:"fast"`) rather
than leaving it as a silent assumption.

## Minor (non-blocking)
- PR is self-flagged as not live-verified against upstream `api.anthropic.com/v1/models` (no
  reachable dario instance from the author's box) — acceptable since both the autodetect and
  baked-fallback paths converge on the same id and the fallback path is fully covered by the new
  tests.

PR #851, opened 2026-07-25
fix(template): re-bake CC 2.1.220 + opus-5 fallback-credit beta (v5.3.1)

askalf:
Fresh live capture-and-bake plus a full live sweep of the whole surface. Two real drifts fixed here, and **two findings I did not fix because they need your call** — those are the important part of this PR, so they're up front.

---

## ⚠️ Finding 1 — CC ships Opus 5 and Sonnet 5 different system prompts. dario sends the base to both.

`systemPromptForModel()` branches on one thing: `fable`. Everything else gets the baked base, which is captured on `claude-opus-4-8`. I probed what CC actually sends per model (same mechanism the bake uses — `ANTHROPIC_MODEL` + loopback MITM, stub key, no OAuth), **two passes each, byte-identical both times**:

| capture model | system prompt (raw chars) | vs opus-4-8 base |
|---|---:|---|
| `claude-opus-4-8` | 6,664 | (reference — what dario bakes) |
| `claude-opus-5` | 9,990 | **DIFFERENT** — +13 lines |
| `claude-sonnet-5` | 28,156 | **DIFFERENT** — +147 lines |
| `claude-fable-5` | 11,084 | different (already has its own baked variant) |

Opus 5's extra content is unmistakably model-specific — it names itself:

```
- You are powered by the model named Opus 5. The exact model ID is claude-opus-5.
- Assistant knowledge cutoff is May 2026.
# Delivering work
# Corrections
Do not call the AgentTool unless the user requested it
```

Sonnet 5's +147 lines are the whole long-form prompt structure (`# System`, `# Doing tasks`, hooks/compaction guidance) — a different prompt *family*, not a few added lines.

**Consequences.** Billing is probably unaffected — `docs/research/system-prompt-classifier-study.md` measured prompt content/length as non-inputs to the classifier, and every variant routed `five_hour`. The real cost is behavioral: a user on Opus 5 through dario is missing the instructions real CC would give it. And it undercuts the fidelity claim for the two most-used models.

**Why I stopped rather than fixing it.** The bundle schema holds exactly two prompts (`system_prompt`, `system_prompt_fable`) and the bake spawns CC once per variant. Adding slots is mechanical, but the runtime half is not: `captureLiveTemplateAsync` does **one** capture at proxy startup, and dario can't spawn CC per model per request. So bundle-side per-model variants and the live-capture path would have different shapes, and picking how they reconcile (capture the N models the operator actually uses? fall back to base and accept the gap? capture lazily on first request per family?) is a design decision with a `_schemaVersion` bump attached. That's yours, not mine.

## ⚠️ Finding 2 — first request to a low-output-cap model returns a 400 to the client

Reproduced on a fresh proxy: a `max_tokens: 8` request for `claude-opus-4-1-20250805` came back
`400 invalid_request_error: max_tokens: 64000 > 32000` — dario had replaced it with `DEFAULT_MAX_TOKENS`. The clamp-and-retry at `proxy.ts:3264` learns and caches the cap (the very next request to that model, the `[1m]` variant, succeeded, and both retries after it returned `"OK"`), so the mechanism works — but the **first** caller eats the 400 rather than being rescued by the retry the comment describes. `opus-4-1` is advertised on `/v1/models`, so a client picker offering it fails once per proxy lifetime. Unrelated to Opus 5 and pre-existing; not mixed into a template PR.

---

## Fixed here

**Template re-bake.** `capture-and-bake --check` exited **2** — genuine shape drift, not a label lag:

```
anthropic_beta added: afk-mode-2026-01-31   (CC opted into a feature flag)
fable system-prompt variant drift (9200 -> 9222 chars)
```

Re-baked; `--check` now exits **0**. This is the same-binary remote-config class the checker exists for: CI reported zero shape drift on the v5.2.21 label refresh earlier the *same day*, so the flag flipped between those two runs. (Note `drift:wire` reported `drift: false` throughout — it strips model-conditional betas by design, so afk-mode is invisible to it. The two checks are complementary, not redundant.)

**Opus 5 was missing `fallback-credit-2026-06-01`.** v5.3.0 — mine — asserted opus-5 keeps the unchanged base set. Live capture disproves it:

| model | fallback-credit |
|---|---|
| `claude-opus-5` | **yes** |
| `claude-fable-5` | **yes** |
| `claude-opus-4-8` | no |
| `claude-sonnet-5` | no |
| `claude-haiku-4-5` | no |

It tracks the two models whose classifiers can refuse and route to a fallback. `betaForModel` now applies the fable transform to opus-5. Matched on `opus-5` specifically, not `opus-5+` — a future opus-6 probably inherits it, but that's a guess and the drift runner should be what tells us.

**The drift runner was blind to the flagship.** `check-wire-drift.mjs` probed opus-4-8 / sonnet-5 / haiku / fable — every family *except* `claude-opus-5`. That's why the omission shipped. Added; drift-free across all five with the fix in.

**Live E2E ran on stale models.** The obedience probes used opus-4-8 and sonnet-4-6, neither of which is what `opus`/`sonnet` resolve to now. Now opus-5 / sonnet-5.

## Verification

- `npm test` — 118/118
- `capture-and-bake --check` — exit 0 (was 2)
- `drift:wire` — `drift: false`, 5 models incl. opus-5
- **live E2E against a repo-build proxy on :3458 — 17/17**, including obedience on Opus 5 and Fable, streaming, tool use, OpenAI-compat, rate-limit headers
- **live 21-id model sweep — 20/21**, the one failure being Finding 2
- OAuth unrotated across all of it (credentials mtime unchanged, expiry only ticked down)

**Flaky test, disclosed:** twice this session a full run reported `# fail 1` that re-runs did not reproduce and that I could not attribute to a file (the failing subtest name didn't survive in my captured output). A 5-run loop to catch it was still going when I opened this; I'll comment with the culprit if it reproduces. Not caused by this diff — the first occurrence was on a docs-only tree.

**Merging this ships it** (5.3.0 -> 5.3.1).


askalf:
Closing the loop on the flaky-test note in the PR body: **it did not reproduce.** The catch-loop ran the full suite 5 more times, all `118/118`.

So the session record is 2 unattributable `# fail 1` runs out of ~12, then 12 consecutive clean runs. I could not capture the failing subtest name either time, so I have no culprit to name.

One correlation worth writing down for whoever sees it next, offered as a hypothesis rather than a finding: **both failures landed within a minute of live-proxy activity** (a `dario proxy` on :3456 / :3458 had just been running and killed). The suite runs `--test-concurrency=8` and several tests bind sockets, so a `TIME_WAIT` collision against a just-released port is a plausible mechanism — and it would explain why the 12 clean runs all came later, after sockets drained. Untested; I'm not filing it as a bug on that basis.

Practical read: if CI ever shows a lone unexplained failure here, re-run before investigating, and don't run the suite immediately after a live proxy on the same box.

sprayberry-redline (approved):

## What I checked
- `gh pr diff` (all 9 changed files) and surrounding context for `betaForModel` in `src/proxy.ts` (cloned read-only).
- CI check-rollup: all 12 checks green (`test`, `compat`, `wire-drift`, `build (18/20/22)`, CodeQL, etc.) — no pending or failing checks.

## Findings
None blocking.

**Correctness — verified, not a bug.** `src/proxy.ts:464`:
```js
} else if (m.includes('fable') || /opus-5(?!\d)/.test(m)) {
```
The negative lookahead correctly scopes the match: `claude-opus-5[1m]` matches (next char `[` isn't a digit), a hypothetical future `claude-opus-50`/`claude-opus-51` does *not* match (next char is a digit), and dated suffixes like `opus-5-20260101` still match (next char `-`). This lines up with the comment's stated intent ("matched on `opus-5` specifically rather than `opus-5+`") and the new test coverage in `test/fable-beta.mjs`/`test/beta-matrix.mjs` exercises both the plain and `[1m]` cases plus a double-insert idempotency check. I traced this looking for an over/under-match bug and didn't find one.

## What's good
- The PR body separates "fixed here" from two explicitly-deferred findings (per-model system-prompt divergence, first-request 400 on cap learning) with clear reasoning for *why* they're out of scope for this PR rather than silently ignoring them — good discipline.
- Live-capture evidence (CC 2.1.220) is cited inline at each behavioral change (`proxy.ts` comment block, CHANGELOG, test comments), and the new wire-drift entry (`claude-opus-5` added to `MODELS`) closes the exact gap that let the omission ship unnoticed in 5.3.0 — a real regression-prevention improvement, not just a fix.
- Test changes are proportionate to the behavior change: golden matrix assertion switched to compare against `FABLE` on purpose (not recomputed independently) so a future upstream split of the two families is caught rather than masked.

No changes requested.

PR #852, opened 2026-07-25
auto-rebake: template drift detected 2026-07-25

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-25T16:44:00+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.3.2` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — worth a closer read

- **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.220) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.220, 27 tools, 6171 char system prompt
[bake] scrubbed:
[bake]   tools: 27 → 27 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6171 → 4754 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] previous baked template: CC v2.1.220 captured 2026-07-25T15:56:45.086Z, 33 tools, 4776 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4754 → fable 9200 chars
[bake] check: drift detected — 1 differing slot (verdict: moderate):
[bake] **Verdict:** 🟡 Moderate — worth a closer read
[bake] 
[bake] - **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)
[bake] 
[bake] check: per-slot detail:
[bake]   • anthropic_beta removed: afk-mode-2026-01-31
[bake] wrote drift-summary.md for workflow embedding
[bake] check: fable system-prompt variant drift (9222 → 9200 chars).
[bake] check: bundled template is stale relative to live CC. Run `node scripts/capture-and-bake.mjs` to re-bake.
```

### Validation

[...]

PR #854, opened 2026-07-25
fix(template): per-model system prompts for opus-5 and sonnet-5 (v5.4.0)

askalf:
Fixes **Finding 1** from #851. Three defects that turned out to be one mechanism — and the root cause is worse than what I reported there, because it was breaking fable *today*, not just leaving Opus 5 unhandled.

Finding 2 (single-pass 400 recovery) is a separate concern in the proxy retry path and follows in its own PR — it's a restructure of the hottest path and doesn't belong in a template commit.

---

## What was wrong

**1. Opus 5 and Sonnet 5 got the base prompt.** `systemPromptForModel()` branched on `fable` and nothing else. Measured on CC 2.1.220, `ANTHROPIC_MODEL` + loopback capture, **two byte-identical passes each** (raw chars):

| capture model | prompt | vs base |
|---|---:|---|
| `claude-opus-4-8` (the bake base) | 6,664 | reference |
| `claude-opus-5` | 9,990 | +13 lines, incl. `# Delivering work`, `# Corrections` |
| `claude-sonnet-5` | 28,156 | +147 lines — the long-form family (`# System`, `# Doing tasks`) |
| `claude-fable-5` | 11,084 | already had a variant |

**2. A live capture silently wiped the baked variants — this was live breakage.** `loadTemplate()` returned *either* the live cache *or* the bundle. A live capture is one `claude --print` on one model, so it can never carry variants. A fresh cache therefore reverted every model-specific prompt to the base, which made the baked fable variant **inert on exactly the machines that have CC installed**. Reproduced before the fix:

```
base  len: 10006 (live capture in play)
fable len: 10006
fable variant: LOST -- falling back to base
```

So the fable variant only ever worked in no-CC deployments (the Hetzner container), where the bundle is used. That's the opposite of where you'd expect it to work.

**3. The captured base was machine-specific.** Look at that `10006` — an unpinned `claude --print` uses whichever model the user set as their CC *default*. This box defaults to Opus 5, so the "shared base" *was* the Opus 5 prompt, ~50% larger than opus-4-8's. The bake has always pinned `ANTHROPIC_MODEL` for exactly this reason; the runtime capture never did.

## The fix

- **`system_prompt_variants` map** keyed by family token (`fable`, `opus-5`, `sonnet-5`). The bake captures one per model, stores it only when it differs from the base after scrubbing, and keeps the previous variant on capture failure or residual-path detection — same guard rails the fable path already had.
- **`loadTemplate` merges** the bundle's variants onto a live-captured template. Variants the live template already carries **win**, so a future per-model live capture supersedes the bake with no further change here.
- **`TEMPLATE_BASE_MODEL`** is now one exported constant that both the bake and the runtime capture import, so the two paths cannot drift. A caller-supplied `ANTHROPIC_MODEL` still wins — that's how the bake captures each variant.
- **Legacy `system_prompt_fable`** is still read via `promptVariantsOf()`, so an older live cache already on a user's disk keeps working. The bake no longer writes it (no point duplicating ~9KB in the bundle).

## Verification

Selection, against the real bundle:

```
base              10006  base          <- pre-pin cache, deliberately left in play for this run
claude-opus-4-8   10006  base
claude-opus-5      8110  opus-5 variant
claude-opus-5[1m]  8110  opus-5 variant
claude-sonnet-5   13442  sonnet-5 variant
claude-fable-5     9222  fable variant
claude-sonnet-4-6 10006  base
claude-haiku-4-5  10006  base
claude-opus-50    10006  base          <- `-5` match is bounded
```

Then with a **pinned** fresh capture, both halves together:

```
fresh live capture base chars: 6764      <- was 10006 unpinned
base             6764
fable variant    9222 (intact)
opus-5 variant   8110 (intact)
sonnet-5 variant 13442 (intact)
```

That's the whole fix in one frame: deterministic base, and three variants surviving a live capture that previously would have flattened all of them to the base.

- `npm test` — **119/119** (new `test/prompt-variants.mjs`: legacy fold, map precedence, merge direction, pin sanity)
- `capture-and-bake --check` correctly reported the two new variants as drift (`0 → 8110`, `0 → 13442`) before the bake, exercising the generalized drift path
- `per-model-system-prompt.mjs` extended: 36 assertions incl. bounded `-5` matching (`opus-50` → base) and fable winning over the `-5` arms

**Two notes on assertions I got wrong on the way**, since they're instructive for anyone extending this: the self-naming line (*"powered by the model named Opus 5"*) is in the raw capture but **stripped by the scrubber**, so it can't be a marker; and *"You are an interactive agent..."* is shared by the base and every variant, so it can't be either. The markers used are section headers that survive scrubbing and are genuinely exclusive (`# Delivering work`, `# Doing tasks`).

## Not changed

`_schemaVersion` stays at 3. The addition is additive and the reader tolerates its absence in either direction (new code + old cache via the legacy fold; old code + new bundle can't happen, since the bundle ships with the code).

**Merging this ships it** (5.3.1 → 5.4.0 — minor, because what dario sends on the wire changes for two model families).

**Flaky test, still disclosed:** a lone unattributable `# fail 1` has now appeared **three** times this session against ~15 runs, never reproducing on re-run and never with a captured subtest name. First occurrence was on a docs-only tree, so it predates all of this. Correlation noted in #851: every occurrence closely followed live-proxy activity on the same box, which fits a `TIME_WAIT` collision under `--test-concurrency=8`. Unproven.


sprayberry-redline (dismissed):

**Verdict:** Clean — no blocking issues found; the fix and its tests are solid.

## What I checked
- `gh pr diff` in full (9 files, +284/-? across CHANGELOG, package metadata, `scripts/capture-and-bake.mjs`, `src/cc-template-data.json`, `src/cc-template.ts`, `src/live-fingerprint.ts`, and two test files).
- CI rollup: all required checks (`CodeQL`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `actionlint`, `analyze`, `triage`, `validate-package-json`) pass; `compat`/`wire-drift` were still pending at review time and `auto-merge` is skipping — none of these are blocking findings, just not yet resolved.

## Correctness walkthrough
- `src/cc-template.ts:159-163` bounds the new family match with `/opus-5(?!\d)/` / `/sonnet-5(?!\d)/`, and the fable branch is checked first — verified against the added test cases (`claude-opus-50 → base`, `claude-sonnet-51 → base`, `claude-fable-5` still wins over the `-5` arms). The lookahead correctly excludes a hypothetical `opus-50` while still matching `claude-opus-5[1m]` (next char `[` isn't a digit).
- `scripts/capture-and-bake.mjs:75-124`: the refactor from a single `fableVariant` to a `variants` map preserves the original fallback semantics per branch — capture failure, scrub-residual, and catch-error paths all call `keepPrevious()` to carry the prior baked variant forward; the "matches base" path intentionally does not (no variant is needed when it matches base), which mirrors the pre-existing behavior.
- `src/live-fingerprint.ts:220-227` (`withBundledVariants`): merge order is `{ ...promptVariantsOf(bundled), ...promptVariantsOf(live) }`, so a live-captured variant wins over the bundle — matches the stated intent ("future per-model live capture supersedes the bake") and is covered by `test/prompt-variants.mjs`'s "a variant present on the live template wins over the bundle" case.
- `promptVariantsOf` correctly folds the legacy `system_prompt_fable` slot in under `fable` only when the map doesn't already have that key, and only when the legacy string is non-empty — preserves the old `.length > 0` guard that used to gate `CC_SYSTEM_PROMPT_FABLE`.
- The dropped `sonnet-5 → base` assertion in `test/per-model-system-prompt.mjs` is not a coverage regression — it's superseded by the new `sonnet-5 → sonnet-5 variant` assertion added right below in the same file, consistent with the behavior change.

## What's good
- Root-caused three real, related defects (missing per-model variants, live-cache clobbering the bake, machine-dependent base capture) with concrete before/after char counts, and locked each down with a dedicated test (`test/prompt-variants.mjs` is new, `test/per-model-system-prompt.mjs` extended).
- `TEMPLATE_BASE_MODEL` is now a single exported constant imported by both the bake and the runtime capture, closing off the drift class described in the PR body.

askalf:
Rebased onto master (merge, not rebase) after the auto-rebake bot landed **#852 (v5.3.2)**. Three conflicts, resolved as:

- **`package.json`** → `5.4.0` (ahead of the bot's 5.3.2).
- **`CHANGELOG.md`** → both entries kept, 5.4.0 above 5.3.2.
- **`src/cc-template-data.json`** → **re-baked from scratch**, not hand-merged. A template bundle is a captured artifact; merging two different capture sessions line-by-line would produce a bundle that never came off any real CC. And keeping the bot's version would have shipped this PR **inert** — its bundle has no `system_prompt_variants`, so every model would have fallen back to the base.

Re-baked bundle: base 4776, fable 9222, opus-5 8110, sonnet-5 13442. Suite 119/119 on the merged tree.

---

## ⚠️ What the conflict exposed: the bake is not reproducible across machines

The bot's #852 bundle came back with **base 4754 / fable 9200** — which are exactly the pre-#851 numbers. My bakes on this box consistently produce **4776 / 9222**. Same CC version (2.1.220), same pinned capture model, and a consistent **22-char** delta in both the base and the fable variant.

So CI and this box capture measurably different prompts, and the drift-watch treats the difference as drift in whichever direction it last saw. That means after this PR lands, the next scheduled `capture-and-bake --check` on CI will very likely flag "drift" again and auto-rebake back to 4754 — and then a local bake flips it forward. **A rebake ping-pong that generates a release each way.**

Two things worth separating:

- This PR's base-model pin (`TEMPLATE_BASE_MODEL`) removes one axis of that variance — the captured base no longer depends on the operator's configured default model, which was a 3,000+ char swing on this box. Real improvement, and independent of the 22 chars.
- It does **not** explain the remaining 22. Candidates I have not chased: a remote-config flag that differs per environment (the same class as the `afk-mode` flip in #851), or something host-shaped surviving the scrub differently. Both would be a bug in "the bundle is a canonical artifact" rather than in this PR.

I'd rather flag this than let it churn quietly. If the ping-pong shows up, the cheap mitigation is to make the drift check tolerant of a small system-prompt delta (or pin the bake to one environment); the real fix is finding the 22 chars. Happy to take that as a follow-up — it wants a diff of the two captures side by side, which needs a CI capture artifact I don't have locally.

sprayberry-redline (approved):

## Context
This PR's own prior review (by this same bot, at commit `81a75f9`) was `DISMISSED` — GitHub does this automatically when new commits land on a reviewed PR. I diffed that reviewed commit against the current head (`fd3cc1e`) directly: the only changes since are a merge-conflict resolution in `CHANGELOG.md`/`package.json` (identical content, just re-applied against a newer `master`) and a re-capture timestamp bump in `src/cc-template-data.json` (`_captured` changed, no prompt content changed). `src/proxy.ts` and `test/proxy-400-recovery.mjs` only show up in the branch-to-branch compare because they came in from `master` via the merge commits — they are **not** part of this PR's diff against its base and are out of scope here. So this is a fresh, from-scratch review of the same substantive change.

## What I checked
- `gh pr diff` in full (9 files: `CHANGELOG.md`, `package.json`/`package-lock.json`, `scripts/capture-and-bake.mjs`, `src/cc-template-data.json`, `src/cc-template.ts`, `src/live-fingerprint.ts`, and two test files).
- CI rollup (`gh pr view --json statusCheckRollup`): every check is `SUCCESS` — `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `actionlint`, `analyze` (CodeQL), `compat`, `wire-drift`, `validate-package-json`. `auto-merge` is `SKIPPED` (not blocking, this isn't a bot PR).

## Correctness walkthrough
- `src/cc-template.ts:161-162`:
  ```
  if (/opus-5(?!\d)/.test(m)) return CC_SYSTEM_PROMPT_OPUS5;
  if (/sonnet-5(?!\d)/.test(m)) return CC_SYSTEM_PROMPT_SONNET5;
  ```
  The negative lookahead correctly excludes `claude-opus-50` (next char after `opus-5` is a digit → falls through to base) while still matching `claude-opus-5[1m]` (next char is `[`, not a digit). The `fable` branch is checked first, so a hypothetical `fable-5` model can't be swallowed by the `-5` arms. Backed by the new assertions in `test/per-model-system-prompt.mjs` (`opus-50 → base`, `sonnet-51 → base`, `fable-5 still wins`).
- `src/live-fingerprint.ts` — `promptVariantsOf()` folds the legacy `system_prompt_fable` slot under `fable` only when the map doesn't already carry that key and the legacy string is non-empty (preserves the old `.length > 0` guard). `withBundledVariants()` merges as `{ ...promptVariantsOf(bundled), ...promptVariantsOf(live) }` — live wins over bundle, matching the stated "future live capture supersedes the bake" intent, and covered by `test/prompt-variants.mjs`'s "live template wins over the bundle" case.
- `scripts/capture-and-bake.mjs` — the refactor from a single `fableVariant` to a `variants` map/loop preserves the three original fallback paths (capture failure, scrub-residual, catch) all calling `keepPrevious()` to carry the prior baked variant forward, while the "matches base" path intentionally stores nothing — consistent with pre-existing per-model semantics, just generalized.
- `test/per-model-system-prompt.mjs` drops the old `sonnet-5 → base` assertion, but it's superseded in the same diff by `sonnet-5 → sonnet-5 variant` further down — not a coverage regression, just an updated expectation matching the behavior change.
- `src/cc-template-data.json` change is only the `_captured` timestamp plus the (unshown-in-diff-summary but expected) variant map payload — consistent with a rebake, no logic in this file.

## What's good
- Three related defects (missing per-model variants, live-cache clobbering the bake, machine-dependent base capture) are each locked down with a dedicated, well-targeted test (new `test/prompt-variants.mjs`, extended `test/per-model-system-prompt.mjs`).
- `TEMPLATE_BASE_MODEL` is now a single exported constant imported by both the bake and the runtime capture, closing the drift class the PR describes.
- CHANGELOG and version bump (5.3.2 → 5.4.0) accurately reflect the wire-behavior change for two model families.

No blocking findings — approving.

PR #855, opened 2026-07-25
fix(proxy): make 400 recovery multi-pass, not one-shot

askalf:
Fixes **Finding 2** from #851. Companion to #854 (Finding 1), kept separate because this touches the proxy's hottest path and should be reviewable and revertable on its own.

## The bug

Each of dario's five 400 remediations — effort level, effort param, max_tokens cap, `anthropic-beta` flag, long-context — retried **inline, exactly once**, then fell out of the `if / else if` chain. A model that trips *two* of them gets one fixed and the other forwarded to the client.

`opus-4-1` trips two: it predates `output_config.effort` **and** caps `max_tokens` at 32000. Reproduced on a fresh proxy with `--verbose`:

```
#0 effort parameter unsupported by claude-opus-4-1-20250805 — retrying without output_config.effort
#1 400          <- client sees "max_tokens: 64000 > 32000"
```

The effort branch consumed the single retry; the max_tokens rejection went straight out. Request 2 then succeeded, because by then effort was cached-stripped up front so max_tokens got the retry — which is why this looked like a warm-up quirk rather than a bug. But `opus-4-1` is advertised on `/v1/models`, so a client picker offering it fails once per proxy lifetime, and the code comment claimed the retry made it transparent.

## The fix

Mutate `finalBody` / `headers` as before, then **`continue dispatchLoop`** instead of fetching inline. The loop head already re-sends the request and re-runs pool accounting, and the new response re-enters the same recovery chain — so successive distinct rejections are each remediated inside one client request. `MAX_RECOVERY_PASSES = 4` bounds it (opus-4-1 needs 2) so a rejection dario can't actually fix can't spin.

This **deletes five duplicated `fetch` + pool-accounting blocks** — net **−39 lines** (45 added, 84 removed). The two header-based remediations now mutate `headers` in place instead of building a local `retryHeaders`, which is what lets the loop head pick the change up (it re-derives `outboundHeaders` from `headers` every pass).

Re-dispatching rather than adding an inner loop also means pool accounting, sticky-session rebinding, and the 429 failover path keep their single existing implementation instead of gaining a second copy.

## Verification

**Acceptance — the exact reproduction, now on one request:**

```
#0 effort parameter unsupported by claude-opus-4-1-20250805 — retrying without output_config.effort
#0 max_tokens 64000 exceeds claude-opus-4-1-20250805 cap 32000 — retrying clamped
#1 200
```

Both remediations fire under the same request id (`#0`), and the client's **first** call returns 200 with `"OK"`.

- `npm test` — 118/118
- **live E2E — 17/17** against a repo-build proxy through the refactored path (streaming, tool use, OpenAI-compat, per-model obedience, rate-limit headers)

## Verification gap, stated plainly

**There is no hermetic regression test for the multi-pass sequencing**, and I did not add one. The recovery chain lives inside the request handler and talks to a hardcoded `ANTHROPIC_API` const; testing it in-process needs either an upstream-base override or extraction of the handler. Adding an env var that redirects all upstream traffic, purely for testability, is a production footgun I'm not willing to introduce as a side effect of a bug fix.

The durable answer is dependency injection of the upstream base / fetch for this handler — which also matches the standing preference for injectable deps over live smoke. That's a self-contained follow-up; happy to do it next if you want the coverage locked in. Until then this fix rests on the live acceptance test above, which is decisive but manual.

**No version bump — deliberately.** #854 is still open and bumps to 5.4.0 from the same base; a second bump here would guarantee the same `package.json` / `CHANGELOG` conflict I hit on #846 today. Merge order can stay free, and whichever release follows carries this. Say the word if you'd rather I bump it once #854 lands.


askalf:
**The verification gap in the PR body is now closed** — pushed as a second commit rather than left as a follow-up.

`ProxyOptions.fetchImpl` injects the fetch used for upstream calls, mirroring `CatalogDeps.fetchImpl` in `model-catalog.ts`. Still an option and **not** an env var: an env var that silently redirects upstream traffic is a footgun on a proxy holding someone's subscription credentials. The loopback self-health check is deliberately not routed through it (it isn't upstream); the presence ping is.

Worth noting the refactor paid for itself here — **the seam is one line because there's now one upstream call site.** Before it there were six (five inline retries plus the loop head), each needing to be threaded.

`test/proxy-400-recovery.mjs` is fully hermetic: scripted fake upstream, `upstreamApiKey` for x-api-key mode, and `noClaudeAuth` so the pool is never loaded — dario's own startup line confirms *"the Claude refresh token is never touched."* Three blocks, 14 assertions:

1. **The exact bug.** effort-param 400 → max_tokens 400 → 200. The client's *first* request gets 200, upstream saw 3 calls, and each pass's body is asserted individually: pass 2 dropped `effort` but still carried `max_tokens: 64000`; pass 3 clamped to 32000 with `effort` still gone. Before the fix this returned the 400.
2. **The bound.** A rejection reporting an ever-lower cap, so every pass is genuinely remediable — the chain stops at exactly `1 + MAX_RECOVERY_PASSES` upstream calls and forwards the 400 rather than spinning.
3. **No speculative retries.** A non-remediable 400 is forwarded on the first pass, exactly one upstream call, error text preserved.

One thing that cost a debugging round and is now commented in the test: each block needs its **own model id**, because `effortSupportByModel` / `maxTokensCapByModel` are module-level and persist across requests by design. Sharing an id lets block 1's cached cap pre-clamp block 2 up front, and block 2 then silently stops exercising the bound (it saw 1 upstream call instead of 5) while still "passing" its first assertion.

Suite is **119/119**.

---

**Correction on the flaky test, and it's a correction against myself.** I finally caught it: `not ok 44 - issue-29-tool-translation.mjs`.

That **disproves the `TIME_WAIT`/port-collision hypothesis** I recorded on #851 and #854. That file binds no port, starts no server, and does no I/O — it's a pure tool-translation unit test. It passes 3/3 in isolation and only fails under `--test-concurrency=8`, so it's contention of some other kind (timing under load, or shared state through the driver's subprocess spawn). Three further full runs after the catch were clean, so I have a name but not a root cause. Please disregard the port-collision theory in the other two PRs.

sprayberry-redline (changes requested):

**Verdict:** one correctness bug in the new multi-pass recovery chain — the two header-mutating branches can stomp each other's fix when both fire in the same request. Everything else (the `continue dispatchLoop` refactor, the pass bound, the new test) reads well.

## Findings

### 1. [Correctness] Stale `beta` read lets a later header-mutating recovery pass undo an earlier one — `src/proxy.ts:3157`/`3161` vs `3313`/`3317`

The anthropic-beta-rejection branch:
```
const reducedBeta = beta.split(',').filter((t) => t.length > 0 && !set!.has(t)).join(',');
...
headers['anthropic-beta'] = reducedBeta;
...
continue dispatchLoop;
```
and the long-context branch:
```
const reducedBeta = beta.split(',').filter((t) => !LONG_CONTEXT_BETAS.has(t)).join(',');
...
headers['anthropic-beta'] = reducedBeta;
...
continue dispatchLoop;
```
Both compute `reducedBeta` from the **outer, request-scoped `beta` string** (fixed before `dispatchLoop` starts), not from the *current* `headers['anthropic-beta']` that the previous pass just wrote. Under the old code this was safe — the if/else chain only ever ran once per request, so only one of these branches could fire. Under this PR's new multi-pass model (`continue dispatchLoop`), both branches can fire in sequence within the *same* request, and each recomputes from the stale original `beta`, discarding whatever the other branch already stripped.

**Concrete scenario:** a request carries an experimental `anthropic-beta` flag `X` the account doesn't have, *and* `context-1m`, which the same account also can't use.
- Pass 1: upstream 400s on flag `X` → `betaRejectedFlags` branch fires, `set` caches `X`, `headers['anthropic-beta']` becomes `beta` minus `X`. `continue dispatchLoop`.
- Pass 2: upstream now 400s on `context-1m` → `isLongContextError` branch fires, computing `reducedBeta = beta.split(',').filter(t => !LONG_CONTEXT_BETAS.has(t))` from the **original** `beta` — which still contains `X`. `headers['anthropic-beta']` is overwritten back to a string that reintroduces `X`.
- Pass 3: upstream 400s on `X` again (wasted round trip re-fixing something already fixed). `betaRejectedFlags` fires again; `set` already has `X` cached so it correctly strips it this time.

With `MAX_RECOVERY_PASSES = 4` this particular case still resolves (3 of 4 passes used), but it burns an extra round-trip the PR's own design is meant to avoid, and a request stacking a *third* distinct rejection (e.g. also an effort or max_tokens clamp) on top of this pair can exhaust the bound and get a spurious 400 forwarded to the client — the exact failure this PR sets out to fix.

```suggestion
// Derive from the currently-active beta string, not the request-scoped
// original, so a later recovery pass doesn't undo an earlier one:
const activeBeta = headers['anthropic-beta'] ?? beta;
const reducedBeta = activeBeta.split(',').filter((t) => !LONG_CONTEXT_BETAS.has(t)).join(',');
```
(and the mirror change in the `betaRejectedFlags` branch, reading from `headers['anthropic-beta']` instead of `beta`).

## Test coverage
The new `test/proxy-400-recovery.mjs` covers the effort+max_tokens double-trip well (both mutate `finalBody`, not `headers`, so they don't hit the bug above), plus the pass-bound and non-remediable-400 cases. It does not cover a case where **two header-mutating** branches (anthropic-beta-rejected + long-context) fire in the same request — which is exactly the combination that triggers the bug above and would have caught it.

## What's good
- The `continue dispatchLoop` refactor is a real simplification: it deletes five duplicated fetch+pool-accounting blocks and gives pool accounting/sticky-session rebinding a single implementation instead of a second copy per remediation branch, matching the PR's stated goal.
- `MAX_RECOVERY_PASSES` is a sane, well-justified bound and the new test explicitly exercises it (5 calls = 1 + 4, then forwards the 400).
- `fetchImpl`/`upstreamFetch` is a clean DI seam, deliberately not env-var-based, with a clear rationale in the comment (credential-redirect footgun).
- CI: build/actionlint/validate-package-json are green; `analyze`, `test`, `compat`, `docker-cap-drop-smoke`, `wire-drift` were still pending at review time — no failing checks observed.

## What I checked
`gh pr diff`, the full changed block in `src/proxy.ts` (dispatch loop, all five recovery branches, `upstreamFetch` wiring) via a shallow clone, the new test file in full, and the CI check-rollup. I did not run the suite locally (per policy — CI is the source of truth here) and did not trace call sites of `startProxy`/`fetchImpl` outside this diff.

askalf:
**Finding 1 confirmed and fixed — you were right, and it was a bug this PR introduced.**

I verified the mechanism before accepting it: `beta` is assigned at `proxy.ts:2923–2972`, all **before** `dispatchLoop`, and never reassigned inside it. Safe while the chain ran once per request; not safe once `continue dispatchLoop` lets both header-mutating branches fire in sequence.

Then I reproduced it on the wire. Rejecting `advisor-tool-2026-03-01` (a flag dario really sends, and not a long-context beta), followed by a long-context rejection:

| pass | `advisor-tool-2026-03-01` |
|---|---|
| 1 | present |
| 2 | **stripped** — the beta-rejection branch worked |
| 3 | **back again** — the long-context branch recomputed from the original `beta` |

Both branches now derive from `headers['anthropic-beta'] ?? beta`. Your suggested shape, applied to both sites.

**Regression test added for exactly the gap you identified** — the header+header combination. You were right that the existing blocks couldn't have caught it: both mutate `finalBody`, not `headers`.

Worth recording that **the test passed against the broken build twice** before it was actually testing anything, because I got it wrong in two ways:

1. **Rejecting a flag dario never sends.** My first version rejected `experimental-flag-x`, which isn't in the outbound set — so there was nothing to reintroduce and the assertion was vacuous. The rejected flag has to be one that's genuinely on the wire.
2. **`orderHeadersForOutbound` returns an array of `[k, v]` tuples, not an object.** Reading it as an object gave `undefined`, and `!(undefined ?? '').includes(x)` is `true`, so every beta assertion passed while asserting nothing. The fake now handles both shapes and matches case-insensitively.

I only caught both because I insisted on seeing the test fail against the unfixed build first. Red-then-green earned its keep here.

Suite is 119/119. Also fixed the CI-vs-local template churn this surfaced — see the #854 thread.

sprayberry-redline (approved):

**Verdict:** the correctness bug from my prior CHANGES_REQUESTED review is fixed and covered by a new targeted regression test — approving.

## Re-review context

This is a re-review of head `8c4b223`. My prior review (submitted 2026-07-25T17:23:12Z, commit `2a549df`) flagged one correctness bug: the `anthropic-beta`-rejection branch and the long-context branch both recomputed `reducedBeta` from the request-scoped, never-reassigned `beta` string instead of the currently-active header, so under the new multi-pass `continue dispatchLoop` model, a second header-mutating pass could silently undo the first pass's fix.

## Verification of the fix

Both branches now read from the live header instead of the stale original:

`src/proxy.ts` (anthropic-beta-rejected branch):
```
const activeBeta = headers['anthropic-beta'] ?? beta;
const reducedBeta = activeBeta.split(',').filter((t) => t.length > 0 && !set!.has(t)).join(',');
...
headers['anthropic-beta'] = reducedBeta;
```

`src/proxy.ts` (long-context branch):
```
const activeBeta = headers['anthropic-beta'] ?? beta;
const reducedBeta = activeBeta.split(',').filter((t) => !LONG_CONTEXT_BETAS.has(t)).join(',');
...
headers['anthropic-beta'] = reducedBeta;
```

Tracing the scenario from my original finding (a request that trips both the beta-flag rejection and the long-context rejection) through this code: pass 1 strips the rejected flag and writes it into `headers['anthropic-beta']`; pass 2 now derives `activeBeta` from that already-reduced header (not the original `beta`), so stripping the long-context token no longer reintroduces the flag pass 1 removed. The bug is gone.

## Test coverage

`test/proxy-400-recovery.mjs` adds a new case, `'two HEADER-mutating rejections do not undo each other (reviewer catch, #855)'`, that scripts exactly this sequence (flag rejection on call 1, long-context rejection on call 2) and asserts:
- resolution in 3 upstream calls with no wasted re-fix pass,
- pass 3's beta header contains neither the originally-rejected flag nor `context-1m-2025-08-07`.

This is the precise regression the original bug needed and would have caught it.

## CI

`build (18/20/22)`, `test`, `actionlint`, `validate-package-json`, `wire-drift`, and `docker-cap-drop-smoke` are all green. `analyze` (CodeQL) and `compat` were still in progress at review time — no failing checks observed.

## What's good
- Fix is minimal and targeted: only the two `reducedBeta` derivations changed, no unrelated churn.
- The new test is scoped exactly to the fixed defect and uses the real header path rather than a synthetic shortcut (comment calls out that the rejected flag must be one dario actually sends, to avoid a vacuous pass).
- Explanatory comments at both call sites make the invariant (`headers['anthropic-beta']` is the live value, `beta` is fixed at request scope) explicit for future readers.

## What I checked
`gh pr view`/`gh pr diff` for the current head (`8c4b223`), the full diff of both header-mutating branches, the new test case in `test/proxy-400-recovery.mjs`, and the CI check-rollup. I did not re-run the suite locally (per policy) and did not re-verify the effort/max_tokens branches beyond confirming they're unchanged from my prior pass (they mutate `finalBody`, not `headers`, so they were never in scope for this bug).

PR #849, opened 2026-07-25
chore(template): label refresh _version -> v2.1.220 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-25T07:08:54+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.220** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.220`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.220`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.2.21` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.220. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #840, opened 2026-07-23
chore(template): label refresh _version -> v2.1.218 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-23T19:08:34+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.218** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.218`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.218`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.2.18` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.218. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #828, opened 2026-07-21
chore(template): label refresh _version -> v2.1.216 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-21T07:26:53+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.216** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.216`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.216`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.2.15` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.216. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #820, opened 2026-07-19
chore(template): label refresh _version -> v2.1.215 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-19T07:21:24+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.215** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.215`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.215`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.2.13` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.215. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #856, opened 2026-07-25
fix(template): stop replaying the bake host's OS/arch onto every request (v5.3.3)

askalf:
## The bug

`header_values` captured `x-stainless-os` / `x-stainless-arch`, and `proxy.ts` overlaid `header_values` on top of the values it had **just computed correctly** for the running process — so the captured value won:

```js
'x-stainless-os': OS_NAME,          // correct for THIS process… (proxy.ts:1669)
…
staticHeaders[k] = v;               // …then clobbered by the bake host (proxy.ts:1693)
```

The bundle had been baked on **Windows** for several releases (#820 / #828 / #840 / #849 / #851), so the **Linux production box announced `x-stainless-os: Windows` on every upstream call.** For a proxy whose entire purpose is wire fidelity, the fallback template was itself the tell.

## And it drove a rebake ping-pong

`cc-drift-template-watch` captures live on the **Linux** runner every 30 min and diffs against the bundle. A Windows-baked bundle therefore read as permanent, unclearable drift:

| | captured | OS | tools | PowerShell |
|---|---|---|---|---|
| #851 (was in bundle) | 15:56Z | **Windows** | 30 | yes |
| #852 (auto-rebake) | 16:43Z | **Linux** | 27 | no |
| #854 (open PR) | 16:48Z | **Windows** | 30 | yes |

It auto-rebaked to Linux (#852), and the next Windows-side bake (#854) would have flipped it straight back — forever. **Neither capture was wrong.** These two keys just have no business being baked: they describe the bake host, not CC's wire shape.

## Fix — both sides, mirroring the `x-api-key` precedent from v3.19.2

- `extractStaticHeaderValues` no longer **stores** them → new captures are clean.
- The overlay no longer **replays** them → every already-baked template and warm cache **self-heals without waiting for a re-bake**.

The overlay moved out of `proxy.ts` into `overlayTemplateHeaderValues()` in `cc-template.ts` as a pure function — same rationale as `orderHeadersForOutbound`: testable without standing up the proxy.

The two keys are also stripped from the shipped bundle here, so the watch converges immediately rather than needing one more rebake to settle.

`header_order` still lists both keys. The **order** CC emits headers in is wire shape and is not host-specific; only the captured **values** are dropped, and the proxy sets them per-process exactly as before.

## Verification

Ran **#852's real Linux-captured headers** (from the Hetzner box) through the fixed extractor:

```
in : accept, anthropic-version, user-agent, x-app, x-stainless-arch=x64,
     x-stainless-lang, x-stainless-os=Linux, x-stainless-package-version, …
out: accept, anthropic-version, user-agent, x-app,
     x-stainless-lang, x-stainless-package-version, …

host keys dropped : true
CC keys preserved : true
```

A Linux bake and a Windows bake now produce **identical** `header_values` — which is precisely the property that ends the ping-pong.

**A live capture on the box was deliberately not run.** Spawning `claude` there can rotate the fleet's CC OAuth credential, which has caused incidents before, and the same proof was available from data already captured.

| | |
|---|---|
| `test/template-header-overlay.mjs` (new) | **13 assertions** — Windows-baked bundle vs Linux runtime, arm64 mac vs x64-baked bundle, case-insensitive keys, `x-api-key` regression, degenerate inputs |
| `test/live-fingerprint.mjs` | **+5** — host keys excluded from capture, **and** `x-stainless-lang` / `runtime` / `runtime-version` still captured so the fix can't over-reach |
| full suite | **119/119 green** |

## Relationship to #854

#854 is currently conflicting against master (it and #852 both touch `CHANGELOG.md`, `package.json`, `src/cc-template-data.json`). This PR does **not** resolve that — it removes the reason the conflict was contentious. Once this lands, whichever host bakes the template no longer changes these keys, so #854 can be resolved on its actual merits: keep its `system_prompt_variants` map, and re-bake the variants on Linux so the prompt bodies match production.


sprayberry-redline (dismissed):

**Verdict:** No blocking issues found — clean, well-scoped fix with good regression coverage. Approving.

## What I checked
- `gh pr diff` for all 8 changed files.
- Cloned the head branch read-only to confirm surrounding context in `proxy.ts` (lines ~1657-1701), `live-fingerprint.ts` (`STATIC_HEADER_EXCLUDE`, ~line 710-750), and the bundled `cc-template-data.json`.
- CI rollup: `test`, `build (18/20/22)`, `CodeQL`, `analyze`, `wire-drift`, `docker-cap-drop-smoke`, `actionlint`, `validate-package-json`, `triage` all `SUCCESS`. Only `compat` is still pending (no conclusion yet) — not a red flag, just unresolved at review time.

## Findings
None. Specifically verified:
- **Extraction side** (`src/live-fingerprint.ts:730-731`): `'x-stainless-os'` and `'x-stainless-arch'` are added to `STATIC_HEADER_EXCLUDE`, alongside the existing `x-api-key` precedent — new captures will no longer store the bake host's OS/arch.
- **Replay side** (`src/cc-template.ts:290-320`, new `overlayTemplateHeaderValues`): skips the same two keys (plus `x-api-key`) via `NEVER_REPLAY_HEADER_VALUES`, using `k.toLowerCase()`. Since `extractStaticHeaderValues` already lowercases stored keys (`out[lk] = v`), the lookup is consistent — no case-mismatch gap.
- **Call site** (`src/proxy.ts:1687-1701`): `staticHeaders` sets `x-stainless-arch: arch` / `x-stainless-os: OS_NAME` correctly for the running process *before* the overlay runs, and the overlay now mutates in place while skipping those two keys, so the runtime-correct values survive the overlay for both fresh and cached templates. Matches the PR's stated fix precisely.
- **Bundled data** (`src/cc-template-data.json`): the two keys are removed from the shipped `header_values`, consistent with the "self-heal without a re-bake" claim — no stale Windows values ship in this PR.
- **`header_order`** is untouched, matching the PR's explicit statement that only captured values (not wire order) are host-specific.
- **Tests**: new `test/template-header-overlay.mjs` exercises the pure `overlayTemplateHeaderValues` function directly (Windows-baked bundle vs Linux runtime, arm64 vs x64-baked, case-insensitivity, `x-api-key` regression, degenerate inputs); `test/live-fingerprint.mjs` gains assertions that the two host keys are excluded from capture while the CC-determined stainless keys (`lang`/`runtime`/`runtime-version`) are still captured. This is exactly the coverage the changed logic needs, and CI confirms both suites pass.

## Minor
None worth blocking on — the changelog and version bump are mechanical and consistent with the diff.

## What's good
- The fix addresses both sides of the bug (capture and replay) so already-baked bundles/caches self-heal immediately rather than needing a re-bake, which is explicitly verified with a real captured-header run in the PR description.
- Extracting `overlayTemplateHeaderValues` as a pure function mirrors the existing `orderHeadersForOutbound` pattern and makes the skip logic unit-testable without standing up the proxy — good reuse of an established idiom in this codebase.
- New tests target the exact regression (host-key leakage) and pin that non-host stainless keys still get captured, guarding against over-reach.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped fix with strong regression coverage. Approving.

## Context
This PR's head moved (merge of master/v5.4.0 into the fix branch, commit `9206d2a`) after an earlier
approving review on `4da5fcc` was dismissed by the push. I re-reviewed against the current head. The
merge only pulls in unrelated `master` changes (per-model system prompt variants); it does not touch
the header-overlay fix, and `gh pr diff` against the current base is unchanged in substance.

## What I checked
- `gh pr diff 856` (8 files, +216/-9) against the current head `9206d2a`.
- `gh pr checks` — full rollup is now green, including `compat`, which was still pending on the prior pass.
- Cloned the branch read-only to confirm the fix's three sites still line up after the merge:
  `src/live-fingerprint.ts:818-820`, `src/cc-template.ts:294-327`, `src/proxy.ts:1687-1701`.

## Findings
None. Specifically verified against the current head:
- **Extraction** (`src/live-fingerprint.ts:818-820`): `'x-stainless-os'` / `'x-stainless-arch'` are in
  `STATIC_HEADER_EXCLUDE`, so `extractStaticHeaderValues` never stores them — new captures are clean.
- **Replay** (`src/cc-template.ts:308-312`, `overlayTemplateHeaderValues`): iterates `header_values` and
  skips both keys (plus `x-api-key`) via `k.toLowerCase()` against `NEVER_REPLAY_HEADER_VALUES`. Since
  `extractStaticHeaderValues` stores keys already lowercased, the lookup is consistent — no case gap in
  the actual data path, and `test/template-header-overlay.mjs` additionally proves the case-insensitive
  skip holds even for a hypothetical uppercase template key.
- **Call site** (`src/proxy.ts:1687-1701`): `staticHeaders` sets `x-stainless-arch`/`x-stainless-os` from
  the live process (`arch`, `OS_NAME`) *before* `overlayTemplateHeaderValues(staticHeaders, CC_TEMPLATE.header_values)`
  runs, and the overlay skips exactly those two keys — so the runtime-correct values survive the overlay
  for both a fresh v2 template and an already-baked/stale one. Matches the stated fix.
- **Bundled data** (`src/cc-template-data.json`): both keys removed from `header_values`; `header_order`
  (line ~1406-1413) still lists them, consistent with the PR's claim that only captured *values*, not wire
  *order*, are host-specific.
- **Tests**: `test/template-header-overlay.mjs` (new, 13 assertions) directly exercises the pure overlay —
  Windows-baked bundle vs Linux runtime, arm64 mac vs x64-baked bundle, case-insensitivity, the `x-api-key`
  regression, and degenerate inputs (`undefined`/`{}`/new key). `test/live-fingerprint.mjs` gains 5
  assertions confirming the two host keys are excluded from capture while `x-stainless-lang`/`runtime`/
  `runtime-version` are still captured, guarding against over-reach. CI's `test` job is green.
- **CI rollup**: `test`, `build (18/20/22)`, `CodeQL`, `analyze`, `wire-drift`, `docker-cap-drop-smoke`,
  `actionlint`, `validate-package-json`, and `compat` all pass on the current head; only `auto-merge` shows
  `skipping`, which is expected pre-approval gating, not a failure.

## What's good
- Fixes both sides of the bug (stop storing + stop replaying) so already-baked bundles and warm in-memory
  caches self-heal immediately rather than needing a rebake — verified in the PR description against a
  real captured Linux header set.
- Extracting `overlayTemplateHeaderValues` as a pure function mirrors the existing `orderHeadersForOutbound`
  pattern, making the skip logic unit-testable without standing up the proxy.
- New tests target the exact regression (host-key leakage) and pin that legitimate CC-determined stainless
  keys are still captured/replayed, so the fix can't silently over-reach.

PR #857, opened 2026-07-25
auto-rebake: template drift detected 2026-07-25

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-25T18:25:49+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.4.2` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — worth a closer read

- **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.220) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.220, 27 tools, 6180 char system prompt
[bake] scrubbed:
[bake]   tools: 27 → 27 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6180 → 4754 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] previous baked template: CC v2.1.220 captured 2026-07-25T17:32:17.394Z, 33 tools, 4776 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4754 → fable 9200 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4754 → opus-5 8088 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4754 → sonnet-5 13448 chars
[bake] check: drift detected — 1 differing slot (verdict: moderate):
[bake] **Verdict:** 🟡 Moderate — worth a closer read
[bake] 
[...]

PR #859, opened 2026-07-25
fix(bake): re-derive tool_names after the preservation merges (v5.4.2)

askalf:
Fixes the correctness issue the fleet reviewer raised on #857.

## The bug

`tool_names` is **derived** from `tools` in both places that build a template — `scrub-template.ts:50` and `live-fingerprint.ts:757` — so equality is the contract, not a coincidence.

`capture-and-bake` then merges other-platform and interactive-only tools into `scrubbed.tools` *after* `scrubTemplate()` has already derived the names, and nothing re-synced them.

| bundle | `tool_names` | `tools` | in `tools` only |
|---|---|---|---|
| **master today** | 30 | 33 | `AskUserQuestion`, `EnterPlanMode`, `ExitPlanMode` |
| a Linux bake (#857) | 27 | 33 | + `Glob`, `Grep`, `PowerShell` |

A consumer trusting `tool_names` as the inventory concludes those tools aren't in the toolset while their full definitions sit right there in `tools`.

## One correction to the review

The reviewer framed this as something #857 introduces. It doesn't — **master already ships the 30-vs-33 form**. #857 only widens it to 27-vs-33, because a Linux capture is missing more tools for platform-preservation to re-add. So this fixes a live inconsistency rather than preventing a new one, and #857 was the messenger.

Everything else in the review was right, including that the fix belongs in the bake script rather than the generated artifact.

## Fix

- **Source:** re-derive `scrubbed.tool_names` from `scrubbed.tools` after *both* preservation merges.
- **Artifact:** the shipped `cc-template-data.json` re-synced to 33/33, so the fix isn't waiting on the next bake.

Deriving rather than hand-maintaining is the point — this is the **second** time these lists diverged. The CHANGELOG's *"Template tool-list drift from the community tool-mapping PR"* is the first, same shape: a change updated `tools` and left `tool_names` behind.

## Verification

`test/template-invariants.mjs` now asserts it against the **real bundled artifact**, both directions plus equal length. Confirmed it fails on the pre-fix bundle:

```
FAIL every tool definition has a tool_names entry:
     in tools but not tool_names: AskUserQuestion, EnterPlanMode, ExitPlanMode
FAIL the two lists are the same length: tool_names=30 tools=33
229 pass, 2 fail
```

and passes after the re-sync (231 pass, 0 fail). A test that only ever passes wouldn't tell you it was pinning anything.

Full suite **121/121**.

## Sequencing with #857

#857 (the Linux re-bake carrying the Linux-captured prompt variants) should be re-baked on top of this so its artifact comes out consistent, rather than merged as-is. Once this lands I'll re-dispatch `cc-drift-template-watch`; the bot de-dups on an open rebake PR, so #857 needs closing first for a fresh one to open.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped fix. Approving.

## What I checked
- `gh pr diff 859` — the full unified diff (6 files, +70/-3).
- CI check-rollup via `gh pr checks 859`: all 12 applicable checks green (`build`, `test`, `compat`, `wire-drift`, CodeQL, etc.); `auto-merge` is `skipping` (expected, gated separately).
- Shallow-cloned the branch to confirm the derivation contract the PR describes: `src/scrub-template.ts:50` (`cloned.tool_names = cloned.tools.map((t) => t.name)`) and `src/live-fingerprint.ts:757` (`tool_names: tools.map((t) => t.name)`) do use `tools` as the source of truth for `tool_names`, so the fix's premise — that `capture-and-bake` needs to re-derive `tool_names` after mutating `tools` — checks out.
- Confirmed `package.json`'s `build` script (`tsc && cp src/cc-template-data.json dist/`) runs before `test` in CI (`ci.yml:33`), so `test/template-invariants.mjs` reading `dist/cc-template-data.json` will see the freshly-copied, already-resynced `src/cc-template-data.json` — the new test isn't checking a stale artifact.

## Correctness
`scripts/capture-and-bake.mjs` +209-215: the fix adds
```js
scrubbed.tool_names = scrubbed.tools.map((t) => t.name);
```
placed after both the other-platform-tool merge and the interactive-only-tool merge, which is exactly where the stated bug (`tool_names` computed before, then `tools` mutated after, with nothing to re-sync) would be fixed. This is a minimal, correct fix for the described class of bug — a straightforward re-derivation rather than a hand-maintained parallel list, which also directly addresses the "second time these diverged" root cause.

## Test coverage
`test/template-invariants.mjs` new block asserts both directions (`tools` → `tool_names` and back) plus equal length, against the real bundled `dist/cc-template-data.json` rather than a synthetic fixture — good, since the bug was specifically an artifact-vs-derivation drift that a synthetic test could miss. The PR body's before/after output (`229 pass, 2 fail` → `231 pass, 0 fail`) demonstrates the test fails on the pre-fix bundle and passes after, which is the standard I'd want for a regression test.

## What's good
- `src/cc-template-data.json` is updated in the same PR to be consistent with the fix immediately, not left waiting on the next bake.
- CHANGELOG entry accurately reflects the diff and links back to the earlier occurrence of the same drift, giving good context for future readers.
- Version bump (5.4.1 → 5.4.2) and lockfile are consistent with a patch release.

No further concerns — the diff is small, the fix matches its stated cause, and it's exercised by a test that demonstrably pins the invariant against the real artifact.

PR #861, opened 2026-07-25
fix(tui): clip + scroll the Config tab so it fits the body region

jerzydziewierz:
## What does this PR do?

Fixes a rendering bug where the Config tab overflows the terminal and the top of the panel — the `Config` title and the first several fields — scrolls off the top of the screen, leaving the tab looking like it starts above the visible area.

Two independent causes, both in `src/tui/tabs/config.ts`:

**Rows are built wider than the terminal.** `valueW` was `cols - labelW - 6`, so label + value already consumed nearly the full width; `hint` was then appended *after* that padding with no truncation. Measuring the real render, the widest row is 126 / 146 / 166 / 206 chars at 80 / 100 / 120 / 160 cols. The overflow is width-independent — the hint column grows with the terminal — which is why resizing wider never helps. Fourteen of the 21 lines wrap onto a second physical row, so the body occupies 35 physical rows.

**The tab ignores its row budget.** `render(state, dimv)` accepts `dimv` but never reads `dimv.rows`; all 17 fields render unconditionally with no viewport. `renderTui` then measures body height with `body.split('\n').length` (`src/tui/tui-app.ts:282`), which counts *logical* lines and is blind to wrapping, so it sees 21 and believes it fits. It also only pads a short body (`if (bodyLines < bodyRows)`) and never truncates a long one. Since `App.redraw` writes the frame after a bare `clearScreen` with the cursor at home, the surplus rows push the head of the panel off the top of the alt-screen.

The fix clamps the value column, truncates each row to `cols` and pads it back, and windows `FIELDS` around the selection using the same arithmetic `hits.ts` already uses (`clampVisibleStart`) for its buffer. A `n / total` counter appears in the title only when the list actually overflows. A final `lines.slice(0, dimv.rows)` is a floor for terminals too short for even one field plus the prompt.

Config is the only tab with this bug; `hits.ts` already budgets correctly. Note that `09d8eec` (`pool.strategy`) added a 17th field, which makes the overflow slightly worse on current master.

I deliberately did **not** touch `tui-app.ts`. Its logical-line height check and pad-only behaviour are a latent weakness any tab could hit, but fixing it isn't needed here and would widen the diff — happy to file that separately if you'd like.

## How to test

```
npm run build
node dist/cli.js tui     # press `c`, then arrow up/down
```

Before: the title and first fields are scrolled off the top, and hint text wraps onto a second line. After: the panel starts at the top, hints are clipped at the right edge, and the list scrolls with the selection.

Automated coverage is in `test/tui-tabs.mjs` — it sweeps six terminal geometries, walks the selection through every field, and asserts that no row exceeds `cols` and the body stays within its `rows - 5` budget. Verified the test actually pins the bug: reverting the `config.ts` change fails 300 assertions, e.g. `40x8 idx=16: body within budget :: lines=21 budget=3`.

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes (offline regression test, no credentials required) — 118/118


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean fix, well tested, approving.

**What I checked**
- `gh pr diff` for `src/tui/tabs/config.ts` and `test/tui-tabs.mjs`.
- Traced the row-budget accounting across `tui-app.ts` (`renderTui` passes `{cols, rows: rows-5}` to `tab.render`) and confirmed the tab's internal `listRows = dimv.rows - 5` is a *separate* reservation for its own title/blank/prompt chrome lines (1+1+listRows+1+1or2 ≈ `dimv.rows`), not a double-subtraction bug — the total pushed lines stay within `dimv.rows`.
- Compared the new `startIdx`/`endIdx` windowing in `config.ts` against the existing `clampVisibleStart` in `src/tui/tabs/hits.ts:255` — same centering formula (`selectedIdx - floor(listRows/3)`, clamped to `[0, total-listRows]`), reused rather than reinvented.
- Read `truncate`/`pad`/`visibleWidth` in `src/tui/render.ts` — `valueW = Math.max(6, Math.min(16, w - labelW - 4))` stays positive at all tested widths (40–200 cols), and the row is both `truncate(..., w)` and `pad(..., w)`'d, so the final visible width is exactly `w`.
- CI rollup (`gh pr checks`/`statusCheckRollup`): all green — build (18/20/22), test, CodeQL, actionlint, docker-cap-drop-smoke, validate-package-json.
- New test (`test/tui-tabs.mjs`) sweeps 6 terminal geometries (down to 40x8) and walks the full field list, asserting no row exceeds `cols` and body stays within `rows-5`; PR description states reverting the fix fails 300 of these assertions, which is credible given the assertions directly target the two root causes described.

No correctness, security, reuse, or coverage issues found in the changed lines. Reasonable to approve as-is.

**What's good**
- Root-caused both independent causes (unbounded value column, no row viewport) rather than patching one symptom.
- Reused the existing `clampVisibleStart` windowing pattern from `hits.ts` instead of inventing new logic.
- Test coverage directly targets the failure mode across a realistic size matrix, not just a single geometry.

PR #863, opened 2026-07-25
fix(tui): close every SGR attribute truncate() opens

askalf:
## What does this PR do?

`truncate()` (`src/tui/render.ts`) counted visible width correctly but exited its walk the moment `visible === target` and returned `out + ellipsis` — **every escape sequence past the cut point was discarded**, including the `\x1b[22m` / `\x1b[39m` that `dim()` and `fg()` append. A cut landing inside a styled span left that attribute switched on.

The leak escaped the frame, not just the line. `App.redraw` writes `clearScreen + frame` (`src/tui/app.ts:215`) and `\x1b[2J\x1b[H` does not reset SGR; `renderFooter` ends on plain text, so nothing closed the attribute before the frame did. The dim carried into the *next* frame's header and tab strip, clearing only at the `dim('─')` separator on row 3 — and every frame re-leaked it.

**Surfaced by #861.** That fix made Config rows `truncate(… + dim(hint), w)`. At 80 columns the value column clamps to 16, so the row prefix is 44 and hints get 32 visible characters — **7 of the 14 hints clip on a default-width terminal.** The underlying bug predates #861: `renderFooter` already leaked cyan at 11 widths between 4 and 60.

`hits.ts:213` was checked and is **unaffected** — swept columns 20–160 with a populated buffer, zero leaks, because its cut always lands clear of the colored status column.

## The fix

Replay the clipped remainder's own SGR codes:

```ts
const tail = text.slice(i).match(/\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]/g);
return out + ellipsis + (tail ? tail.join('') : '');
```

Not a blanket `reset`, which would look simpler but breaks the selected row: `config.ts` wraps the already-truncated row in `inverse(...)`, and an inner `\x1b[0m` would end the highlight before the row's trailing padding, dropping it off the right edge. Openers whose content was clipped arrive paired with their closers, so they render as a no-op.

## How to test

```
npm run build
node dist/cli.js tui     # press `c`
```

Before: the Config hints clip and everything after the first clipped row renders dim, including the header and tab strip. After: attributes close at the cut.

## Tests

`test/tui-tabs.mjs` gains an SGR-balance pass that replays each line's codes and asserts nothing is left active at its end:

- a **control assertion** that the checker flags a known-unbalanced string, so the suite cannot pass vacuously
- direct `truncate()` cuts at six widths through a trailing `dim()` span
- the `inverse()`-preservation case that rules out the blanket-reset fix
- a Config sweep over six geometries with a full selection walk
- a `renderFooter` sweep across widths 4–60

Verified it pins the bug: reverting `render.ts` fails **8 assertions** (134 leaking Config lines at 80×24, 151 at 60×20, 11 leaking footer widths). Full suite is 121/121 with the fix.

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes — 121/121
- [x] Version bumped (5.4.3 → 5.4.4) + CHANGELOG section, validated with `extract-release-notes.mjs 5.4.4` and `check-package-json.mjs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.4] - 2026-07-25
+
+- **`truncate()` dropped the closing code of any style span it cut through.** The walk stopped the moment it had emitted `maxWidth` visible characters and returned immediately, so every escape sequence past the cut point was discarded — including the `\x1b[22m` / `\x1b[39m` that `dim()` and `fg()` append. A clipped span left its attribute switched on.
+
+- **The leak escaped the line, and the frame.** `App.redraw` writes `clearScreen + frame`, and `\x1b[2J\x1b[H` does not reset SGR; `renderFooter` ends on plain text, so nothing closed the attribute before the frame did. The dim carried into the *next* frame's header and tab strip, clearing only at the `dim('─')` separator on row 3 — and every frame re-leaked it.
+
+- **Surfaced by the Config viewport fix (#861).** That change made Config rows `truncate(… + dim(hint), w)`; at 80 columns the value column clamps to 16, leaving hints 32 characters, so 7 of the 14 clip on a default-width terminal. The underlying bug predates it — `renderFooter` already leaked cyan at 11 widths between 4 and 60. `hits.ts` was checked across columns 20–160 and is unaffected: its cut always lands clear of the colored status column.
+
+- **Fixed by replaying the clipped remainder's own SGR codes**, not by appending a blanket reset. `config.ts` wraps the truncated row in `inverse()`, and an inner `\x1b[0m` would end the highlight before the row's trailing padding, dropping it off the right edge of the selected row. Openers whose content was clipped arrive paired with their closers, so they render as a no-op.
+
+- **Tests:** `test/tui-tabs.mjs` gains an SGR-balance pass — a control assertion so the checker cannot pass vacuously, direct `truncate()` cuts at six widths, the `inverse()`-preservation case that rules out the reset, a Config sweep over six geometries with a full selection walk, and a `renderFooter` width sweep. Confirmed to fail 8 assertions against the pre-fix build (134 leaking Config lines at 80×24, 11 leaking footer widths) and pass after.
+
 ## [5.4.3] - 2026-07-25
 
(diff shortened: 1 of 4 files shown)


sprayberry-redline (approved):

## What I checked
- `src/tui/render.ts` diff: the `truncate()` change appends the clipped remainder's own SGR codes instead of a blanket reset.
- Correctness of the fix: the walk loop only ever advances `i` past a *complete* escape sequence (`if (m) { out += m[0]; i += m[0].length; continue; }`, render.ts:157-158, unchanged by this PR) or a single plain character. That means when the loop exits, `text.slice(i)` always starts either mid-plain-text or exactly at the start of an escape sequence — never mid-sequence — so the new line:
  ```ts
  const tail = text.slice(i).match(/\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]/g);
  return out + ellipsis + (tail ? tail.join('') : '');
  ```
  correctly captures every remaining SGR code (opens and closes alike) without splitting one. Since `dim()`/`fg()`/`inverse()` always emit paired open+close (render.ts:88-106), any opener whose content got clipped arrives paired with its own closer in the tail, so it's a no-op as the PR description claims — verified against `dim`/`fg`/`inverse` definitions directly.
- Test coverage (`test/tui-tabs.mjs`): includes a control assertion that the balance-checker can actually fail (guards against a vacuous pass), direct multi-width `truncate()` cuts through a `dim()` span, an `inverse()`-preservation case, a `ConfigTab` sweep across six geometries, and a `renderFooter` width sweep. This is solid coverage for the fixed path.
- CI: most checks were still pending at review time (`build`, `test`, `compat`, `analyze`); the checks that had completed (`build (20)`, `actionlint`, `triage`, `validate-package-json`) were green. Not blocking on the rest, per review policy — will not hold up the verdict, but flagging that the `test` job hadn't reported yet when this review was submitted.

## What's good
- Root-caused the bug precisely (loop exits without flushing the SGR state of the remainder) and picked the narrower fix (replay remainder codes) over a blanket reset, with a concrete counterexample (`inverse()`-wrapped selected row) for why reset would be wrong.
- Version bump + changelog match the checklist.

No correctness, security, or test-coverage issues found in the changed lines.

PR #864, opened 2026-07-26
fix(tui): bound every Hits row to the terminal width

### LINKED ISSUE #862: fix(tui): Hits tab chrome rows overflow the terminal width
## Summary

Several rows in the Hits tab are pushed with no width bound, so they exceed `cols` and wrap onto a second physical line. This is the same failure mode #861 just fixed for the Config tab — the data rows here are already truncated (`hits.ts:211-213`) and the detail pane is bounded by `renderKvRow`, but the tab's *chrome* is not.

Verified on `master` @ `65eac7e`.

## Rows that overflow

| Row | Visible width | Wraps when |
|---|---|---|
| Halt banner, line 2 | 106 | `cols < 106` |
| Halt banner, line 1 | 105 | `cols < 105` |
| Empty-state "Waiting for requests…" | 67 | `cols < 67` |
| Column header (`time model in out lat st`) | 56 | `cols < 56` |
| "Use ↑↓ to select a request for details." | 41 | `cols < 41` |

The column header is the one that's always present: `colTime 9 + colModel 18 + colIn 8 + colOut 7 + colLat 7 + colStatus 5 = 54`, plus the two-space prefix = 56 fixed. Any terminal under 56 columns wraps it on every frame.

The halt banner is the worst case, and it's the one that matters most — it's pinned to the top so it stays visible while scrolling (#288), and its width scales with the account and model strings. With a realistic account it measures 105/106, so it wraps on a standard 80-column terminal every time it appears, i.e. exactly when the user needs to read it.

## Why it matters

`renderTui` measures body height with `body.split('\n').length` (`src/tui/tui-app.ts:282`), which counts logical lines and is blind to wrapping. `App.redraw` then writes the frame after a bare `clearScreen` with the cursor at home (`src/tui/app.ts:215`), so every surplus physical row pushes the head of the panel off the top of the alt-screen. Same mechanism described in #861.

## Repro

```js
import { HitsTab } from './dist/tui/tabs/hits.js';
import { visibleWidth } from './dist/tui/render.js';

const s = HitsTab.initialState();
s.subscribed = true;
s.buffer = [/* any RequestRecord */];
s.halt = {
  since: Date.now(), cooldownUntil: Date.now() + 1800000,
  request: { claim: 'overage', model: 'claude-opus-4-5-20260101', account: 'someone@example.com' },
};

HitsTab.render(s, { cols: 80, rows: 19 })
  .split('\n')
  .forEach((l, n) => { if (visibleWidth(l) > 80) console.log(n, visibleWidth(l)); });
// → 1 105
// → 2 106
```

## Suggested fix

Wrap the unbounded `lines.push(...)` calls in `truncate(…, w)`, the same way #861 did for Config. The column header would also benefit from clamping its fixed column constants on narrow terminals, since truncation alone drops the rightmost headers while the data rows below them still render.

The halt banner needs a judgement call rather than a blind truncate — clipping it hides the account and the resume instructions. Reflowing it into two shorter pinned lines, or moving the detail into the pane below, would keep it readable at 80 columns.

askalf:
Closes #862.

## What does this PR do?

The Hits tab's data rows were already truncated to `w - 2` and the detail pane is bounded by `renderKvRow` — but the tab's **chrome** was pushed unbounded: title, column header, empty-state copy, scroll hint, and the halt banner. Those rows wrapped onto a second physical line.

That matters because `renderTui` measures body height with `body.split('\n').length` (`src/tui/tui-app.ts:282`), a logical count blind to wrapping, and `App.redraw` writes the frame after a bare `clearScreen` with the cursor at home (`src/tui/app.ts:215`). Every surplus physical row pushes the head of the panel off the top of the alt-screen — the same mechanism as the Config overflow fixed in 5.4.3.

## Measured, columns 30–160

| Row | Width | Wrapped when |
|---|---|---|
| Halt banner, line 2 | 106 | `cols < 106` |
| Halt banner, line 1 | 105 | `cols < 105` |
| SSE-error hint | 106 | `cols < 106` |
| Empty-state copy | 67 | `cols < 67` |
| Column header | 56 (fixed) | `cols < 56` |
| No-selection hint | 41 | `cols < 41` |

The SSE-error line is worse than its measured width suggests — it interpolates `state.connectionError`, which is upstream-supplied and unbounded.

## The halt banner

This is the row that actually hurt. It's pinned visible while scrolling (#288) and its width scales with the account and model strings, so it wrapped on a standard 80-column terminal **every time it appeared** — exactly when the user needs to read it.

Clipping it would hide the account and the resume instructions, so both lines are reflowed most-actionable-first instead:

```
  ⚠ HALTED  overage · 17:46:40 · opus-4-5-20260101 · thomas@sprayberrylabs.com   (78)
  → 503 until R or dario resume · auto-resume in 30m                             (52)
```

Down from 105/106. The account sits last so a narrow terminal clips the least-critical field, and the resume path fits outright. Everything is still truncated to `w` as a hard floor.

The column header now truncates to `w - 2` — the same budget the data rows use — so the columns stay aligned rather than the header running two characters wider than the rows beneath it.

## Tests

`test/tui-tabs.mjs` sweeps all six Hits states (connecting, waiting, SSE error, populated, halted, no-selection) across columns 30–160, asserting no row exceeds `cols`. Plus two assertions that the halt banner fits 80 **and** isn't clipped to get there — checking only "not clipped" would have passed against the pre-fix build, since nothing was truncating.

Verified it pins the bug: reverting `hits.ts` fails **8 assertions**. Full suite 121/121 with the fix.

## How to test

```
npm run build
node dist/cli.js tui     # press `h`, then resize narrow
```

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes — 121/121
- [x] Version bumped (5.4.4 → 5.4.5) + CHANGELOG section, validated with `extract-release-notes.mjs 5.4.5` and `check-package-json.mjs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.5] - 2026-07-25
+
+- **The Hits tab pushed its chrome rows with no width bound.** Data rows were already truncated to `w - 2` and the detail pane is bounded by `renderKvRow`, but the title, column header, empty-state copy, scroll hint and halt banner were not — so they wrapped onto a second physical line. `renderTui` measures body height with `body.split('\n').length` (`tui-app.ts:282`), a logical count blind to wrapping, and `App.redraw` writes after a bare `clearScreen` with the cursor at home, so each surplus physical row pushed the head of the panel off the top of the alt-screen. Same mechanism as the Config overflow fixed in 5.4.3.
+
+- **Measured across columns 30–160:** the halt banner was 105 and 106 wide, the SSE-error hint 106 (its upstream error text is unbounded entirely), the empty-state copy 67, the column header a fixed 56, and the no-selection hint 41.
+
+- **The halt banner was the one that mattered.** It is pinned visible while scrolling (#288) and its width scales with the account and model strings, so it wrapped on a standard 80-column terminal every time it appeared — precisely when the user needs to read it. Rather than clip it, both lines are reflowed most-actionable-first and now measure **78 and 52** at 80 columns: claim, time, short model and account on the first; the resume path (`R` / `dario resume`) and auto-resume countdown on the second. The account sits last so a narrow terminal clips the least-critical field, and the resume instructions fit outright.
+
+- **The column header now truncates to `w - 2`**, the same budget the data rows use, so the columns stay aligned instead of the header running two characters wider than the rows beneath it.
+
+- **Tests:** `test/tui-tabs.mjs` sweeps all six Hits states — connecting, waiting, SSE error, populated, halted, no-selection — across columns 30–160 and asserts no row exceeds `cols`, plus that the halt banner fits 80 *without* being clipped to get there. Confirmed to fail 8 assertions against the pre-fix build and pass after.
+
+- Closes #862.
+
 ## [5.4.4] - 2026-07-25
 
 - **`truncate()` dropped the closing code of any style span it cut through.** The walk stopped the moment it had emitted `maxWidth` visible characters and returned immediately, so every escape sequence past the cut point was discarded — including the `\x1b[22m` / `\x1b[39m` that `dim()` and `fg()` append. A clipped span left its attribute switched on.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.4.4",
+  "version": "5.4.5",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/tui/tabs/hits.ts
+++ b/src/tui/tabs/hits.ts
@@ -140,15 +140,16 @@ export const HitsTab: Tab<HitsState> = {
     const listRows = Math.max(3, totalRows - detailRows - 2);
 
     if (state.buffer.length === 0) {
-      lines.push(' ' + brand('Hits') + dim('  — live request stream'));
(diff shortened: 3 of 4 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found - clean, well-tested fix.

## What I checked
- `gh pr diff` for `src/tui/tabs/hits.ts` and `test/tui-tabs.mjs`.
- Pulled surrounding context from `src/tui/render.ts` (`truncate`/`visibleWidth`) to confirm the ANSI-aware truncation helper is pre-existing and used correctly, not reimplemented.
- Confirmed the new header truncation budget (`w - 2` at `src/tui/tabs/hits.ts:199`, `lines.push(truncate('  ' + dim(...), w - 2))`) matches the budget the data rows already use (`truncate(row, w - 2)` at `hits.ts:218-220`), so columns stay aligned as the PR claims.
- CI check-rollup: all green - `test`, `build (18/20/22)`, `compat`, `CodeQL`, `docker-cap-drop-smoke`, `actionlint`, `validate-package-json`, `triage` all pass.

## What's good
- Every previously-unbounded chrome line (title, empty-state copy, SSE-error hint, halt banner, column header, scroll hint, no-selection hint) is now wrapped in `truncate(..., w)` or `w - 2`, directly addressing the wrap-pushes-panel-off-top mechanism described in the PR body.
- `test/tui-tabs.mjs` sweeps all six Hits states across columns 30-160 and asserts no row exceeds `cols`, plus dedicated assertions that the halt banner fits 80 columns *without* being clipped - a meaningfully strong test that would fail against the pre-fix build (verified by the author: 8 assertions fail on revert).
- The halt-banner reflow (most-actionable-first, account last) is a sensible UX call given the width constraint, and is exercised by the new tests.

No correctness, security, reuse, or coverage issues found in the changed lines. Nothing further to flag.

PR #860, opened 2026-07-25
auto-rebake: template drift detected 2026-07-25

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-25T19:34:15+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.4.3` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — worth a closer read

- **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.220) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.220, 27 tools, 6169 char system prompt
[bake] scrubbed:
[bake]   tools: 27 → 27 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6169 → 4754 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] previous baked template: CC v2.1.220 captured 2026-07-25T17:32:17.394Z, 33 tools, 4776 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4754 → fable 9200 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4754 → opus-5 8088 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4754 → sonnet-5 13448 chars
[bake] check: drift detected — 1 differing slot (verdict: moderate):
[bake] **Verdict:** 🟡 Moderate — worth a closer read
[bake] 
[...]

PR #865, opened 2026-07-26
fix(scrub): canonicalize ~/.claude/projects paths so the bake is host-portable (v5.4.6)

askalf:
Root-causes and fixes the rebake ping-pong. **The 22 bytes were a scrubbed filesystem path.**

## Finding

#854 (my bake, base **4776**) was followed three hours later by `#860 auto-rebake: template drift detected` (CI's bake, base **4754**) — each host reporting the other as drift, exactly the churn I flagged on #854.

Isolated by diffing the two committed bundles. **One line differed, in each direction:**

```
+ MINE: C:\Users\user\.claude\projects\C--Users-user-project\memory\     60 bytes
- CI:   /root/.claude/projects/project/memory/                           38 bytes
```

Same logical path. The identity rules did their job — no username, no real project slug, nothing leaked — but they preserved the **host's path shape**, and that shape differs by 22 bytes between Windows and Linux. It appears once in the base prompt and once in the fable variant, which is the entire delta in both.

So `computeDrift` was correct to report a byte difference. The bug is that the difference existed at all: **the bake was not host-portable**, so two bakers on different platforms could never agree, indefinitely.

## Fix

`scrubText` now canonicalizes every `~/.claude/projects/...` path to one placeholder:

| host form | becomes |
|---|---|
| `C:\Users\bob\.claude\projects\C--Users-bob-p\memory\` | `/home/user/.claude/projects/project/memory/` |
| `/root/.claude/projects/-root-p/memory/` | `/home/user/.claude/projects/project/memory/` |
| `/home/bob/.claude/projects/-home-bob-p/memory/` | `/home/user/.claude/projects/project/memory/` |
| `/Users/bob/.claude/projects/-Users-bob-p/memory/` | `/home/user/.claude/projects/project/memory/` |

Two details that matter:

- **Trailing segments are consumed, not just the prefix.** `memory\` and `memory/` are the same *length* but different *bytes*, so prefix-only normalisation would still have drifted.
- **Anchored on the already-scrubbed home forms.** The canonicalizer only matches paths whose username has already been masked, so it cannot make an *unscrubbed* path look clean — `findUserPathHits` must keep failing those, and the bake's residual-path guard depends on it.

Same class as **#856** (stopped replaying the bake host's OS/arch onto every request). That was the header half; this is the prompt half.

## Verification

Re-baked, so the committed bundle is already canonical — base **4759**, fable 9205, opus-5 8093, sonnet-5 13442 — and `capture-and-bake --check` **exits 0**.

Cross-platform identity asserted directly: all four platform forms collapse to a single byte-identical string (`new Set([...]).size === 1`), idempotent, no residual user paths, and a non-project Windows path still masks the username while keeping its shape.

**121/121** suite.

**Five existing assertions changed**, all deliberately: they encoded the old platform-preserving contract — one literally read *"Windows backslash projects form untouched by the POSIX rule"*, which is precisely the behaviour that caused this. Username masking on **non-project** paths is untouched and still asserted. Each updated assertion carries a comment recording what it used to claim and why it flipped.

## Expect one more rebake, then stability

The canonical form is `/home/user/...`, which matches neither of the two bundles that were fighting. CI's next `--check` will therefore see drift once, re-bake to the canonical form, and match every host from then on. That single rebake is expected, not a regression.

## Two notes on how this was found

**My first diff was wrong.** It compared my capture against my *working tree* bundle rather than CI's committed one and reported a nonsense 279-byte delta. The real diff needed `git show <the auto-rebake commit>:src/cc-template-data.json`.

**A capture-integrity observation, worth its own issue.** During that detour, one capture returned a prompt containing a line from **my own session's** system prompt — text that has no business in a captured CC prompt. The capture reads the first `/v1/messages` to arrive at a loopback MITM, so a foreign process's request can apparently be captured instead, which would silently poison a bundle. It reproduced once in six captures and I cannot trigger it on demand, so this is an observation rather than a diagnosis — but if real it matters, because a bake could ship an operator's session prompt to npm.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — the canonicalization logic checks out on manual trace; approving.

## What I checked
- `src/scrub-template.ts`: the new `canonicalizeClaudeProjectPaths` + `CLAUDE_PROJECT_PATH` regex, which runs after `USER_PATH_PATTERNS` (correctly anchored on already-identity-masked forms per the comment at `src/scrub-template.ts` — the anchor alternation `[A-Za-z]:[\\/]Users[\\/]user|\/root|\/home\/user|\/Users\/user` can't match an unscrubbed path).
- Manually traced 4 of the added test cases through the regex + split/slice logic (Windows flattened slug, `/root` POSIX slug with trailing separator, idempotency on the canonical form, and the already-de-identified Windows form with `C--Users-user-project`) — each produces the exact expected output in `test/scrub-template.mjs`.
- `test`, `CodeQL`, `build (18/20/22)`, `compat`, `wire-drift`, and `docker-cap-drop-smoke` checks are all green on `statusCheckRollup` (headRefOid `a0cf96b`). PR description claims 121/121 suite pass, consistent with the green `test` check.
- Version bump (`package.json`, `package-lock.json` 5.4.5→5.4.6) and `CHANGELOG.md` entry are consistent with each other.
- `src/cc-template-data.json` diff is re-baked/captured template data (not hand-written logic) and the visible memory-path line now reads the intended canonical `/home/user/.claude/projects/project/memory/` form — consistent with the stated fix.

## What's good
- The five changed test assertions each carry an inline comment explaining what the old assertion encoded and why it flipped (e.g. `test/scrub-template.mjs` line documenting "WAS: asserted the POSIX rule must not touch the Windows form... now the opposite") — good practice for a behavior-reversing test change.
- The regex is deliberately anchored on already-masked forms so it can't launder an unscrubbed path into looking clean, and the trailing-segment handling (`slice(1)` after splitting on `/projects/`) correctly handles both the Windows flattened-slug and POSIX slug shapes despite them using different placeholder text.

## Minor (non-blocking)
- The capture-integrity issue mentioned in the PR description ("a foreign process's request can apparently be captured instead") is explicitly called out as an observation, not part of this diff — worth its own issue as the author says, not a finding against this PR.

PR #866, opened 2026-07-26
fix(tui): stop the TUI crashing and overflowing on small terminals

askalf:
Findings from a full TUI audit. The harness drove the **real** `startTuiApp()` through a fake TTY against a local stub proxy — real sockets, real SSE, real raw-mode key parsing, real tab mount/unmount — and audited every frame the app actually wrote to stdout, across six tabs and twelve terminal geometries. Navigation keys only, so nothing wrote config or POSTed `/admin/resume`.

Five defects, in severity order.

## 1. Analytics crashed the TUI on a narrow terminal

Bar widths are computed as (terminal width − label columns), which goes negative on a small terminal. `String.repeat()` throws `RangeError` on a negative count, and the throw propagated out of `App.redraw` rather than degrading:

```
RangeError: Invalid count value: -1
    at progressBar (src/tui/render.ts:205)
    at Object.render (src/tui/tabs/analytics.ts:92)
    at renderTui (src/tui/tui-app.ts:202)
```

**Analytics threw at every width ≤ 31 columns** (`-7` at 20 cols through `-1` at 31). `progressBar` now floors width at 0 and tolerates a non-finite value, so the bar collapses to empty and the surrounding row still renders.

## 2. Three tabs rendered taller than the terminal — at the default 80×24

Measured with data loaded:

| Tab | Rows rendered | vs 80×24 |
|---|---|---|
| Status | 27 (fixed) | **+3** |
| Analytics | 28 (fixed) | **+4** |
| Hits | scales, overshoots | **+3** |
| Config / Accounts / Backends | fits | ✅ |

`App.redraw` writes the frame from row 1 of the alt-screen with no scrollback, so the overflow scrolls off the **top**. On Analytics at 80×24 the four lost rows are:

```
LOST row 0: "┌ dario v5.4.5 ──────────  http://127.0.0.1:3456"
LOST row 1: " Status  Config  Analytics  Hits  Accounts  Backends "   ← the tab strip
LOST row 2: "─────────────────────────────────────────────────────"
LOST row 3: " Analytics  — last 60 min"                               ← the panel title
```

The tab strip is the one row telling the user which tab is active, so the overflow cost exactly the wrong four rows — the same failure mode as #861/#862, one level up.

**Root cause:** `status.ts`, `analytics.ts` and `backends.ts` contain zero references to `dimv.rows`; they render every panel unconditionally. And `renderTui` only padded a short body (`if (bodyLines < bodyRows)`) — it never truncated a long one, so nothing caught it. Hits scales but reserves 11 rows of non-list chrome while rendering 15 (the halt banner isn't in the arithmetic); with a fuller buffer it overflows even at 120×40.

`renderTui` now clips the body to its budget, keeps header/tab strip/footer pinned, and marks the clip with a `… N more rows — resize for the rest` note so the loss is visible. A final slice to `rows` backstops it unconditionally.

**This is a floor, not a fix for the tabs.** A tab that overflows here is still choosing what to drop by accident of ordering — Status silently loses its Overage-guard panel. Per-tab budgeting (collapse panels, paginate) is the real answer and is worth doing separately; I didn't want to bury it in this diff.

## 3. Accounts pushed unbounded rows

Its column header measures 68, and the disk-fallback branch interpolates `~/.dario/accounts/<alias>.json` — unbounded by alias length. Rows are now bounded at the push site rather than at each call site: hand-auditing 15 separate pushes is precisely how this was missed while the neighbouring Hits tab was being fixed in #864.

## 4. The header bar never shrank

`renderHeader` clamps its dash filler at 0 but still emitted both ends full-length, overflowing below ~46 columns — a threshold that moves with the version string and proxy URL. It now drops the status (the proxy URL, also shown on the Status tab) before clipping the brand, which is the row's only identity.

## 5. "detected 4m ago ago"

`formatAgo()` already ends in "ago"; the template appended a second one.

## Tests

New `test/tui-frame.mjs`, 200 assertions, covering the frame-level invariants no individual tab guarantees:

- no frame taller than the terminal and no row wider than it — six tabs × twelve geometries down to 24×6, with every tab in its **loaded** state (the empty states already fit, so an empty fixture would assert nothing)
- header, tab strip and footer all survive at the sizes that used to overflow
- clipping is announced when it happens, and absent when it doesn't
- `progressBar` degrades at negative, zero and non-finite inputs
- Analytics renders at every width from 20 to 40

`renderTui` is exported for this. **Each of the five fixes was reverted individually and confirmed to fail** — `RangeError` for progressBar, header 39-wide at 30 cols, `detected 8764h ago ago`, Accounts 68-wide, and Status 27/24 + Analytics 28/24 + Hits 42/40.

Full suite: 122/122.

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes — 122/122
- [x] Version bumped (5.4.5 → 5.4.6) + CHANGELOG section, validated with `extract-release-notes.mjs 5.4.6` and `check-package-json.mjs`


sprayberry-redline (dismissed):

**Verdict:** No blocking issues found — approving. Solid, well-tested fix.

## What I checked
- Full diff across `layout.ts`, `render.ts`, `tabs/accounts.ts`, `tabs/status.ts`, `tui-app.ts`, and the new `test/tui-frame.mjs`.
- Traced the clip math in `renderTui` (`src/tui/tui-app.ts:268-315`): `bodyRows = Math.max(1, rows - 5)`, the `hidden = bodyLineArr.length - bodyRows + 1` count, and the final `out.join('\n').split('\n').slice(0, rows).join('\n')` backstop — the arithmetic is consistent (e.g. 25 body lines against a 24-row budget yields `hidden=2`, `slice(0, 23)` + 1 note line = 24 total, matching the budget).
- Checked `progressBar` (`src/tui/render.ts:200-211`) — now floors width at 0 and guards non-finite input before `String.repeat()`, which is the actual root cause of the `RangeError` crash described in the PR body.
- Checked `renderHeader`'s fallback path (`src/tui/layout.ts:24-33`) and `truncate()`'s ANSI-aware walk (`src/tui/render.ts:144-`) — truncation counts visible chars correctly and preserves/flushes SGR codes.
- `accounts.ts`'s `push` helper correctly routes every `lines.push` through `truncate(s, w)`, closing the unbounded-row gap described (column header width 68, unbounded alias interpolation).
- CI check-rollup: all 10 checks green (build ×3, test, docker-cap-drop-smoke, compat, CodeQL ×2, actionlint, validate-package-json) as of the current head SHA `eac5e32`.

I did not find a scenario where the changed lines produce an incorrect result — the edge cases I traced (0-width clip budget, small-terminal header collapse, ANSI-embedded truncation) all resolve as intended.

## What's good
- Each of the five fixes is paired with an inline comment explaining the specific failure mode (not just what changed but why), and the PR description states each fix was individually reverted and confirmed to fail under the new test — good discipline.
- `test/tui-frame.mjs` asserts the actual invariant that matters (no frame taller/wider than the terminal) across 6 tabs × 12 geometries down to 24×6, rather than just unit-testing helpers in isolation.

## Minor (non-blocking)
- `renderTui`'s clip backstop assumes `bodyRows = Math.max(1, rows - 5)` keeps total frame height within `rows`, which holds for all tested geometries (min 24×6) but not for a hypothetical `rows < 6` terminal, where the footer could be sliced away by the final backstop. Not exercised by the new tests and not a realistic terminal size — flagging only as a documentation note, not a fix requirement.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — approving. Solid, well-tested defensive fix set.

**What I checked**
- Full diff (`layout.ts`, `render.ts`, `tabs/accounts.ts`, `tabs/status.ts`, `tui-app.ts`, new `test/tui-frame.mjs`).
- Traced the frame-clip arithmetic in `renderTui` (`src/tui/tui-app.ts:263-311`): `bodyRows = Math.max(1, rows - 5)`, and when `bodyLineArr.length > bodyRows`, `hidden = bodyLineArr.length - bodyRows + 1` with `slice(0, bodyRows - 1)` + note. For example at `bodyRows=1` (very small terminal), `slice(0,0)` correctly yields just the note line and `hidden` correctly counts all original lines as dropped — no off-by-one found.
- `progressBar` (`src/tui/render.ts:203-212`): width is now floored/clamped (`Math.max(0, Math.floor(width))`) and `value` guarded for non-finite before the `.repeat()` calls that previously threw `RangeError` on negative width — confirmed the fix addresses the crash described in the PR body (bar width derived as `terminal width − label columns` going negative, per call sites in `tabs/analytics.ts`).
- `renderHeader` (`src/tui/layout.ts:24-32`): computes the full string first, falls back to `truncate(...)` only if it still overflows — no case where a narrow width bypasses the fallback.
- `accounts.ts`: all pushes now routed through a single `push()` wrapper that truncates to `w`, bounding the previously-unbounded disk-fallback path row.
- `status.ts`: single-line dupe-"ago ago" fix, matches the described bug.
- CI: `gh pr checks 866` — all reported checks pass (build ×3, test, CodeQL, actionlint, docker-cap-drop-smoke, validate-package-json); `compat` is still pending at review time (non-blocking, noted per policy rather than waited on).

**What's good**
- New `test/tui-frame.mjs` asserts the actual invariant that matters (frame never exceeds terminal rows/cols) across 12 geometries × 6 tabs, plus targeted checks for each of the five fixes (each reportedly reverted individually and confirmed to fail per the PR description).
- Comments are non-obvious and explain real constraints (e.g. why `truncate` replays SGR codes on cut, why `progressBar` collapses instead of clamping upward).
- The PR body is explicit about the fix being a "floor, not a fix" for per-tab budgeting — appropriately scoped and doesn't overreach into unrelated refactors.

No minor nits worth blocking on either — the diff is otherwise clean.

PR #867, opened 2026-07-26
fix(test): isolate the live-template cache so tests stop poisoning each other

askalf:
Root-causes and fixes the intermittent `issue-29-tool-translation.mjs` failure — and closes a hazard that was worse than the flake.

## The flake

`issue-29-tool-translation.mjs` failed roughly **1 run in 5** under `--test-concurrency=8`, passed every time in isolation, and never named itself in the summary. I chased a `TIME_WAIT`/port-collision theory first; that was **wrong** — the file binds no socket and does no I/O.

The actual cause is shared mutable state on disk:

- `test/live-fingerprint.mjs` writes **fake** templates into the **real** `~/.dario/cc-template.live.json` to exercise `loadTemplate`'s accept/reject rules. One fake has a single tool: `tools: [{ name: 'Bash' }]`.
- `test/issue-29-tool-translation.mjs` imports `dist/cc-template.js`, whose **module init** calls `loadTemplate()` and reads that same file.
- Under `--test-concurrency=8` those are concurrent **processes**, so issue-29 can initialise while a fake is on disk and evaluate its Bash↔process mapping against a 1-tool template.

Reproduced deterministically by writing that exact fake and running the test:

```
no live cache (bundled template)  exit=0   54 pass, 0 fail
fake live cache mid-suite         exit=1   48 pass, 6 fail
```

## The worse problem underneath

The test backs the real cache up and restores it on exit — so **if it ever dies between the write and the restore, the operator's own dario keeps serving `FAKE LIVE SYSTEM PROMPT` and a 1-tool template** until someone deletes the file by hand. A unit test should not be able to do that.

## Fix

`src/live-fingerprint.ts` resolves the cache path through `liveCachePath()`, overridable with `DARIO_LIVE_TEMPLATE_CACHE`. The test points it at a per-process temp directory and never touches `~/.dario` again.

Two deliberate details:

- **Resolved per call, not as a module-level `const`.** ESM imports are hoisted, so a test setting the env var in its body would run *after* a const had already captured the real path.
- **A temp directory keeping the real basename** (`<tmp>/dario-live-template-test-<pid>/cc-template.live.json`), so the existing quarantine helpers — which match `cc-template.live.json.corrupt-*` — keep working untouched, and two concurrent runs can't collide. My first attempt used a temp *filename*, which broke those helpers.

Production behaviour is unchanged unless the env var is set.

## Verification

- `live-fingerprint.mjs` **70/70** against the temp cache
- Full suite **121/121, six consecutive runs** — the flake reproduced ~1 in 5 before, so six clean runs is a meaningful signal rather than a single lucky pass
- Temp dir is removed on exit, so runs don't accumulate directories

## Not fixed here

Seven other suites also touch `~/.dario` (`atomic-write`, `credential-freshness`, `durable-token-persist`, `ensure-login-in-pool`, `mirror-login-to-credentials`, `oauth-detector`, `pool-reconcile`). They operate on credentials and accounts rather than the template cache, so none of them can poison `cc-template.js` the way this did — but they share the same *class* of hazard with each other, and the same override pattern would fix it if another flake ever points there. Out of scope for this PR.

No version bump, deliberately: #865 is open and bumps to 5.4.6 from the same base, and a second bump would guarantee the `package.json`/`CHANGELOG` conflict. This rides the next release.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean fix, approving.

## What I checked
- `gh pr diff 867` (54 additions / 9 deletions across `src/live-fingerprint.ts` and `test/live-fingerprint.mjs`)
- Cloned the head (`80e4820`) to confirm `liveCachePath()` has no top-level/module-init call site in `src/live-fingerprint.ts` (only invoked inside `readLiveCache`/`writeLiveCache`/`quarantineCorruptCache`), which is the load-bearing claim in the PR description (the ESM-hoisting fix only works if the path is resolved lazily, not captured at import time) — confirmed.
- CI rollup: `gh pr checks 867` — all checks pass (`test`, `build (18/20/22)`, `CodeQL`, `compat`, `docker-cap-drop-smoke`, etc.)

## Findings
None blocking.

**Minor:** `test/live-fingerprint.mjs:54` — `const { readFileSync } = require('node:fs');` inside `restoreCache()`. This is pre-existing code untouched by this diff, and `require` isn't defined in an `.mjs` (ESM) module, so this branch would throw if it ever executed. In practice the PR's own change makes this branch unreachable: `LIVE_CACHE` now always points at a freshly-created per-process temp dir (`TEST_CACHE_DIR`), so `existsSync(LIVE_CACHE)` at backup time is always false, `BACKUP` is never written, and `restoreCache`'s `if (existsSync(BACKUP))` branch containing the `require` call can never be taken. Not a regression from this PR and not blocking — just noting it's now dead code, worth a follow-up cleanup if anyone touches this file again.

## What's good
- Root cause is well-isolated (module-init read of shared mutable state across concurrent test processes) and the reproduction numbers in the PR body (54/0 vs 48/6) back it up.
- The lazy-resolution fix (`liveCachePath()` as a function, not a module-level `const`) correctly avoids the ESM-hoisting trap described — verified no other call site in `src/live-fingerprint.ts` reads the cache path at import time.
- Keeping the real basename inside a per-pid temp *directory* (rather than a temp filename) is a nice touch that keeps the `*.corrupt-*` quarantine glob matching intact and avoids concurrent-run collisions.
- Cleans up its own temp dir on exit (`rmSync(TEST_CACHE_DIR, { recursive: true, force: true })`).
- Correctly scoped: doesn't attempt to fix the other seven suites that touch `~/.dario`, and explains why they're a different hazard class.

PR #871, opened 2026-07-26
fix(tui): give each tab a real row budget

### LINKED ISSUE #868: fix(tui): give each tab a real row budget instead of relying on the frame clip
Follow-up to #866, which stopped the TUI from writing frames taller than the terminal. That fix is a **floor, not a layout**: `renderTui` clips the body to its row budget and pins the header, tab strip and footer, so the chrome the user navigates by survives. But a tab that overflows is still choosing what to drop by accident of ordering.

## What's actually missing

`status.ts`, `analytics.ts` and `backends.ts` contain **zero references to `dimv.rows`**. They render every panel unconditionally and hand back whatever height that comes to — a fixed 27 rows for Status and 28 for Analytics, regardless of the terminal.

The consequence at a default 80×24, where the body budget is 19 rows:

- **Status** renders Proxy, Config, Models and Overage-guard in that order. Overage-guard is last, so it is what gets clipped — the halt banner, the cause, the auto-resume countdown and the manual-resume instructions. That's the loudest, most time-sensitive panel in the tab, and it's the one that disappears.
- **Analytics** loses its trailing panels the same way.

So the tab that tells you the proxy has halted hides that fact first, on the most common terminal size. #866 makes the loss visible (`… N more rows — resize for the rest`) rather than silent, which is strictly better than scrolling the tab strip off the top, but it isn't the right answer.

## What the right answer looks like

Each tab should size its own content to `dimv.rows` and decide what to drop, rather than being clipped from the bottom:

- **Priority ordering** — render the panels that matter most first when space is short. Overage-guard, when halted, should outrank the Models list.
- **Collapse rather than clip** — a panel that doesn't fit could render a one-line summary (`Models: 5 advertised`) instead of vanishing.
- **Paginate or scroll** the panels that are genuinely long, the way `config.ts` (#861) and `hits.ts` window their lists around a selection via `clampVisibleStart`.

`config.ts` is the model to copy: it takes `dimv.rows`, computes a row budget, windows its field list around the selection, and shows an `n / total` counter when the list overflows.

## Scope

Four tabs, independent of each other, each a self-contained change:

- [ ] `status.ts` — priority ordering, with Overage-guard promoted when `halted`
- [ ] `analytics.ts` — collapse per-model / per-account tables to summaries when short
- [ ] `backends.ts` — currently fits at every size measured, but has no budget, so it will regress the moment a panel is added
- [ ] `hits.ts` — reserves 11 rows of non-list chrome while rendering 15; the halt banner isn't in the arithmetic and the detail pane is 8 rows against 9 reserved. With a full buffer it overflows even at 120×40.

The frame-level invariants in `test/tui-frame.mjs` already assert the property from the outside, so each tab can be fixed and verified independently without weakening the backstop in #866.

askalf:
Closes #868.

## Why this, when #866 already stopped the overflow

5.4.7 made `renderTui` clip an over-long body so the header and tab strip survive. That's a **floor, not a layout** — it drops whatever sorts last. On Status that was the Overage-guard panel, so at a default 80×24 the tab that tells you the proxy has **HALTED** hid the halt, the cause, the auto-resume countdown and the resume instructions. The `… N more rows` note made the loss visible, not acceptable.

## The mechanism

New `src/tui/panels.ts`. A panel declares:

| field | meaning |
|---|---|
| `lines` | full rendering |
| `collapsed` | shorter stand-in, usually one summary row |
| `priority` | lower = more important; drives what degrades first |
| `required` | never dropped (may still collapse) |

`fitPanels` degrades **least-important-first** in two passes — collapse everything offering a shorter form, then drop what isn't required — and preserves **display order**. Priority decides what degrades, never where a panel appears: users navigate by position, so resorting panels under pressure would be its own kind of surprise.

## Per tab

**Status** — Proxy is `required`; Overage-guard is `required` at top priority *when halted* and collapses to one line when idle. Models is the longest and least time-critical (a static catalog between releases), so it collapses to `5 advertised` first and drops last. Confirmed: HALTED and `dario resume` survive every budget from 40 rows down to 14.

**Analytics** — the rate-limit gauges and the overage row are `required`; the per-model breakdown and billing buckets collapse to counts. Was a fixed 28 rows against a 19-row budget.

**Hits** — reserved 11 rows of chrome while rendering up to 15. The pinned halt banner was **absent from the arithmetic entirely** and the detail pane is 8 rows, not the 9 reserved. Now derived from what the render actually emits; on a terminal too short for both, the detail pane yields to the request list rather than the reverse — the list is the tab's reason to exist.

**Backends** — had no budget at all, plus a 70-wide column header and data rows interpolating `baseUrl` (unbounded). Bounded at the push site, list windows to the height, and a truncated list says how many are hidden.

## Two incidental bugs, both found by the tests

- **The overage gauge rendered as `Overa…`.** Its label column was 6 — one narrower than the word — so `pad` truncated it and left it flush against its bar: `  Overa…██░░░…`. The row that means "investigate immediately" was the one row that didn't read. Gauge rows now share an 8-wide label column.
- **Two footers were unbounded** (`Updated <ago>` on Analytics, `Last refresh: <ago>` on Status). They'd been invisible because an over-long body was clipped before reaching them — fixing the height exposed the width.

## Tests

New `test/tui-budget.mjs`, 79 assertions:

- `fitPanels` directly — collapse-before-drop, `required` surviving at the lowest priority, display order preserved, over-budget required panels returned rather than truncated
- per tab: the body fits **and** the rows worth looking at survive — HALTED + resume path on Status, the overage row on Analytics, the halt banner on Hits, width *and* height on Backends

Each of the four tabs was reverted individually and confirmed to fail: **status 5, analytics 9, hits 12, backends 14**. Notably, reverting `analytics.ts` fails "overage row survives" at *every* size — that's the `Overa…` clip, not a budget issue.

`test/tui-frame.mjs` gains an assertion that **80×24 no longer needs the frame clamp at all**, and its clip test moves to 80×12, where Status's required panels genuinely can't fit. The 5.4.7 backstop is unchanged and still correct — it just isn't reached at ordinary sizes any more.

Full suite: 123/123.

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes — 123/123
- [x] Version bumped (5.4.8 → 5.4.9) + CHANGELOG section, validated with `extract-release-notes.mjs 5.4.9` and `check-package-json.mjs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,26 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.9] - 2026-07-26
+
+- **Every tab now spends its own row budget instead of being clipped from the bottom.** 5.4.7 stopped the app writing frames taller than the terminal, but that fix is a floor: `renderTui` clips whatever sorts *last*. On Status that was the Overage-guard panel — so the one tab that tells you the proxy has HALTED hid the halt, the cause, the auto-resume countdown and the resume instructions, on a default 80×24 terminal. Closes #868.
+
+- **New `src/tui/panels.ts`.** A panel declares its full `lines`, an optional shorter `collapsed` form, a `priority`, and whether it is `required`. `fitPanels` degrades least-important-first — collapse everything that offers a shorter form, then drop what isn't required — and preserves *display* order, since users navigate by position and reordering panels under pressure would be its own surprise.
+
+- **Status** keeps Proxy and, when halted, Overage-guard; Models collapses to a count (`5 advertised`) then drops, and Config collapses to one row. HALTED and the resume path survive at every budget from 40 rows down to 14.
+
+- **Analytics** keeps the rate-limit gauges and the overage row — the "investigate immediately" signal — and collapses the per-model breakdown and billing buckets to summaries. It was a fixed 28 rows against a 19-row body budget.
+
+- **Hits** reserved 11 rows of chrome while rendering up to 15: the pinned halt banner was missing from the arithmetic entirely, and the detail pane is 8 rows, not the 9 reserved. The reservation is now derived from what the render actually emits, and on a terminal too short for both, the detail pane yields to the request list rather than the other way round.
+
+- **Backends** had no budget at all and pushed unbounded rows — a 70-wide column header and data rows interpolating `baseUrl`. Rows are bounded at the push site, the list windows to the available height, and a truncated list says how many backends are hidden.
+
+- **The overage gauge was mislabelled `Overa…`.** Its label column was 6 characters — narrower than the word "Overage" — so `pad` truncated it and left it flush against its bar. The gauge rows (5h / 7d / Overage) now share an 8-wide label column, so the row that means "investigate immediately" reads unambiguously.
+
+- **Two footers were unbounded** (`Updated <ago>…` on Analytics, `Last refresh: <ago>…` on Status). They had been invisible because an over-long body was clipped before reaching them; now that the tabs fit and reserve their footers, they render — and are bounded.
+
+- **Tests:** new `test/tui-budget.mjs` (79 assertions) covers `fitPanels` directly — collapse-before-drop, required panels surviving at the lowest priority, display order preserved — then asserts per tab that the body fits AND that the rows worth looking at survive: HALTED and the resume path on Status, the overage row on Analytics, the halt banner on Hits, width and height on Backends. Each of the four tabs was reverted individually and confirmed to fail (status 5, analytics 9, hits 12, backends 14 assertions). `test/tui-frame.mjs` gains an assertion that 80×24 no longer needs the frame clamp at all.
+
 ## [5.4.8] - 2026-07-26
 
 - **Template rebake** — re-captured `src/cc-template-data.json` after cc-drift-template-watch detected wire-fingerprint drift against a live CC capture. Bundled fallback template now matches the current CC wire shape.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.4.8",
+  "version": "5.4.9",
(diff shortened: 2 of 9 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found -- clean, well-tested refactor. Approving.

## What I checked
- Full diff (`src/tui/panels.ts` new module, `analytics.ts`, `backends.ts`, `hits.ts`, `status.ts`, `test/tui-budget.mjs`).
- `fitPanels` degrade-order logic: sorts least-important-first via `(b.priority - a.priority) || (b.i - a.i)`, collapses before dropping, never drops `required` panels, and preserves display order by iterating the original `panels` array when flattening output (`src/tui/panels.ts:108-126`) -- matches the stated intent.
- `hits.ts` chrome-row arithmetic against the lines actually pushed later in the same function -- the reserved row counts (title, banner rows, blank, header, scroll-hint, detail pane rows + separator) match what's rendered.
- `backends.ts`'s hidden-count math: `hidden = state.backends.length - shown.length`, then on truncation it overwrites the last *shown* row with a "more backends" message -- correctly accounts for the row it just overwrote in addition to the truly-hidden ones (`src/tui/tabs/backends.ts`).
- CI check-rollup: all green (CodeQL, build x3, compat, test, docker-cap-drop-smoke, triage, validate-package-json).

## What's good
- Each degrade decision (what collapses vs. drops vs. is required) is justified per-tab with a concrete rationale in comments, and the changelog/PR description ties each to a specific prior failure (the mislabeled gauge, unbounded footers, the Hits arithmetic gap) -- easy to verify against the code.
- New `test/tui-budget.mjs` asserts behavior, not just "fits", and the PR description states each tab was reverted individually and its corresponding assertions fail -- good regression coverage for exactly this bug (#868).

## Minor (non-blocking)
- `backends.ts`'s header/rule rows are truncated to `w` while most other per-tab rows elsewhere use `w - 4`/`w - 2` for the indent; pre-existing, not introduced by this diff, but worth a follow-up glance if column alignment ever looks off on narrow terminals.

PR #873, opened 2026-07-26
feat(tools): land the TUI audit harness, fix the Analytics title overflow it found

askalf:
Lands the harness that found the v5.4.7 and v5.4.9 bugs, so it stops living on one box — and fixes the bug it found on its first in-repo run.

## `tools/tui-audit`

```bash
npm run build
npm run audit:tui              # or: npm run audit:tui -- --strict
```

Drives the **real** `startTuiApp()` through a fake TTY against a local stub proxy — real sockets, real SSE, real raw-mode key parsing, real tab mount/unmount — and audits every frame the app actually writes:

- **row width** — no line wider than `cols`
- **frame height** — no more physical rows than the terminal has
- **SGR balance** — per line *and* cumulatively across the frame

Twelve geometries, 200×50 down to **24×8**. Output to `.tui-audit/` (gitignored): stripped frame dumps per geometry plus `findings.json`.

The stub serves everything `ProxyClient` calls — `/health`, `/v1/models`, `/admin/resume`, `/analytics`, `/accounts`, and the `/analytics/stream` SSE feed with an `overage_halt` event.

**Not part of `npm test`** — ~25s and needs `dist/` built. `test/` keeps the fast pure-render assertions; this is the slow integration pass for render/layout changes. `--strict` exits non-zero on findings; the default is 0, because a finding usually needs a judgement call about which tab gives up what.

**Safety:** navigation keys only (Tab, arrows) — never `s`/`d`/`r`/`R`/Enter, so a run can't write config or POST `/admin/resume`. Stub binds **39456**, not 3456, so it can't collide with a real `dario proxy`.

## The bug it found

` Analytics  — last 60 min` is **25 characters against 24 columns**. The title is a `required` panel, so the row budget never clips it, and it had no `truncate` of its own — the one row in that tab still unbounded after #871.

## Why CI didn't already catch it

Worth recording, because it's the second time these two approaches have covered for each other:

`test/tui-frame.mjs` sweeps to 24×6 and asserts no row exceeds `cols` — it *should* have caught this. It didn't, because at six rows the body budget is one row, so the frame clamp swaps the title for the `… more rows` note **before its width can be measured**. At 24×8 there's room for the title to render, and overflow.

The reverse happened during the v5.4.7 audit: this harness missed `progressBar()` throwing `RangeError` on a negative width (Analytics crashed at any width ≤ 31) because its narrowest geometry was 40 columns — the fixture sweep caught that one.

Live runs are better evidence for what real data does; fixture sweeps are better for edge geometry. Neither alone has been sufficient. Both README and CHANGELOG record this so the next person doesn't assume one covers the other.

`test/tui-budget.mjs` now checks Analytics at 24/28/32/40 columns × 8 rows, so CI covers this class without needing the harness. Confirmed to fail on revert: `widest=25`.

## Verification

- `npm test` — **123/123**
- `npm run audit:tui` — **0 findings** across all twelve geometries (2 findings before the fix, both the title row at 24×8)

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes — 123/123
- [x] For changes that touch `proxy.ts`, `cc-template.ts`, or streaming behavior: n/a
- [x] No new runtime dependencies added (harness is zero-dep, stdlib only)
- [x] Version bumped (5.4.9 → 5.4.10) + CHANGELOG — required because this carries a runtime change to `analytics.ts`, not just tooling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ credentials.json
 drift-summary.md
 drift-output.txt
 label-target.txt
+.tui-audit/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.10] - 2026-07-26
+
+- **The TUI audit harness now lives in the repo** as `tools/tui-audit`, run with `npm run audit:tui`. It drives the real `startTuiApp()` through a fake TTY against a local stub proxy — real sockets, real SSE, real raw-mode key parsing, real tab mount/unmount — and audits every frame the app actually writes for row width, frame height and SGR balance, across twelve geometries from 200×50 down to 24×8. The stub serves every endpoint `ProxyClient` calls, on port 39456 so it can never collide with a real `dario proxy`. Navigation keys only, so a run can't write config or POST `/admin/resume`.
+
+  This is deliberately not part of `npm test` — it takes ~25s and needs a build. `test/` keeps the fast pure-render assertions; this catches what only appears when the whole loop runs against moving data.
+
+- **The Analytics title row overflowed a very narrow terminal** — ` Analytics  — last 60 min` is 25 characters against 24 columns. It's a `required` panel, so the row budget never clips it, and it had no `truncate` of its own. Found by the harness on its first in-repo run.
+
+  Worth recording why `test/tui-frame.mjs` missed it despite sweeping to 24×6: at six rows the body budget is one row, so the frame clamp swaps the title for the `… more rows` note before its width can be measured. At 24×8 there's room for the title to render — and overflow. `test/tui-budget.mjs` now checks Analytics at 24/28/32/40 columns by 8 rows, so CI covers it without needing the harness.
+
 ## [5.4.9] - 2026-07-26
 
 - **Every tab now spends its own row budget instead of being clipped from the bottom.** 5.4.7 stopped the app writing frames taller than the terminal, but that fix is a floor: `renderTui` clips whatever sorts *last*. On Status that was the Overage-guard panel — so the one tab that tells you the proxy has HALTED hid the halt, the cause, the auto-resume countdown and the resume instructions, on a default 80×24 terminal. Closes #868.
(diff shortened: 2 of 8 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean fix backed by a well-reasoned new test harness and CI is green.

**What I checked**
- `src/tui/tabs/analytics.ts`: the title row fix wraps the existing `truncate()` helper (already used elsewhere in this file, imported from `render.ts`) around the previously-unbounded title line at `analytics.ts` — `lines.push(truncate(' ' + brand('Analytics') + dim(...), w));`. `truncate()` correctly measures visible width (ANSI-aware) and re-flushes any open SGR codes from the clipped tail, so this doesn't reintroduce the "unclosed dim() bleeds into next line" class of bug the codebase already guards against.
- `test/tui-budget.mjs`: adds a regression check across `[24, 28, 32, 40]` cols × 8 rows asserting `widest <= cols` for the Analytics tab — directly covers the reported overflow and the PR states it fails on revert (`widest=25`), which is a good, verifiable regression guard.
- `tools/tui-audit/{audit.mjs,stub-proxy.mjs}`: new integration harness is read-only from a safety standpoint — it only ever sends Tab/arrow keys (`inp.send('\t')`, `inp.send('\x1b[B')`, `inp.send('\x1b[A')`), never the `s/d/r/R/Enter` keys that would write config or hit `/admin/resume`. Stub binds `39456` (not `3456`), avoiding collision with a real running proxy. Cleanup via `close()` clears its SSE interval timers and closes connections — no obvious leak.
- CI check-rollup: build (18/20/22), test, docker-cap-drop-smoke, CodeQL, actionlint, validate-package-json all green; `compat` was still in progress at fetch time but not a blocker signal from what's completed.
- Version bump (5.4.9 → 5.4.10) and CHANGELOG entry are present and match the diff's own checklist.

**Minor**
- `audit.mjs`'s `runGeometry` relies on fixed `wait()` sleeps (900ms/400ms/60ms/350ms) to sequence key sends against async network/SSE responses — inherently a bit timing-fragile in CI environments under load, though this harness is explicitly excluded from `npm test` for exactly this kind of reason (documented in the README), so it's a reasonable tradeoff rather than a defect.

Nothing here blocks merge.

PR #874, opened 2026-07-26
fix(scrub): catch instruction-file prose by its wrapper, not by the heading

### LINKED ISSUE #872: bake: the capture accepts any request on its MITM port with no proof it came from the spawned CC
The bake captures the **first** request that reaches its loopback MITM and treats it as CC's. Nothing proves that request came from the child the bake spawned. Once during the 5.4.x work a capture came back carrying text from the operator's own interactive session — prose that has no business in a captured CC system prompt.

**This is an observation, not a diagnosis.** It happened once in roughly six captures, I could not trigger it again, and I did not keep the bad artifact. Filing it because the blast radius is disproportionate to the effort of guarding against it.

## Why it matters

`src/cc-template-data.json` is not a local scratch file — it is committed, published to npm, and pushed to GHCR. The capture is the sole input to it. A capture that picks up somebody else's context can silently bake an operator's private prompt into a public package.

## The current artifact is clean — verified

Scanned `master`'s bundle (163,040 bytes; base 4759 plus fable 9205 / opus-5 8093 / sonnet-5 13448) for markers that could only come from an operator's own context. Two needles hit, both benign:

| hit | where it actually lives |
|---|---|
| `operator` ×5 | CC's own **PowerShell** tool description (`Pipeline chain operators && and \|\| are NOT available…`) |
| `interactively` ×1 | CC's own **Workflow** tool description (`interactively-authenticated MCP servers…`) |

No operator-private text is in the shipped bundle. This issue is about closing a hole, not cleaning up a leak.

## Two candidate mechanisms

**(a) No provenance check.** `runCapture` in `src/live-fingerprint.ts` binds `server.listen(0, '127.0.0.1')` and accepts any request whose URL merely `includes('/v1/messages')`. First one in wins and settles the promise. The ephemeral port makes a collision unlikely, not impossible, and nothing downstream can tell a foreign request from the child's.

**(b) Inherited context.** The child is spawned with `env: { ...process.env }` and an inherited cwd, so anything CC loads from user- or project-level instruction files is part of what it emits. `extractTemplate` keeps only `system_prompt` and `tools`, and instruction files arrive elsewhere in the body — which is very likely why this didn't reach the bundle. That's an incidental boundary, not an enforced one; it holds until `extractTemplate` starts keeping one more field.

Note that scrubbing cannot cover this. `scrubText` de-identifies — usernames, home paths, project slugs — and has no notion of provenance. Foreign prose containing no user paths passes `findUserPathHits` completely clean, which is exactly the case that would slip through.

## Proposed guards

Cheapest first. The third is the only deterministic one, and worth adding on its own merits even if the other two are judged unnecessary.

1. **Context-bleed deny-list (deterministic, do this one).** CC wraps user/project instruction files in a stable preamble — `Contents of …CLAUDE.md`, `IMPORTANT: These instructions OVERRIDE any default behavior`. A capture whose prompt contains that wrapper is carrying somebody's private instructions; reject it and fail the bake loudly. No heuristics, no false positives on a genuine CC prompt.

2. **Structural anchor assertion.** Require the captured system prompt to contain CC's known anchors before accepting it. Catches wholesale substitution, not partial bleed.

3. **Provenance nonce.** The capture already injects a sentinel key (`sk-dario-fingerprint-capture`) when the operator has none — make it a per-capture random value and require the captured request's `x-api-key` to match. Caveat: when a real `ANTHROPIC_API_KEY` is present the code passes it through deliberately, so this needs a fallback for that path rather than unconditionally overriding the operator's auth. Worth designing carefully rather than rushing, since changing what the child authenticates with could change what CC sends.

All three are local to `runCapture` / `extractTemplate` and none change the proxy's runtime behaviour.

## Not asking anyone to reproduce it

Guard 1 is worth landing whether or not the original observation is ever reproduced — it fails closed, costs one string check, and turns a silent bake into a loud abort. If it never fires, that is a fine outcome.

Labelled `pinned` deliberately: this describes unfinished work, and an unlabelled sibling issue was closed as COMPLETED earlier today with no comment and no implementing commit.

askalf:
Closes #872 — guard 1, the deterministic one. Guards 2 and 3 are not in this PR; see the bottom.

## What was actually missing

Less than #872 assumed. The scrub already handles this class: `removeHostContextSections` strips CC's `# claudeMd`, `# auto memory`, `# userEmail`, `# currentDate` and `# Environment` sections, and `findUserPathHits` flags any that survive so the release gate fails rather than shipping a leak (#642-audit).

The gap is that **both halves read the same heading list**. Rename or reshape the heading and `removeSection` strips nothing while the heading detector flags nothing — two silent failures stacked — and once paths are scrubbed the leftover prose carries no user path for the other detectors to see. The existing code comment already worried about "a renamed heading"; nothing acted on it.

That is the shape of what #872 observed: prose with no path in it, arriving somewhere the section detector wasn't looking.

## The change

`findUserPathHits` now also matches the wrapper prose itself, so the heading no longer has to be intact:

- `Codebase and user instructions are shown below`
- `IMPORTANT: These instructions OVERRIDE any default behavior`
- `Contents of <path>.md` (on one line)
- `user's private global instructions` — straight and typographic apostrophe
- `auto-memory, persists across conversations`
- `Here are useful instructions from`

Extending that function rather than adding a parallel one means both existing gates arm themselves with no new call sites: the bake fails hard on `JSON.stringify(scrubbed)` (capture-and-bake.mjs:119) and `check-cc-drift` fails the release (check-cc-drift.mjs:361). The variant path keeps its existing warn-and-keep-previous behaviour, which is right — a variant is optional, the base is not.

Both failure messages said "residual user paths". A hit is no longer necessarily a path, and mis-naming one is what sent the original investigation looking at model pricing, so they now say "host content" and name the three possibilities.

## Marker selection was measured, not guessed

I probed every candidate against the real baked template before committing to any. Two obvious ones are disqualified:

| candidate | agent_identity | system_prompt | fable | opus-5 | sonnet-5 | tool descriptions |
|---|---|---|---|---|---|---|
| `CLAUDE.md` | 0 | **1** | **1** | **1** | **1** | **3** |
| `<system-reminder>` | 0 | **2** | **1** | **1** | **1** | **1** |

CC's own system prompt and tool descriptions mention both. Either as a marker would have failed **every** bake from the first run — a guard that fires on legitimate content is worse than no guard. `test/context-bleed.mjs` pins them as explicit non-markers so nobody "tightens" the list and breaks the pipeline.

## Deliberately not in `extractTemplate`

`extractTemplate` serves the live-capture path too, and `refreshLiveFingerprintAsync` writes the capture to `~/.dario/cc-template.live.json` **unscrubbed on purpose** — the live fingerprint exists to replay the operator's own CC install faithfully, host context included. Gating there would break that by design. Only the bake publishes, so only the bake gates.

## Verification

`test/context-bleed.mjs`, 21 assertions in four groups: the prose is caught; the renamed-heading case is caught by the content markers and provably *not* by the heading detector; every baked slot plus the whole serialized bundle is clean; and the excluded markers stay excluded.

Full suite **124/124**.

Ran a real `capture-and-bake --check` with the guard armed, on a machine that has both a user-level and a project-level `CLAUDE.md`. **The guard stays silent** — because CC delivers instruction files in `messages` and the bake only keeps `system[1]` and `system[2]`. That is the same boundary #872 identified as the reason the original bleed never reached a bundle, now confirmed from the other direction: no false positive on a live capture that has plenty of instruction content available to bleed.

## Two things this does not do

**Guard 3 (the provenance nonce) is not here.** #872 already noted the caveat: when a real `ANTHROPIC_API_KEY` is present the capture passes it through deliberately, so a nonce needs a fallback rather than unconditionally overriding the child's auth, and changing what the child authenticates with can change what CC sends. That deserves its own PR with its own live check, not a rider on this one. Guard 2 (anchor assertion) only catches wholesale substitution, which the tool-count and block-shape checks in `extractTemplate` already largely cover.

**The bundled template is stale and I did not re-bake it.** That `--check` run reports genuine drift — `afk-mode-2026-01-31` back on (it is remote-config controlled and flips), sonnet-5 variant 13448 → 13442, and +279 chars in the base prompt. Re-baking here would collide with the cc-drift bot, which CLAUDE.md says not to compete with. `src/cc-template-data.json` is untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.11] - 2026-07-26
+
+- **A bake can no longer ship instruction-file prose that the section strip missed.** `scrubText` already removes CC's host-context sections and `findUserPathHits` already flags any that survive, but both read the same heading list — so a heading CC renames or reshapes strips nothing and flags nothing, and once paths are scrubbed the leftover prose carries no user path for the other detectors to catch. That is two silent failures stacked, and it is the shape of what dario#872 saw: one capture in six came back carrying another session's instruction text. `findUserPathHits` now also matches the wrapper prose itself — the `Codebase and user instructions are shown below` and `IMPORTANT: These instructions OVERRIDE` sentences, the `Contents of <path>.md` line, the private-instruction and auto-memory annotations — so the heading no longer has to be intact for the leak to be caught. Both existing gates pick it up for free: the bake fails hard on the serialized template and `check-cc-drift` fails the release.
+
+  Marker selection was empirical, not guesswork. Bare `CLAUDE.md` and `system-reminder` are deliberately NOT markers — CC's own system prompt mentions both (2 hits) and its tool descriptions mention them again (3 and 5), so either would have failed every bake from the first run. `test/context-bleed.mjs` pins that, asserts every baked slot is clean, and asserts the renamed-heading case that the heading detector provably misses.
+
+  Not wired into `extractTemplate`, on purpose: that function also serves the live-capture path, and the live fingerprint keeps host context deliberately — it exists to replay the operator's own CC install faithfully. Only the bake publishes, so only the bake gates. Verified against a real `--check` run on a machine that has both a user-level and a project-level CLAUDE.md: the guard stays silent, because CC delivers instruction files in `messages` and the bake only keeps `system`.
+
 ## [5.4.10] - 2026-07-26
 
 - **The TUI audit harness now lives in the repo** as `tools/tui-audit`, run with `npm run audit:tui`. It drives the real `startTuiApp()` through a fake TTY against a local stub proxy — real sockets, real SSE, real raw-mode key parsing, real tab mount/unmount — and audits every frame the app actually writes for row width, frame height and SGR balance, across twelve geometries from 200×50 down to 24×8. The stub serves every endpoint `ProxyClient` calls, on port 39456 so it can never collide with a real `dario proxy`. Navigation keys only, so a run can't write config or POST `/admin/resume`.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.4.10",
+  "version": "5.4.11",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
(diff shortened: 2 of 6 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-tested fix. Approving.

**What I checked**
- `gh pr diff 874` (full unified diff, all 6 changed files).
- CI check-rollup (`statusCheckRollup`): all checks (build ×3, test, CodeQL, compat, actionlint, docker-cap-drop-smoke) completed with `SUCCESS`; `auto-merge` skipped as expected for a non-bot-flagged PR.
- Author identity: `askalf` / Thomas Sprayberry (`is_bot: false`) — distinct from the reviewer bot identity, so no self-approval conflict.

**Correctness**
The core change in `src/scrub-template.ts` adds `INSTRUCTION_INJECTION_MARKERS`, a bounded set of non-global regexes matched against the serialized template, e.g.:

```
{ label: 'instruction-file heading', re: /\bContents of [^\n]{1,300}\.md\b/ },
```

The `[^\n]{1,300}` bound avoids catastrophic backtracking/ReDoS risk, and the loop in `findUserPathHits` (`test/context-bleed.mjs` diff, `src/scrub-template.ts` new loop at end of the function) correctly pushes every matching marker without mutating shared regex state (no `g` flag, so no `lastIndex` bugs across repeated calls).

The straight/typographic apostrophe alternation (`user['’]s private global instructions`) and the disqualification of `CLAUDE.md` / `<system-reminder>` as markers (verified against the real baked template, per the PR body's probe table) show the marker set was chosen empirically rather than guessed — the negative-case tests in `test/context-bleed.mjs` section 4 back this up.

The decision to extend `findUserPathHits` rather than add a parallel function means both existing gates (`capture-and-bake.mjs:119`, `check-cc-drift.mjs:361`) pick up the new detection with no new call sites — reduces surface area for a wiring mistake.

**Test coverage**
`test/context-bleed.mjs` (135 lines, new) covers: positive detection on a realistic wrapped sample, the renamed-heading case the old heading-only detector would miss, zero false positives across every baked slot (base/variants/tool descriptions/full serialized bundle), and the deliberately-excluded markers. `package.json`'s `test:serial` script is updated to include the new test file in the run order.

**What's good**
- Error/warning messages updated in `capture-and-bake.mjs` and `check-cc-drift.mjs` to say "host content" instead of "residual user paths," accurately reflecting the widened detection scope.
- Clear documentation in the new JSDoc explaining why the guard is deliberately *not* wired into `extractTemplate` (live-capture path intentionally preserves host context by design).
- PR body includes verification against a real `capture-and-bake --check` run, confirming no false positive in practice.

No blocking findings — approving.

PR #875, opened 2026-07-26
feat(capture): prove the captured request came from the CC we spawned

askalf:
Completes the last guard from #872. Guard 1 landed in #874; #872 closed with it, so this is the remainder rather than a new issue.

## The hole

`runCapture` accepted the first request whose URL merely contained `/v1/messages`, with nothing tying it to the CC child it had just spawned. The port is ephemeral so a collision is improbable rather than impossible — but "improbable" is the whole of the defence, and once a foreign request was captured nothing downstream could tell it from the child's.

## The nonce goes in the URL, not the API key

#872 proposed a nonce in `ANTHROPIC_API_KEY` and flagged the caveat: the capture passes a real key through deliberately, so overriding it could change what CC sends. I probed the installed CC before writing anything, and the measurement redirected the design:

```
base url given to CC:  http://127.0.0.1:60370/dario-capture-probe-abc123
  request 1: /dario-capture-probe-abc123/api/hello
  request 2: /dario-capture-probe-abc123/v1/messages?beta=true
  authorization: Bearer sk-an…      x-api-key: (absent)
```

Two results:

1. **CC honours a path segment in `ANTHROPIC_BASE_URL`.** It appends its own paths beneath it, and probes `/api/hello` first — which still 404s here, as before.
2. **On a subscription install CC authenticates with `authorization: Bearer`, and sends no `x-api-key` at all.** A key-borne nonce would therefore do nothing on OAuth installs while changing the auth path for API-key ones. Changing what the child authenticates with is exactly what a fingerprint capture must not do.

So the nonce rides in the path: `ANTHROPIC_BASE_URL=http://127.0.0.1:<port>/<nonce>`, and `isOwnCaptureRequest` requires that nonce as a **prefix** plus `/v1/messages`. Nothing about the child's auth changes.

Fails closed. No nonce, no capture — `runCapture` returns null and the bake exits non-zero, which is the right outcome for a request we cannot attribute. A `/v1/messages` that arrives without the nonce now logs, because a silent 404 there is precisely how a foreign capture went unnoticed the first time.

The nonce is also stripped from the recorded `CapturedRequest.path`. Nothing bakes that field today, but a per-capture random value has no business travelling further than it must.

## Guard 2 is not here, on purpose

The anchor assertion #872 listed third is now redundant rather than deferred. `extractTemplate` already requires two system blocks, a non-empty `agent_identity` and `system_prompt`, and at least one non-`mcp__` tool. Guard 2 existed to catch wholesale substitution by some other client — and guard 3 makes that unreachable by construction, because a request that isn't ours is never captured at all. What remains is "our own child sent something structurally odd", which the existing non-empty checks cover and the drift detector reports on. Adding it would be ceremony.

## Verification

`test/capture-provenance.mjs`, 14 assertions. The ones that matter beyond the happy path:

- a bare `/v1/messages` with no nonce is rejected — the #872 case
- `/someone-else/<nonce>/v1/messages` is rejected, so the nonce appearing *anywhere* in the path is not enough; anything that could echo it back cannot defeat the check
- another capture's nonce is rejected
- an empty nonce accepts nothing, rather than degrading into matching `//v1/messages`
- `/<nonce>/api/hello` and `/<nonce>/v1/models` are not captures

**A real `capture-and-bake --check` captured all four prompts through the nonce path** — base 5038, fable 9205, opus-5 8093, sonnet-5 13442, CC v2.1.220. No null capture, no foreign-request warnings. That was the make-or-break test: if CC had not honoured the subpath, every capture would have timed out and the bake would have broken.

Full suite green.

## The bundle is still stale, deliberately

That `--check` reports genuine drift — `afk-mode-2026-01-31` back on (it is remote-config controlled and flips), sonnet-5 variant 13448 → 13442, +279 chars in the base prompt. The re-bake is a separate PR on top of this one, so the bundle is produced by the final guard code rather than an intermediate state. `src/cc-template-data.json` is untouched here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.12] - 2026-07-26
+
+- **The capture now proves the request came from the CC child it spawned.** `runCapture` accepted the first request whose URL merely contained `/v1/messages`, with nothing tying it to the child — the ephemeral port was the whole of the defence, and once a foreign request was captured nothing downstream could tell it from the child's (dario#872). `ANTHROPIC_BASE_URL` now carries a per-capture nonce and the MITM only accepts a path with that nonce as its prefix; anything else 404s, and a `/v1/messages` without it logs rather than passing unnoticed. Fails closed — no nonce, no capture, and the bake exits non-zero.
+
+  The nonce rides in the URL rather than in `ANTHROPIC_API_KEY`, which #872 had proposed. Measured first: CC honours a path segment in `ANTHROPIC_BASE_URL` (it requests `/<nonce>/v1/messages?beta=true`, probing `/<nonce>/api/hello` first), and on a subscription install it authenticates with `authorization: Bearer` and sends no `x-api-key` at all — so a key-borne nonce would do nothing there while changing the auth path for API-key installs. Changing what the child authenticates with is the one thing a fingerprint capture must not do. Verified with a real `--check`: all four prompts captured through the nonce path (base 5038, fable 9205, opus-5 8093, sonnet-5 13442).
+
+  The third guard #872 listed, a structural anchor assertion, is redundant rather than deferred: `extractTemplate` already requires two system blocks, a non-empty identity and prompt, and at least one non-`mcp__` tool, and the nonce makes wholesale substitution unreachable — a request that is not ours is never captured at all.
+
 ## [5.4.11] - 2026-07-26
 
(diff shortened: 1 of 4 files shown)


sprayberry-redline (approved):

**What I checked**
- `src/live-fingerprint.ts`: the new `isOwnCaptureRequest(url, nonce)` requires the nonce as a path *prefix* (`url.startsWith(`/${nonce}/`)`) AND `/v1/messages` present in the URL. Traced this against the new test cases in `test/capture-provenance.mjs` — prefix-only match (`/someone-else/<nonce>/v1/messages` → rejected), empty-nonce degenerate case (`nonce.length === 0` short-circuits to `false`, so it can't collapse into matching `//v1/messages`), and another capture's nonce being rejected all check out against the implementation.
- The path-stripping fix at `src/live-fingerprint.ts` (`path: (req.url ?? '/v1/messages').replace(`/${nonce}`, '')`) only strips the leading `/${nonce}` segment since the nonce is guaranteed to be a prefix by the guard above it — no double-strip or mid-string collision risk since the nonce is a fresh random value per capture.
- `test/capture-provenance.mjs` is a plain `.mjs` file under `test/`, and `test/all.test.mjs`'s driver auto-wraps every `test/*.mjs` file (except explicitly opted-out E2E files) as a subtest — so this is exercised by `npm test` without needing an explicit registration, and it's also added directly to the `test:serial` chain in `package.json`.
- CI check-rollup is fully green (CodeQL, build ×3 node versions, compat, test, docker-cap-drop-smoke, actionlint, validate-package-json all `pass`).
- Fail-closed behavior (no nonce/URL → `false` → 404, foreign `/v1/messages` logged rather than silently dropped) matches the stated intent and is covered by the "fails closed" test group.

**What's good**
- Good adversarial test coverage: prefix-vs-substring nonce placement, empty nonce, another capture's nonce, and CC's actual probe shapes (`?beta=true`, `/count_tokens`) are all exercised.
- The PR body documents the key design decision (nonce in URL path vs. `ANTHROPIC_API_KEY`) with a measured rationale rather than an assumption, which matches what the diff actually does.

No correctness, security, reuse, or coverage issues found in the changed lines.

PR #869, opened 2026-07-26
auto-rebake: template drift detected 2026-07-26

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-26T04:51:01+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.4.8` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — worth a closer read

- **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.220) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.220, 27 tools, 6176 char system prompt
[bake] scrubbed:
[bake]   tools: 27 → 27 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6176 → 4759 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] previous baked template: CC v2.1.220 captured 2026-07-26T01:33:28.732Z, 33 tools, 4759 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4759 → fable 9205 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4759 → opus-5 8093 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4759 → sonnet-5 13448 chars
[bake] check: drift detected — 1 differing slot (verdict: moderate):
[bake] **Verdict:** 🟡 Moderate — worth a closer read
[bake] 
[...]

PR #876, opened 2026-07-26
chore(template): re-bake against CC v2.1.220 — --check exits 0

askalf:
Re-bake against CC v2.1.220. `--check` reports **no drift** and exits 0.

Baked with the #874 and #875 guards both in the tree — deliberately last, so the capture behind this bundle was provenance-checked by the nonce and the instruction-file detector was armed and silent throughout. That was the whole reason for holding it rather than baking earlier in the sequence.

## What moved

Four insertions, four deletions, one file.

| slot | before | after |
|---|---|---|
| base prompt | 4759 | **4759 — unchanged** |
| `anthropic_beta` | no afk-mode | **+ `afk-mode-2026-01-31`** |
| sonnet-5 variant | 13448 | **13442** |
| fable / opus-5 variants | 9205 / 8093 | unchanged |

Canonicalization from #865 holds: three canonical `/home/user/.claude/projects/project` paths, zero Windows / `/root` / `/Users` forms, zero username leaks. So this bundle is byte-reproducible on CI's Linux host and there is no repeat of the #854 ↔ #860 ping-pong.

## The base prompt was never drifting

Successive `--check` runs disagreed on the **raw** captured length — 7037, 6799, 6733, 6715 — which reads like an unstable prompt, and I initially called it flapping. Eight consecutive captures then came back byte-identical:

```
sample 1..8: raw=6715 scrubbed=4759 tools=30 afk-mode=true
distinct scrubbed base lengths: [4759]
```

The raw movement is `# gitStatus` — branch, modified files, recent commits — changing as I committed through the session. That section is exactly what the scrub removes, and the scrubbed output has matched the bundle the whole time. My flapping call was wrong; measuring it was still worth the ten minutes, because it is what turned the re-bake from a coin-flip into a known three-slot change.

One observation stays unexplained: a single earlier `--check` scrubbed to 5038 (+279 chars, a `# Context management` paragraph). It has not reproduced across the ~14 captures since. Recording it rather than dressing it up.

## afk-mode is deliberately not suppressed

That flag genuinely flaps — absent when #869 re-baked at 04:51Z, present, absent, then present in 8/8 samples. Four state changes in eleven hours, so the drift gate **will** fire on this slot again.

It would be easy to add it to `MODEL_CONDITIONAL_BETAS` and silence that. It would also be wrong, and the distinction is the point:

- those betas are ones **dario itself appends per-request** (`context-1m` rides `[1m]`, `fallback-credit` rides fable), so comparing them against a base that never carried them is a false positive we manufactured — #484
- `afk-mode` is one **CC sends**, so its presence is real wire shape

`betaForModel`'s own contract already settles it: *"afk-mode-2026-01-31 is remote-config controlled and can flip within a Claude Code version, so its presence is a property of `base` (the loaded template), not of this function — the template refresh keeps it current."* The runtime already absorbs the consequence — afk-mode is rejected on Max 5x, and `unavailableBetas` parses the flag out of the 400, strips it, retries once and caches it per account — and the inverse is free, since *"removing a beta can never provoke an upstream 400."*

So a future drift alarm on this slot is the detector doing its job on a genuinely changing value, not the host-dependent ping-pong #865 eliminated. Suppressing it would mean lying about the wire.

## Verification

`--check` exit 0, no drift. Full suite 125/125. Bundle scanned for residual host content — clean on every slot including tool descriptions and the whole serialized file.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean re-bake, approving.

**What I checked**
- Diff (`gh pr diff 876`): three files — `CHANGELOG.md`, `package.json`, `src/cc-template-data.json`.
- `package.json` version bump (`5.4.12` → `5.4.13`) matches the new `## [5.4.13]` CHANGELOG entry — consistent.
- `src/cc-template-data.json` changes are captured template data (timestamp, one tool description string, `anthropic_beta` gaining `afk-mode-2026-01-31`, one prompt-variant string) — not hand-written logic, so there's no control-flow to break.
- CI check-rollup is fully green: `CodeQL`, `wire-drift`, `compat`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `validate-package-json`, `actionlint`, `triage`, `analyze` all pass. `wire-drift` passing is the relevant signal for "no unintended drift in the captured bundle."
- PR body's own reasoning about the `afk-mode` beta and the raw-vs-scrubbed length discrepancy is internally consistent with the diff shown (the beta string literally gains `,afk-mode-2026-01-31` at end).

**What's good**
- Changelog entry gives a clear, falsifiable account of what moved and why (three slots), makes the reasoning for *not* suppressing the flapping `afk-mode` beta explicit, and is corroborated by the diff itself.

No correctness, security, reuse, or coverage findings on the changed lines — this is a mechanical data re-bake with matching version/changelog and a green CI rollup.

PR #877, opened 2026-07-26
fix(config): make the documented env off-switches actually switch off

askalf:
Started as a README accuracy pass. Turned up a bug on the way, so it fixes that too.

## The bug: two documented env off-switches did nothing

`cli.ts` has advertised these since v4.1:

```
//   DARIO_OVERAGE_GUARD=off            → env equivalent of --no-overage-guard
//   DARIO_OVERAGE_NOTIFY=off           → env equivalent of --no-overage-notify
```

Neither worked. `parseBooleanEnv` returns `true` or `undefined` and **never `false`**:

```
parseBooleanEnv("off")   = undefined
parseBooleanEnv("0")     = undefined
parseBooleanEnv("false") = undefined
```

Both feed `flag ?? env ?? fileCfg ?? true`, so `off` fell through to the default and **the overage guard stayed enabled**. No warning, no log line. The people it hit are the ones with no alternative: a CLI flag was the only working way to disable it, which rules out Docker, Compose, k8s and systemd — exactly the deployments that reach for an env var.

The truthy-only behaviour is deliberate, not an accident — `test/strict-template-flags.mjs` asserts `parseBooleanEnv('off') === undefined`. So this adds `parseTriStateEnv` and uses it **only** at those two sites, rather than widening the shared parser and flipping four existing assertions. The other three call sites (`DARIO_STEALTH`, `DARIO_NO_LIVE_CAPTURE`, `DARIO_STRICT_TEMPLATE`) read their env with `||`, where `false` and `undefined` are indistinguishable, so they neither need the change nor notice it.

Unrecognised values still return `undefined` at both parsers, so a typo defers to the config file and the default instead of being guessed either way. `test/strict-template-flags.mjs` now covers that too: 16 assertions → 33.

## New: docs/configuration.md

**29 environment variables were documented nowhere** — not the README, not `docs/`. The whole overage-guard set, the request-queue knobs, template fidelity, the pacing axes, and the env forms of flags the README already documents. The README's own "Overage guard" section named none of its four variables.

Grouped by task rather than by flag, because the audience is the deployment that can't pass a flag. It also states the precedence order and which vars are not part of the supported surface.

**Deliberately not a second copy of `commands.md`.** That page is already billed as the "full flag/env reference" and covers 29 vars of its own, so I measured the overlap before writing:

| | count |
|---|---|
| in both | 10 |
| only in `commands.md` (OAuth overrides, session rotation, TLS, CORS, `PACE_*`) | 19 |
| only in the new page | 29 |

Neither was a superset. I cut all 10 duplicates out of the new page and cross-linked both ways, so **overlap is now zero** and no default is stated twice. Two copies of a default is one that goes stale.

## Four stale README numbers

| claim | actual |
|---|---|
| per-model beta sets "opus 9, sonnet 8, haiku 6" | **opus 10, sonnet 9**, haiku 6 |
| "~22k lines across **49** files" | **23,833 lines across 52** |
| "**113 test files**" | **132** in `test/`, **125** under the parallel runner |
| "0 runtime dependencies" | correct, re-verified |

The beta counts moved **today** — 5.4.13's re-bake added `afk-mode` to the base set, lifting every per-model set by one. That line now says the counts track the baked base rather than freezing a number, so the next base change doesn't re-stale it. The test-file line now distinguishes 132 files from 125 runner subtests, the delta being the e2e/compat/stealth suites that opt out and have their own entry points.

Also documented the six drift and audit runners that existed in `package.json` and nowhere else: `drift:wire`, `drift:sdk`, `audit:tui`, `check:overage`, `stress`, `cch:calibrate`.

## What I checked and did not change

The model lineup is current — Fable 5 / Opus 5 / Sonnet 5 / Haiku 4.5, `[1m]` variants, the `opus48`/`47`/`46`/`sonnet46` pins. The TUI section is accurate: its six tabs match `src/tui/tabs/` exactly and the mock already shows opus-5 / sonnet-5 / haiku-4-5. Capabilities and the billing-split section hold up.

One correction I made mid-write, worth recording because it nearly shipped: my first draft listed `DARIO_DIR`, `DARIO_DNS_RESULT_ORDER`, `DARIO_MODEL_CATALOG_TTL_MS`, `DARIO_CCH` and `DARIO_USAGE_PORT` as "not part of the supported surface". Checking each individually, `DARIO_DIR` **is not an environment variable at all** — it's a `const` in `accounts.ts` that my `DARIO_[A-Z_]+` inventory matched by name — and the other four are real knobs with real defaults that now have proper sections. Only `DARIO_LIVE_TEMPLATE_CACHE` and `DARIO_TEST_URL` are genuinely test-only.

Full suite 125/125.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.14] - 2026-07-26
+
+- **`DARIO_OVERAGE_GUARD=off` and `DARIO_OVERAGE_NOTIFY=off` now actually turn those features off.** `cli.ts` has documented both as the env equivalents of `--no-overage-guard` / `--no-overage-notify` since v4.1, but `parseBooleanEnv` returns `true` or `undefined` and never `false` — so `off` fell straight through `flag ?? env ?? fileCfg ?? true` and the guard stayed enabled, silently. Container deployments were the only ones affected, and the worst ones to affect: a flag was the sole working way to disable it. New `parseTriStateEnv` handles `off|0|false|no` as well as `on|1|true|yes`, and is used only at those two sites. `parseBooleanEnv` keeps its truthy-only contract — the three call sites that read it (`DARIO_STEALTH`, `DARIO_NO_LIVE_CAPTURE`, `DARIO_STRICT_TEMPLATE`) use `||`, where `false` and `undefined` are indistinguishable, and `test/strict-template-flags.mjs` pins that behaviour. An unrecognised value still yields `undefined` at both, so a typo defers to the config file and the default instead of guessing.
+
+- **New [`docs/configuration.md`](docs/configuration.md).** 29 environment variables that appeared in neither the README nor `docs/`: the whole overage-guard set, the request-queue knobs, template fidelity, the pacing axes, and the env forms of flags the README already documents. Grouped by task rather than by flag, because the audience is Docker / Compose / k8s / systemd, where a flag is not available. Also states the precedence order and which variables are not part of the supported surface. `commands.md` remains the per-flag reference and the two cross-link; nothing is documented in both, since two copies of a default is one that goes stale.
+
+- **Corrected four stale self-referential claims in the README.** Per-model `anthropic-beta` counts were opus 9 / sonnet 8 / haiku 6 and are now 10 / 9 / 6 — they moved with 5.4.13's re-bake, which added `afk-mode` to the base, so the line now says the counts track the baked base instead of freezing a number. Source is 23,833 lines across 52 files, not ~22k across 49. `test/` holds 132 files of which 125 run under the parallel runner, not 113. Zero runtime dependencies re-verified and unchanged. Also documented the six drift and audit runners (`drift:wire`, `drift:sdk`, `audit:tui`, `check:overage`, `stress`, `cch:calibrate`) that existed in `package.json` and nowhere else.
+
 ## [5.4.13] - 2026-07-26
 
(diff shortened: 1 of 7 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-tested fix + docs pass. Approving.

**What I checked**
- `src/cli.ts`: new `parseTriStateEnv` (src/cli.ts:831-846) correctly returns `false` for `off|0|false|no`, `true` for `on|1|true|yes`, and `undefined` otherwise (typo/unset defers to config file/default rather than guessing). It's wired in at exactly the two call sites that need a real `false` — `overageGuardEnabledFromEnv` and `overageNotifyFromEnv` (src/cli.ts:608, src/cli.ts:626) — both consumed via `?? … ?? true` chains, which is exactly the shape that silently swallowed `undefined` before. The other three `parseBooleanEnv` call sites (stealth, no-live-capture, strict-template) are untouched, matching the PR's own reasoning that those are read with `||` where `false`/`undefined` are indistinguishable.
- `test/strict-template-flags.mjs`: new block asserts off/0/false/no/OFF (case)/whitespace → false, on/1/true/yes → true, and unset/empty/typo → undefined, plus the actual bug scenario chained through a `?? … ?? true` default. `parseBooleanEnv`'s existing truthy-only contract assertions are untouched, so the shared parser's behavior for the other three call sites isn't disturbed.
- Docs (`README.md`, `docs/configuration.md`, `docs/commands.md`, `CHANGELOG.md`): cross-links are two-way and the PR body's own overlap audit (10 shared vars cut from the new page) checks out against what's rendered in the diff — no numbers in the diff contradict each other.
- CI rollup (`gh pr checks`): `build`, `test`, `docker-cap-drop-smoke`, `CodeQL`, `actionlint`, `validate-package-json` all green; `compat` was still `IN_PROGRESS` at review time — not a blocking signal, just noting it wasn't observed to completion.

**Minor**
- `package.json` version bump to `5.4.14` matches the new `CHANGELOG.md` entry — consistent.

No correctness, reuse, or efficiency concerns in the changed lines.

PR #878, opened 2026-07-26
auto-rebake: template drift detected 2026-07-26

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-07-26T16:45:15+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.4.15` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — worth a closer read

- **anthropic_beta removed:** `afk-mode-2026-01-31` (CC opted out of a feature flag)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.220) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.220, 27 tools, 6168 char system prompt
[bake] scrubbed:
[bake]   tools: 27 → 27 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6168 → 4759 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] previous baked template: CC v2.1.220 captured 2026-07-26T15:47:41.811Z, 33 tools, 4759 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4759 → fable 9205 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4759 → opus-5 8093 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4759 → sonnet-5 13448 chars
[bake] check: drift detected — 1 differing slot (verdict: moderate):
[bake] **Verdict:** 🟡 Moderate — worth a closer read
[bake] 
[...]

PR #880, opened 2026-07-26
fix(drift): stop the two rebake loops — suppress afk-mode, gate non-Linux bakes

askalf:
Two slots were re-baking on a loop. They turned out to have different causes and need different fixes — only one of them is suppressible.

Bundle changes ride #878, which merged first and is already the Linux-canonical state. This PR is mechanism only.

## afk-mode: suppressed

`afk-mode-2026-01-31` is remote-config gated, not version gated. On 2026-07-26 it moved four times in twelve hours:

| time | state |
|---|---|
| 04:51Z | off — auto-rebake #869 |
| 15:45Z | on — 8/8 captures, baked into 5.4.13 |
| 16:45Z | off — auto-rebake #878 |

Each flip opened a rebake PR, bumped a version, and cut a full release: multi-arch GHCR build, Sigstore attestation, npm publish, box autodeploy. Release churn on Anthropic's remote-config cadence.

New `REMOTE_CONFIG_CONDITIONAL_BETAS` strips it from the baked base and excludes it from the drift comparison on both sides. Verified against `computeDrift`:

| case | drift entries |
|---|---|
| present → absent | **0** |
| absent → present | **0** (symmetric) |
| a genuine new beta | **1**, reported as an addition |

**A separate set from `MODEL_CONDITIONAL_BETAS`, not a quiet widening of it**, because the reasons differ and the distinction is the whole argument. Those are flags *dario* appends per-request, so comparing them against a base that never carried them is a false positive we manufactured (#484). This one *CC sends* — suppressing it is a real, if small, divergence from wire fidelity, taken because the alternative is a release per flip.

Safe in both directions, and this is checked in the code rather than assumed: `betaForModel` documents that its per-family shape holds whether or not the base carries afk-mode (anchor-relative inserts degrade to append/no-op), `removing a beta can never provoke an upstream 400`, and afk-mode is the side with a cost — it is rejected on Max 5x, absorbed by `unavailableBetas` at one 400 per account. A base without it is strictly cheaper than a base that oscillates.

## sonnet-5: not suppressible — it was bake-host divergence

I went looking for a second oscillator and found something else. Diffing master's variant against #878's:

```
master (Windows) : ", Glob, Grep"     "the Glob or Grep"
pr878  (Linux)   : (removed)          "`find` or `grep` via the Bash tool"
```

CC's sonnet-5 prompt **names the tools that exist on the bake host**. Six bytes, one slot; base, `agent_identity`, fable and opus-5 are byte-identical across both bakes. So a Windows bake and CI's Linux bake each report the other as drift, indefinitely.

Third channel for the same disease: **#854 → #860 was paths, #856 was the OS/arch headers, this is prose that mentions the tool list.**

Unlike those two it cannot be normalised. The correct text depends on which tools the *caller* has, which the bundle cannot know, and rewriting CC's prose would be inventing wire shape instead of replaying it. `PLATFORM_ONLY_TOOLS` already unions the tools *array* — both bundles carry 33 names including Glob/Grep/PowerShell — but nothing can union a sentence.

So the fix is procedural, mirroring the existing `--allow-older-cc` guard: a write bake from a non-Linux host refuses with an actionable message, `--allow-non-linux-bake` overrides for a deliberate local bake, and **`--check` is not gated at all** — detecting drift from any platform is useful and harmless; only *writing* the bundle is restricted.

Verified both directions on this Windows box:

```
write bake  -> refused, exit 1, points at cc-drift-template-watch
--check     -> exit 0, unaffected
```

And the guard's value shows in the current state: `--check` from here against #878's Linux bundle now reports **exactly one** differing slot, sonnet-5, correctly attributed to my host — with afk-mode absent from the report entirely. Previously that would have been an invisible ping-pong; now it is a local-only artifact I am prevented from baking in.

## Two assertions updated, deliberately

Cases 34 and 38 both used afk-mode as their stand-in for "a real base beta" — precisely the role it no longer plays. Both now use `advisor-tool-2026-03-01`, which still is one, so each case tests what it was written to test. Each carries a note recording what it previously claimed. Case 38 also gains the inverse assertion, that afk-mode *is* stripped.

Full suite 125/125; `bake-drift-report` 111 assertions.

## Still open, and not fixed here

The base prompt occasionally scrubs to **5038 instead of 4759** — a +279 char `# Context management` paragraph. Two occurrences today against **20 consecutive clean captures** (8 in one sampler run, 12 in another), so roughly 1-in-10 and not reproducible on demand. It will occasionally trigger a CI rebake.

I also want to correct something I wrote in 5.4.13: I called the base prompt stable on the strength of 8 identical samples. That was a clean streak, not stability. Filing the observation separately rather than leaving it in a changelog entry as settled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.16] - 2026-07-26
+
+- **`afk-mode-2026-01-31` no longer drives a release every time Anthropic flips it.** It is remote-config gated, not version gated, and on 2026-07-26 it moved four times in twelve hours: off when #869 re-baked at 04:51Z, on for 8/8 captures and baked into 5.4.13, off again in #878 at 16:45Z. Each flip opened a rebake PR, bumped a version, and cut a full release — multi-arch GHCR build, Sigstore attestation, npm publish, box autodeploy. New `REMOTE_CONFIG_CONDITIONAL_BETAS` strips it from the baked base and excludes it from the drift comparison on both sides, so neither state reads as drift. A genuine base-beta add or removal still surfaces — asserted, not assumed.
+
+  Deliberately a separate set from `MODEL_CONDITIONAL_BETAS` rather than a quiet widening of it, because the reasons differ: those are flags dario appends per-request, so comparing them against a base that never carried them is a false positive we manufactured (#484). This one CC sends. Suppressing it is a real, if small, divergence from wire fidelity, taken because the alternative is a release per flip. Safe in both directions: `betaForModel` documents that its per-family shape holds whether or not the base carries afk-mode, `removing a beta can never provoke an upstream 400`, and afk-mode is the side with a cost — it is rejected on Max 5x, which `unavailableBetas` absorbs at one 400 per account.
+
+- **A bake now refuses to write the bundle from a non-Linux host.** The sonnet-5 prompt variant names the tools that exist where the capture ran: Windows yields `, Glob, Grep` and `the Glob or Grep`, Linux yields `` `find` or `grep` via the Bash tool``. Six bytes, one slot, every other slot byte-identical — so a Windows bake and CI's Linux bake each report the other as drift and re-bake forever. Third channel for the same disease: #854 -> #860 was paths, #856 was the OS/arch headers, this is prose that mentions the tool list.
+
+  Unlike those two it cannot be normalised. The correct text depends on which tools the caller has, which the bundle cannot know, and rewriting CC's prose would be inventing wire shape rather than replaying it — `PLATFORM_ONLY_TOOLS` already unions the tools array, but nothing can union a sentence. So the fix is procedural, mirroring the existing `--allow-older-cc` guard: `--allow-non-linux-bake` for a deliberate local bake, and `--check` is not gated at all, since detecting drift from any platform is useful and only writing is harmful.
+
+  Two existing assertions used afk-mode as their stand-in for "a real base beta" and now use one that still is. Each carries a note on what it previously claimed.
+
 ## [5.4.15] - 2026-07-26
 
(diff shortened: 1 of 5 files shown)


sprayberry-redline (approved):

## What I checked
- `scripts/drift-report.mjs`: new `REMOTE_CONFIG_CONDITIONAL_BETAS` set, and `stripModelConditionalBetas`/`computeDrift` now filter both `MODEL_CONDITIONAL_BETAS` and the new set symmetrically on both `prev`/`now` sides (`scripts/drift-report.mjs:277-279`: `.filter((b) => !MODEL_CONDITIONAL_BETAS.has(b) && !REMOTE_CONFIG_CONDITIONAL_BETAS.has(b))`). This correctly makes afk-mode-present→absent and absent→present both report zero drift while a genuine new beta (e.g. `brand-new-2026-09-01`) still surfaces — verified against the new test in `test/bake-drift-report.mjs` (case 22) and by reading the filter logic directly.
- `scripts/capture-and-bake.mjs`: the new non-Linux write guard (`if (!CHECK_MODE && process.platform !== 'linux') { ... }`) is placed after the older-CC-version check and before the actual bake write, mirrors the existing `--allow-older-cc` pattern, and correctly leaves `--check` ungated — matches the PR's stated intent and the "write bake refused / --check unaffected" verification in the PR description.
- Test updates (case 34, case 38) correctly swap `afk-mode` out for `advisor-tool-2026-03-01` as the "real beta" stand-in now that afk-mode is suppressed, and add explicit assertions that afk-mode *is* stripped (case 38) and that suppression is symmetric (case 22) — the updated assertions test what they were originally meant to test.
- CI check-rollup: all checks green (build 18/20/22, test, CodeQL, compat, actionlint, docker-cap-drop-smoke).

## What's good
The two independent root causes (remote-config churn vs. bake-host divergence) are kept as separate mechanisms with separate fixes rather than conflated, and the PR is explicit about which parts are safe to automate (afk-mode suppression) vs. which require a process guard because the correct behavior is host-dependent (sonnet-5 prose). Reasoning for directional safety (removing a beta can't provoke a 400) is documented inline next to the code it justifies.

No further concerns from the diff.

PR #882, opened 2026-07-26
feat(pkg): ship docs/ in the package, minus the maintainer runbook

askalf:
`docs/configuration.md` shipped in 5.4.14 and was reachable **only on GitHub** — the wrong place for a config reference whose entire audience is someone running the installed package in a container, where a flag isn't available.

## What ships, what doesn't

`files` gains `docs`, with one exclusion.

| | |
|---|---|
| cost | 1.5 MB → **1.8 MB** unpacked, 108 → 131 files (~270 kB) |
| excluded | `docs/recovery.md` |

**Why that one is excluded.** It's a maintainer runbook — *"things automation can't fix"*, including recovering OAuth credentials on the release host — and it hardcodes that host's **root SSH target and key filename**. It's already public in this repo, so this isn't novel exposure. But an npm tarball is a different *distribution*: mirrored, scraped, pulled onto machines that never asked for it, and **immutable** once published. Publishing a production SSH target to every install buys a user nothing.

## Verified rather than assumed

`npm pack --dry-run` confirms npm honours the `!` negation in `files`:

- `docs/recovery.md` — **0 occurrences**
- every other doc present, including `configuration.md` (6.9 kB)

I audited the contents **before** shipping rather than trusting that public-on-GitHub meant safe-to-publish. No real credentials: the `sk-ant-api03-...` and `$GHCR_PAT` matches are placeholders inside command examples. No host paths, no usernames.

## One risk I can't fully retire

npm's content scanner has blocked `@askalf` publishes before, and 22 docs discussing OAuth interception, fingerprint replay and stealth pacing widen that surface. If the publish leg 403s at PUT, that's the scanner and not this change — the existing chase-an-allowlist thread applies.

## Separate question, not actioned here

Whether `docs/recovery.md` belongs in a **public repo** at all is worth deciding on its own. This PR only stops it being republished to npm; it doesn't move it.

Suite 125/125.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.17] - 2026-07-26
+
+- **`docs/` now ships in the npm package.** `docs/configuration.md` landed in 5.4.14 and was reachable only on GitHub, which is the wrong place for a config reference whose whole audience is someone running the installed package in a container. The `files` allowlist gains `docs`, costing about 270 kB against a 1.5 MB unpacked package.
+
+  **`docs/recovery.md` is deliberately excluded.** It is a maintainer runbook — "things automation can't fix", including recovering OAuth credentials on the release host — and it hardcodes that host's root SSH target and key filename. It is already public in this repo, so this is not novel exposure, but an npm tarball is a different distribution: mirrored, scraped, pulled onto machines that never asked for it, and immutable once published. Publishing a production SSH target to every install buys a user nothing. Verified with `npm pack --dry-run`: 0 occurrences of `docs/recovery.md`, every other doc present.
+
+  Audited the rest before shipping rather than after. No real credentials: the `sk-ant-api03-...` and `$GHCR_PAT` hits are placeholders in command examples. No host paths, no usernames.
+
 ## [5.4.16] - 2026-07-26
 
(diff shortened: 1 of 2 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues — clean, well-scoped packaging change.

## What I checked
- `package.json:16-21` — adds `"docs"` and `"!docs/recovery.md"` to the `files` allowlist. npm requires negation patterns to appear after the positive pattern they narrow, which is the case here (`"docs"` then `"!docs/recovery.md"`), so the exclusion is ordered correctly.
- `CHANGELOG.md` — new `[5.4.17]` entry accurately describes the change and the rationale for excluding `docs/recovery.md` (it hardcodes the release host's SSH target/key filename).
- `version` bump `5.4.16` → `5.4.17` is consistent with the changelog entry.
- CI rollup (`gh pr checks`): all completed checks are green (`build (18/20/22)`, `actionlint`, `triage`, `validate-package-json`); `test`, `analyze`, `compat`, `docker-cap-drop-smoke` were still in progress at review time with no failures — nothing here is affected by a 2-line `files` array change, so I did not wait on them.
- PR description states the author verified the exclusion with `npm pack --dry-run` (0 occurrences of `docs/recovery.md`, all other docs present) — this is exactly the right way to validate a `files`-allowlist change, since npm's glob/negation semantics are easy to get subtly wrong.

## What's good
- The PR explicitly separates "what ships" from "what's excluded and why," with a concrete security rationale (production SSH target should not be immutably published to every npm install) rather than blanket-shipping the whole `docs/` tree.
- Flags the pre-existing npm content-scanner risk (unrelated to this diff) instead of silently hoping it doesn't recur.
- Scopes out the separate question of whether `docs/recovery.md` belongs in a public GitHub repo at all — correctly not conflated with this PR.

No correctness, security, reuse, or test-coverage issues found in the diff.

PR #884, opened 2026-07-26
chore: remove two exported test seams that no test calls

askalf:
Audited the whole package for dead code. The headline is that there is very little — one real finding, two symbols removed.

## What the audit checked

| check | result |
|---|---|
| `src/` modules | **52, all reachable** from `index.ts` or `cli.ts` — 0 dead files |
| `scripts/` | **19, all referenced** by `package.json` or a workflow |
| test files | **132**; runner skips **7** (documented live-integration + perf); **125 run** — reconciles exactly with the suite count |
| exports with no external caller | **83**, of which **80 are used inside their own file** |

## The finding

`_resetVersionCacheForTest` (`src/version.ts`) and `_resetInstalledVersionProbeForTest` (`src/live-fingerprint.ts`) clear module-level memoization between tests. **Nothing has ever called either** — one occurrence each in the entire tree, the declaration itself.

They're redundant *by construction*, not by neglect: `test/all.test.mjs` spawns each test file as its own **process**, so module state can't leak between files, and within one file there's nothing to reset.

The reason to delete them is that **their names assert tests depend on them.** A seam claiming coverage it doesn't have is worse than no seam. If the suite is ever consolidated into a single process, they come back.

## The number worth being careful about

**80 of the 83 uncalled exports are live code that's merely over-exported.** `BG` is used at `render.ts:93`; `refreshStatus` at `status.ts:81`. My first pass measured "not referenced *outside* its own file" and would have reported all 83 as dead — a spot-check caught it. Dropping their `export` keyword is a lint nit and I'm not attempting it here.

## One dead symbol kept on purpose

`clearLineRight` in `src/tui/render.ts` is genuinely unused. It stays: a one-line ANSI primitive between `clearScreen` and `reset` — both used — in a module whose entire job is ANSI primitives. It misrepresents nothing, and the first person needing to clear a line re-adds it. Deleting what lies is worth doing; deleting what's merely unused-but-coherent isn't.

## Verification

- suite **125/125**
- both removed symbols absent from `dist/`
- the cache variables they touched are still referenced 3× each, so nothing was orphaned by the removal
- re-ran the audit after the change: 1 remaining dead symbol, the one kept deliberately


sprayberry-redline (dismissed):

**Verdict:** No blocking issues — clean, well-scoped dead-code removal. Approving.

## What I checked
- Diff: `CHANGELOG.md`, `package.json` (version bump 5.4.17→5.4.18), `src/live-fingerprint.ts`, `src/version.ts`.
- Verified the PR's central claim by grepping a full shallow clone: `_resetVersionCacheForTest` and `_resetInstalledVersionProbeForTest` have zero references anywhere in the tree outside their own declarations (`src/version.ts:30`, `src/live-fingerprint.ts:1046` pre-removal) — the removal is safe, they were truly uncalled.
- CI check-rollup: 12/12 green (CodeQL, build ×3, compat, docker-cap-drop-smoke, test, validate-package-json, etc.); `auto-merge` shows `skipping` (expected — gated on this review).
- `src/version.ts` diff (`-cached = null;` and the `_resetVersionCacheForTest` export) and `src/live-fingerprint.ts` diff (`-_installedVersionProbe = { value: null, cached: false };` and the `_resetInstalledVersionProbeForTest` export) only remove the dead functions — the underlying `cached`/`_installedVersionProbe` state variables they touched remain otherwise untouched and still used by the live code paths.

## What's good
- The PR body's own audit methodology (reachability sweep, cross-checking exports used only inside their own file vs. genuinely dead) is sound and the one finding kept out of scope (`clearLineRight`) is reasoned rather than reflexively deleted — good judgment on where to stop.
- Changelog entry is detailed and matches the actual diff.

No further findings — this is a minimal, correct, low-risk change.

sprayberry-redline (approved):

PR #886, opened 2026-07-26
feat(wire): forward a genuine CC client's identity headers instead of replaying a template

### LINKED ISSUE #885: wire: passthrough rebuilds genuine CC's headers instead of forwarding them
Two findings from a wire-fidelity audit, both about the **header** axis. Neither contradicts #813 — that audit established header *order* is a `fetch()` no-op and that the JA3/order bind is real. This is about which headers get sent at all, and it is orthogonal to order.

## 1. Passthrough discards genuine CC's own headers and replays a reconstruction

On the passthrough path — the one taken when `isGenuineCCClient` says the caller really is Claude Code, including its Task/Agent sub-agents — dario builds the upstream header set from scratch (`proxy.ts` ~L3029):

```js
const headers = {
  ...staticHeaders,                      // passthrough: accept + Content-Type ONLY
  ...upstreamAuthHeaders(...),
  'x-claude-code-session-id': outboundSessionId,   // dario's, not the client's
  'anthropic-version': passthrough ? (req.headers['anthropic-version'] ?? '2023-06-01') : '2023-06-01',
  'anthropic-beta': beta,
  'x-client-request-id': randomUUID(),
  'x-stainless-timeout': '600',
};
```

There is no `...req.headers`. `anthropic-version` is the **only** client header that survives. Everything the real CC client sent — its actual `x-stainless-*` values, its real `user-agent`, its own request id, and anything the SDK adds conditionally — is dropped and replaced with template-derived or synthesized values.

That is inverted. The template exists to *synthesize* CC's shape for clients that aren't CC. On the path where we are holding CC's genuine bytes, synthesis is strictly worse than forwarding: a live capture is at best a good imitation of the thing we already have in hand.

`"byte-faithful passthrough"` is accurate about the **body** — and #813 credited that fidelity specifically. It does not currently extend to headers, and the log line at L2779 (`system + tools forwarded verbatim`) is careful to say body parts only.

**Not a blanket "forward everything".** Four categories must still be dario's:

| category | why it can't be forwarded |
|---|---|
| auth (`authorization` / `x-api-key`) | must become the pool account's credential |
| `x-claude-code-session-id` | session rotation is a deliberate feature |
| transport (`host`, `connection`, `content-length`, `accept-encoding`) | owned by the HTTP stack |
| `anthropic-beta` | merged with operator pins + per-account rejection cache |

The proposal is narrower: on the genuine-CC path, **forward the client's own CC-identity headers** (`x-stainless-*`, `user-agent`, `x-app`, `anthropic-dangerous-direct-browser-access`) instead of overlaying template values. Those are exactly the headers the template is trying to reproduce.

## 2. The capture cannot see first-party-conditional headers

`x-client-request-id` is emitted by dario on every request and appears **0 times** in the baked template — not in `header_values` (correctly excluded, it rotates) and not in `header_order`, which is unfiltered (`extractHeaderOrder` records every raw name, lowercased and deduped, with no exclusion list).

That reads like an extra header dario invents. It is not. The string is present in the installed CC 2.1.220 binary, and dario's own history explains it: `3a8fda9` removed it, `cbe20d6` restored it with *"sent for firstParty + api.anthropic.com"*.

The bake captures CC against a **loopback MITM**, so any header the SDK gates on the base URL being first-party Anthropic is **structurally invisible to the capture**. Consequences:

- `header_order` is incomplete by construction, so comparing dario's emitted set against it produces false "extra header" findings. It produced one for me before I checked.
- more importantly, **no watcher can see drift on those headers.** `check-wire-drift.mjs` spawns CC against loopback too, and it only asserts two signals anyway — the per-model beta set and the `cch` gate. A first-party-conditional header being added or dropped upstream would go unnoticed indefinitely, which is the same silence that let the beta drift through before #528.

Worth at least a comment at the `header_order` definition recording that it is loopback-scoped and therefore not a complete picture.

## Method note

Candidate header names were extracted from CC's binary and diffed against `header_order`. That surfaced ~60 names, and **most are noise** — response headers dario already parses (`anthropic-ratelimit-unified-*`), other-SDK identifiers (`anthropic-sdk-python`), even font names (`anthropic-sans`). A string in a 259 MB binary is not evidence of emission. Of the plausible request-header candidates, `x-claude-code-agent-id` and `x-claude-code-parent-agent-id` are the ones I could not rule out; if CC sends those on sub-agent requests, finding 1 would fix them for free by forwarding, which is another argument for forwarding over enumerating.

askalf:
Closes #885 finding 1. Crosschecked live, red-green verified both directions.

## The measurement, before any code changed

Ran the real proxy with `ProxyOptions.fetchImpl` recording the upstream request, then POSTed a genuine-CC-shaped body whose headers carried sentinel values:

```
forwarded=0  replaced=12  dropped=1
```

Every CC-identity header the client sent was replaced with a template value; `x-claude-code-agent-id` was dropped outright. After the fix: **13/13 forwarded, 0 replaced, 0 dropped.**

## Why forwarding beats synthesising here

The template exists to *synthesise* CC's shape for clients that **aren't** CC. On the passthrough path `isGenuineCCClient` has already established the caller **is** Claude Code, so its headers are the authentic article — a capture is at best a good imitation of bytes we already hold.

It also fixes something the capture structurally cannot: the bake runs against a **loopback MITM**, so any header the SDK gates on a first-party `api.anthropic.com` base URL is invisible to it (#885 finding 2). `x-claude-code-agent-id` is in CC's binary and has never appeared in a bake. Forwarding gets it right without anyone enumerating it — which is why the implementation uses **prefix rules** (`x-stainless-`, `x-claude-code-`, `x-client-`) rather than an exact allowlist.

## What dario still owns

Spread order encodes the precedence — template, then client, then dario:

| kept dario's | why |
|---|---|
| `authorization` / `x-api-key` | must become the pool account's credential |
| `x-claude-code-session-id` | session rotation is a feature, not an accident |
| `anthropic-beta` | merged set: operator pins + per-account rejection cache |
| body framing + transport | owned by the proxy and the HTTP stack |

`x-client-request-id` and `x-stainless-timeout` prefer the client's real value and fall back to a synthesised one.

**Non-CC clients are unaffected** — they keep the full template rebuild, which is what they need. And the gate reads the request **body**, so a client can't self-authorise into forwarding by spraying `x-stainless-*` headers.

## The test has two layers and the second is load-bearing

**My first version of this fix typechecked, passed every unit test, and forwarded nothing.** It gated on `passthrough` — a *startup CLI flag* — instead of the per-request genuine-CC signal, which is block-scoped where `buildCCRequest` destructures it. `staticHeaders` is built once at startup and then has `overlayTemplateHeaderValues` applied unconditionally, which is how template values reach genuine-CC requests in the first place.

Only a live request through the assembled proxy revealed `forwarded=0`. So the test keeps both layers:

1. pure unit tests over `forwardClientCCIdentityHeaders` — allow/deny, prefix rules, array values, blank rejection, case-insensitivity
2. a live sentinel request through `startProxy`

**Red-green verified in both directions:** reverting the gate to the buggy variable turns exactly 7 assertions red while all 27 unit assertions still pass — which is the trap, stated as a test.

Suite **126/126**.

## Docs

`docs/wire-fidelity.md` gains the axis, and its closing line no longer implies the roadmap covers everything. Header **order** is deliberately absent — `fetch()` re-normalises it and the JA3/order bind is real (#813) — and the honest framing is byte-identical **body**, structurally close **headers**, not packet-identical.


sprayberry-redline (approved):

## What I checked
- `src/cc-template.ts`: new `forwardClientCCIdentityHeaders` (allow/deny-list + prefix rules), verified the never-forward set correctly excludes `x-claude-code-session-id`, auth headers, and transport/framing headers.
- `src/proxy.ts`: the fix's core claim — that `genuineCCRequest` (proxy.ts:2783, set from the per-request `genuineCC` returned by `buildCCRequest`) is distinct from `passthrough` (proxy.ts:1260, the startup CLI flag `opts.passthrough`) — holds up. The forwarding gate at proxy.ts:3045 `(passthrough || genuineCCRequest)` correctly covers both the startup-flag path and per-request genuine-CC detection, which is exactly the bug the PR describes fixing (a prior version only gated on the flag and forwarded nothing).
- Header precedence: `forwardedIdentity` is spread after `staticHeaders` but before `upstreamAuthHeaders`/session-id/beta assignment, so dario-owned values (auth, session id, beta set) still win even if a forwarded key collided — confirmed by reading the object literal at proxy.ts:3048-3060.
- `test/passthrough-header-forwarding.mjs`: two layers as described — pure unit tests over the allow/deny logic, plus a live `startProxy` request with sentinel header values captured via a fake upstream `fetchImpl`. This is exactly the kind of test that would have caught the original bug (unit tests alone passed against the buggy `passthrough`-gated version).
- `src/version.ts` / `src/live-fingerprint.ts`: confirmed `_resetVersionCacheForTest` and `_resetInstalledVersionProbeForTest` have no remaining callers anywhere in the tree after removal.
- CI: `gh pr checks` shows all checks passing (test, build×3, CodeQL, compat, docker-cap-drop-smoke, etc.); only `wire-drift` is still pending, which isn't blocking.

## Minor
- `x-stainless-retry-count` is now forwarded from the client verbatim to the real upstream (proxy.ts via `forwardClientCCIdentityHeaders`); this is called out and justified in the PR body/test comments ("real count beats a hardcoded 0") and the forwarding gate already requires body-based genuine-CC detection, so it doesn't cross a new trust boundary — not blocking, just worth having on record.

## What's good
Excellent PR discipline: a documented red-green verification (reverting the gate turns exactly 7 assertions red), before/after measurements from a real proxy run, and a docs update (`wire-fidelity.md`) that stays honest about scope (explicitly disclaims header-order fidelity). The changelog is unusually thorough for a header-forwarding fix.

Issue #528, opened 2026-06-15
CCH calculation problem


lwsh123k:
Before running the code, I noticed that the calculation of cch is random. I also looked at the related questions on GitHub, where through observation, you found that the calculation of this variable is a random value.

However, based on my search, this is not the case. Here are the links that I think are quite useful for reference:
[https://a10k.co/b/reverse-engineering-claude-code-cch.html](https://a10k.co/b/reverse-engineering-claude-code-cch.html)
[https://github.com/freecodexyz/free-code/pull/9](https://github.com/freecodexyz/free-code/pull/9)

askalf:
Thanks for the references — fair catch, and you're right that the inline comment is misleading. Here's what I found trying to verify it end-to-end, because the result is a bit more nuanced than "random vs not random."

**You're correct that `cch` isn't truly random.** The a10k write-up is solid — the author caught the actual hash being written under a debugger, so on the build they analyzed it's a deterministic `xxHash64` over the request body, not an RNG. Our `// random … (Claude Code does the same)` comment overstates what we actually know, and I'll fix it.

**But I couldn't reproduce the published algorithm against a *current* Claude Code build**, and that matters before changing `computeCch`. I pointed the genuine `claude` binary (v2.1.177) at a loopback collector, captured the exact wire bytes of its `/v1/messages` POSTs, restored the `cch=…` field to the `cch=00000` placeholder in place, and recomputed:

`xxHash64(body_with_cch=00000, seed=0x6E52736AC806831E) & 0xFFFFF` → 5-hex

Across three live requests:

| captured `cch` | recomputed (published algo) |
| --- | --- |
| `e9bfb` | `6e8c1` |
| `848d9` | `cc95a` |
| `0212a` | `cd55b` |

No match. To rule out my own error:
- bodies were byte-complete (`Content-Length` == received) and valid JSON;
- the hash math checks out — validated against known `XXH64` test vectors and cross-checked with two independent implementations (`xxhash-wasm` and `hash-wasm`, which agree);
- I swept seed endianness, pre-image scope (`cch=00000` / unmodified / field removed), `XXH64` vs `XXH3`, and low/high/substring bit-extraction. Nothing reproduced the captured value.

[...]

PR #887, opened 2026-07-26
feat(drift): watch the header set, header values and body key order

askalf:
`check-wire-drift.mjs` asserted exactly **two** signals — the per-model beta transform and the `cch` gate. A header CC starts sending, a value it nudges, or a body key reorder was invisible. That's the silence that hid the beta drift before #528, on three more axes.

## The evidence was already being thrown away

The capture read `req.headers` and the full body, then kept only beta/billing/version/cch. It now also records the wire header order from `rawHeaders`, the header values, and the body's top-level keys.

## dario's side is measured, not computed

The real proxy runs in-process with an injected `ProxyOptions.fetchImpl` recording the outbound request. Deriving the expected set from the pure helpers would miss call-site additions — **which is exactly how #886 produced a header fix that passed every unit test and emitted nothing.**

The probe sends a **plain** body on purpose, so dario takes the template-**rebuild** path. That's where it synthesises CC's shape, so that's where synthesis can drift; the passthrough path forwards the client's own headers and has nothing to drift against.

Excluded from comparison, each for a stated reason: auth (pool credential substitution), transport (HTTP stack), per-call rotating ids (presence meaningful, value not), and `anthropic-beta` (already asserted exactly by check 1).

## Red-verified — and two attempts failed first, instructively

| axis | red test | result |
|---|---|---|
| `header.set` | remove `x-app` from the overlay source | fires, names `x-app` |
| `header.values` | corrupt `x-stainless-lang` | fires: `CC=js dario=RED-TEST-WRONG` |
| `body.order` | swap `model`/`messages` | fires with the full before/after |

**Attempt 1 failed:** breaking the *hardcoded* header fallbacks changed nothing — drift stayed false — because `overlayTemplateHeaderValues` runs afterwards and heals them. That independently confirms the external audit's triage that those literals are fallback-only and unreachable whenever a template loads.

**Attempt 2 failed worse:** the body-order check **silently did nothing.** dario passes the outbound body as a `Uint8Array`, so parsing it as a string threw and left the key list null, and the check skipped without a word. An unreadable body is now a reported `body.order.unmeasured` finding — a check that cannot measure must not report agreement. That's the same defect class this script exists to catch, found in the script itself.

## Green

`drift=false`, 0 findings, 21 header names and 10 body keys captured and surfaced in the report payload. Suite **126/126**.


sprayberry-redline (dismissed):

## What I checked
- Full diff (`CHANGELOG.md`, `package.json` version bump, `scripts/check-wire-drift.mjs`, +178/-3).
- CI check-rollup: all 14 checks green (CI test/build×3/docker-smoke, CodeQL×2, compat, wire-drift, actionlint, validate-package-json, spam-triage).
- Cross-referenced `startProxy`/`ProxyOptions.fetchImpl` in `src/proxy.ts:821-860` to confirm the new `emitViaDario()` probe (script lines around `startProxy({ port, host, upstreamApiKey, noClaudeAuth, fetchImpl })`) uses the injected fetch seam as documented, and that `fetchImpl` is a real, supported test hook rather than something bypassing the intended proxy path.
- Traced the three new comparisons end to end:
  - `header.set`: `missing = opus.headerNames.filter((h) => !WIRE_COMPARE_SKIP.has(h) && !(h in emitted.headers))` — correctly flags a CC header dario never emits.
  - `header.values`: compares `opus.headerValues?.[h]` vs `emitted.headers[h]`, skipping `WIRE_COMPARE_SKIP` entries (auth, transport, rotating ids, `anthropic-beta` already covered by check 1) — sound.
  - `body.order`: `shared`/`ourShared` filter each side's key list down to the keys both share while preserving each side's own order, then compares the joined strings — correctly detects a genuine reorder of common keys without false-flagging on keys only one side has.
- The "unmeasured" fallback (`!Array.isArray(emitted.bodyKeys) || !Array.isArray(opus.bodyKeys)`) reports a `body.order.unmeasured` finding instead of silently treating an unreadable body as agreement — this is exactly the fix described for the PR's own "Attempt 2 failed worse" note, and it reads correctly in the diff.

## What's good
- The PR body documents two red-verified failed attempts and how each was fixed (fallback-healing false negative, `Uint8Array` body parse throwing and going unnoticed) — that diagnostic honesty is reflected accurately in the code (the `try/catch` around body decode at the `emitViaDario` body-parsing block, and the `unmeasured` finding path).
- `WIRE_COMPARE_SKIP` documents *why* each header is excluded (auth substitution, transport-owned, per-call rotation, already covered elsewhere) rather than excluding silently.

## Minor (non-blocking)
- The in-process probe binds a hardcoded port (`39877`); on a self-hosted runner with concurrent job runs this could theoretically collide and produce a flaky "could not measure" low-severity finding rather than a false pass, so it fails safe — not worth blocking on.

No correctness bugs, security issues, or missing test coverage found in the changed lines.

sprayberry-redline (approved):
_Automated review from the Sprayberry Labs fleet code reviewer._

## What I checked
- Full diff (`scripts/check-wire-drift.mjs`, `CHANGELOG.md`, `package.json` version bump).
- PR's own CI check-rollup: all 11 checks pass, including a dedicated `wire-drift` job.
- Traced the new header-set/value comparison (`scripts/check-wire-drift.mjs`) and the body-key-order comparison logic against the capture/emit data shapes to confirm they can't silently no-op:
  - `capture()` now unconditionally records `headerNames`/`headerValues`/`bodyKeys`, so the `opus && opus.headerNames` guard before the new checks only skips when capture itself already failed (which already produces a `severity: 'high', category: 'capture'` finding) — no new silent-skip path introduced.
  - `emitViaDario()`'s body decode explicitly reports `body.order.unmeasured` rather than silently treating an undecodable body as "no drift", consistent with the PR's stated intent and the "Attempt 2 failed" note in the description.
  - `WIRE_COMPARE_SKIP` exclusions (auth, transport-owned headers, per-call rotating ids, `anthropic-beta`) each have a stated, checkable reason in-line.

## Minor (non-blocking)
- `scripts/check-wire-drift.mjs:18-30` (top-of-file docblock): the new items are inserted as "3. Header SET..." / "4. Top-level body key ORDER..." *before* the existing "2. `cch` billing-integrity token..." entry, leaving the doc list out of numeric order (`1, 3, 4, 2`). Purely cosmetic — doesn't affect behavior — but worth a quick reorder next touch.

## What's good
- The PR description documents two failed verification attempts and why they failed (hardcoded-fallback healing, silent body-decode skip) — that's exactly the kind of adversarial self-check that gives confidence the new assertions actually fire when they should, and I was able to independently confirm the described "must-report-when-unmeasured" behavior in the diff.
- Real proxy is exercised in-process via injected `fetchImpl` rather than re-deriving expected output from the same pure helpers being tested — avoids the class of bug called out from #886.

PR #888, opened 2026-07-27
perf: stop spawning `claude --version` three times at startup

askalf:
Profiled rather than guessed, and the profile changed what was worth doing.

## The finding

**720 ms — 12.7% of all CPU — in blocking `spawnSync`**, every bit of it inside `startProxy`, asking the same question of a 259 MB binary three separate ways:

| caller | samples | spawns |
|---|---|---|
| `findClaudeBinary` → `probeOneVersion` | 158 | `claude --version` per candidate |
| `probeInstalledCCVersion` | 142 | `claude --version` on the winner |
| `detectCliVersion` | 138 | its own `execSync('claude --version')` |

`findClaudeBinary` is now memoised per process — the installed binary can't change mid-process, and it had four call sites each re-enumerating and re-probing. `detectCliVersion` reuses the already-memoised probe, which is also **more correct**: that probe honours `DARIO_CLAUDE_BIN` and picks the newest candidate instead of trusting whatever bare `claude` the shell resolves.

```
startProxy      1277 ms -> 983 ms    (4 runs each, ~23%)
spawnSync CPU    720 ms -> 499 ms
```

## What I deliberately did NOT optimise

With an instant upstream, a request costs **3.4 ms** of dario CPU — 270 req/s on a 65 kB rebuild, 219 req/s on a 105 kB passthrough. Against a completion that takes seconds, that's a fraction of a percent of end-to-end latency. Micro-optimising the request path would be theatre, so I didn't.

**The most useful thing the profiling established isn't a code change.** The first benchmark read **2.0 req/s and p50 = 508 ms on both paths** despite a 40 kB payload difference. Identical cost across two unrelated code paths isn't code — it's a fixed floor. It was the pacing floor: **99.3% of request latency is the deliberate inter-request pace**, already tunable via `--pace-min` / `DARIO_PACE_MIN_MS`. Throughput here is a setting, not an optimisation. Anyone "optimising" dario's request path without turning pacing off first is measuring a constant.

## Also: the suite no longer depends on the operator's live cache

Found because my own benchmark broke three tests. Running the proxy refreshes `~/.dario/cc-template.live.json` with a **headless** capture, which omits the interactive-only tools CC never sends from `--print`. Three suites then fail on a machine where dario has recently run and pass everywhere else.

**I nearly misattributed it to the perf change.** They reproduced on unmodified `master`, and redirecting the cache turned `7 passed, 2 failed` into `9 passed, 0 failed`.

`test/all.test.mjs` now pins `DARIO_LIVE_TEMPLATE_CACHE` for every spawned child at a nonexistent path, so `loadTemplate` falls back to the bundled snapshot deterministically. **#867 closed the write half** — tests no longer poison the real cache. This is the read half: a test whose result changes because you started the proxy an hour ago isn't measuring the code.

Verified **126/126 with the polluted cache still in place** — the fix is what makes that pass.

## Remaining spawn cost, and why it stays

The residual 499 ms is two probes that legitimately run once: candidate comparison in `findClaudeBinary` (needed to pick the newest of several installs) and `probeBunVersion` (the TLS-class check). Both are once-per-process now. Making them async would move startup off the critical path further, but that changes ordering guarantees around the drift banner and `--strict-tls`, which deserves its own change rather than riding a memoisation fix.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-documented perf fix. Approving.

## What I checked
- `src/live-fingerprint.ts`: `findClaudeBinary` now memoizes the resolved binary path per process in `_claudeBinCache`, short-circuiting before touching the cache when `DARIO_CLAUDE_BIN` is set (`src/live-fingerprint.ts:719-722`) — so a test that flips the env override is never served a stale cached value, matching the PR's own claim.
- `src/proxy.ts:125-134`: `detectCliVersion` now calls the already-memoized `probeInstalledCCVersion()` instead of its own `execSync('claude --version')`, and the now-unused `execSync` import was removed. `probeInstalledCCVersion` (`src/live-fingerprint.ts:946-950`) was already memoized pre-PR, so this is a straightforward reuse, not new caching logic.
- `test/all.test.mjs`: each spawned child test now gets `DARIO_LIVE_TEMPLATE_CACHE` pinned to a nonexistent path under a fresh `mkdtempSync` dir (`test/all.test.mjs:73-80`), and `liveCachePath()` (`src/live-fingerprint.ts:251-255`) reads that same env var per-call (not hoisted to a module const), so the override is honored correctly by every child process.
- CI check-rollup is fully green (12/12: CodeQL, actionlint, analyze, build×3, compat, docker-cap-drop-smoke, test, triage, validate-package-json, wire-drift).

## Minor (non-blocking)
- `_claudeBinCache` is keyed on a fixed `key: ''` (`src/live-fingerprint.ts:709,721,724`) that never varies — the override path returns before ever consulting the cache, so the `key` field is currently dead weight. A plain `{ value } | null` would express the same invariant with less indirection, but this is cosmetic.
- The new `_resetClaudeBinCacheForTest` export (`src/live-fingerprint.ts:711-713`) isn't referenced by any test in this diff — harmless if intended for a follow-up, but currently untested/unused.

## What's good
- The CHANGELOG entry is unusually thorough and honest about what was *not* optimized (the request path) and why, backed by real measurements.
- The test-cache fix is a well-reasoned "read half" companion to #867, with a clear repro (`7 passed, 2 failed` → `9 passed, 0 failed`) and verification that the underlying suites still pass with the polluted cache in place (126/126).

PR #889, opened 2026-07-27
sec: write dario's state files with explicit restrictive modes

askalf:
Deep security pass over dario's state directory. **Credentials were already correct** — `durableWriteFile` opens with `0o600`, and the Linux deployment confirms `-rw-------` on every `credentials.json`. Three writes passed no mode and therefore landed `0644`.

> Stacked on #888 so it inherits the test-isolation fix — without it, three template suites fail on any machine where dario has recently run. Merge #888 first; base retargets to master automatically.

## What was wrong

| write | contents | was | now |
|---|---|---|---|
| `~/.dario` (`mkdirSync`) | the state dir itself | **755** | **700** |
| `cc-template.live.json` | the **unscrubbed** live capture | 644 | 600 |
| `--log-file` stream | redacted metadata | 644 | 600 |

**The directory is the one that matters.** Whichever code path creates `~/.dario` first decides its permissions, and `atomicWriteJson` runs at *startup*, before any credential write. The credential paths create their dirs `0700`; this one could get there first and land `755` — which is exactly what the Windows box shows. A `755` parent is what turns the `0644` files inside from untidy into readable.

**The live cache isn't incidental data.** Verified on the Linux deployment to contain `# Environment` (cwd, OS, platform) and `gitStatus:` (branch, modified files, recent commits). Structurally it can also carry `# claudeMd`, `# userEmail` and `# auto memory` — the scrub list exists because CC emits them, and the live path deliberately does *not* scrub, since it exists to replay the operator's own install faithfully.

The log file is defence-in-depth only: `writeLogLine` runs `redactSecrets` and `-vv` bodies go to stdout, so records are metadata — but they still show which account served which request and when.

## What the audit found clean

Recording this rather than leaving an implication of a broader problem:

- both auth gates compare with `timingSafeEqual` behind a length guard, and **fail closed** with no token configured
- request bodies bounded at **10 MB** with an explicit rejection
- proxied paths allow-listed, **default-deny 403**, against SSRF
- `shell: true` only on the Windows `.cmd` path CVE-2024-27980 requires, and every such site **rejects shell metacharacters** first
- redaction tested against a corpus of real shapes — `sk-ant-api/oat/ort`, JWT triples, `Bearer` raw and quoted — **caught every one**. The four it misses are dario's *own* secrets, and no code path emits them
- `doctor --json` run with sentinel values in `DARIO_API_KEY` and `DARIO_ADMIN_TOKEN` emitted **neither**, so the documented *"safe to paste into bug reports"* claim holds under test

## Two notes on verification honesty

**Windows proved nothing.** It doesn't enforce POSIX modes, so the 644 I first observed there was a platform artifact — including on `credentials.json`, which is genuinely 600 on Linux. The finding comes from the Linux deployment, not my laptop. I nearly reported "credentials are world-readable", which would have been false and alarmist.

**My first mode unit-test captured nothing.** `dist` uses named ESM imports, so patching `fs` on the namespace object doesn't rebind the module's local reference — zero calls recorded, while my `exit=$?` read a pipe's status and printed success. The modes are verified present in the compiled output and the write path round-trips; the behavioural check belongs on Linux after deploy, and I'll do it there.

Suite **126/126**.


sprayberry-redline (dismissed):

**Verdict:** No blocking issues found — approving. One minor note below.

## What I checked
- `gh pr diff 889` (5 files, +101/-12) and surrounding context in `src/live-fingerprint.ts` / `src/proxy.ts` via a shallow clone of the head branch.
- CI rollup: `actionlint`, `compat`, `triage`, `wire-drift` all **pass**; `auto-merge` skipped (expected, not required).
- Verified `mkdirSync(..., { recursive: true, mode: 0o700 })` actually sets 0700 on newly-created intermediate/leaf dirs (reproduced locally) — the security fix in `atomicWriteJson` (`src/live-fingerprint.ts`) does what the comment claims.
- Traced the new `findClaudeBinary`/`probeInstalledCCVersion` memoization: `test/find-claude-binary.mjs` and `test/oauth-detector.mjs` don't exercise the memoized paths directly (they test `enumerateClaudeCandidates`/`findInstalledCC` standalone or skip on CI), so the process-level cache introduced here isn't hit by existing tests in a way that could regress them.
- `test/all.test.mjs`'s new `DARIO_LIVE_TEMPLATE_CACHE` pin to a non-existent path inside a fresh `mkdtempSync` dir correctly forces `loadTemplate`'s bundled-snapshot fallback per the PR's own description.

## Minor (non-blocking)
`src/live-fingerprint.ts:723-733` (`_claudeBinCache` / `findClaudeBinary`):
```
let _claudeBinCache: { key: string; value: string | null } | null = null;
...
function findClaudeBinary(): string | null {
  const override = process.env.DARIO_CLAUDE_BIN;
  if (override) return override;
  if (_claudeBinCache && _claudeBinCache.key === '') return _claudeBinCache.value;

  const resolved = resolveClaudeBinaryUncached();
  _claudeBinCache = { key: '', value: resolved };
  return resolved;
}
```
The comment says the cache is "keyed on the override so a test flipping `DARIO_CLAUDE_BIN` is not served a stale answer," but `key` is always the literal `''` — the actual override-safety comes entirely from the early `if (override) return override;` return, not from any key comparison. The `key` field never varies, so it's dead weight that could mislead a future reader into thinking a real cache invalidation exists there.

```suggestion
// src/live-fingerprint.ts
let _claudeBinCache: string | null | undefined;

function findClaudeBinary(): string | null {
  const override = process.env.DARIO_CLAUDE_BIN;
  if (override) return override;
  if (_claudeBinCache !== undefined) return _claudeBinCache;
  return (_claudeBinCache = resolveClaudeBinaryUncached());
}
```

## What's good
- The mode fixes are narrowly scoped, each carries a clear rationale in-line, and the changelog is unusually rigorous about what was and wasn't verified (including honestly flagging a Windows false-positive and a broken first unit test, rather than glossing over it).
- The perf memoization is backed by real before/after numbers (1277ms → 983ms startup) and correctly reasons about staleness (binary can't change mid-process).
- The test-isolation fix (`DARIO_LIVE_TEMPLATE_CACHE` pin) fixes a genuine, previously-observed flake (3 suites failing depending on machine state) with a hermetic fallback rather than a workaround.

Note: `gh pr view` reports `mergeStateStatus: DIRTY` since this is stacked on #888 (now merged) — likely just needs a rebase against current `master`, not a code issue.

sprayberry-redline (approved):

**Verdict: re-approving.** No blocking findings — the merge commit (800b5c6) that resolved the master conflict is a clean, mechanical resolution; the substantive hardening from the original review is unchanged.

## What I checked
- Full PR diff (4 files, 30/-4): `package.json`, `CHANGELOG.md`, `src/live-fingerprint.ts`, `src/proxy.ts`.
- `package.json:3` — version bumped to `5.4.22` (correct next value after master's `5.4.21`).
- `CHANGELOG.md:14` — new `## [5.4.22] - 2026-07-27` entry correctly stacked above master's `[5.4.21]`, with a fresh `## [Unreleased]` retained at the top.
- `src/live-fingerprint.ts:440` — `mkdirSync(dirname(targetPath), { recursive: true, mode: 0o700 })`: the `~/.dario` state dir now created `0700` instead of the previous mode-less (effectively `0755`) call. This is the fix that matters most per the PR body — whichever writer creates the directory first sets its permissions, and this path runs at startup before any credential write.
- `src/live-fingerprint.ts:446` — `writeFileSync(tmp, JSON.stringify(data, null, 2), { mode: 0o600 })`: the unscrubbed live-capture temp file now created `0600`; `renameSync` preserves that mode onto the final path.
- `src/proxy.ts:1402` — `createWriteStream(logFilePath, { flags: 'a', mode: 0o600 })`: log stream now created `0600`. Comment correctly notes the mode only applies on creation, so a pre-existing log file's mode is untouched — accurate description of Node's `fs` semantics with the `'a'` flag.
- CI: all 12 checks green on head `800b5c6` — `compat` and `wire-drift` (the two called out for post-merge verification) both pass, plus build/test/CodeQL/actionlint/docker-cap-drop-smoke.
- Confirmed I am not the PR author (`askalf`/Thomas Sprayberry) — no self-approval conflict.

## What's good
The three mode changes are minimal, targeted, and each carries an inline comment explaining *why* (not just what), which matches the PR body's reasoning. The changelog entry is unusually rigorous about documenting what was audited and found already-correct, avoiding implying a broader problem than exists.

No new commits beyond the conflict-resolution merge changed any reviewable logic, so this re-approval covers the same content as the original review.

PR #890, opened 2026-07-27
feat(drift): trip a loud tripwire when a base capture matches the #881 anomaly

### LINKED ISSUE #881: bake: the base prompt intermittently captures 279 extra chars (5038 vs 4759)
The scrubbed base system prompt is usually 4759 chars. Roughly one capture in ten it comes back **5038** — a +279 char `# Context management` paragraph that is not in the baked bundle.

## Evidence

Two occurrences on 2026-07-26, against **20 consecutive clean captures** in between (8 in one sampler run, 12 in another, all `raw=6715/6813 scrubbed=4759`). Both 5038 readings happened inside a `capture-and-bake` invocation; every one of the 20 clean readings came from a base-only sampler. That correlation is suggestive but well short of a cause — n=2.

The extra text begins:

> When you have enough information to act, act. Do not re-derive facts already established in the conversation, re-litigate a decision the user has already made, or narrate options you will not pursue.

## Why it is worth tracking rather than ignoring

Two reasons, in order of confidence.

**It causes occasional rebake churn.** The base is compared byte-for-byte, so a 5038 capture on CI's 30-minute drift check opens a rebake PR, bumps a version and cuts a release. That is the same churn #880 just removed for `afk-mode`, arriving through a different slot. Unlike `afk-mode` this one is not a flag we can classify and strip — it is prose in the middle of the prompt.

**It is the last unexplained residue of #872.** The capture-provenance nonce (#875) proves the request came from the CC child we spawned, so this is not a foreign process being captured. And the guard from #874 stays silent on it, so it does not match any instruction-file wrapper. Both of those narrow it considerably: whatever produces the paragraph, it is our own child sending it. But "our own child sometimes sends 279 extra chars" is still not understood, and I would rather have it written down than rediscovered.

## What would settle it

Capture with the surrounding conditions recorded, not just the length — cwd, the full env delta, whether it is the first spawn of a bake run versus a later variant spawn, and elapsed time since the previous spawn. If it correlates with being the first capture in a multi-capture run, that points at CC session state rather than at remote config.

## Not blocking

`--check` on a 5038 capture correctly reports drift, and a rebake from Linux produces a valid bundle either way. The cost is release churn, not incorrectness.

## Correcting the record

I described the base prompt as stable in the 5.4.13 changelog on the strength of eight identical samples. Eight identical samples of a value that varies about one time in ten is a clean streak, not stability, and the claim was too strong.

askalf:
Follow-up to #881. **Tripwire only — this does not fix #881 and deliberately does not close it.**

## The gap

#881 records that roughly **1 local capture in 10** returns the scrubbed base system prompt at **5038 chars instead of 4759** — an extra `# Context management` paragraph beginning "When you have enough information to act, act." that is not in the baked bundle. 26h of CI data says it has never reached the Linux runner.

The harm if it ever does is specific and silent: `cc-drift-template-watch.yml` exits 2, auto-opens a rebake PR, and that PR looks exactly like a genuine CC prompt change. Merging it bakes a bad capture into the shipped wire-shape contract and cuts a release for an upstream change that never happened. Nothing in the current path distinguishes that from real drift.

## What this adds

`detectIssue881Residue(captured, bundled)` in `scripts/drift-report.mjs`, fires on either of two clauses, **both relative to the bundle**:

1. **marker** — the marker sentence is in the capture and absent from the bundle. Primary signal; survives the paragraph's wording drifting in length.
2. **length** — the capture is exactly 5038 against a bundle of exactly 4759, the pair documented in #881. Backstop for the sentence being reworded while the byte count still lands on the known pair.

Relative-to-the-bundle is what keeps this from becoming permanent noise: if the paragraph is ever legitimately baked in, the bundle gains the marker (and stops being 4759), both clauses go false, and the detector **self-disarms with no code change**.

## Specificity — the harder half

The ticket's constraint was to alert on the #881 signature, *not* on any base-length change, since false-flagging every change defeats the tripwire.

- The genuine **4754 → 4759** step #881 itself cites trips neither clause and flows through as ordinary drift. Covered by a test.
- `# Context management` **alone is not usable** as the discriminator. I checked `src/cc-template-data.json` before choosing: the current 4759-char bundle *already contains that heading* (it does not contain the marker sentence). Matching on the heading would have flagged every clean run. The marker sentence is the part that only appears in 5038 readings.

## How the signal surfaces

On a hit the bake writes the `::warning title=dario#881::` annotation **straight to stderr, not through `log()`** — a `[bake] ` prefix would stop Actions parsing line 1 as a workflow command, and the annotation is the whole point. The workflow's existing `cat drift-output.txt` puts it on the run summary, and that same stream is already embedded verbatim in the drift issue and PR bodies.

It also drops `issue-881-tripwire.txt` (sibling of `drift-summary.md` / `label-target.txt`, gitignored) so the workflow tests for a file instead of grepping prose. On a hit the workflow:

- titles the PR **`[#881 residue?] auto-rebake: ...`** — the PR list, merge-queue view and email notification show the title and nothing else;
- adds an **`issue-881-residue`** label, created on demand via `gh label create --force` (rare enough that pre-creating it in repo settings would just rot; same create-then-use shape the other watchers use);
- prepends a `[!CAUTION]` block to the body telling the reviewer to re-dispatch the watcher, and to **comment the outcome on #881 either way** — hit or false positive, that correlation is what the issue is open for.

The detector runs in **both** `--check` and the real bake, and the two verdicts are **OR'd**. They are separate live captures of an intermittent fault, so they can disagree, and a clean `--check` followed by a 5038 bake is the genuinely dangerous ordering — the bake's capture is the one the PR actually ships.

## Detection only, deliberately

Nothing here strips the paragraph, rewrites it, or suppresses the drift it causes. #881 defers the bake-time normalisation on the grounds that the call "should not be taken on a hunch", and this does not take it — it makes sure the hunch gets data instead of a merged rebake.

## Test plan

`test/bake-drift-report.mjs` gains sections 40 and 41 (21 new assertions), picked up automatically by `test/all.test.mjs`:

- both clauses fire; each reports its own `reason`
- heading-only does **not** fire; 4759-vs-4759 does not fire; 5038 against a non-4759 bundle does not fire
- the genuine 4754 → 4759 step is **not** flagged; an unrelated 9000-char change is **not** flagged
- self-disarm: once the paragraph is in the bundle, it stops firing
- `undefined` / non-string inputs do not throw
- line 1 of the warning starts at index 0 with `::warning ` (the no-prefix invariant), names 881, carries the issue URL and the captured length

Run locally: **132 pass, 0 fail** (was 110). `node scripts/check-package-json.mjs` → OK.

Version bumped `5.4.22 → 5.4.23` with a CHANGELOG entry, since dario needs an in-PR bump or nothing ships.

Refs #881.

### DIFF
--- a/.github/workflows/cc-drift-template-watch.yml
+++ b/.github/workflows/cc-drift-template-watch.yml
@@ -121,6 +121,21 @@ jobs:
           node scripts/capture-and-bake.mjs --check 2> drift-output.txt
           code=$?
           echo "exit_code=$code" >> "$GITHUB_OUTPUT"
+
+          # dario#881 tripwire. capture-and-bake writes issue-881-tripwire.txt
+          # only when the scrubbed base capture carries the known +279-char
+          # residue. The loud ::warning:: annotation itself rides in
+          # drift-output.txt and reaches the run summary via the `cat` below —
+          # this just records the hit as a step output so the rebake step can
+          # label the PR it opens. 26h of CI data says this has never fired on
+          # the runner; if it does, the rebake is #881 residue, not CC drift.
+          if [ -f issue-881-tripwire.txt ]; then
+            echo "issue_881=true" >> "$GITHUB_OUTPUT"
+            echo "::notice title=dario#881::tripwire hit on the --check capture — see the warning annotation and https://github.com/askalf/dario/issues/881"
+          else
+            echo "issue_881=false" >> "$GITHUB_OUTPUT"
+          fi
+
           if [ $code -eq 0 ]; then
             echo "No drift detected."
             cat drift-output.txt
@@ -159,6 +174,8 @@ jobs:
           # v4.6.4). Setup: docs/drift-monitor.md "Optional: PAT for
           # downstream workflow triggers".
           GH_TOKEN: ${{ secrets.DARIO_DRIFT_BOT_PAT || secrets.GITHUB_TOKEN }}
+          # dario#881 tripwire verdict from the --check capture (see below).
+          CHECK_ISSUE_881: ${{ steps.check.outputs.issue_881 }}
         run: |
           set -eo pipefail
 
@@ -179,6 +196,22 @@ jobs:
 
           # Run the real bake. Writes src/cc-template-data.json.
           node scripts/capture-and-bake.mjs
+
+          # dario#881 tripwire, second read. The bake above is a SEPARATE live
+          # capture from the --check one, and the anomaly is intermittent — so
+          # the two can disagree, and the bake's verdict is the one that
+          # describes the bundle this PR actually ships. Rewritten (or removed)
+          # by every capture-and-bake run, so this is fresh. OR them: either
+          # capture carrying the residue is enough to mark the PR, because
+          # either way a human is about to look at a base-prompt diff whose
+          # provenance is in doubt.
+          issue_881_bake=false
+          [ -f issue-881-tripwire.txt ] && issue_881_bake=true
+          issue_881=false
+          if [ "$issue_881_bake" = "true" ] || [ "$CHECK_ISSUE_881" = "true" ]; then
+            issue_881=true
+          fi
+          echo "issue_881=$issue_881" >> "$GITHUB_OUTPUT"
 
           # Sanity: did the bake actually change the bundle?
           if git diff --quiet -- src/cc-template-data.json; then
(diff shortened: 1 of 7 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, narrowly-scoped detection-only tripwire with solid test coverage.

## What I checked
- `gh pr diff` for all 7 changed files (workflow, `.gitignore`, `CHANGELOG.md`, `package.json`, `scripts/capture-and-bake.mjs`, `scripts/drift-report.mjs`, `test/bake-drift-report.mjs`).
- Traced the new `TRIPWIRE_881` write/read path against the surrounding (unchanged) context in `capture-and-bake.mjs` — confirmed `prev` (the bundle comparator) is loaded from `OUT` *before* the tripwire check runs and before `writeFileSync(OUT, ...)` later in the file, so the marker/length comparison is always against the pre-bake bundle, matching the PR's "relative to the bundle" design.
- Verified the workflow's two-read OR logic (`.github/workflows/cc-drift-template-watch.yml`): `steps.check.outputs.issue_881` is set unconditionally near the top of the "Run drift check" step (before the exit-code branching), so it's always populated by the time the `rebake` step (gated on `exit_code == '2'`) reads it via `CHECK_ISSUE_881`. No missing-output race.
- CI check-rollup: all green (CodeQL, actionlint, build ×3, test, compat, docker-cap-drop-smoke, validate-package-json).

## Notes (non-blocking)
- The marker sentence (`ISSUE_881_MARKER`) is matched verbatim; if the anomalous paragraph's wording ever drifts without also landing on the exact 5038/4759 length pair, the tripwire would silently miss it. The PR is upfront about this being a documented tradeoff (length clause as backup), so not flagging as a bug — just worth knowing if #881's root cause ever surfaces a differently-worded variant.
- `extra_label=()` relies on the script not using `set -u`; under the existing `set -eo pipefail` (no `-u`), expanding the empty array in `"${extra_label[@]}"` is safe. Confirmed correct as written, just noting I checked it since empty-array expansion is a common bash footgun.

## What's good
- Detection is genuinely detection-only — no stripping/rewriting of the anomalous paragraph, consistent with #881's deferral of any bake-time normalization.
- Self-disarm property (both clauses go false once the paragraph is legitimately baked in) is a nice design choice that avoids permanent noise, and is covered by its own test.
- Test suite (`test/bake-drift-report.mjs` sections 40–41) exercises both clauses, the specificity requirement (genuine 4754→4759 drift must NOT fire), self-disarm, and defensive non-string/undefined inputs — good coverage for a detector whose whole job is precision.

PR #893, opened 2026-08-01
docs(README): correct the line-count claim to ~24k and guard it in CI

askalf:
`src/` is **23,986 lines across 52 files**. The README hero line claimed `~22k` — a 9% understatement of the one number that makes "audit it yourself" a credible invitation rather than a slogan.

### Why a guard and not just another correction

This is the second time the number has gone stale, and the first correction is instructive. The 5.4.15 changelog entry already contains the right measurement:

> Source is 23,833 lines across 52 files, not ~22k across 49.

That was accurate when written. But `git log -L 21,21:README.md` shows the hero line untouched since #717, so the correction reached the changelog and never reached the claim a reader actually sees. #582 had refreshed the same number once before that. A hardcoded figure nobody re-measures only ever drifts in the flattering direction — the repo grows, the claim doesn't, and the understatement quietly becomes the marketing.

### The check

`scripts/check-readme-line-count.mjs` measures `src/**/*.ts` on every PR and fails if the claim is more than **1000 lines** off, printing the exact value to write:

```
FAIL: README claims ~22k lines; src/ actually has 23,986 across 52 files (off by +1986).
Update the hero line in README.md to ~24k.
```

It also fails if the sentence is deleted, so the guard cannot outlive the thing it guards. Tolerance is deliberately a full thousand: tight enough that the claim never misleads by a round number, loose enough that ordinary PRs do not trip it.

It runs inside the existing `validate-package-json` job rather than a new one — same class of cheap metadata assertion, no `npm install`, and no new required status check to add to branch protection.

### Verification

All three paths exercised locally:

| case | result |
|---|---|
| corrected README (`~24k`) | `ok  README ~24k vs actual 23,986 lines across 52 files (-14, tolerance 1000)` — exit 0 |
| stale claim (`~22k`) | exit 1, message above |
| claim deleted | exit 1, tells you to remove the check too |

No source changes, so no version bump — this rides the next release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,12 @@ jobs:
       - name: package.json is valid JSON and canonically formatted
         run: node scripts/check-package-json.mjs
 
+      # Kept in this job rather than a new one so it does not add a required
+      # status check to branch protection -- it is the same class of cheap
+      # metadata assertion, and it needs no npm install.
+      - name: README line-count claim still matches src/
+        run: node scripts/check-readme-line-count.mjs
+
   build:
     needs: validate-package-json
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ checklist.
 
 ## [Unreleased]
 
+- **The README's "~22k lines you can read in a weekend" was understating by 9%, and CI now measures it.** `src/` is **23,986 lines across 52 files**; the hero line said `~22k`. It now says `~24k`.
+
+  The reason this needs a guard rather than another correction is that it is the second one. The 5.4.15 entry below records "Source is 23,833 lines across 52 files, not ~22k across 49" — an accurate measurement, logged in this file, that never reached the line it was about. `git log -L 21,21:README.md` shows the hero line untouched since #717, so the correction landed everywhere except the claim a reader actually sees. Before that, #582 had already refreshed the same number once.
+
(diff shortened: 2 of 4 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues — clean, well-scoped docs + CI guard. Approving.

## What I checked
- `gh pr diff 893` (77 additions / 1 deletion across 4 files).
- `scripts/check-readme-line-count.mjs` logic, read in full — regex match, line-count math, exit codes for the three cases (corrected / stale / claim-deleted).
- Confirmed the CI wiring (`.github/workflows/ci.yml:20-25`) adds the new step to the existing `validate-package-json` job rather than a new required check.
- Spot-checked the repo: `src/` does contain 52 `.ts` files, consistent with the PR's claim.
- CI check-rollup: all checks (`validate-package-json`, `build`, `test`, `CodeQL`, `actionlint`, etc.) pass on this head SHA.

## Findings
None blocking. Two minor observations, not action items:

- `scripts/check-readme-line-count.mjs:22-27` — `tsFiles()` recurses through all of `src/` with no exclusion list (e.g. `.d.ts`, test fixtures). Not a bug today since the count matches the PR's stated 23,986/52 figures, but worth knowing if `src/` later grows a generated- or vendored-code subtree, since the tolerance (±1000) would absorb small drift anyway.
- The new CI step (`.github/workflows/ci.yml:20-25`) has no comment issue — the inline rationale for keeping it in the existing job is clear and matches the PR description.

## What's good
- The guard fails closed in both directions: stale claim *and* claim-deletion both exit 1 with an actionable message, so the check can't silently go stale itself (`scripts/check-readme-line-count.mjs:34-41`, `:52-58`).
- Tolerance (±1000 lines) is a reasonable balance — tight enough to catch real drift, loose enough not to trip on routine PRs.
- No new required status check added to branch protection; rides the existing `validate-package-json` job with no `npm install`, keeping CI cost near zero.
- Changelog and README are updated consistently with the script's own output format.

PR #899, opened 2026-08-02
docs: surface the Own Your Stack family above the fold, add plumbline

askalf:
## What

Two README-only changes to make the Own Your Stack family findable from dario:

1. **Header strip.** The `## Own Your Stack` section sits at the very bottom of a ~370-line README — below License — where the ~21k monthly npm installs' worth of readers never scroll. This adds one centered `<sub>` line to the header naming four siblings (truecopy · strongroom · fieldpass · plumbline) with an anchor to the full family section.
2. **Adds plumbline to the family list** — public since July, it was missing from the section (slotted after fieldpass: "own your agent oversight").

## Why

dario's README is the highest-traffic surface the family has; the family section exists but is effectively invisible at its current position. One quiet line above the fold, no layout changes, no nagware.

## Notes

- Docs-only: no version bump, so nothing ships per the release gate; the npm README updates at the next publish.
- Diff is 3 added lines, nothing removed.


sprayberry-redline (approved):

PR #900, opened 2026-08-02
fix(doctor): name a remedy that actually works for identity drift

askalf:
`dario doctor`'s Identity warning named a command that cannot work, in the exact situation where the user is already stuck.

## The bug

The Identity check fires when a pool account's stored `deviceId`/`accountUuid` no longer matches `~/.claude.json` — the state where non-Haiku requests come back 401 from Anthropic. Its remedy read:

> re-run `dario accounts add <alias>` to refresh the stored snapshot

Two things are wrong with that, and either alone is fatal:

1. `accounts add` **exits 1 on an alias that already exists** (`cli.ts`, "Account "<alias>" already exists. Remove it first"). A drifted alias exists by definition, so the suggested command can never succeed.
2. Its default path **runs a full OAuth browser flow**, not a re-snapshot of the current credentials.

Hit live on a box whose `login` alias had drifted: the doctor said to re-add, the CLI refused, and there was no route forward from the warning itself.

## The fix

The remedy now splits by alias kind, because the two cases genuinely differ:

- **The reserved `login` alias** needs only `dario accounts remove login`. It is back-filled from whatever credentials are current, so it re-materializes on the next `dario login` / `dario proxy`.
- **A user-added alias** needs `dario accounts remove <alias>` then `dario accounts add <alias>`, in that order — and there the add re-running OAuth is correct by design rather than by accident.

The warning also now leads with the consequence (401 on non-Haiku) and closes with the fix, instead of burying the 401 after a dead-end instruction.

## Why the test didn't catch it

`test/doctor-identity-drift.mjs` asserted `detail.includes('dario accounts add')` — which the **correct** instruction also contains. It passed on the broken message and would have passed on the fixed one, so it never had discriminating power.

It now asserts the ordering (`remove` before `add`) and adds a case for a drifted `login` alias, checking it is never routed through a bare re-add. Verified the new assertions actually discriminate: reverting only `src/doctor.ts` and rebuilding fails **5 assertions**; with the fix, 38/38 pass.

## Verification

- `npm run build` — clean
- `npm test` — **126/126 pass**
- `node scripts/check-package-json.mjs` — OK
- `node scripts/extract-release-notes.mjs 5.4.24 --file CHANGELOG.md` — returns the section (release body will not be empty)

Version bumped to **5.4.24** in-PR per the release gate, with the CHANGELOG's `[Unreleased]` section promoted to `[5.4.24]` per the file's own convention — which also ships the README line-count guard from #893 that has been sitting unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.24] - 2026-08-02
+
+- **`dario doctor` told you to run a command that cannot work.** The Identity check fires when a pool account's stored `deviceId`/`accountUuid` no longer matches `~/.claude.json` — the state where non-Haiku requests come back 401 — and its remedy read "re-run `dario accounts add <alias>` to refresh the stored snapshot". `accounts add` exits 1 on an alias that already exists, telling you to remove it first, and its default path runs a full OAuth browser flow rather than re-snapshotting anything. The single instruction the warning gave was a dead end for every user who reached it, in the one situation where they were already stuck.
+
+  The remedy now splits by alias kind, because the two cases genuinely differ. The reserved `login` alias needs only `dario accounts remove login`: it is back-filled from whatever credentials are current, so it re-materializes on the next `dario login` / `dario proxy`. A user-added alias needs `dario accounts remove <alias>` then `dario accounts add <alias>`, in that order, and that add re-runs OAuth for the account by design rather than by accident.
+
+  The unit test covering this line asserted the broken advice. It checked that the detail string contained `dario accounts add` — which the *correct* instruction also contains — so it would have passed either way and never had a chance of catching this. It now asserts the ordering (remove before add) and that a drifted `login` is never routed through a bare re-add.
+
 - **The README's "~22k lines you can read in a weekend" was understating by 9%, and CI now measures it.** `src/` is **23,986 lines across 52 files**; the hero line said `~22k`. It now says `~24k`.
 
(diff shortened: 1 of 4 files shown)


sprayberry-redline (approved):

**What I checked**
- `gh pr diff` for all 4 changed files (`src/doctor.ts`, `test/doctor-identity-drift.mjs`, `CHANGELOG.md`, `package.json`).
- CI check-rollup: all green (CodeQL, build×3, test, docker-cap-drop-smoke, compat, analyze, validate-package-json).

**Findings**

None blocking. The core fix (`src/doctor.ts:199-219`) correctly splits the remedy by alias kind:
```
const loginDrifted = driftedAliases.includes(MIGRATED_LOGIN_ALIAS);
const otherDrifted = driftedAliases.filter((a) => a !== MIGRATED_LOGIN_ALIAS);
```
This is a real bug fix — the old detail string unconditionally told every drifted user to run `dario accounts add <alias>`, which the PR body documents exits 1 on an existing alias. The new logic routes the reserved `login` alias to a bare `remove` and any other alias to `remove` then `add`, matching the stated CLI behavior.

**What's good**
- The test rewrite in `test/doctor-identity-drift.mjs` genuinely improves discriminating power: it now asserts ordering (`remove` before `add`) via `indexOf` comparison, and adds a dedicated case asserting a drifted `login` alias is never routed through `accounts add login`. The PR body's claim that reverting only `src/doctor.ts` fails 5 assertions is a good verification practice.
- `MIGRATED_LOGIN_ALIAS` is imported from `./accounts.js` rather than hardcoding `"login"` again, avoiding a second source of truth.
- Version bump and CHANGELOG entry follow the repo's stated release-gate convention.

No test-coverage gaps identified for the changed logic — both alias-kind branches (login-only, other-alias, and by implication the untouched aligned/no-drift path) are covered by the existing and new test cases.

PR #904, opened 2026-08-03
feat(oauth): DARIO_IGNORE_CC_CREDENTIALS — run dario next to a live Claude Code session without rotating its OAuth

askalf:
## The problem

When dario runs on the **same machine/account** as an interactive `claude` session, `loadCredentials()` reads dario's own `~/.dario/credentials.json`, CC's `~/.claude/.credentials.json`, **and** the OS keychain, then uses the *freshest*. The moment CC's token is fresher — e.g. right after you log into CC — dario grabs that same token, refreshes it, Anthropic rotates it, and the interactive session still holding the old copy starts 401ing.

That's the "dario broke my Claude Code login" failure, and until now there was **no config that keeps the subscription AND stops it**:

- `ANTHROPIC_UPSTREAM_API_KEY` isolates, but forwards via `x-api-key`, bypassing OAuth/Max + the pool entirely → **retail per-token billing**, which defeats the entire purpose of running dario.
- The credential source list `[darioPath, ccPath]` + keychain was **hardcoded with no opt-out**.

## The fix

`DARIO_IGNORE_CC_CREDENTIALS=1` (accepts `1`/`true`/`yes`/`on`) narrows the source set to dario's **own** `~/.dario/credentials.json` only — never the CC file **or** the keychain (modern CC stores its token there, so isolating from the interactive session means skipping both).

Result: dario stays on the Max subscription (**$0**) and can never rotate a live `claude` session's token.

- `ignoreCcCredentials(env)` and `credentialSourcePlan(ignoreCc)` are pure + exported so the isolation guarantee is testable without real credentials on disk (same pattern as `pickFreshestCredentials`).
- **Default off** — unset behaves exactly as before (reads all sources, picks freshest). Purely additive, no behaviour change for anyone not setting it.
- Surfaced in `dario config` (OAuth section) and the proxy startup banner so it's visible which mode is active.
- Requires a prior `dario login` when set; the banner says so.

## Test plan

- [x] 27 new assertions (`test/ignore-cc-credentials.mjs`): env parsing across `1/true/yes/on` + falsey/whitespace/missing cases; both source plans; and the core property — the isolated plan reaches **neither** the CC file **nor** the keychain, while the default still includes a CC source.
- [x] Full suite **127/127**
- [x] `tsc --noEmit` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.25] - 2026-08-03
+
+- **`DARIO_IGNORE_CC_CREDENTIALS` — run dario next to a live `claude` session without rotating its OAuth.** When dario runs on the same machine/account as an interactive Claude Code session, `loadCredentials()` reads dario's own `~/.dario/credentials.json`, CC's `~/.claude/.credentials.json`, AND the OS keychain, then uses the *freshest*. The moment CC's token is fresher (e.g. right after you log into CC), dario grabs that same token, refreshes it, Anthropic rotates it, and the interactive session still holding the old copy starts 401ing — the "dario broke my Claude Code login" failure.
+
+  Set `DARIO_IGNORE_CC_CREDENTIALS=1` (accepts `1`/`true`/`yes`/`on`) and dario uses ONLY its own `~/.dario/credentials.json` — from a prior `dario login` — never the CC file or keychain. This keeps dario on the Max subscription ($0) while guaranteeing it can never touch the interactive session's token. It is env-only, default off (unchanged behaviour), and surfaced in `dario config` and the proxy startup banner. The existing `ANTHROPIC_UPSTREAM_API_KEY` also isolates, but bypasses OAuth/Max onto retail per-token billing — which defeats the purpose of running dario at all; this flag is the isolation path that preserves the subscription.
+
 ## [5.4.24] - 2026-08-02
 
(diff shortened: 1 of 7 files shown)


sprayberry-redline (approved):

## What I checked
- `gh pr diff` for all 7 changed files (`CHANGELOG.md`, `package.json`/`package-lock.json` version bump, `src/config-report.ts`, `src/oauth.ts`, `src/proxy.ts`, `test/ignore-cc-credentials.mjs`).
- CI check-rollup: all 12 checks green (CodeQL, actionlint, analyze, build ×3, compat, docker-cap-drop-smoke, test, triage, validate-package-json, wire-drift).

## Findings
None blocking. The change is additive and gated:
- `ignoreCcCredentials(env)` (`src/oauth.ts`) is a pure env parser defaulting to `false` when the var is unset — confirmed the historical behavior (reading dario's file + CC's file + keychain, picking freshest) is preserved when the flag isn't set.
- `credentialSourcePlan(ignoreCc)` cleanly narrows `filePaths`/`readKeychain` based on the flag, and `loadCredentials()` (`src/oauth.ts:466-483`) correctly gates both the file loop and the keychain read (`if (plan.readKeychain) { ... }`) on the same plan — so setting the flag skips the CC file *and* the keychain, matching the stated isolation guarantee (no path where one is skipped but not the other).
- `src/proxy.ts:8` import and the new banner line at `src/proxy.ts:1273` are in an `else if` after the existing `upstreamApiKey` banner check, so the two banners can't both print — consistent with existing style.
- `src/config-report.ts` surfaces the new setting in the OAuth report section without touching unrelated rows.
- Test file adds focused, deterministic coverage of the pure functions (env parsing across truthy/falsey/whitespace values, both source plans) without needing credentials on disk.

## What's good
- Default-off, purely additive — no behavior change for anyone not setting the var.
- The isolation guarantee (skip CC file *and* keychain together) is tested directly against the pure `credentialSourcePlan`/`ignoreCcCredentials` functions rather than only indirectly through `loadCredentials`.
- Version bump and CHANGELOG entry are consistent with `package.json`/`package-lock.json`.

PR #910, opened 2026-08-04
fix(proxy): request-queue slot leak on a connected-but-not-reading client (#905)

### LINKED ISSUE #905: proxy: request queue never drains — all requests 504 queue-timeout for hours while /health stays green
## Summary

On a single-user, low-concurrency setup, the request queue stops admitting requests and never recovers. Every `POST /v1/messages` sits until it 504s with `queue-timeout` (dario#80), for hours at a time, while `GET /health` keeps returning `{"status":"ok","oauth":"healthy"}` in ~2ms the entire time. Only `systemctl --user restart dario` clears it; after a restart the very next request returns 200 in ~1.9s.

Four separate episodes in the last week, totalling ~14h of downtime.

## Error

```
HTTP 504: dario queue timeout — request waited longer than 120000ms for a
concurrency slot. Tune --queue-timeout, or reduce client-side concurrency. (dario#80)
```

## Environment

- dario **5.4.24** (also seen on 4.8.x — see below)
- Node v22.22.0, Ubuntu, kernel 6.8, `dario proxy` as a systemd **user** service
- Pool of 1 account, OAuth subscription billing (`five_hour`), `Model: passthrough`
- `DARIO_QUEUE_TIMEOUT_MS=120000`; `--max-concurrent` / `--max-queued` left at defaults (10 / 128)
- Client is a self-hosted Discord gateway (Hermes) calling `model=claude-opus-5`. dario logs `detected hermes-style text-tool protocol — auto-enabling preserve-tools for this client` at startup.

## Occurrences

Counted from the client's own error log (each event is one 504 returned to the client):

| Date (UTC) | First | Last | Span | 504s | `queueTimeoutMs` in force |
|---|---|---|---|---|---|
| 2026-07-29 | 15:32:47 | 15:32:47 | — | 2 | 60000 (default) |
| 2026-07-30 | 13:16:00 | 15:48:02 | 2h32m | 151 | 60000 (default) |
| 2026-08-03 | 04:16:20 | 13:13:49 | 8h57m | 84 | 300000 |
| 2026-08-04 | 00:16:05 | 03:06:49 | 2h50m | 31 | 120000 |

Two things this rules out:

- **Not the queue-timeout setting.** It reproduces at 60s (default), 300s, and 120s. Requests never get a slot no matter how long they're willing to wait.
- **Not a regression in a recent release.** The 07-29/07-30/08-03 episodes were on 4.8.x; 08-04 was on 5.4.24, upgraded the previous afternoon. `src/request-queue.ts` has only 2 commits (last `ded77615`, 2026-04-24), so both versions run identical queue code.

## Why this looks like a real bug rather than genuine overload

The client is a single-user Discord gateway. Normal concurrency is 1–2 in-flight requests; it is not plausible that 10 slots were legitimately occupied continuously for nearly 9 hours on 08-03. Whatever is holding `active` at `maxConcurrent` never drains until the process is restarted.

I want to be careful not to over-claim a mechanism: I read the compiled `dist/request-queue.js` and `dist/proxy.js` on 5.4.24, and **the obvious suspects look correct** —

- `decideAdmit()` / `release()` logic is sound; `release()` decrements and hands the slot to the next FIFO entry.
- `queue.acquire()` and `queue.release()` are properly paired: `release()` sits in a `finally`, and the early `return`s on the body-size and pool-empty paths are inside the `try` that `finally` covers.

So I can't point at a leak in the accounting, and I'm not asserting one. Reporting the observable behaviour instead.

One thing that stood out while reading, offered only as a possible lead: in `dist/proxy.js` (5.4.24) `queue.acquire()` is at line 2033, but `upstreamTimeout` is not armed until line 2746. Anything that stalls in that window holds a slot with no `UPSTREAM_TIMEOUT_MS` covering it. Most awaits in that range are bounded, so this may well be nothing — but it is the one stretch where a hang would be unbounded.

## The part that would actually help everyone: make the queue observable

There is currently **no way to see `active` / `queued` from outside the process.** `src/request-queue.ts` documents `snapshot()` as *"Snapshot of queue state — exposed for /analytics + tests"*, but the `/analytics` payload on 5.4.24 contains no queue fields — I checked:

```
$ curl -s localhost:3456/analytics | jq 'keys'
["allTime","perAccount","perModel","predictions","utilization","window"]
```

The practical consequence is that `/health` is actively misleading during this failure. It returns `status: ok` with a healthy OAuth countdown while every real request is failing, so nothing that monitors dario can detect this state — I only found it by sending a real inference call and watching it hang.

**Request:** surface `queue.snapshot()` (`active`, `queued`, `maxConcurrent`, `maxQueued`) on `/health` and/or `/analytics`. That would let anyone hitting this confirm slot exhaustion vs. an upstream stall in one curl, and would let me hand you actual numbers from the next episode instead of inference.

Happy to run any diagnostics on the next occurrence — it recurs every couple of days here, so it's reproducible on this box even though I don't have a minimal repro.

## Workaround, for anyone finding this issue

`/health` cannot detect it. A watchdog has to send a real inference request:

```bash
curl -sf -m 45 http://localhost:3456/v1/messages \
  -H 'content-type: application/json' -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"claude-haiku-4-5-20251001","max_tokens":4,
       "messages":[{"role":"user","content":"ping"}]}' | grep -q '"content"' \
  || systemctl --user restart dario
```

askalf:
## Summary

Fixes #905. Diagnosis, confirmed with a reproduction, not a hunch:

A streaming client that stays **connected but stops reading** (a wedged consumer, or a dead peer behind an open TCP window) can permanently hold a request-queue concurrency slot:

1. `res.write()` returns `false` once the client's socket buffer fills, and the SSE forwarding loop parks in a drain wait.
2. Pre-fix, that wait resolved only on `res` `'drain'`/`'close'` — events a silent-but-alive client never fires.
3. The 5-minute upstream timeout still fires and calls `upstreamAbort.abort()`, but nothing is awaiting `reader.read()` at that moment, so the abort has no effect on the parked handler.
4. The handler's `finally` — where `queue.release()` lives — never runs. The slot leaks until process restart.

Enough of these accumulate over hours/days and pin `active` at `maxConcurrent`; every subsequent request then queues and 504s `queue-timeout`, while `/health` stays green throughout (it's an unrelated fast path). This matches all four episodes in #905 exactly, including `/health` staying healthy and only a restart clearing it.

The original report also correctly ruled out the queue accounting itself (`decideAdmit`/`release()` pairing is sound) and asked for `queue.snapshot()` to actually be wired into `/health`/`/analytics` — it had documented itself as "exposed for /analytics + tests" since dario#80 but never was.

## Fix

- `waitForClientDrain` (new, `src/stream-drain.ts`) resolves the drain wait on `res` `'drain'`, `res` `'close'`, **or the upstream `AbortSignal` firing** — the same signal the existing timeout/client-close/SSE-overflow paths already abort through. Once resolved, the next `reader.read()` rejects on the aborted body (a mechanism those other abort reasons already rely on), normal error teardown runs, and the slot releases. Worst-case slot hold time is now bounded by the upstream timeout instead of unbounded.
- `/health` and `/analytics` both now include `queue: { active, queued, maxConcurrent, maxQueued }` for authenticated/internal callers.
- `ProxyOptions.upstreamTimeoutMs` (new, optional) makes the previously-hardcoded 5-minute constant injectable, so the repro doesn't need to wait 5 real minutes.

## Verification

- `test/queue-slot-leak.mjs` (new): spins a real `startProxy()` with a scripted infinite-SSE upstream, connects a raw-socket client that reads the first bytes then pauses forever, and asserts (a) the slot is held mid-stream, (b) it releases once the upstream timeout fires despite the silent client, and (c) a follow-up request is served 200 rather than 504. Also asserts the `/health`/`/analytics` queue fields.
- `test/stream-drain.mjs`: 5 new cases for `waitForClientDrain` — resolves on drain, on close, on abort (the fix), immediately if already aborted, and no listener accumulation across 50 repeated waits on one signal.
- **Confirmed the new test actually catches the bug**: ran it against unmodified `master` (git stash A/B) and it fails in exactly the way #905 describes (slot stuck at `active:1`, follow-up request 504s `queue-timeout`). Then confirmed it passes with the fix applied.
- Full suite: 126/128 passing. The 2 failures (`template-interactive-tools.mjs`, `tool-advertise-respects-client.mjs`) are pre-existing — verified failing identically on unmodified master via the same A/B stash check, unrelated to this change.
- `npm run build` clean, `npm run audit` 0 vulnerabilities.

Bumps 5.4.27 → 5.4.28 per the in-PR version-bump convention (a PR that doesn't touch `package.json` merges but ships nothing).

## Test plan

- [x] `npm run build`
- [x] `npm test` (126/128, 2 pre-existing unrelated failures)
- [x] `node test/queue-slot-leak.mjs` — reproduces the wedge and confirms the fix
- [x] `node test/stream-drain.mjs`
- [x] `npm run audit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [5.4.28] - 2026-08-04
+
+- **Fixed: request-queue slot leak on a connected-but-not-reading streaming client (#905).** A client that stayed connected but stopped reading mid-stream (dead peer behind an open TCP window, a wedged consumer) could permanently hold one of the `--max-concurrent` slots. `res.write()` returning `false` parked the SSE forwarding loop in a drain wait that resolved only on `res` `'drain'`/`'close'` — events a silent-but-alive client never fires. The 5-minute upstream timeout still fired and aborted the upstream fetch, but nothing was awaiting `reader.read()` at that moment, so the request handler's `finally` (where `queue.release()` lives) never ran. Each such client permanently ate one slot; enough of them over hours/days pinned `active` at `maxConcurrent`, and every subsequent request queued until it 504'd with `queue-timeout` — while `/health` stayed green throughout, because it's a separate fast path uninvolved in the wedge. Only a full process restart cleared it.
+
+  Fix: the drain wait (`waitForClientDrain` in `src/stream-drain.ts`) now also resolves when the upstream `AbortSignal` fires — the same signal the existing 5-minute timeout, client-disconnect, and SSE-overflow paths already abort through. Once the wait resolves, the next `reader.read()` rejects on the aborted body (already relied on by those other abort reasons), normal error teardown runs, and the slot releases — bounding worst-case slot hold time to `UPSTREAM_TIMEOUT_MS` instead of forever.
+
+- **`/health` and `/analytics` now surface the request-queue snapshot (#905).** `src/request-queue.ts`'s `snapshot()` had documented itself as "exposed for /analytics + tests" since the queue was introduced (dario#80), but was never actually wired into either endpoint — there was no way to see `active`/`queued` from outside the process, so slot exhaustion was invisible short of sending a real inference request and watching it hang. Both endpoints now include `queue: { active, queued, maxConcurrent, maxQueued }` for authenticated/internal callers; `active === maxConcurrent` with `queued > 0` sustained is the slot-exhaustion signature.
+
 ## [5.4.27] - 2026-08-04
 
 - **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.221` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.221 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "5.4.25",
+  "version": "5.4.28",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "5.4.25",
+      "version": "5.4.28",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.4.27",
+  "version": "5.4.28",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/health-response.ts
+++ b/src/health-response.ts
@@ -31,6 +31,13 @@ export interface HealthStatusLike {
   sessions?:
     | { mode: 'pool'; stickyBindings: number }
     | { mode: 'single'; active: number };
+  /**
+   * Request-queue snapshot (dario#905), surfaced to internal callers only.
+   * `active === maxConcurrent` with `queued > 0` for a sustained period is
+   * the slot-exhaustion signature — before this field existed, that state
+   * was invisible from outside the process while every request 504'd.
+   */
+  queue?: { active: number; queued: number; maxConcurrent: number; maxQueued: number };
 }
 
 export interface HealthResponse {
@@ -144,6 +151,7 @@ export function buildHealthResponse(
         expiresIn: s.expiresIn,
         requests: requestCount,
         ...(s.sessions ? { sessions: s.sessions } : {}),
+        ...(s.queue ? { queue: s.queue } : {}),
         ...(s.refreshFailures ? { refreshFailures: s.refreshFailures } : {}),
       }
     : liveness;
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -920,6 +920,12 @@ interface ProxyOptions {
   /** Max ms a queued request waits before it times out with 504. Default 60000. dario#80. */
   queueTimeoutMs?: number;
   /**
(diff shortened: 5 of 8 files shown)


sprayberry-redline (approved):

**What I checked**
- `src/stream-drain.ts`: `waitForClientDrain` settles on `drain`, `close`, or the upstream `AbortSignal` firing, resolves immediately if already aborted, and unconditionally removes all three listeners (`drain`, `close`, `abort`) in its single `done` callback — including the abort-listener cleanup that the pre-existing `drain`/`close`-only version didn't need. `test/stream-drain.mjs` directly exercises the immediate-abort case and a 50-iteration repeated-wait loop against one shared `AbortSignal` with a `MaxListeners` warning assertion, which is the right test for exactly this kind of listener leak.
- `src/proxy.ts`: the wiring at `waitForDrain = () => waitForClientDrain(res, upstreamAbort.signal)` (was previously local `res`-only `drain`/`close` handling) plugs into the existing `upstreamAbort` controller that the timeout (`upstreamTimeout = setTimeout(...)`), client-close, and SSE-overflow paths already share, so the abort path correctly routes back through the existing `catch`/`finally` (`queue.release()` at the `finally` block) rather than adding a second teardown path.
- `UPSTREAM_TIMEOUT_MS` → `upstreamTimeoutMs` (injectable via new `ProxyOptions.upstreamTimeoutMs`) is a mechanical, consistently-applied rename across all use sites (fetch timeout, both `forwardToOpenAI` call sites, the 504 message) — no site was missed.
- `/health` and `/analytics` `queue` field additions stay behind the existing auth gates (`checkAuth` before `/analytics`; `shouldDiscloseHealthInternals`'s existing authenticated/loopback/cf-ray logic for `/health`), so this doesn't introduce a new disclosure path.
- `test/queue-slot-leak.mjs` reproduces the wedge with a scripted infinite-SSE upstream and a raw-socket client that stops reading, and asserts the slot releases once the injected short timeout fires and that a follow-up request succeeds — a real regression test for the actual bug, not just the helper in isolation.
- CI rollup is all-green (build ×3 node versions, test, compat, CodeQL, wire-drift, docker-cap-drop-smoke, actionlint, validate-package-json).

**What's good**
The abort-signal arm is the minimal fix for the actual root cause (nothing awaits `reader.read()` while parked in the old drain-only wait), reuses the controller/signal the codebase already threads through every other abort path instead of adding a new one, and the PR ships both a unit test for the primitive and an end-to-end repro for the leak itself.

No minor notes beyond the above — this is a clean, well-tested fix.

PR #916, opened 2026-08-05
feat(admin): bulk re-auth — start-needed + batch complete (#913)

### LINKED ISSUE #913: auto-login
I have no idea if it's possible :)

but you think it will be possoble for an headless setup to 'simplify' o 'reduce' the number of iteration required to auth an user?

idea 1:
new api called
/admin/getAllAccountThatNeedARefresh (lol, ok name suck)
return a list of account that need a new login (oauth token, some errors) with already the link for unattended login presented. in this way, i call 1 api and i got the list of all redirect

idea 2:
new api called
/admin/completeAll
get a param of array with alias: token: and update in 1 shot all the account
work like "complete" but for an array of user

idea 3:
will be possible to add a url for callback?
imagine if my dario account is in
docker.local:123
can be possible to generate redirect url to point to doker.local:123 (reachable from my pc obviusly) so i don't have to do the complete step? this will meld perfectly with idea1, i got a list of accounts, kick on all link, make sure to use the right subscription, and it's done
idea3 can be exploited with an env variabl elike "DARIO_DOMAIN" or "DARIO_CALLBACK"?

Thankyou

askalf:
## Summary

Closes #913 (community feature request, "auto-login"). Three ideas were proposed; two are implemented, one turned out to already be solved.

- **`POST /admin/login/start-needed { threshold? }`** — finds every pool account whose `consecutiveAuthFailures` has crossed `threshold` (default 3) and starts a fresh login for each, returning every `authorize_url` in one response.
- **`POST /admin/login/complete`** now also accepts a batch body (`{ "items": [{alias,code}, ...] }`) alongside the existing single-item form. Per-item errors don't abort the batch — `results[i].status` is `"added"` or `"error"` per item.
- **`GET /admin/accounts`** now also surfaces `consecutive_auth_failures`.

**Threshold, not the raw `auth-cooldown` boolean — verified empirically, not assumed.** I built a small harness against the real `AccountPool` (not just read the code) and confirmed a single 401 already puts an account into the *same* `auth-cooldown` status for 60s as a genuinely dead refresh token — the two are indistinguishable by that field alone. `consecutiveAuthFailures` climbing past a few retries is the only signal that actually separates "blip" from "needs a human": it grows unbounded while the cool-down *window* caps at 30min, and a single success anywhere resets the whole streak to 0 regardless of depth. Filtering `start-needed` on the raw cooldown boolean would have false-positived on every account that failed once a minute ago.

**Rate-limited per account acted on, not per HTTP request.** Checked `rate-limit.ts`: `tryRemove()` consumes one token per call. A naive "gate once, then loop internally" design would let one request move N accounts' credentials for the price of one mutation-bucket token — defeating the throttle's purpose. Both bulk endpoints consume the bucket once per item they actually act on, and report `truncated: true` + whatever succeeded so far if the bucket runs dry mid-batch, rather than borrowing against it.

**Idea 3 (configurable callback domain / `DARIO_DOMAIN`) is moot — confirmed, not assumed.** Called `startAddAccount()` for real and inspected the live `authorize_url`: `redirect_uri` is Anthropic's own hosted `platform.claude.com` page, not anything dario serves. There's no dario-hosted callback in this flow to redirect elsewhere — it already works from any browser on any machine, which is the entire point of the manual-paste flow. Left a comment on #913 explaining this.

## Implementation notes

`doStartLogin`/`doCompleteLogin` factor the single-item route logic out of the handlers so the bulk endpoints reuse it exactly (same `MAX_PENDING` guard, same audit calls, same pending-login bookkeeping) instead of duplicating it.

One small behavior change riding along, called out explicitly: a failed `completeAddAccount` is now audited (`login_complete ok:false`) — previously it only hit the outer catch-all silently. Needed for a batch run's per-item failures to be discoverable in the audit trail at all; without it a 10-item batch with 3 silent failures would be undebuggable after the fact.

## Test plan
- [x] 28 new tests in `test/admin-api.mjs` (85 total, 0 fail): threshold filtering (below/at/above, custom threshold), no-op outside pool mode, rate-limit truncation on both bulk endpoints, batch per-item error reporting, `MAX_BATCH_ITEMS` cap, audit trail for the new failure-audit behavior, single-item form unchanged.
- [x] Confirmed the new test suite fails appropriately against pre-refactor behavior isn't applicable here (new endpoints), but re-verified all *existing* 57 tests still pass unmodified against the refactor.
- [x] `npm run build` (tsc) clean.
- [x] Full suite: 126/128 pass. The 2 failures (`template-interactive-tools.mjs`, `tool-advertise-respects-client.mjs`) reproduce identically on unmodified `master` (verified via `git stash`) — pre-existing CC-template drift, unrelated to this change.
- [x] `docs/admin-api.md` + `CHANGELOG.md` updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+- **Admin API: bulk re-auth (#913).** `POST /admin/login/start-needed` finds every pool account whose `consecutiveAuthFailures` has crossed a threshold (default 3 — a single 401 already shows the same `auth-cooldown` status for 60s, so the raw cool-down flag alone can't tell a blip from a dead refresh token) and starts a fresh login for each in one call. `POST /admin/login/complete` now also accepts a batch body (`{ "items": [{alias,code}, ...] }`) alongside the existing single-item form, completing several pending logins per request. Both bulk endpoints rate-limit per account acted on, not per HTTP call, and report `truncated: true` if the mutation bucket runs dry mid-batch. `GET /admin/accounts` now also surfaces `consecutive_auth_failures`. See `docs/admin-api.md`.
+
 ## [5.4.29] - 2026-08-05
 
 - **CC drift patch** — `SUPPORTED_CC_RANGE.maxTested` bumped `2.1.221` → `2.1.222` for CC v2.1.222. Auto-drafted by `cc-drift-watch.yml`. Template re-capture, if needed, is auto-handled by `cc-drift-template-watch.yml`.
--- a/docs/admin-api.md
+++ b/docs/admin-api.md
@@ -88,17 +88,88 @@ All endpoints accept the token as `authorization: Bearer <token>` or
 | Method + path | Body | Returns |
 |---|---|---|
 | `POST /admin/login/start` | `{ "alias"?: string }` | `{ alias, authorize_url, expires_at, instructions }` |
+| `POST /admin/login/start-needed` | `{ "threshold"?: number }` | `{ started: [...], count, truncated? }` |
 | `POST /admin/login/complete` | `{ "alias": string, "code": string }` | `{ alias, status: "added", expires_at }` |
+| `POST /admin/login/complete` (batch) | `{ "items": [{ "alias", "code" }, ...] }` | `{ results: [...], count, truncated? }` |
 | `GET /admin/accounts` | — | `{ accounts: [...], count }` |
 | `DELETE /admin/accounts/<alias>` | — | `{ alias, removed }` (`404` if no such alias) |
 
 `GET /admin/accounts` is the monitoring surface: each entry carries the
 persisted metadata (`alias`, `scopes`, `expires_in_ms`) **plus live pool
 status whenever pool mode is active** — `util5h` / `util7d` utilization,
-representative `claim` (e.g. `five_hour`), routing `status`, and
-`request_count`. It's the admin-token-gated equivalent of the proxy-key-gated
-`GET /accounts` pool view; a headless operator needs only the admin token to
-watch headroom.
+representative `claim` (e.g. `five_hour`), routing `status`,
+`request_count`, and `consecutive_auth_failures`. It's the admin-token-gated
+equivalent of the proxy-key-gated `GET /accounts` pool view; a headless
+operator needs only the admin token to watch headroom.
+
+## Bulk re-auth, in one round-trip
+
+For a pool with several accounts, the round-trip of "notice one's broken,
+`/start` it, `/complete` it" per account doesn't scale. Two endpoints collapse
+that:
+
+```bash
+# 1. One call: every account that's actually stuck, each with a ready link.
+curl -s -X POST -H "$ADMIN" "$BASE/admin/login/start-needed"
+# -> { "started": [
+#      { "alias": "acct-3", "authorize_url": "...", "expires_at": "...",
+#        "consecutive_auth_failures": 5 },
+#      { "alias": "acct-7", "authorize_url": "...", "expires_at": "...",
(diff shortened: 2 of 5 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-tested change. Two non-blocking observations below.

## What I checked
- Full diff (`src/admin-api.ts`, `src/proxy.ts`, `test/admin-api.mjs`, docs/changelog).
- CI rollup: all 13 checks green (`test`, `build (18/20/22)`, `CodeQL`, `compat`, `docker-cap-drop-smoke`, `wire-drift`, etc.) — `gh pr checks 916`.
- Cross-referenced `consecutiveAuthFailures`/`authCooldownMs` against `src/pool.ts` to confirm the threshold semantics described in the PR body actually match the pool's cooldown accounting.
- Traced the rate-limit gating change: `isMutation` in `src/admin-api.ts` no longer includes `/admin/login/complete` at the top-level pre-parse gate — verified it's still throttled per-item inside `doCompleteLogin`'s caller, both for the single-item form (`src/admin-api.ts`, the `sendThrottled(...)` call right before `doCompleteLogin`) and the batch loop (`if (wait > 0) { truncated = true; break; }`), so the "N accounts for one token" bypass the PR body calls out is actually closed.

## Findings

**Minor — bulk-endpoint throttling isn't audited, unlike the single-item path.**
`src/admin-api.ts`, inside the `start-needed` and batch-`complete` loops:
```
const wait = deps.rateLimit?.('mutation') ?? 0;
if (wait > 0) { truncated = true; break; }
```
This sets `truncated` and returns 200, whereas the single-item guard path calls `sendThrottled(...)`, which calls `deps.audit?.({ action: 'rate_limited', ... })`. A bulk request that gets throttled mid-batch leaves no `rate_limited` audit event, only the response body's `truncated: true` — an operator scanning the audit log won't see it. Not a correctness bug (the throttle itself works), just an observability gap worth a follow-up if the audit log is relied on for this.

**Minor — a mutation-bucket token can be spent on a no-op start.**
`src/admin-api.ts`, `start-needed` loop:
```
const wait = deps.rateLimit?.('mutation') ?? 0;
if (wait > 0) { truncated = true; break; }
const result = await doStartLogin(c.alias, now, deps, remote);
```
`doStartLogin` can still return `ok: false` (e.g. `pendingLogins.size >= MAX_PENDING` for a brand-new alias), but the token was already consumed before that check. In the unlikely case a bulk `start-needed` run hits `MAX_PENDING`, it burns bucket budget on candidates it ends up skipping rather than reserving that budget for accounts it actually starts. Low real-world impact (`MAX_PENDING` is 64 and pools are typically much smaller) — noting for awareness, not blocking.

## What's good
- The threshold-vs-cooldown-boolean distinction is exactly right: `pool.ts` shows `isInAuthCooldown` gates purely on `lastAuthFailureAt` + `authCooldownMs(consecutiveAuthFailures)`, so a single failure and a dead token really are indistinguishable by `status` alone — `NEEDS_LOGIN_THRESHOLD` is a legitimate fix for that, not a false claim.
- `doStartLogin`/`doCompleteLogin` extraction is a clean, low-risk refactor — same guards, same audit calls, same pending-login bookkeeping, just callable per-item.
- The now-audited failed token exchange (`login_complete ok:false`) is a real, useful behavior change and is explicitly called out in the PR body and covered by a new test.
- Per-item rate-limiting inside the bulk handlers (rather than gating once at the top) correctly prevents the "one HTTP request moves N credentials for one token" issue the PR describes.
- Extensive, specific test coverage (threshold boundaries, no-pool no-op, truncation on both endpoints, batch per-item errors, `MAX_BATCH_ITEMS` cap, audit trail change, single-item form unchanged) — I didn't spot an untested code path in the diff.

PR #922, opened 2026-08-07
feat(health): prove dario can serve, not just that it looks configured (#905)

askalf:
## The gap

dario's two health surfaces are both structural:

| Endpoint | Answers |
|---|---|
| `GET /livez` | Is the HTTP server accepting connections? |
| `GET /health` | Do credentials and the pool look serviceable? |

Neither proves a request can complete. That is exactly the shape #905 reported: `/health` returning `{"status":"ok"}` in 2ms while every `POST /v1/messages` 504'd, for ~14h across four episodes. #910 fixed that specific drain bug and wired the queue snapshot in, but the *class* is still uncovered — and the reporter's remediation is still an external watchdog sending real inference calls and restarting the service. The user is writing the liveness check dario should own.

There is also a failure `/health` **provably** cannot see: an account whose token is unexpired and refreshable (so `status: healthy`) but whose `accountUuid` has drifted, so upstream 401s every request. Structural inspection says fine. Only a round-trip says otherwise.

## What this adds

**`GET /health?probe=1`** — a minimal `max_tokens: 1` request to Anthropic using the same auth the request path uses (pool bearer, or `x-api-key` in upstream-API-key mode), folded into the response. A failed probe forces 503 even when the structural read is clean, so existing status-code-only monitors start catching outages they previously missed.

```json
{
  "status": "degraded",
  "oauth": "valid",
  "probe": { "ok": false, "reason": "auth-rejected", "status": 401,
             "latencyMs": 233, "ageMs": 4812, "model": "claude-haiku-4-5" },
  "queue": { "active": 10, "queued": 4, "maxConcurrent": 10,
             "stalledSince": 1754790000000, "stalledForMs": 28800000 }
}
```

**`queue.stalledSince` / `stalledForMs`** — the wedge signal, separate from the probe and free.

## Design notes worth reviewing

**Why a probe costs nothing unless you ask.** Three guards: opt-in (a plain `/health` never spends a token, so docker healthchecks keep their current zero cost); gated to trusted callers; cached + single-flighted (`DARIO_PROBE_TTL_MS`, default 60s), so polling every second bills at most once per minute.

**The gate is stricter than #642 disclosure, on purpose.** The disclosure rule treats `authenticated` as sufficient — and `authenticateRequest` returns `true` when no `DARIO_API_KEY` is configured. Harmless for a read-only field; not harmless for an action that spends money, since an unkeyed dario published through a Cloudflare tunnel would expose `?probe=1` as a button any anonymous caller could press to bill the operator. `shouldRunServingProbe` refuses anything carrying `cf-ray` regardless of what `authenticated` says. It can only ever deny, never widen.

> This was found by the E2E test, not by reading. Pre-existing note, unchanged here and **not** fixed by this PR: on an unkeyed proxy the same short-circuit means `oauth`/`expiresIn`/`requests` are disclosed to tunnel callers too. Worth its own issue — I didn't want to change #642 semantics inside a feature PR.

**429 and 529 are not outages.** `ok` answers "would a restart or an alert help?", not "did it return 200". A throttled dario is working; a watchdog that restarts on it just thrashes — the same lesson `/livez` already encodes in its docstring after a shared-refresh-family outage restart-looped for 4h+. Auth rejection, 5xx, network failure and timeout set `ok: false`.

**The probe never takes a concurrency slot.** One that queued behind real traffic would report unhealthy during a legitimate burst and hand a watchdog a reason to restart a busy-but-fine dario. The concurrency axis is covered by `stalledSince` instead.

**Why stall, not saturation.** A busy dario runs at its cap with a backlog all day and is healthy — depth alone cannot distinguish it from the #905 wedge, since both sample as `active === maxConcurrent, queued > 0`. Turnover can. `stalledSince` is reset by any *release* and deliberately **not** by arrivals: refreshing on arrival would let a steady arrival rate mask a permanent wedge. `queue-stall.mjs` pins that false positive explicitly.

**What it does not claim.** The probe covers the credential/network axis; `stalledSince` covers concurrency. Neither exercises the CC transform path, and the module header says so rather than implying end-to-end coverage it doesn't have.

## Tests

- `test/serving-probe.mjs` (51) — verdict policy, caching, single-flight, bounded token acquisition, never-throws. Injected fetch, no network.
- `test/queue-stall.mjs` (30) — wedge vs busy on an injected clock, including the 8h-wedge single-sample case and the busy-but-flowing false positive.
- `test/health-probe-e2e.mjs` (29) — real `startProxy()`, counts actual upstream calls to pin both silent failure directions. This is what caught the `authenticateRequest` short-circuit.
- `test/health-response.mjs` extended 74 → 95.

Full suite: **129 pass / 131**. The two failures (`template-interactive-tools`, `tool-advertise-respects-client`) reproduce identically on pristine `origin/master` (`f3c9377`) on this machine and are unrelated — verified in a clean worktree before opening this.

Version bumped 5.4.31 → **5.5.0** in-PR per the release gate; `[Unreleased]` promoted.

Addresses #905.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,22 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.0] - 2026-08-07
+
+- **`/health` can now prove dario actually serves, not just that it looks configured (#905).** Both existing health surfaces are structural: `/livez` says the HTTP server is accepting connections, `/health` says credentials and the pool look serviceable. Neither can catch the shape that produced ~14h of downtime in #905 — dario green from the outside while every real request failed — which is why that reporter's remediation was an external watchdog sending real inference calls. That check is now in-house.
+
+  `GET /health?probe=1` sends a minimal `max_tokens: 1` request to Anthropic using the same auth the request path uses (pool bearer, or `x-api-key` in upstream-API-key mode) and folds the verdict into the response. A failed probe forces 503 even when the structural read is clean, so a status-code-only uptime monitor starts seeing outages it previously could not — including the case `/health` provably cannot see: an unexpired, refreshable token whose accountUuid has drifted, so upstream 401s every request.
+
+  Three guards, because a probe is a real billed request: it is **opt-in** (a plain `/health` never spends a token, so existing docker healthchecks and monitors keep costing nothing); it is honoured only for callers passing a gate **deliberately stricter than the #642 disclosure rule** — that rule treats "authenticated" as sufficient and `authenticateRequest` returns true when no `DARIO_API_KEY` is configured at all, which is harmless for a read-only field but would turn `?probe=1` on an unkeyed tunnel-published dario into a button any anonymous caller could press to bill the operator; `shouldRunServingProbe` therefore refuses anything arriving with `cf-ray` regardless of what `authenticated` says, and can only ever deny, never widen; and results are **cached and single-flighted** (`DARIO_PROBE_TTL_MS`, default 60s), so a monitor polling every second still costs at most one probe per minute.
+
+  Verdict semantics answer "would a restart or an alert help?", not "did it return 200". Rate-limited (429) and upstream-overloaded (529) stay `ok: true` — dario is working, and a watchdog that restarts on them just thrashes, the same lesson `/livez` already encodes. Auth rejection, 5xx, network failure and timeout set `ok: false`. The probe deliberately does not route through dario's own proxy path and never takes a concurrency slot: one that queued behind real traffic would report unhealthy during a legitimate burst and hand a watchdog a reason to restart a busy-but-fine dario.
+
+  Knobs: `DARIO_PROBE_MODEL` (default `claude-haiku-4-5`), `DARIO_PROBE_TTL_MS`, `DARIO_PROBE_TIMEOUT_MS`. See `docs/usage.md`.
+
+- **Queue slot exhaustion is now distinguishable from a busy queue (#905).** #910 put `active`/`queued` on `/health`, which made saturation visible but not diagnosable: a 14h wedge and a healthy one-second burst produce an identical sample (`active === maxConcurrent, queued > 0`). The distinguishing property is turnover, not depth — a busy dario runs at its cap with a backlog all day and is fine, while the #905 wedge held slots for hours with no release at all.
+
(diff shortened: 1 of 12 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-reasoned change with strong test coverage. Approving.

## What I checked
- Full diff (`src/health-response.ts`, `src/proxy.ts`, `src/request-queue.ts`, `src/serving-probe.ts`) plus the queue-stall state machine and the probe-gating logic, read against the surrounding unchanged code in a shallow clone.
- CI check-rollup: all 12 checks green (`test`, `build (18/20/22)`, `CodeQL`, `analyze`, `compat`, `docker-cap-drop-smoke`, `wire-drift`, etc.).

## Correctness
- `RequestQueue.updateStall()` (`src/request-queue.ts`): traced the stall-clock invariant across `acquire()` (admit/enqueue/timeout paths) and `release()`. Arrivals only *set* `stalledSince` when it's currently `null` (never reset an in-progress stall); `release()` unconditionally clears then re-evaluates, so sustained turnover never accumulates age while a genuine wedge does. This matches the documented invariant and I couldn't find a state transition that breaks it.
- `shouldRunServingProbe` (`src/health-response.ts`): the `viaCfRay` short-circuit is evaluated *before* `discloseInternals`, and is a pure deny-only refinement over the existing #642 disclosure gate — it can't be tricked into widening access via the auth/loopback checks.
- `buildHealthResponse`'s `dead = structurallyDead || s.probe?.ok === false` correctly only forces degraded when a probe actually ran and failed; `s.probe` being `undefined` (no probe requested/gated) leaves the existing structural verdict untouched.
- `getServingProbe`'s module-level `cache`/`inflight` singleton is assigned synchronously before any `await`, so concurrent same-tick callers correctly share one in-flight request rather than racing — no double-billing under JS's single-threaded execution model.
- Probe timeout (`AbortController`) is armed before token acquisition, and `Promise.race` against the abort signal prevents a hung `getToken()` from wedging `inflight` forever, exactly as the header comment claims.

## What's good
- Design rationale is unusually well-documented inline (why 429/529 aren't failures, why the probe skips the concurrency slot, why stall tracks turnover not depth) and matches the actual implementation.
- Four new/extended test files (`serving-probe.mjs`, `queue-stall.mjs`, `health-probe-e2e.mjs`, `health-response.mjs`) covering verdict policy, caching/single-flight, the wedge-vs-busy distinction, and the real gate bug the E2E test caught (`authenticateRequest` short-circuit). Good coverage for new behavior.
- Explicitly scopes what the probe does *not* claim to cover (CC transform path) rather than overselling it.

No local test run performed per policy — CI rollup is all-green and is the authoritative signal here.

PR #925, opened 2026-08-08
fix(health): an unkeyed proxy must not disclose OAuth internals to tunnel callers

askalf:
## The gap

`shouldDiscloseHealthInternals` grants the internal view to any caller `authenticateRequest()` accepts:

```ts
if (opts.authenticated) return true;   // ← returns before cf-ray is consulted
if (opts.viaCfRay) return false;
return opts.loopback;
```

And `authenticateRequest()` short-circuits:

```ts
if (!apiKeyBuf) return true;           // no DARIO_API_KEY configured
```

That short-circuit is deliberate and stays — the common setup is loopback-only, and requiring a key there would break `dario doctor` and every docker healthcheck. But it makes `authenticated` **vacuous** on an unkeyed proxy: *every* caller satisfies it, so the auth branch returns before `cf-ray` is ever reached.

Net effect: an unkeyed dario published through a Cloudflare tunnel hands `oauth`, `expiresIn`, `requests`, `sessions`, `queue` and `refreshFailures` to anyone who asks.

#642 closed this gate's *spoofable-header* direction — a caller can't grant itself the internal view by omitting `cf-ray`. This is the same fail-open re-entering through a side door: the auth branch simply never reaches the check.

## The fix

```ts
if (opts.authenticated && opts.keyConfigured) return true;
if (opts.viaCfRay) return false;
return opts.loopback;
```

The auth branch is now reachable only by a caller that actually presented the operator's secret. Unkeyed proxies fall through to the transport rules — where **bare loopback is still trusted**, which is why docker `HEALTHCHECK`, `dario doctor` and the TUI are unchanged.

`keyConfigured` is a **required** field rather than an optional with a default: for a security predicate, every call site should be forced to state it. One call site, in `proxy.ts`.

## Behaviour change

If you monitor `/health` **through a tunnel on a proxy with no `DARIO_API_KEY`**, you now get `{"status":"ok"}` instead of the full body. Set `DARIO_API_KEY` and send it, or query from loopback, to keep the detail.

The HTTP status (200/503) is unchanged for every caller, so status-code uptime checks are unaffected — that invariant is stated in the module docstring and is why the split is safe to tighten.

Documented in `docs/usage.md` with a who-sees-what table and an upgrade note.

## Also

`shouldRunServingProbe` (added in #922) justified itself partly by this bug, so its docstring is refreshed rather than left citing a fixed premise. Its remaining job is narrower and still real: refusing tunnel callers that *did* present a key, since a leaked key on a per-call billed endpoint needs only one request per cache TTL to keep running. It would have refused the unkeyed caller either way — the two are the same defence at different depths.

## Tests

- Existing gate suite now states `keyConfigured` explicitly; all six original assertions unchanged in meaning.
- New suite for the unkeyed shapes: tunnel denied, tunnel+loopback-socket (CF sidecar) denied, LAN denied, **bare loopback still granted** — the docker/doctor regression case.
- Property check: no transport shape gains disclosure by removing the key, i.e. `keyConfigured` can only ever narrow.
- `health-probe-e2e.mjs` flips from asserting the gap to asserting it's closed (real `startProxy()`, unkeyed, `cf-ray` set → single-key body), and adds a case proving an unkeyed proxy still serves its own healthcheck in full.

`health-response.mjs` 95 → 101, `health-probe-e2e.mjs` 29 → 35. Full suite **129/131** — the two failures (`template-interactive-tools`, `tool-advertise-respects-client`) reproduce identically on pristine `origin/master` on this machine and are unrelated.

Version 5.5.0 → **5.5.1** in-PR per the release gate.

Follow-up to #642. Found by the end-to-end test added in #922.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.1] - 2026-08-07
+
+- **Security: an unkeyed proxy no longer discloses OAuth internals to tunnel callers (#642 follow-up).** `shouldDiscloseHealthInternals` granted the internal view to any caller `authenticateRequest()` accepted — and `authenticateRequest()` short-circuits to `true` when no `DARIO_API_KEY` is configured. That short-circuit is deliberate and stays (the common setup is loopback-only, and requiring a key there would break `dario doctor` and every docker healthcheck), but it made `authenticated` **vacuous** on an unkeyed proxy: every caller satisfied it, the auth branch returned before `cf-ray` was ever consulted, and an unkeyed dario published through a Cloudflare tunnel handed its OAuth status, token countdown, request volume, session counts, queue snapshot and refresh-failure count to anyone who asked. #642 closed the spoofable-header direction of this gate; this was the same fail-open re-entering through a side door.
+
+  The gate now requires `authenticated && keyConfigured`, so the auth branch can only be taken by a caller that actually presented the operator's secret. Unkeyed proxies fall through to the transport rules, where **bare loopback is still trusted** — docker `HEALTHCHECK`, `dario doctor` and the TUI are unchanged — and the tunnel is not. `keyConfigured` is a required field rather than an optional with a default, so every call site has to state it.
+
+  Behaviour change worth knowing before upgrading: if you monitor `/health` **through a tunnel on a proxy with no `DARIO_API_KEY` set**, you will now get the liveness verdict only (`{"status":"ok"}`) instead of the full body. Set `DARIO_API_KEY` and send it to keep the detail, or query from loopback. The HTTP status (200/503) is unchanged either way, so status-code uptime checks are unaffected.
+
+  Found by the end-to-end test added in 5.5.0 (`test/health-probe-e2e.mjs`), which counted the disclosure while pinning the serving probe's own gate.
+
 ## [5.5.0] - 2026-08-07
 
 - **`/health` can now prove dario actually serves, not just that it looks configured (#905).** Both existing health surfaces are structural: `/livez` says the HTTP server is accepting connections, `/health` says credentials and the pool look serviceable. Neither can catch the shape that produced ~14h of downtime in #905 — dario green from the outside while every real request failed — which is why that reporter's remediation was an external watchdog sending real inference calls. That check is now in-house.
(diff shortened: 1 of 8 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped security fix. Approving.

**What I checked**
- `gh pr diff` for all 8 changed files (logic in `src/health-response.ts` + `src/proxy.ts`, tests in `test/health-*.mjs`, docs/changelog/version bump).
- CI check-rollup: all checks green (`build` on Node 18/20/22, `test`, `CodeQL`, `compat`, `wire-drift`, `docker-cap-drop-smoke`, `actionlint`, `validate-package-json`).

**Correctness**

The core fix in `src/health-response.ts`:
```ts
-  if (opts.authenticated) return true;
+  if (opts.authenticated && opts.keyConfigured) return true;
   if (opts.viaCfRay) return false;
   return opts.loopback;
```
correctly closes the described gap: `authenticateRequest()` returns `true` for every caller when no `DARIO_API_KEY` is configured, so the old `authenticated` check alone was vacuous and returned before `cf-ray` was ever consulted. Requiring `keyConfigured` too means the auth branch is only reachable by a caller that actually presented a configured secret; unkeyed proxies now correctly fall through to the transport rules (tunnel → denied, bare loopback → still granted).

The call site in `src/proxy.ts`:
```ts
+        // Without this, `authenticated` is vacuously true on an unkeyed proxy
+        // and the tunnel check below is never reached — see the gate's docs.
+        keyConfigured: apiKeyBuf !== null,
```
correctly derives `keyConfigured` from whether the key buffer exists, matching `authenticateRequest`'s own short-circuit condition. Making the field required (not optional/defaulted) means the compiler forces every call site to state it — verified there's only the one call site, and it's updated.

`shouldRunServingProbe` is untouched functionally (docstring-only change), and correctly still refuses tunnel callers unconditionally regardless of the `keyConfigured` fix — the PR body's claim that "it would have refused the unkeyed caller either way" checks out from the existing (unchanged) logic.

**Test coverage** is solid: new cases cover the unkeyed+tunnel gap directly, unkeyed+loopback (CF sidecar) staying denied, unkeyed+bare-loopback staying granted (the docker/doctor regression case), and a property check that `keyConfigured` can only narrow disclosure, never widen it. The e2e test flips from asserting the gap exists to asserting it's closed.

**What's good:** thorough docstring updates explaining the vacuous-auth trap and why the fix routes through transport rules rather than just demanding a key; changelog and usage-doc updates include a clear "who sees what" table and upgrade note for the behavior change.

Issue #636, opened 2026-07-02
admin inconsistence


matteo-rama:
hi,

doing some test with v4.8.116 and i am seeing a bit of inconsistence

starting with admin mode i got
```
2026-07-02T10:04:42.991503393+02:00 [dario] model catalog fetch failed: Not authenticated. Run `dario login` first. — serving baked list
2026-07-02T10:15:12.419699785+02:00 [dario] admin-audit: login_start alias=fbk-arch ok=true status=200 from=10.111.3.7
2026-07-02T10:17:49.331730278+02:00 [dario] admin: pool hot-reloaded — 1 account
2026-07-02T10:17:49.331776678+02:00 [dario] admin-audit: login_complete alias=fbk-arch ok=true status=200 from=10.64.183.167
```

row "model catalog fetch failed: Not authenticated. Run `dario login` first. — serving baked list" shouldn't be there? it's confusing imho

running the cli i get

```
┌ dario v4.8.116 ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── http://127.0.0.1:3456 ┐
 Status     Config     Analytics     Hits     Accounts     Backends                                                                                                                                                                                    
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 Proxy
  Status:               unreachable — is `dario proxy` running?       
```

but why proxy is not running? it's actually running, maybe because i didn't do login? but i don't want to do login :)

/admin/accounts -> ok, output seems fine

/status -> no 
```
{
  "authenticated": false,
  "status": "none"
}
```

health -> no 
```
{
  "status": "degraded",
  "oauth": "none",
  "requests": 0
}
```
/accounts -> yes

does it make sense to you this kind of report? some api i think aren't easy to understand or consistent imho, health isn't degraded for me at this moment

askalf:
Thanks — this was a real gap, and your report nailed all four symptoms of it. The legacy surfaces (`/status`, `/health`, the TUI probe, and the model-catalog fetch) all read the single-account `credentials.json`, which a login-less admin/pool setup deliberately doesn't have — so the pool served your requests while everything else reported the proxy as dead.

Fixed in **[v4.8.117](https://github.com/askalf/dario/releases/tag/v4.8.117)** (#637), on npm now:

[...]

PR #926, opened 2026-08-08
chore(ci): dogfood the serving probe in our own health watcher

askalf:
## The gap this closes

We shipped `/health?probe=1` and `queue.stalledForMs` in 5.5.0 to detect a proxy that looks healthy but cannot serve. **Our own production watcher was still alerting on `oauth != healthy` alone** — a purely structural read.

Concretely, `cc-oauth-health.yml` would have sat green through:

- all ~14h of **dario#905** (`/health` returning `{"status":"ok"}` in 2ms while every request 504'd)
- an **accountUuid-drifted account** — token unexpired, refresh fine, upstream 401ing everything

That's the definition of not dogfooding: we built the instrument and left our own monitor on the old signal.

## Three axes

| axis | source | catches |
|---|---|---|
| `oauth != healthy` | unchanged | credential death, the 2026-06-23 split |
| `probe.ok == false` | `/health?probe=1` | drift, upstream rejection, network — the #905 *shape* |
| `queue.stalledForMs > 5m` | queue snapshot | the #905 *wedge* |

Axes 2 and 3 are independent **on purpose**. The probe never takes a concurrency slot — one that queued behind real traffic would page during a legitimate burst — so it structurally cannot see a wedge. Axis 3 covers that. The end-to-end simulation confirms it: the wedge scenario fires while the probe still reports `served`.

The alert now names the failing axis in the title and selects a matching runbook, so `not serving (auth-rejected)` sends you to account drift rather than to the OAuth credential runbook.

## Cost and access

One `max_tokens:1` haiku request per run. The `docker exec` loopback path is exactly what dario requires before it will spend a token, so this monitor — and only this monitor — gets the serving verdict; dario's TTL cache means nothing is billed twice.

## Why the logic left the YAML

**A watcher's failure mode is silence.** If the extraction quietly stops matching — a field renamed, a body shape changed — the workflow reports "all axes healthy" forever, and the first anyone hears about it is the next outage nobody got paged for. That is the same class of bug as #905 itself.

Logic that can only be exercised by firing the real workflow against a real broken proxy never gets exercised. So parsing + verdict now live in `scripts/health-verdict.sh` (dependency-free POSIX sh) and `test/health-verdict.mjs` drives it directly.

**32 assertions**, including the ones that matter most:

- **Throttles must NOT page.** 429/529 stay `ok: true`; if this ever alerted, the fleet would get restart-looped during a rate-limit window — the precise failure `/livez` exists to avoid.
- **Precedence.** A dead credential also fails the probe; the alert must name the root cause, not the symptom.
- **Version tolerance.** Against a pre-5.5.0 container the new fields are absent — logged as "not measured", never as a failure. It starts working on its own once the box updates.
- **Scope-anchored matching**, so a decoy `"ok":false` elsewhere in the body cannot flip the verdict.
- **A stripped public body reads as `unreachable`, not `healthy`** — if the watcher ever loses its loopback vantage point it must alert, not go quiet. That's the 2026-07-02 blindness that #636 caused, re-pinned as a test.

Also verified by simulating the workflow's actual shell block over six real body shapes (green 5.5.2, pre-5.5.0, container down, wedge, drift, throttled) — all six produced the correct verdict and the correct file/stay-quiet decision.

## No version bump

`scripts/` and `.github/` are outside the published `files`, so nothing ships. CHANGELOG entry under `[Unreleased]`.

Full suite 130/132 — the two failures (`template-interactive-tools`, `tool-advertise-respects-client`) reproduce identically on pristine `origin/master` on this machine and are unrelated.


sprayberry-redline (changes requested):

**Verdict:** 1 blocking CI finding (new high-severity CodeQL alert introduced by this PR) — everything else in the diff reads well.

## Findings

### 1. [Blocking] New high-severity CodeQL alert on `test/health-verdict.mjs:25`
CI's CodeQL check on this PR reports:
> "Clear-text logging of sensitive information — This logs sensitive data returned by an access to oauth as clear text."

Line in question (`test/health-verdict.mjs:25`):
```js
const check = (name, cond, detail) => {
  if (cond) { console.log(`  OK ${name}`); pass++; }
  else { console.log(`  FAIL ${name}${detail ? ' :: ' + detail : ''}`); fail++; }
};
```
This is a *new* alert (`gh api .../check-runs` shows `CodeQL` → `conclusion: failure`, "1 new alert including 1 high severity security vulnerability", `annotations_count: 1`), triggered because several call sites pass `v.oauth` (derived from parsing the health body's `oauth` field) as the `detail` argument, e.g. `test/health-verdict.mjs:336`:
```js
check('oauth healthy', v.oauth === 'healthy', v.oauth);
```
so on a failing assertion the value flows into `console.log`. In practice `oauth` here is only ever a status enum (`healthy`/`broken`/`unreachable`), not a raw credential, so this looks like a CodeQL false-positive on the field name — but per the fleet's CI-authoritative policy, a failing/blocking check the PR introduces is a real, citable finding regardless. This needs one of: dismissing the alert in code scanning with a documented false-positive rationale, or tweaking the log line to not echo `v.oauth` directly, before this is mergeable.

```suggestion
// Option A: dismiss the CodeQL alert with a "false positive" rationale
// (oauth here is a status enum, not a credential) via the Security tab.
// Option B, if you'd rather silence it in code:
const check = (name, cond, detail) => {
  if (cond) { console.log(`  OK ${name}`); pass++; }
  else { console.log(`  FAIL ${name}`); fail++; } // drop `detail` from the log line
};
```

## What's good
- The oauth/probe/queue precedence logic in `scripts/health-verdict.sh` matches its own test coverage exactly (`OAuth wins over probe and queue`, `probe wins over queue`), and the throttle/version-tolerance cases are well covered.
- Extracting the parsing out of the YAML `run:` block into a testable POSIX-sh script is a solid simplification — the prior logic was untestable in place.
- `queue.stalledForMs` threshold is `-gt`, matching the "strictly greater" test case, and the loopback-only probe URL is a literal (no injection risk from `$body`/`$URL`).
- All other CI checks (`build`, `test`, `docker-cap-drop-smoke`, `actionlint`, `validate-package-json`) are green on this head SHA.

## What I checked
- Full diff (`gh pr diff 926`), both new files (`scripts/health-verdict.sh`, `test/health-verdict.mjs`) in full, and the modified workflow YAML.
- PR's CI check-rollup via `gh api .../check-runs` (did not run tests locally).
- PR author (`askalf`) is distinct from this reviewer bot's own identity (`sprayberry-reviewer`) — no self-review conflict.

askalf:
Good catch on the blocking check — resolved as **Option A (dismiss with rationale)**, alert #44, reason `false positive`. CodeQL rerun triggered, since a dismissal doesn't auto-green the failed check-run.

Taking Option B instead would have made the code worse in two ways, so I want the reasoning on the record rather than just the outcome:

**It's a name heuristic, not a data-flow finding.** `js/clear-text-logging` fires because a local is called `oauth`. The value it holds is `/health`'s oauth **status enum** — `src/health-response.ts:101` types it `'healthy' | 'broken' | 'none'`, plus `'unreachable'` synthesised by the verdict script when the body is empty. It is never the token. That's deliberate: #642 stripped `lastRefreshError` off `/health` precisely so no raw string can ride that field, and the countdown is a formatted duration.

Verified rather than assumed, in the test file itself:

```
$ grep -o '"oauth":"[^"]*"' test/health-verdict.mjs | sort -u
"oauth":"broken"
"oauth":"healthy"
```

Both are string literals defined in the same file. There is no credential anywhere in it, so there is no secret to leak in any execution.

**Dropping `detail` from the log would cost real diagnostic value.** It only prints on failure, and it's what turns `FAIL axis names OAuth` into `FAIL axis names OAuth :: not serving (auth-rejected)`. For a watcher whose failure mode is silence, a test that says *what* it got when it breaks is the point.

**Renaming to dodge the heuristic would be the contortion the repo's own FP posture rejects.** `oauth` is the accurate name — it's literally the health body's `oauth` field. This is the same disposition as the existing dismissals (`js/insufficient-password-hash` on high-entropy bearer tokens, `js/superfluous-trailing-arguments` on the generic map helper): honest dismissal with an audit trail beats bending correct code around a pattern-match.

Everything else in your review I agree with, and thanks for confirming `$body` can't reach the probe URL — that was the injection surface I was most careful about.

askalf:
Pushed `fb7c73f` — which also clears the stale `CHANGES_REQUESTED` (the CodeQL finding it blocked on was dismissed as a false positive and the check is green; a new commit is what retriggers review).

The new commit is not a formality. While re-reading the alert body I'd added, I noticed this repo is **public** and these alerts file **public issues**, but the body the watcher fetches is the loopback-only *internal* variant — token countdown, lifetime request volume, session/sticky counts, refresh-failure streak — and it was being embedded verbatim.

Nothing leaked: every alert filed so far (#839, #892, #894, #895, #897) caught a **down** container and printed `<unreachable>`. The path was never exercised with a reachable one. But `oauth: broken` with the container **up** is a normal outcome — #241 exists to detect exactly that — and it would have published all of it.

And `?probe=1` widened it: `probe.detail` carries `errText()` from a network failure, which can name internal hosts. So that part is a defect **this branch introduced**, not just a pre-existing one.

The axis table already carries everything the runbook acts on, so the raw dump was redundant as well as leaky. Verified by rendering the alert against a realistic internal body and asserting none of `requests`, `stickyBindings`, the internal IP, `refreshFailures`, the countdown or `ECONNREFUSED` survive, while the table does.

askalf:
Closing to recreate as a fresh PR — sprayberry-reviewer's dedup won't refile a second review on an already-reviewed PR, and the CHANGES_REQUESTED that blocked this one is resolved (CodeQL alert #44 dismissed as false positive, rerun green) but stuck on a stale commit since approvals-only dismiss-on-push doesn't clear a change-request. Same branch, same head, no code difference — recreated for a clean review pass. See the successor PR for the approval.

PR #929, opened 2026-08-08
chore(ci): dogfood the serving probe in our own health watcher

askalf:
_Recreated from #926 — no code difference, same branch/head. sprayberry-reviewer's dedup won't file a second review on an already-reviewed PR, so #926 was stuck on a stale `CHANGES_REQUESTED` (pinned to an earlier commit) with no path to a fresh review. Closing and reopening gets a clean pass._

## The gap this closes

We shipped `/health?probe=1` and `queue.stalledForMs` in 5.5.0 to detect a proxy that looks healthy but cannot serve. **Our own production watcher was still alerting on `oauth != healthy` alone** — a purely structural read.

Concretely, `cc-oauth-health.yml` would have sat green through:

- all ~14h of **dario#905** (`/health` returning `{"status":"ok"}` in 2ms while every request 504'd)
- an **accountUuid-drifted account** — token unexpired, refresh fine, upstream 401ing everything

That's the definition of not dogfooding: we built the instrument and left our own monitor on the old signal.

## Three axes

| axis | source | catches |
|---|---|---|
| `oauth != healthy` | unchanged | credential death, the 2026-06-23 split |
| `probe.ok == false` | `/health?probe=1` | drift, upstream rejection, network — the #905 *shape* |
| `queue.stalledForMs > 5m` | queue snapshot | the #905 *wedge* |

Axes 2 and 3 are independent **on purpose**. The probe never takes a concurrency slot — one that queued behind real traffic would page during a legitimate burst — so it structurally cannot see a wedge. Axis 3 covers that. The end-to-end simulation confirms it: the wedge scenario fires while the probe still reports `served`.

The alert now names the failing axis in the title and selects a matching runbook, so `not serving (auth-rejected)` sends you to account drift rather than to the OAuth credential runbook.

## Cost and access

One `max_tokens:1` haiku request per run. The `docker exec` loopback path is exactly what dario requires before it will spend a token, so this monitor — and only this monitor — gets the serving verdict; dario's TTL cache means nothing is billed twice.

## Why the logic left the YAML

**A watcher's failure mode is silence.** If the extraction quietly stops matching — a field renamed, a body shape changed — the workflow reports "all axes healthy" forever, and the first anyone hears about it is the next outage nobody got paged for. That is the same class of bug as #905 itself.

Logic that can only be exercised by firing the real workflow against a real broken proxy never gets exercised. So parsing + verdict now live in `scripts/health-verdict.sh` (dependency-free POSIX sh) and `test/health-verdict.mjs` drives it directly.

**32 assertions**, including the ones that matter most:

- **Throttles must NOT page.** 429/529 stay `ok: true`; if this ever alerted, the fleet would get restart-looped during a rate-limit window — the precise failure `/livez` exists to avoid.
- **Precedence.** A dead credential also fails the probe; the alert must name the root cause, not the symptom.
- **Version tolerance.** Against a pre-5.5.0 container the new fields are absent — logged as "not measured", never as a failure. It starts working on its own once the box updates.
- **Scope-anchored matching**, so a decoy `"ok":false` elsewhere in the body cannot flip the verdict.
- **A stripped public body reads as `unreachable`, not `healthy`** — if the watcher ever loses its loopback vantage point it must alert, not go quiet. That's the 2026-07-02 blindness that #636 caused, re-pinned as a test.

Also verified by simulating the workflow's actual shell block over six real body shapes (green 5.5.2, pre-5.5.0, container down, wedge, drift, throttled) — all six produced the correct verdict and the correct file/stay-quiet decision.

## Disclosure fix folded in (found during this PR's own review cycle)

dario is a **public** repo and this workflow files **public issues**, but the body it fetches via `docker exec` is the loopback-only *internal* `/health` variant — token countdown, lifetime request volume, session/sticky counts, refresh-failure streak. The alert was embedding it verbatim.

Nothing has leaked: every alert filed so far (#839, #892, #894, #895, #897) caught a **down** container and printed `<unreachable>`. The path was never exercised with a reachable one — and `oauth: broken` while the container is **up** is a normal outcome (#241 detects exactly that), which would have published the lot.

`?probe=1` widened it further: `probe.detail` carries `errText()` from a network failure, which can name internal hosts. So part of this is a defect introduced by this branch, not only pre-existing.

Fixed: the raw dump is gone. The axis table already carries every value the runbook acts on, so it was redundant as well as leaky; the issue now reports reachability only and points at the box for the full body. Verified by rendering the alert against a realistic internal body (container up, oauth broken, network-error probe) and asserting none of `requests`, `stickyBindings`, the internal IP, `refreshFailures`, the countdown, or `ECONNREFUSED` survive — while the actionable table does.

Same disclosure class as #642 → #925 earlier this session: internals reaching a reader that shouldn't have them.

## No version bump

`scripts/` and `.github/` are outside the published `files`, so nothing ships. CHANGELOG entry under `[Unreleased]`.

Full suite 130/132 on the original commit — the two failures (`template-interactive-tools`, `tool-advertise-respects-client`) reproduce identically on pristine `origin/master` on this machine and are unrelated.


sprayberry-redline (approved):

## What I checked
- Full diff (`.github/workflows/cc-oauth-health.yml`, `scripts/health-verdict.sh`, `test/health-verdict.mjs`, `CHANGELOG.md`).
- CI check-rollup: all checks green on the latest commit (`build`, `test`, `docker-cap-drop-smoke`, `CodeQL`, `actionlint`, `validate-package-json`), run twice (once pre- and once post-fixup commit).
- Traced `scripts/health-verdict.sh`'s field extraction (oauth / probe.ok / probe.reason / queue.stalledForMs) and axis-precedence logic against `test/health-verdict.mjs`'s 32 assertions, including the object-scoped regex (`"probe":{[^}]*}`, `"queue":{[^}]*}`) against the decoy-key test, the throttle-must-not-page case (429/529 stay `ok:true`), and version tolerance (pre-5.5.0 body → fields read as "not measured", never as a failure).
- The disclosure fix: raw `/health` body is no longer embedded in the (public) issue body — replaced with a reachability boolean and an axis table, verified against the PR's own stated test of a realistic internal body.

Nothing in the diff exhibits a correctness bug I could pin to specific lines and reproduce — the fixed-position 5-line stdout contract between `health-verdict.sh` and the workflow's `sed -n 'Np'` reads is internally consistent, and the `case` pattern matching (`OAuth*`, `"not serving"*`, `"queue wedged"*`) correctly matches the axis strings the script emits.

## What's good
- Extracting the parse/verdict logic out of the YAML `run:` block into a testable POSIX-sh script is a solid structural improvement — this class of watcher genuinely fails silently otherwise.
- The disclosure fix (dropping the raw internal `/health` body from a public issue) is a real security improvement folded into the same PR, with a specific test verifying the leaking fields are stripped.
- Good edge-case coverage: throttle-must-not-page, exact-threshold boundary (300000 vs 300001), precedence between axes, and the "stripped public body must still alert" regression test tying back to a prior incident (#636).

## Not verified (scope note)
This workflow only runs on `schedule`/self-hosted runner, so CI doesn't exercise it end-to-end — I relied on the PR author's stated manual simulation (six body shapes through the actual shell block) plus the unit tests, not a live run. Flagging for visibility, not as a blocker.

PR #932, opened 2026-08-08
fix(ci): sparse-checkout must name a directory, not health-verdict.sh

askalf:
The "dario serving health alert (self-hosted)" watcher has been failing at
checkout on every 5-minute run since #929 merged. Run #31239237095:

    fatal: 'scripts/health-verdict.sh' is not a directory; to treat it as a
    directory anyway, rerun with --skip-checks

## Root cause

#929 added `sparse-checkout: scripts/health-verdict.sh` plus
`sparse-checkout-cone-mode: false` — which reads correct, since non-cone mode
does accept file patterns. But the pin is `askalf/checkout-with-retry`, and
that wrapper declares only `ref, repository, token, fetch-depth, path,
sparse-checkout, persist-credentials`. It has no `sparse-checkout-cone-mode`
input, so the value was dropped with a warning that is easy to miss in the log:

    ##[warning]Unexpected input(s) 'sparse-checkout-cone-mode', valid inputs
    are ['ref', 'repository', 'token', 'fetch-depth', 'path',
    'sparse-checkout', 'persist-credentials']

The inner `actions/checkout` therefore ran at its default
`sparse-checkout-cone-mode: true`, where every pattern must be a directory, and
the file path was fatal.

## Fix

Check out the `scripts` directory instead. Nothing else in the repo uses
sparse-checkout, so this is the only occurrence.

The alternative — teaching `checkout-with-retry` to forward
`sparse-checkout-cone-mode` — is a cross-repo change that would still leave
this workflow broken until a new wrapper release is cut and re-pinned here.
Worth doing separately if another caller ever wants file-level patterns; it is
not needed to make the watcher green. The extra files under `scripts/` cost
effectively nothing at `fetch-depth: 1`.

The comment block now records why the pattern must stay a directory, so the
next person does not "helpfully" narrow it back to the single file.

## Test plan

- Not unit-testable — the failure only reproduces on the self-hosted
  `dario-drift` runner during checkout.
- Verification is the next scheduled run (cron `*/5`): it should get past
  checkout, and `$GITHUB_WORKSPACE/scripts/health-verdict.sh` must resolve for
  the probe step. Can be forced early with `workflow_dispatch`.
- Watch for the recovery path too — if the watcher opened a `dario-oauth`
  issue while it was down, a green run should close it.

Source ticket: 00MSJVC71U9CA9ECE3822E8284
Intervention: 00MSJVC1MKCB6055032F55BEC4


sprayberry-redline (approved):

**Verdict:** No blocking issues — approving. Clean, minimal CI fix.

**What I checked**
- `gh pr diff 932` (7 additions / 2 deletions, single file: `.github/workflows/cc-oauth-health.yml`)
- CI check-rollup: all checks green (`analyze`, `actionlint`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `CodeQL`, `validate-package-json`); `auto-merge` is `SKIPPED`, not a failure.
- Author is `askalf`, distinct from the reviewer bot identity — no self-approval collision.

**Findings**
None blocking. The change:
```
-          sparse-checkout: scripts/health-verdict.sh
-          sparse-checkout-cone-mode: false
+          sparse-checkout: scripts
```
correctly addresses the stated root cause: `askalf/checkout-with-retry` doesn't forward `sparse-checkout-cone-mode` to the inner `actions/checkout`, so it always runs in cone mode, where sparse-checkout patterns must be directories — a file pattern like `scripts/health-verdict.sh` fails checkout (`fatal: '...' is not a directory`). Pointing the pattern at the `scripts` directory instead is correct for cone mode and matches the linked failing run. The added comment records the constraint so it isn't "helpfully" narrowed back to a file path later. Downstream usage of `scripts/health-verdict.sh` in the same job is unaffected — the file still exists on disk after checking out the containing directory.

**What's good**
- Root-cause explanation in the PR body is precise and matches the diff.
- Comment added inline prevents regression by a future contributor.
- Scope is minimal — no unrelated changes.

PR #940, opened 2026-08-10
Tighten dario#lock-step: single-source variant families, bake gate, doctor coverage (v5.5.6)

askalf:
Tightens dario#lock-step — the mechanism keeping dario's per-model system prompts (fable / opus-5 / sonnet-5) byte-aligned with CC's — on three axes.

## 1. Single source of truth for variant families

The families lived in two hand-maintained lists: `VARIANT_MODELS` in `scripts/capture-and-bake.mjs` and the hardcoded selection arms in `systemPromptForModel`. That is the exact divergence class that shipped `tool_names` ≠ `tools` twice (#859's second-occurrence note). A family added to one side and not the other either bakes a variant no arm can serve, or adds an arm that silently falls back to the base forever.

Both now derive from `VARIANT_FAMILIES` in `src/live-fingerprint.ts` (next to `TEMPLATE_BASE_MODEL`, which exists there for the same reason): key, capture model, and matcher in one table, matcher-precedence order preserved (fable ahead of the bounded `-5` arms). `test/template-invariants.mjs` asserts the shipped bundle against the table — every family carried and distinct from the base, no orphan keys — so a divergence can't ship quietly a third time.

## 2. The bake refuses silent variant loss

A variant capture that failed with no previous variant to keep used to log a warning and bake anyway. Two consequences:

- baked: the bundle serves that family the shared base prompt with no other symptom (requests still 200);
- in `--check`: the absence reads as "N → 0 chars" variant drift, handing the watch workflow a rebake PR that *strips* the variant.

That outcome is now exit 1 — infra failure, the #635 older-CC-guard philosophy — with `--allow-missing-variant` as the deliberate bypass. Captured-and-matches-base still passes: that's a measured result, and a genuine variant retirement stays reportable drift. Kept-previous variants log a note that the kept text may lag the freshly captured base.

## 3. Degraded lock-step is visible at runtime

`describeTemplate` (doctor Template row + proxy startup line) now appends variant coverage (`variants: fable+opus-5+sonnet-5` / `variants: none`), and `dario doctor` gains a "Prompt variants" row: ok when the loaded template carries every family, warn naming the missing ones. This is the state the live-cache shadow already reached once (verified live back then: `CC_SYSTEM_PROMPT_FABLE === CC_SYSTEM_PROMPT` with a fresh cache present) — it is now visible instead of silent.

## Trade-offs

- **A red watch is now possible where it was green-with-warning.** If a NEW family is added to the table and the drift runner's capture for it fails with nothing previous to keep, `cc-drift-template-watch` goes red (exit 1) until fixed or bypassed. Deliberate: for the three current families the gate cannot fire (all have previous variants to keep), so this only bites on genuinely new families or a wiped bundle.
- **Retiring a variant becomes a reviewed table edit.** If CC ever unifies a family's prompt into the base, the invariants fail on the rebake PR until the family is removed from `VARIANT_FAMILIES`, and doctor warns meanwhile. Chosen over silent disappearance.
- **`describeTemplate` output format changed** (variants suffix). Doctor detail and the startup log line are human-facing; the drift-detection tests cover the new shape.

Version bump to 5.5.6 included (release gate). Within-minor, so autodeploy applies it unattended.

## Tests

- `test/template-invariants.mjs`: new bundled-artifact section (family coverage, base-distinctness, matcher self-consistency, no orphans).
- `test/per-model-system-prompt.mjs`: table-driven routing checks + `missingVariantFamilies(CC_TEMPLATE)` empty on the shipped bundle.
- `test/drift-detection.mjs`: `describeTemplate` variant-coverage lines, including the legacy `system_prompt_fable` fold.
- Full suite: 130/132 files pass; the 2 failures (`template-interactive-tools`, `tool-advertise-respects-client`) reproduce identically on pristine master in a clean worktree and pass standalone — pre-existing suite-level flake, not introduced here (details in comments if needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.6] - 2026-08-09
+
+- **Lock-step tightened: per-model prompt variants get a single source of truth, a bake gate, and doctor coverage (dario#lock-step).** Three loosenesses in the mechanism that keeps dario's per-model system prompts (fable / opus-5 / sonnet-5) byte-aligned with CC's:
+  - **One table drives everything.** The variant families lived in two hand-maintained lists — `VARIANT_MODELS` in `scripts/capture-and-bake.mjs` and the hardcoded selection arms in `systemPromptForModel` — the exact divergence class that shipped `tool_names` ≠ `tools` twice. Both now derive from `VARIANT_FAMILIES` (`src/live-fingerprint.ts`, next to `TEMPLATE_BASE_MODEL` for the same reason): key, capture model, and matcher in one place, matcher-precedence order preserved (fable before the bounded `-5` arms). `test/template-invariants.mjs` asserts the shipped bundle against the table — every family carried and distinct from the base, no orphan variant keys — so a divergence can no longer ship quietly.
+  - **The bake refuses to ship silent variant loss.** A variant capture that failed with no previous variant to keep used to log a warning and bake anyway — the bundle would then serve that family the shared base prompt with no other symptom, and in `--check` the absence masqueraded as "N → 0 chars" variant drift, handing the watch workflow a rebake PR that strips the variant. That outcome is now exit 1 (infra failure, same philosophy as the older-CC guard from #635), with `--allow-missing-variant` as the deliberate bypass. Captured-and-matches-base still passes — that's a measured result, and a genuine variant retirement remains reportable drift. Kept-previous variants now log a note that the kept text may lag the freshly captured base.
+  - **Degraded lock-step is visible at runtime.** `describeTemplate` (doctor's Template row + the proxy startup line) now appends variant coverage (`variants: fable+opus-5+sonnet-5` / `variants: none`), and `dario doctor` gains a "Prompt variants" row: ok when every family in the table is carried by the loaded template, warn naming the missing families — the state a variant-less bundle or a pre-variants live cache shadowing the bundle used to reach invisibly (requests still 200, wrong prompt).
+
 ## [5.5.5] - 2026-08-08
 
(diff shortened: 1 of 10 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean refactor + gate addition, approving.

## What I checked
- `gh pr diff` in full (451 lines across 10 files).
- CI check-rollup: all 13 checks green (CI build×3, test, compat, wire-drift, CodeQL, actionlint, docker-cap-drop-smoke, validate-package-json), none pending/failing.
- Traced the new lock-step gate in `scripts/capture-and-bake.mjs`: for each family in `VARIANT_MODELS`, every code path through the capture loop (fresh capture / matches-base / scrub-residue-found / capture-returns-falsy / capture throws) sets a `variantOutcomes[key]` entry via either the inline assignment or the `keepPrevious()` helper — so `missingVariants`/`keptVariants` at `scripts/capture-and-bake.mjs:334`/`:346` are computed from a fully-populated map, no path leaves a key unset that would silently skip the new `process.exit(1)` gate.
- Confirmed `systemPromptForModel` (`src/cc-template.ts:162-166`) and the bake's `VARIANT_MODELS` (`scripts/capture-and-bake.mjs:92`) both now derive from the single `VARIANT_FAMILIES` table in `src/live-fingerprint.ts`, closing the two-hand-maintained-list divergence the PR describes; matcher precedence (fable before the bounded `-5` arms) is preserved.
- `missingVariantFamilies` (`src/live-fingerprint.ts:238-241`) is used consistently by both `doctor.ts`'s new "Prompt variants" check and the new test in `test/per-model-system-prompt.mjs`.
- New tests (`template-invariants.mjs`, `per-model-system-prompt.mjs`, `drift-detection.mjs`) exercise the new table-driven selection, the bundle-vs-table invariant, and the `describeTemplate` variant-coverage suffix — reasonable coverage for the behavior changed.

## What's good
- The single-source-of-truth table (`VARIANT_FAMILIES`) is a real simplification that removes a documented recurring divergence class (`tool_names` ≠ `tools`, twice).
- The new gate's failure-vs-bypass semantics (`--allow-missing-variant`) mirror the existing `--allow-older-cc` pattern already in the file, keeping the CLI surface consistent.
- Version bump (5.5.5 → 5.5.6) is applied consistently across `package.json`/`package-lock.json`/`CHANGELOG.md`.

No correctness, security, reuse, or efficiency issues found in the changed lines. Nothing here blocks merge.

PR #941, opened 2026-08-10
perf(proxy): skip orchestration scrub on tag-free message blocks (v5.5.7)

askalf:
## What

`sanitizeContent` (the orchestration-tag scrub in `sanitizeMessages`) applied all ~28 tag regexes to every text block on every request — including byte-faithful CC traffic, since `sanitizeMessages` runs before `buildCCRequest`'s genuine-CC branch. Every one of those patterns is anchored on a literal `<` (`<tag…>…</tag>` and `<tag…/>`), so a block with no `<` — the common case: prose user turns, command/tool output — cannot match any of them.

Guarded the loop with a single `indexOf('<')` check. When there's no `<`, the ~28 regex passes are skipped and the block goes straight to the existing trailing whitespace-collapse/trim.

## Why it's safe

The loop is skipped **only** when it provably cannot change the string. The whitespace normalization (`/\n{3,}/g` → `\n\n`, `.trim()`) still runs for every block, so output is byte-identical for all inputs — tag-free, tag-bearing, and code with a bare `<` comparison alike.

## Measured

Benchmarked the current vs guarded `sanitizeContent` over a realistic 8-block message mix (long prose turn, a question, a large tag-free tool_result, one `<system-reminder>`-bearing block, repeated):

```
parity: IDENTICAL across 8 inputs
current: 544.32µs / 8-block request
guarded:  93.01µs / 8-block request
speedup: 5.85x  (83% less CPU on this mix)
```

This is on the fleet-dominant path — a real CC conversation with a long history runs `sanitizeMessages` over every turn before forwarding.

## Tests

`test/sanitize-messages.mjs` gains:
- a **parity check** that re-derives the pre-guard behavior (loop always runs) and asserts byte-for-byte equality against the guarded path via the real `sanitizeMessages` entry point, across tag-free AND tag-bearing inputs (including empty string, whitespace, and generics like `Array<Item>`);
- a guard that tag-free code containing a bare `<` (`Map<string, number>`, `if (a < b)`) is preserved verbatim.

`38 pass, 0 fail` locally. Full suite: 130/132 files pass; the 2 failures (`template-interactive-tools`, `tool-advertise-respects-client`) reproduce identically on pristine master in a clean worktree and are a pre-existing Windows-only suite-parallelism flake — unrelated to this change, which touches neither.

Version bump to 5.5.7 (release gate). Patch → within-minor, so box autodeploy applies it unattended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.7] - 2026-08-09
+
+- **Perf: skip the orchestration-tag scrub on message blocks that can't contain a tag.** `sanitizeMessages` runs on every request (including byte-faithful CC traffic), and `sanitizeContent` applied all ~28 orchestration-tag regexes to every text block unconditionally. Every one of those patterns is anchored on a literal `<`, so a block with none — the common case: prose user turns, command/tool output — cannot match any of them. `sanitizeContent` now guards the loop with a single `indexOf('<')` check and skips straight to the trailing whitespace normalization when there's no `<`. On a realistic message mix this cuts the scrub's CPU ~80% (measured 5.85× on the loop) while producing byte-identical output — the loop is skipped only when it provably cannot change the string, and the whitespace-collapse/trim still runs for every block exactly as before. `test/sanitize-messages.mjs` adds a parity check asserting the guarded path matches the unconditional loop across tag-free and tag-bearing inputs, plus a guard that tag-free code containing a bare `<` comparison is preserved verbatim.
+
 ## [5.5.6] - 2026-08-09
 
 - **Lock-step tightened: per-model prompt variants get a single source of truth, a bake gate, and doctor coverage (dario#lock-step).** Three loosenesses in the mechanism that keeps dario's per-model system prompts (fable / opus-5 / sonnet-5) byte-aligned with CC's:
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.6",
+  "version": "5.5.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "5.5.6",
+      "version": "5.5.7",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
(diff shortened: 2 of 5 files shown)


sprayberry-redline (approved):

## What I checked
- `src/proxy.ts:634-648` (`sanitizeContent`): the new fast path guards the ~28-regex orchestration-tag loop with `if (result.indexOf('<') !== -1)`.
- `buildOrchestrationPatterns` (`src/proxy.ts:619-628`): confirmed every emitted pattern is literally `<`-prefixed —
  ```
  new RegExp(`<${tag}\\b[^>]*>[\\s\\S]*?<\\/${tag}>`, 'gi'),
  new RegExp(`<${tag}\\b[^>]*\\/>`, 'gi'),
  ```
  so a body with no `<` provably cannot match any pattern; the guard cannot change output. The `pattern.lastIndex = 0` reset inside the (now-conditional) loop is unaffected — `String.replace` with a `/g` regex resets `lastIndex` on its own, so no state leaks across calls either way.
- Downstream empty-block/empty-message filters (`sanitizeMessages`, `proxy.ts:687-712`): the trailing `.replace(/\n{3,}/g,'\n\n').trim()` in `sanitizeContent` runs **unconditionally** regardless of the guard, so a whitespace-only, tag-free block collapses to `''` exactly as it did before this change — the existing empty-block/empty-message filtering is untouched and still correct.
- `test/sanitize-messages.mjs:294`: `const got = expected === '' ? '' : viaGuarded;` looked like it might weaken the empty-string assertion, but traced through: when `expected === ''`, `viaGuarded` is already computed as `''` (via the `body.messages.length ? ... : ''` guard a few lines up, since a message whose content scrubs to `''` gets filtered out of `body.messages`). So the substitution is a no-op against an already-correct value, not a hole that hides a real mismatch — dropped as a finding.
- CI check-rollup on head `ead78caa`: all 12 required checks green (CodeQL, actionlint, analyze, build ×3, docker-cap-drop-smoke, test, compat, triage, validate-package-json, wire-drift); `auto-merge` correctly skips for a non-bot author.

## What's good
- The optimization is provably behavior-preserving at the regex-construction level, not just claimed — worth the inline comment explaining why.
- New parity test re-derives the pre-guard behavior and asserts byte-identical output across tag-free, tag-bearing, and edge inputs (empty string, whitespace-collapse, bare `<` in generics/comparisons).
- Version bump + changelog entry included with the perf change.

No correctness, reuse, efficiency, or test-coverage issues rise to blocking. Approving.

PR #951, opened 2026-08-11
ci(liveness): grant actions:write and corroborate the run listing before alarming

askalf:
Two defects surfaced by #950, which fired at 17:14Z against a watcher that had cycled cleanly at 16:28Z.

## Stale read → operator ticket

The alarm read the watcher's last success from `runs?status=success&per_page=1`. That filtered listing served a 54-hour-stale record (`2026-08-09T11:04:47Z`) while the unfiltered listing carried the fresh runs; the 15:17Z liveness cycle right before it had read healthy, and both variants agree again now. One eventually-consistent response was enough to open a ticket.

Now reads the unfiltered listing and takes the newest `success` client-side, then re-reads once after 15s before escalating. Timestamps are uniform ISO-8601 Z, so `max` over the strings orders correctly. Only a *newer* re-read counts, so the correction can't mask a real outage. The healthy path never re-reads and pays no extra time.

Exercised against a stubbed reader:

| first read | re-read | outcome |
|---|---|---|
| 54h (stale) | 46m | healthy → closes the alert |
| 54h | 54h | escalates |
| 20m | — | healthy, no re-read |
| 54h (stale) | 20h | escalates, with the corrected age |

## Self-recovery never ran

The self-recovery `gh workflow run` at the heart of the "try before notifying the operator" path has never fired successfully. The job grants `contents: read` + `issues: write`; a workflow dispatch needs `actions: write`, so it 403'd every time — invisibly, because the dispatch's stderr went to `/dev/null`. Every threshold breach escalated straight to an issue.

Grants `actions: write` and keeps the stderr in the warning text.

## Note on #950 itself

The watcher was never down. #950 closes on its own at the next healthy liveness cycle — this PR is about the alarm not re-firing on the next stale read, and about the recovery path working when there is a real outage to recover from.

### DIFF
--- a/.github/workflows/cc-drift-watcher-liveness.yml
+++ b/.github/workflows/cc-drift-watcher-liveness.yml
@@ -38,6 +38,11 @@ jobs:
     permissions:
       contents: read
       issues: write
+      # Needed by the self-recovery `gh workflow run` below. Without it
+      # the dispatch 403s and every threshold breach escalates straight
+      # to an operator alert — the recovery path had never once fired
+      # before #950 surfaced it.
+      actions: write
     runs-on: ubuntu-latest
     timeout-minutes: 3
     steps:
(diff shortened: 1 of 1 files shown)


sprayberry-redline (approved):

I reviewed this PR, found a blocking defect, and then pushed the fix to this branch myself (`ci(liveness): keep an empty last_success meaning 'never succeeded'`). This approval therefore is not independent review — the same reviewer wrote the code being approved. Approving at the maintainer's direction, and recording that here rather than leaving it implicit.

The finding and the fix, so the record is in one place:

## The defect

`read_last_success()` changed from a server-filtered single record to an unfiltered page filtered client-side:

```
"repos/.../actions/workflows/cc-drift-template-watch.yml/runs?per_page=30"
  --jq '[.workflow_runs[] | select(.conclusion == "success") | .created_at] | max // ""'
```

The downstream handler was unchanged:

```sh
if [ -z "$last_success" ]; then
  echo "No successful watcher runs yet — likely a fresh repo or watcher disabled. Skipping alarm."
  exit 0
fi
```

Under the filtered query `""` could only mean *no successful run had ever existed*, which is why skipping the alarm was safe. Under the unfiltered one it also means **no success within the most recent 30 runs**. `cc-drift-template-watch.yml` runs `41 * * * *`, so that window is roughly 30 hours.

A watcher failing continuously for longer than that — the most severe outage this workflow exists to detect — produced an empty `last_success` and took the "fresh repo" `exit 0` path. The failure mode was inverted: the longer the outage ran, the more certainly the alarm was suppressed. The healthy path was unaffected, so it would have stayed invisible until the day it mattered.

## The fix

`read_last_success()` now falls back to the filtered lookup **only** when the recent page contains no success:

```sh
read_last_success() {
  local newest
  newest=$(gh api ".../runs?per_page=30" --jq '[...] | max // ""')
  if [ -z "$newest" ]; then
    newest=$(gh api ".../runs?status=success&per_page=1" --jq '.workflow_runs[0].created_at // ""')
  fi
  printf '%s' "$newest"
}
```

The recent page stays authoritative whenever it holds a success, so the #950 stale-read fix is untouched and the healthy path makes no extra API call. An empty string recovers its original meaning, so the `exit 0` branch is correct again.

Also added a comment at the `[[ "$recheck" > "$last_success" ]]` comparison noting it is lexicographic and correct only because both operands are uniform ISO-8601 Z — the same property `max` depends on above. The diff justified that for `max` but not for the comparison, and they rely on it equally.

## Unchanged assessment of the rest

The `actions: write` grant and the `dispatch_err` stderr capture are both plainly right; the comment correctly identifies that the missing permission meant the self-recovery path had never once fired, and keeping stderr is what would have surfaced that 403 far earlier. Accepting only a *newer* re-read is the right call and does what the PR body claims — a re-read can move the timestamp forward, never backward, so corroboration cannot mask a real outage.

CI is green on the fix: `analyze`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `validate-package-json` all pass.

PR #955, opened 2026-08-12
chore(template): label refresh _version -> v2.1.228 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-12T05:16:29+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.228** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.228`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.228`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.5.10` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.228. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #954, opened 2026-08-11
chore(cc-drift): v5.5.10 — maxTested → v2.1.228

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.228 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.227` → `2.1.228` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.10`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.10] - 2026-08-11` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.228 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.227). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.227; npm latest is v2.1.228. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.228 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.228 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.228. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.228" step — no local run needed.)
- ✅ **Auto-merges** the moment the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`). `master` requires no review, so no human approval gates it.
[...]

PR #958, opened 2026-08-12
ci(auto-merge): require a bot PR's version to advance past the base tip

askalf:
## The gap

`auto-merge-bot-prs.yml` armed native auto-merge on branch name alone. It had no idea what version a PR carried, so two bot PRs that independently claimed the same version were free to race.

That is not hypothetical — it happened today. #954 (cc-drift `maxTested`) and #955 (template label refresh) were drafted by different watchers, each taking the next patch off `5.5.9`, and both landed on **`5.5.10`** with different dates. Neither watcher could see the other.

Whichever merged *second* was the casualty:

- **equal to the released version** → `cc-drift-auto-release.yml`'s duplicate-tag guard fast-exits green and publishes **nothing**. The run looks healthy. The release silently did not happen — on a chore PR nobody re-reads after merge, which is exactly where it would go unnoticed.
- **lower** → master's version moves *backwards*, below an already-published tag.

Today it only came out right because #954 happened to go green first. That was timing, not a guarantee.

## The fix

A `version_gate` step, run before arming:

| head vs base | head vs tip | result |
|---|---|---|
| unchanged | — | **arm** — PR doesn't touch the version, no hazard |
| changed | `head > tip` | **arm** |
| changed | `head == tip` | **block** — someone else already claimed it |
| changed | `head < tip` | **block** — would move master backwards |
| any read fails | — | **block** |

Comparing against the PR's own base *as well as* the tip is what keeps the common Dependabot action bump unaffected: if the version is untouched, `head == base` and the gate exits early. Only a PR that actually changes the version has to clear the tip.

Comparison is `sort -V`, not string compare — `5.5.99` must not outrank `5.6.0`. Verified across the ordering cases including that one.

All three versions are read through the contents API, since this runner has no checkout.

## Design notes

**Per-PR invariant, not cross-PR ordering.** The workflow never enumerates other open PRs or tries to sequence them. Each PR must simply advance past the tip; correct ordering falls out of that. Less to go wrong, and no dependence on what else happens to be open.

**Fails closed.** If `package.json` can't be read at any of the three refs, auto-merge is not armed. Skipping one arming run costs nothing — `synchronize` re-runs the gate on the next push — while guessing risks precisely the silent no-op release this exists to prevent.

**A blocked PR is stale, not broken.** It needs the branch update it would have needed anyway, and re-arms automatically once its version clears the tip. The `::warning::` names both versions so the reason is visible in the run without digging.

## Scope

Workflow-only. No version bump: `.github/` isn't in the package `files` list, so this ships nothing.


sprayberry-redline (approved):

I wrote this change. This approval is therefore not independent review; approving at the maintainer's direction and recording that here rather than leaving it implicit.

## What it fixes

`auto-merge-bot-prs.yml` armed native auto-merge on branch name alone, with no visibility into the version a PR carried. Two bot PRs that independently claimed the same version were free to race.

That happened today: #954 (cc-drift `maxTested`) and #955 (template label refresh) were drafted by different watchers, each took the next patch off `5.5.9`, and both landed on `5.5.10` with different dates. Whichever merged second was the casualty — equal to the released version means `cc-drift-auto-release.yml`'s duplicate-tag guard fast-exits green and ships nothing (a release that silently did not happen, on a chore PR nobody re-reads), and lower means master's version moves backwards below a published tag.

It only came out right today because #954 happened to go green first. Timing, not a guarantee.

## The invariant

A PR that changes the `package.json` version must carry one strictly greater than the base branch tip. A PR that leaves the version untouched — the common Dependabot action bump — is explicitly unaffected, which is why the gate compares head against the PR's own base as well as against the tip: unchanged version means `head == base`, and it exits early.

Deliberately a **per-PR invariant rather than cross-PR ordering**. The workflow never enumerates other open PRs or tries to sequence them; each PR simply has to advance past the tip, and correct ordering falls out of that. Less to go wrong, and no dependence on what else is open.

## Points worth a second reader's attention

- **`sort -V`, not string comparison.** `5.5.99` must not outrank `5.6.0`. Verified across the ordering cases including that one, plus the equal case, which blocks.
- **Fails closed.** If `package.json` can't be read at any of the three refs, auto-merge is not armed. Skipping one arming run is cheap because `synchronize` re-runs the gate; guessing risks the exact failure the gate exists to prevent. The cost is that a transient contents-API blip delays an arming by one push.
- **The gate applies to the Dependabot path too**, not just the drift-bot branches. A Dependabot PR that bumps `package.json` carries the identical hazard.
- **A blocked PR is stale, not broken.** It needs the branch update it would have needed regardless, and re-arms automatically once its version clears the tip. The `::warning::` names both versions so the reason is legible in the run without digging.

## Verification

`actionlint` passes, along with `analyze`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `validate-package-json`, `triage`.

The gate's own logic was exercised offline against the ordering table above; it has **not** been exercised end-to-end against a live PR, because doing so requires a real bot PR carrying a stale version. The first cc-drift or template-label PR after this merges is the real test. If the gate is wrong in the conservative direction it blocks an arming and someone merges by hand; if it is wrong in the permissive direction the behaviour is exactly today's, so this cannot be worse than the status quo.

Workflow-only, no version bump — `.github/` is not in the package `files` list, so this ships nothing.

sprayberry-redline (approved):

I wrote this change. This approval is therefore not independent review; approving at the maintainer's direction and recording that here rather than leaving it implicit.

## What it fixes

`auto-merge-bot-prs.yml` armed native auto-merge on branch name alone, with no visibility into the version a PR carried. Two bot PRs that independently claimed the same version were free to race.

That happened today: #954 (cc-drift `maxTested`) and #955 (template label refresh) were drafted by different watchers, each took the next patch off `5.5.9`, and both landed on `5.5.10` with different dates. Whichever merged second was the casualty — equal to the released version means `cc-drift-auto-release.yml`'s duplicate-tag guard fast-exits green and ships nothing (a release that silently did not happen, on a chore PR nobody re-reads), and lower means master's version moves backwards below a published tag.

It only came out right today because #954 happened to go green first. Timing, not a guarantee.

## The invariant

A PR that changes the `package.json` version must carry one strictly greater than the base branch tip. A PR that leaves the version untouched — the common Dependabot action bump — is explicitly unaffected, which is why the gate compares head against the PR's own base as well as against the tip: unchanged version means `head == base`, and it exits early.

Deliberately a **per-PR invariant rather than cross-PR ordering**. The workflow never enumerates other open PRs or tries to sequence them; each PR simply has to advance past the tip, and correct ordering falls out of that. Less to go wrong, and no dependence on what else is open.

## Points worth a second reader's attention

- **`sort -V`, not string comparison.** `5.5.99` must not outrank `5.6.0`. Verified across the ordering cases including that one, plus the equal case, which blocks.
- **Fails closed.** If `package.json` can't be read at any of the three refs, auto-merge is not armed. Skipping one arming run is cheap because `synchronize` re-runs the gate; guessing risks the exact failure the gate exists to prevent. The cost is that a transient contents-API blip delays an arming by one push.
- **The gate applies to the Dependabot path too**, not just the drift-bot branches. A Dependabot PR that bumps `package.json` carries the identical hazard.
- **A blocked PR is stale, not broken.** It needs the branch update it would have needed regardless, and re-arms automatically once its version clears the tip. The `::warning::` names both versions so the reason is legible in the run without digging.

## Verification

`actionlint` passes, along with `analyze`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`, `validate-package-json`, `triage`.

The gate's own logic was exercised offline against the ordering table above; it has **not** been exercised end-to-end against a live PR, because doing so requires a real bot PR carrying a stale version. The first cc-drift or template-label PR after this merges is the real test. If the gate is wrong in the conservative direction it blocks an arming and someone merges by hand; if it is wrong in the permissive direction the behaviour is exactly today's, so this cannot be worse than the status quo.

Workflow-only, no version bump — `.github/` is not in the package `files` list, so this ships nothing.

PR #961, opened 2026-08-13
chore(template): label refresh _version -> v2.1.229 (clears sdk-drift)

askalf:

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-13T00:10:19+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.229** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.229`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.229`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.5.13` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.229. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.


sprayberry-redline (approved):

I did not write this change (auto-drafted by `cc-drift-template-watch.yml`), but I ran `gh pr update-branch` on it to clear a `BEHIND` state. Recording that rather than leaving it implicit.

## Verified against the diff

- `_version` → `2.1.229`
- `_supportedMaxTested` → `2.1.229`
- `user-agent` → `claude-cli/2.1.229 (external, sdk-cli)`
- `_captured` and `x-stainless-package-version` untouched
- `package.json` `5.5.12` → `5.5.13`, with a matching CHANGELOG section

Exactly the label-only shape the body claims: three label fields move, the last-real-capture marker and the unrelated stainless pin stay put. Correct minimal diff for a sync with zero wire-shape drift.

## Provenance worth noting

This PR only exists because the runner's own Claude Code was upgraded `2.1.228` → `2.1.229`. `cc-drift-template-watch` captures from the **locally installed** CC, while `check-sdk-drift.mjs` compares the bundle against **npm latest**. With the box lagging npm, the template watcher saw bundle == local, reported no drift, and ran green hourly without ever producing this PR — while #957 kept flagging the mismatch. Green runs that produce nothing are the expensive failure mode for self-hosted watchers, and this was a live instance of it.

Merging this should let the next daily `sdk-drift-watch` (`17 6 * * *`) close #957.

## Version ordering

`5.5.12` → `5.5.13` against a `5.5.12` tip. This is the first PR to run the version rule from the extracted `scripts/version-advances.sh` (#962) rather than the inlined copy, and it produced the same verdict the inline version would have:

```
base(64fa95ac)=5.5.12  head(6360e25d)=5.5.13  tip(master)=5.5.12
Arming: version advances 5.5.12 -> 5.5.13.
```

No collision with anything open, so no repeat of the #954/#955 shape.

sprayberry-redline (approved):

I did not write this change (auto-drafted by `cc-drift-template-watch.yml`), but I ran `gh pr update-branch` on it to clear a `BEHIND` state. Recording that rather than leaving it implicit.

## Verified against the diff

- `_version` → `2.1.229`
- `_supportedMaxTested` → `2.1.229`
- `user-agent` → `claude-cli/2.1.229 (external, sdk-cli)`
- `_captured` and `x-stainless-package-version` untouched
- `package.json` `5.5.12` → `5.5.13`, with a matching CHANGELOG section

Exactly the label-only shape the body claims: three label fields move, the last-real-capture marker and the unrelated stainless pin stay put. Correct minimal diff for a sync with zero wire-shape drift.

## Provenance worth noting

This PR only exists because the runner's own Claude Code was upgraded `2.1.228` → `2.1.229`. `cc-drift-template-watch` captures from the **locally installed** CC, while `check-sdk-drift.mjs` compares the bundle against **npm latest**. With the box lagging npm, the template watcher saw bundle == local, reported no drift, and ran green hourly without ever producing this PR — while #957 kept flagging the mismatch. Green runs that produce nothing are the expensive failure mode for self-hosted watchers, and this was a live instance of it.

Merging this should let the next daily `sdk-drift-watch` (`17 6 * * *`) close #957.

## Version ordering

`5.5.12` → `5.5.13` against a `5.5.12` tip. This is the first PR to run the version rule from the extracted `scripts/version-advances.sh` (#962) rather than the inlined copy, and it produced the same verdict the inline version would have:

```
base(64fa95ac)=5.5.12  head(6360e25d)=5.5.13  tip(master)=5.5.12
Arming: version advances 5.5.12 -> 5.5.13.
```

No collision with anything open, so no repeat of the #954/#955 shape.

PR #960, opened 2026-08-12
chore(cc-drift): v5.5.12 — maxTested → v2.1.229

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.229 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.228` → `2.1.229` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.12`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.12] - 2026-08-12` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.229 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.228). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.228; npm latest is v2.1.229. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.229 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.229 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.229. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.229" step — no local run needed.)
- ✅ **Auto-merges** the moment the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`). `master` requires no review, so no human approval gates it.
[...]

PR #962, opened 2026-08-13
ci(auto-merge): add an hourly sweep and share the version gate

askalf:
## The gap

`auto-merge-bot-prs.yml` triggers on `pull_request_target` **only**, so it depends entirely on webhook delivery.

In the [2026-08-06 Actions incident](https://www.githubstatus.com/) GitHub's webhook throughput fell to roughly **15%**, and GitHub confirmed the missed trigger events **could not be replayed automatically** — recovery meant manually re-running, pushing a commit, or updating the PR. A drift-bot PR whose `opened` event is dropped never gets armed. It just sits open, green and ignored, until a human notices.

Nothing currently catches that. `self-healing-supervisor` retries runs that **failed**; a run that never started is *absent*, not failed.

This is the same silent-no-ship class as the version collision fixed in #958, arriving by a different route.

## The fix

`auto-merge-sweep.yml`, hourly at `27 * * * *` (clear of `cc-drift-watch` `:00`, `cc-drift-auto-release` `:15`, `cc-drift-template-watch` `:41`).

It's the shape `cc-drift-auto-release.yml` already uses — an hourly tick beside the event trigger, with an idempotent gate — applied to the one workflow in the release path that lacked one. It lists open, non-draft, same-repo, **unarmed** drift-bot PRs and arms them. A healthy hour costs a single API listing; `autoMergeRequest == null` does the filtering server-side.

Every candidate goes through the **same version gate** as the event path, so the sweep can't become a way around it.

## Scope: drift-bot branches only

Dependabot is deliberately excluded, for two reasons:

1. Its major-version exclusion comes from `dependabot/fetch-metadata`, which needs the PR event payload and **cannot run on a schedule**. Title-parsing is not an acceptable substitute — a grouped `non-major` title contains the literal substring `major`, which is the exact false positive this workflow's gating was already moved off.
2. Dependabot rebases and recreates its PRs, which re-fires `synchronize` and self-heals the dropped-webhook case anyway. Drift-bot PRs are one-shot, so they're the ones that stay stuck.

## The version rule moved to a shared script

`scripts/version-advances.sh` now holds the rule; both the event path and the sweep call it.

Two copies would have been worse than none — the sweep would silently become a bypass the moment either copy was edited alone.

Extracting it also makes the rule **testable**, which follows the precedent of `scripts/health-verdict.sh` (pulled out of `cc-oauth-health.yml` for exactly this reason). `test/version-advances.mjs` adds 16 assertions, picked up automatically by `test/all.test.mjs`:

| case | expected |
|---|---|
| head == base (Dependabot action bump) | arm |
| head == base while tip advanced | arm |
| head > tip | arm |
| head == tip (the #954/#955 collision) | block |
| head < tip | block |
| `5.5.99` vs a `5.6.0` tip | **block** — `sort -V`, not string compare |
| `5.5.10` vs a `5.5.9` tip | arm |
| any version unreadable | block (fails closed) |

That matters because the gate's blocking direction is one the gate itself now prevents from occurring in the wild. The permissive direction was confirmed live on #960 (`5.5.11 → 5.5.12`, armed); the blocking direction is only ever exercised here.

## Security note

Both workflows `sparse-checkout` only that one script, from the **base branch, never the PR head**. `pull_request_target` runs with write scopes and repository secrets, so checking out and executing head code would be the classic fork-PR RCE. `persist-credentials: false` on both.

## Scope

Workflow + script + test. No version bump — none of these paths are in the package `files` list, so this ships nothing.

### DIFF
--- a/.github/workflows/auto-merge-bot-prs.yml
+++ b/.github/workflows/auto-merge-bot-prs.yml
@@ -114,11 +114,13 @@ jobs:
           # The rule itself lives in scripts/version-advances.sh so this path
           # and the hourly sweep cannot drift apart, and so both directions of
           # it stay unit-testable (test/version-advances.mjs).
+          # head/tail rather than printf+sed: the script's contract is exactly
+          # two lines, and this keeps backslash escapes out of a shell block
+          # nested inside YAML, where one mangled \n silently becomes a real
+          # newline and breaks the parse.
           verdict=$(sh scripts/version-advances.sh "$base_ver" "$head_ver" "$tip_ver")
-          safe=$(printf '%s
-' "$verdict" | sed -n '1p')
-          reason=$(printf '%s
-' "$verdict" | sed -n '2p')
+          safe=$(echo "$verdict" | head -1)
+          reason=$(echo "$verdict" | tail -1)
 
           echo "safe=$safe" >> "$GITHUB_OUTPUT"
           if [ "$safe" = "true" ]; then
--- a/.github/workflows/auto-merge-bot-prs.yml
+++ b/.github/workflows/auto-merge-bot-prs.yml
@@ -75,6 +75,18 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      # Checks out the BASE branch only (actions/checkout's default ref on
+      # pull_request_target), never the PR head. This workflow runs with write
+      # scopes and repository secrets, so checking out and executing head code
+      # would be the classic fork-PR RCE. Sparse: the one script is the whole
+      # need, matching cc-oauth-health.yml.
+      - name: Fetch the version-gate script
+        uses: askalf/checkout-with-retry@744195501c3e2b794c50370b753a7b8c93d084f5  # v1.0.0
+        with:
+          sparse-checkout: scripts/version-advances.sh
+          sparse-checkout-cone-mode: false
+          persist-credentials: false
+
       - name: Check the version advances past the base tip
         id: version_gate
         env:
@@ -86,11 +98,11 @@ jobs:
         run: |
           set -euo pipefail
 
-          # No checkout on this runner, so read package.json at each ref
-          # through the contents API rather than from a working tree.
+          # Read package.json through the contents API rather than from the
+          # checkout: the sparse checkout deliberately holds only the script,
+          # and the three versions live at three different refs anyway.
           version_at() {
-            gh api "repos/${REPO}/contents/package.json?ref=$1" --jq '.content' 2>/dev/null \
-              | base64 -d 2>/dev/null | jq -r '.version // empty'
+            gh api "repos/${REPO}/contents/package.json?ref=$1" --jq '.content' 2>/dev/null               | base64 -d 2>/dev/null | jq -r '.version // empty'
           }
 
           base_ver=$(version_at "$BASE_SHA")     # what this PR branched from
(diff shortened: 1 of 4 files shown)


sprayberry-redline (approved):

I wrote this change. This approval is therefore not independent review; approving at the maintainer's direction and recording it here rather than leaving it implicit.

## What it closes

`auto-merge-bot-prs.yml` triggers on `pull_request_target` only, so it depends entirely on webhook delivery. In the 2026-08-06 Actions incident GitHub's webhook throughput fell to roughly 15%, and the missed trigger events could not be replayed automatically. A drift-bot PR whose event is dropped never gets armed — it sits open, green and ignored. `self-healing-supervisor` does not catch it, because it retries runs that **failed**, and a run that never started is absent rather than failed.

The fix is the shape `cc-drift-auto-release.yml` already proved: an hourly tick beside the event trigger, with an idempotent gate. Applied here to the one workflow in the release path that lacked one.

## Points a second reader should check

- **The sweep cannot become a bypass.** Every candidate runs the same version rule as the event path, now shared via `scripts/version-advances.sh`. Two copies would have been worse than none — either could be edited alone and silently diverge.
- **The Dependabot exclusion is deliberate, not an oversight.** `dependabot/fetch-metadata` needs the PR event payload and cannot run on a schedule, and title-parsing is not a substitute: a grouped `non-major` title contains the substring `major`, the exact false positive this workflow's gating was already moved off. Dependabot also rebases and recreates its PRs, re-firing `synchronize`, so it self-heals the dropped-webhook case. Drift-bot PRs are one-shot, which is why they are the scope.
- **`pull_request_target` + checkout is the fork-RCE shape.** Both workflows sparse-checkout only the one script, from the **base branch, never the PR head**, with `persist-credentials: false`. Worth confirming that reading holds, since this workflow carries write scopes and secrets.
- **Cron slot** `27 * * * *` is clear of `cc-drift-watch` (`:00`), `cc-drift-auto-release` (`:15`) and `cc-drift-template-watch` (`:41`).
- **Cost of a healthy hour** is one `gh pr list` call; `autoMergeRequest == null` filters server-side, and the loop is skipped entirely at zero candidates.

## Verification

`test/version-advances.mjs` adds 16 assertions, auto-discovered by `test/all.test.mjs`; all pass in CI. They cover both directions including `5.5.99` against a `5.6.0` tip (the case a string compare gets backwards), the `head == tip` collision that produced #954/#955, and the fail-closed path on unreadable input.

That coverage is the point of extracting the rule: the gate's blocking direction is one the gate itself now prevents from occurring in the wild, so tests are the only place it is ever exercised. The permissive direction was separately confirmed live on #960 (`5.5.11 → 5.5.12`, armed).

**The sweep's own arming path has not run against a real stranded PR** — that needs a dropped webhook, which cannot be manufactured on demand. The jq candidate filter was dry-run against live repository data and returns cleanly with zero open drift PRs. Worst case if it is wrong: it fails to arm something, which is the status quo it was written to improve on, and the `::notice::` on a successful arm makes any real recovery visible in the run list rather than only in the log.

One honest note on process: the first push failed `actionlint` because a `printf '%s\n'` escape got mangled into a literal newline while rewriting the YAML, breaking the parse. Both callers now read the two-line verdict with `head -1` / `tail -1`, so there is no escape left to mangle. Caught by the required check, as intended.

CI green: `CodeQL`, `actionlint`, `analyze`, `build (18/20/22)`, `docker-cap-drop-smoke`, `test`, `validate-package-json`.

No version bump — workflows, `scripts/`, and `test/` are not in the package `files` list, so this ships nothing.

PR #963, opened 2026-08-13
ci: draft a narrow, owner-triggered Claude job for drift PRs

askalf:
> **Draft — do not merge yet.** This exists to be read before anything is live. It also needs an `ANTHROPIC_API_KEY` secret that does not exist; until it does, the job fails fast with a clear message rather than running half-configured.

## Why

The Claude GitHub App is installed on this account, but **no workflow references it**. Today that is a standing grant of write access to actions, checks, code, discussions, issues, PRs, repository hooks and workflows, in exchange for nothing.

Either wire it to a job worth having, or narrow the install. This is a proposal for the former, scoped as tightly as I could make it while still being useful.

## The one job

Unstick an auto-drafted drift PR: resolve a `CHANGELOG.md` / `package.json` conflict against `master`, or fix a mechanical check failure.

That is the work that keeps needing hand-resolution in a local worktree — six times in one evening recently, every one of them the same shape: two bot PRs drafted off the same release, colliding on the changelog.

Not a general "@claude, do things" integration. Nothing runs on a schedule, on PR open, or on any bot's behalf.

## Security shape

**`issue_comment`, never `pull_request_target`.** `issue_comment` always runs the workflow file from the **default branch**, so a PR cannot alter the rules it will be judged by. `pull_request_target` combined with a head checkout is the fork-PR RCE this repo has deliberately kept shut, and no amount of care inside the job makes that combination safe.

**Four independent gates:**

| # | gate | evaluated |
|---|---|---|
| 1 | comment is on a PR, not an issue | before the job starts |
| 2 | `author_association == 'OWNER'` | before the job starts |
| 3 | head repo is not a fork | before Claude is invoked |
| 4 | head branch is `bot/cc-drift-*` or `bot/template-label-*` | before Claude is invoked |

Gate 4 does double duty. Beyond scope, it means the diff and PR body Claude reads were written by **our own watchers**, not by an arbitrary commenter — which is the realistic prompt-injection path on a public repo.

**`workflows: write` is not granted.** This is mechanical, not advisory: without the scope, the token is *refused* when pushing changes under `.github/workflows`, so the job cannot rewrite CI even if instructed to.

The cost is real and I want it stated plainly: **it cannot fix the actionlint-class breakage** that prompted this discussion. The prompt tells it to stop and comment instead. That limit is the point — an agent reachable from PR comments must not hold the scope that edits the pipeline. `actions` is requested **read-only**, for reading failed run logs.

Top-level `permissions` stays `contents: read` for Scorecard Token-Permissions; the job requests exactly three write scopes and nothing else.

## Cost

Billed to `ANTHROPIC_API_KEY` — metered credits, deliberately **not** the fleet subscription. Two reasons: an invocation cannot eat the pool the fleet runs on, and a capped subscription does not disable it. (`claude_code_oauth_token` is the alternative and would do the opposite on both counts.)

Bounded per run by `--max-turns 15` and `timeout-minutes: 20`; bounded in aggregate by being owner-triggered only.

## The resolution rules are in the prompt on purpose

These are the part that must not be gotten wrong, so they are stated as rules rather than left to judgement:

- **`package.json`: keep the HIGHER version, never resolve downward.** A version below an already-published tag is worse than an unmerged PR — it is the shape that produces a silent no-op release.
- **`CHANGELOG.md`: keep BOTH sides**, newest section on top, descending; never drop an entry for a release that already shipped.
- Anything under `## [Unreleased]` stays unreleased.
- Verify with `scripts/extract-release-notes.mjs` and `scripts/check-package-json.mjs`; if either fails, stop rather than force.

Plus hard limits: no `.github/`, no behaviour changes in `src/`, no approve/merge/close, no touching other PRs, and no commit at all when the right answer is unclear — leave a comment for a human instead.

## What I'd want reviewed

1. Whether gate 4's branch allowlist is the right boundary, or too permissive.
2. Whether `contents: write` on the PR branch is acceptable given gates 1–4, since that is the scope that actually mutates anything.
3. Whether this is worth having at all versus simply narrowing the app install — a defensible answer given the fleet already has a Code Reviewer and a Developer.

### DIFF
--- a/.github/workflows/claude-drift-fix.yml
+++ b/.github/workflows/claude-drift-fix.yml
@@ -1,0 +1,202 @@
+name: Claude drift-PR fix (opt-in, owner-triggered)
+
+# NARROW BY DESIGN. This is not a general "@claude, do things" integration.
+# It does one job: unstick an auto-drafted drift PR — resolve a CHANGELOG /
+# package.json conflict against master, or fix a failing non-workflow check —
+# so the routine bot PRs stop needing hand-resolution in a local worktree.
+#
+# It is OPT-IN per invocation. Nothing runs on a schedule, on PR open, or on
+# any bot's behalf. The only way in is the owner typing the trigger phrase on
+# a qualifying PR, or a manual workflow_dispatch.
+#
+# ── WHY THE TRIGGER IS issue_comment AND NOT pull_request_target ────────────
+# pull_request_target runs with write scopes and repository secrets while the
+# PR head is attacker-controlled on a public repo; combining it with a head
+# checkout is the classic fork-PR RCE, and this repo has deliberately kept
+# that door shut. issue_comment always runs the workflow file from the DEFAULT
+# BRANCH, so a PR cannot alter the rules it will be judged by.
+#
+# ── FOUR INDEPENDENT GATES ─────────────────────────────────────────────────
+#   1. comment is on a pull request, not an issue
+#   2. comment author_association == OWNER
+#   3. PR head repo == this repo (never a fork)
+#   4. PR head branch is one of our own bot prefixes
+# Gates 1–2 are evaluated before any job starts; 3–4 before Claude is invoked.
+# Restricting to bot branches also means the diff and PR body Claude reads were
+# written by our own watchers rather than by an arbitrary commenter, which is
+# the realistic prompt-injection path on a public repo.
+#
+# ── WHAT IT DELIBERATELY CANNOT DO ─────────────────────────────────────────
+# `workflows: write` is NOT granted. That is mechanical, not advisory: without
+# the scope the token is refused when pushing changes under .github/workflows,
+# so this job cannot rewrite CI even if instructed to. The cost is real — it
+# cannot fix the actionlint-class breakage that started this conversation, and
+# it is told to stop and comment instead. Accepting that limit is the point:
+# an agent reachable from PR comments must not hold the scope that edits the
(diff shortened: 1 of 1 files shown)


askalf:
Closing unmerged — the job this was built for turned out not to need a model.

#964 encodes the same conflict resolution deterministically: keep the higher `package.json` version, keep both CHANGELOG sides newest-first. The input shape is fixed by our own drift bots, so there is no unpredictability for an agent to absorb. 42 assertions, and an end-to-end run against a real `git merge` that reproduces the hand resolution exactly.

That end-to-end also made the argument for itself. It caught two defects the unit tests missed — a CRLF blind spot that made the resolver report "no conflict" on a genuinely conflicted file, and an `## [Unreleased]` entry getting buried under a released section. Both are the kind of thing a regression test pins down permanently and a prompt does not.

What would have been left for an agent is novel CI failures, and the permission boundary proposed here already ruled out the largest class of those: without `workflows: write` it could not have fixed the `actionlint` breakage that prompted the discussion.

So the remaining value did not justify a comment-reachable job holding `contents: write`, and the Claude GitHub App install is being narrowed rather than wired.

The design notes stay useful if this is ever revisited — in particular:

- `issue_comment` rather than `pull_request_target`, so a PR cannot alter the rules it is judged by
- the four independent gates, and why the branch allowlist limits scope rather than establishing trust
- that `contents: write` here was only acceptable because the `master` ruleset has `bypass_actors: []`; that coupling would need documenting before anything like this goes live

PR #952, opened 2026-08-11
fix(probe): stop check-overage-live reporting CLEAN for a model that was never served

askalf:
## The bug

`scripts/check-overage-live.mjs` can print

```
✅ CLEAN — all models billed to the subscription plan
```

and exit `0` for a model that was **never served at all**.

`classify()` ended with:

```js
if (SUBSCRIPTION_CLAIMS.has(r.claim || '')) return r.httpStatus === 429 ? 'WINDOW' : 'OK';
if (r.httpStatus === 429) return r.rlStatus === 'rejected' ? 'REJECTED' : 'WINDOW';
```

A 429 that carries **no `anthropic-ratelimit-unified-representative-claim` header** misses the first branch and falls through to the second, landing on `WINDOW` — whose legend asserts "subscription 429 (rate cap, not a breach)". Nothing verified a subscription claim. `WINDOW` isn't counted as a breach, so the run reports clean.

That matters because this script exists to answer one question — *is traffic billing to the subscription?* — and it was answering "yes" from the absence of evidence.

## How it showed up

Two models, one account, one second apart:

| model | claim | overage-util | old outcome |
|---|---|---|---|
| `claude-sonnet-5` | `five_hour` | – | OK |
| `claude-fable-5` | **– (no claim header)** | – | WINDOW → folded into CLEAN |

On a Pro-plan account, `claude-fable-5` comes back as a bare 429: no claim, no rate-limit headers, and **no dario request record at all** (`dario log: 1 request records` for 2 requests). Fable is extra-usage-only there, and with extra usage not enabled the request is rejected rather than billed. Sonnet on the same account in the same second billed `five_hour`/`subscription` normally.

So the run measured nothing for fable and still exited 0.

## The fix

1. **`classify()`** — a claimless 429 is now `UNSERVED`, not `WINDOW`:

```js
if (r.httpStatus === 429) {
  if (r.rlStatus === 'rejected') return 'REJECTED';
  return r.claim ? 'WINDOW' : 'UNSERVED';
}
```

`WINDOW` now means what its legend always claimed: a 429 that actually carried a subscription claim.

2. **Per-model `measured` flag** — true only if at least one request returned a billing claim. Billing classification is the whole output, so "no claim ever returned" is no data.

3. **Three-way verdict, distinct exit code:**

| condition | verdict | exit |
|---|---|---|
| overage/api/halt | ❌ BREACH | 1 |
[...]

PR #964, opened 2026-08-13
ci: resolve release conflicts deterministically instead of by hand

askalf:
## Why this is a script and not a prompt

Four PRs needed hand-resolution in a worktree on 2026-08-12 — #952, #955, #960, #961 — and every one came out the same two ways:

- **`package.json`** — keep the **higher** version.
- **`CHANGELOG.md`** — keep **both** sides, version sections newest-first, unreleased bullets on top.

That is a rule, not a judgement. The input shape is fixed by our own drift bots, so the unpredictability an LLM is good at isn't present.

Getting it wrong is quiet in both directions. A version resolved **downward** lands below an already-published tag, and `cc-drift-auto-release`'s duplicate-tag guard then fast-exits **green** having shipped nothing — a release that silently did not happen, on a chore PR nobody re-reads after merge. Dropping a CHANGELOG side loses an entry for a release that did ship.

## What it refuses

Refusal is the default for anything outside the known shape:

- a conflict in a file other than `package.json` / `CHANGELOG.md`
- a `package.json` conflict touching more than the lone `"version"` line
- a version that isn't `X.Y.Z`
- markers surviving its own output

Nothing is written unless **every** requested file resolved cleanly, so a refusal leaves the working tree exactly as it was found. Refusing costs a hand-resolution; guessing costs a silent no-op release.

## Two defects the unit tests missed

Both were found by running against a **real `git merge`** rather than hand-written fixtures. 27 assertions were green at the time.

**CRLF.** Git on Windows leaves markers as `"<<<<<<< HEAD\r"`. In JavaScript the `.` metacharacter won't consume a carriage return, and `$` without the `m` flag won't match before one — so `/^<{7} .*$/` failed on every marker and the script reported **"no conflict" on a genuinely conflicted file**. That is the worst outcome available to it: the caller proceeds and commits the markers. Now normalized on the way in, with the file's own line endings restored on the way out (an LF file stays LF).

**Unreleased bullets left outside the region.** Git doesn't always put both sides' bullets inside the conflict. One landed after the `>>>>>>>` line, and emitting the merged version sections first buried an `## [Unreleased]` entry under `## [5.5.13]`. Loose text directly after a region is now hoisted back to the section that was open before it.

## Verification

**42 assertions**, auto-discovered by `test/all.test.mjs`: both resolution directions, the `5.5.99` vs `5.6.0` ordering case, identical-sections collapsing, CRLF *and* LF, line-ending preservation in both directions, and every refusal path.

**End-to-end against a real two-file conflict** — branch at `5.5.12`, ours `5.5.14`, master `5.5.13`:

```
conflicted: CHANGELOG.md package.json
package.json: resolved
CHANGELOG.md: resolved

version: 5.5.14
markers: CHANGELOG.md:0 package.json:0

## [Unreleased]
- **check-overage-live.mjs no longer reports a false CLEAN.**
## [5.5.14] - 2026-08-13
## [5.5.13] - 2026-08-13
## [5.5.12] - 2026-08-12

check-package-json OK
extract-release-notes OK
```

Identical to what was produced by hand, and it passes the repo's own validators.

## Usage

```
git merge origin/master            # conflicts
node scripts/resolve-release-conflicts.mjs
node scripts/check-package-json.mjs && node scripts/extract-release-notes.mjs <version>
git add -A && git commit
```

## Not wired into CI yet — deliberately

This is the resolver only. Calling it from `auto-merge-sweep.yml` would change that workflow from *arming* PRs to *pushing commits* to them, which is a real escalation in what the hourly job can do. Worth deciding separately rather than folding into this PR.

Standalone it already removes the manual editing, which was most of the cost.

`scripts/` and `test/` aren't in the package `files` list, so this ships nothing and carries no version bump.

### DIFF
--- a/scripts/resolve-release-conflicts.mjs
+++ b/scripts/resolve-release-conflicts.mjs
@@ -1,0 +1,262 @@
+#!/usr/bin/env node
+// Resolve the ONE conflict shape our release automation produces.
+//
+// Two bot PRs drafted off the same release both add a section under
+// `## [Unreleased]` and both bump `package.json`, so merging master into
+// either one conflicts in exactly two files, in exactly two ways. Every such
+// conflict resolves the same way, and getting it wrong has a specific cost:
+// a version below an already-published tag makes cc-drift-auto-release's
+// duplicate-tag guard fast-exit green having shipped nothing.
+//
+// That is a rule, not a judgement, so it lives here rather than in a prompt.
+//
+//   package.json  keep the HIGHER version. Never downward.
+//   CHANGELOG.md  keep BOTH sides. Version sections descending, newest first;
+//                 loose text (bullets under `## [Unreleased]`) stays on top.
+//
+// REFUSES anything outside that shape — a conflict in another file, a
+// package.json conflict touching more than the version line, a version that
+// is not X.Y.Z. Refusing costs a hand-resolution; guessing costs a silent
+// no-op release.
+//
+// Usage:
+//   node scripts/resolve-release-conflicts.mjs [file...]   # default: both
+// Exit 0 = every conflict resolved (or none present), files rewritten.
+// Exit 1 = refused; the working tree is left exactly as it was found.
+
+import { readFileSync, writeFileSync, existsSync, realpathSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+const START = /^<{7} .*$/;
+const MID = /^={7}$/;
+const END = /^>{7} .*$/;
+
+// CRLF is not cosmetic here. A git checkout on Windows leaves markers as
+// "<<<<<<< HEAD\r", and in JavaScript `.` will not consume a carriage return
+// while `$` (no `m` flag) will not match before one — so every regex above
+// silently fails and the file reads as conflict-free. That is the worst
+// possible failure mode for this script: reporting success while leaving
+// markers in package.json. Normalize on the way in and restore on the way
+// out, so the file keeps whatever convention it already had.
+const detectEol = (text) => (text.includes('\r\n') ? '\r\n' : '\n');
+const toLf = (text) => text.replace(/\r\n/g, '\n');
+const restoreEol = (text, eol) => (eol === '\r\n' ? text.replace(/\n/g, '\r\n') : text);
+
+/** Parse "1.2.3" into comparable parts. Returns null when not X.Y.Z. */
+export function parseVersion(v) {
+  const m = /^(\d+)\.(\d+)\.(\d+)$/.exec((v ?? '').trim());
+  return m ? [Number(m[1]), Number(m[2]), Number(m[3])] : null;
+}
+
+/** Semver compare. Numeric per component, so 5.5.99 < 5.6.0. */
+export function compareVersions(a, b) {
+  const pa = parseVersion(a), pb = parseVersion(b);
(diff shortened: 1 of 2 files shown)


sprayberry-redline (approved):

I wrote this change. This approval is therefore not independent review; approving at the maintainer's direction and recording it here rather than leaving it implicit.

## What it replaces

Four PRs needed hand-resolution in a worktree on 2026-08-12 (#952, #955, #960, #961), and every one resolved the same two ways: keep the higher `package.json` version, keep both CHANGELOG sides newest-first. The input shape is fixed by our own drift bots, so this is a rule rather than a judgement, and it does not need a model to apply it.

Both failure directions are quiet, which is the argument for encoding it. A version resolved downward lands below an already-published tag, and `cc-drift-auto-release`'s duplicate-tag guard then fast-exits **green** having shipped nothing. Dropping a CHANGELOG side loses an entry for a release that did ship. Neither produces a red run.

## Worth a second reader's attention

- **The refusal set is the safety property.** Another file, a `package.json` conflict touching more than the lone version line, a non-`X.Y.Z` version, or markers surviving its own output all cause a refusal, and nothing is written unless every requested file resolved cleanly. Confirm that the partial-write guard actually holds — it is what makes a refusal safe to retry by hand.
- **Line-ending handling is load-bearing, not cosmetic.** See below.
- **The hoisting rule** — loose text directly after a conflict region belongs to the section open before it — is the least obvious part. It is correct for the conflicts our bots produce (where that section is always `## [Unreleased]`), and would need revisiting if the tooling ever conflicts elsewhere in the file.

## Two defects that unit tests did not catch

Both surfaced only when it was run against a real `git merge`. 27 assertions were green at that moment, which is the useful part of the story.

**CRLF.** Git on Windows leaves markers as `"<<<<<<< HEAD\r"`. In JavaScript the dot metacharacter will not consume a carriage return, and `$` without the `m` flag will not match before one, so `/^<{7} .*$/` failed against every marker and the script reported **"no conflict" on a genuinely conflicted file** — the worst outcome available to it, since the caller then commits the markers. All fixtures had been LF.

**Unreleased bullets outside the region.** Git does not always place both sides' bullets inside the conflict; one landed after the `>>>>>>>` line, and emitting the merged version sections first buried an `## [Unreleased]` entry under `## [5.5.13]`.

Both are fixed and both carry regression tests.

## Verification

42 assertions, auto-discovered by `test/all.test.mjs`, covering both resolution directions, `5.5.99` vs `5.6.0` ordering, identical-section collapsing, CRLF and LF including preservation in both directions, and every refusal path.

End-to-end against a real two-file conflict (base `5.5.12`, ours `5.5.14`, master `5.5.13`): resolves to `5.5.14`, restores the unreleased bullet above the version sections, leaves zero markers, and passes `check-package-json` plus `extract-release-notes`. Byte-for-byte the resolution produced by hand earlier that evening.

## Scope

Deliberately not wired into `auto-merge-sweep.yml`. Doing so would change that job from arming PRs to pushing commits to them, which is a real escalation for an unattended hourly workflow and belongs in its own decision. Standalone the script already removes the manual editing.

`scripts/` and `test/` are not in the package `files` list, so this ships nothing and carries no version bump. CI green: `CodeQL`, `actionlint`, `analyze`, `build (18/20/22)`, `docker-cap-drop-smoke`, `test`, `triage`, `validate-package-json`.

PR #973, opened 2026-08-14
chore(template): label refresh _version -> v2.1.232 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-14T05:17:50+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.232** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.232`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.232`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.5.15` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.232. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #966, opened 2026-08-13
chore(cc-drift): v5.5.14 — maxTested → v2.1.231

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.231 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.229` → `2.1.231` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.14`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.14] - 2026-08-13` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.231 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.229). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.229; npm latest is v2.1.231. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.231 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.231 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.231. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.231" step — no local run needed.)
- ✅ **Auto-merges** the moment the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`). `master` requires no review, so no human approval gates it.
[...]

PR #978, opened 2026-08-14
ci(cc-drift): close the drift PRs a new one supersedes

askalf:
The watcher runs hourly and already skips re-drafting when a PR for its **own** branch is open. It never looked at the older sibling branches — so whenever CC moved again before the previous drift PR merged, that PR stayed open indefinitely.

That is not untidy, it is wrong. An older drift PR carries a **lower** `maxTested` and an **older** package version, so merging it walks `master` backwards.

## The worked example: #966

| | maxTested | version |
|---|---|---|
| `master` (after #970, #973) | `2.1.232` | 5.5.15 |
| #966 proposed | `2.1.231` | 5.5.14 |

#966 sat open for a day while #970 — the same drift bump regenerated against a newer upstream — merged, and #973 took master to v5.5.15. It went `CONFLICTING`, **still collected an approving review** from the reviewer agent, and had to be closed by hand, along with the reviewer ticket that had been tracking it as genuine outstanding work.

The rate-cap made it visible, but nothing about it required a cap: any hour where CC moves twice before a merge produces the same stale pair.

## Change

After opening the new PR and enabling auto-merge, the workflow asks `scripts/close-superseded-drift-prs.mjs` which open `bot/cc-drift-v*` PRs the new one supersedes, then comments and closes each with `--delete-branch`.

Two deliberate choices:

- **Ordering.** The tidy-up runs *after* the new PR is queued, so a failure here can never leave the repo with nothing to ship.
- **Failure handling.** Every `gh` call is `|| true`, and a malformed PR listing exits 0 closing nothing. Tidying is not worth failing a run whose real work already succeeded.

## Tests

21 assertions in `test/close-superseded-drift-prs.mjs`, auto-discovered by `test/all.test.mjs`.

The selection is pure and split from the `gh` calls, the same shape as `scripts/resolve-release-conflicts.mjs`. The tests weight the dangerous direction — closing too little costs a manual close, closing a **newer** PR discards the fix that should ship:

- a newer open PR is never closed, even when one is also older
- never closes its own branch, or an equal-version PR
- ignores `bot/template-rebake-*` and human branches
- rejects version strings it cannot parse component-wise, so `bot/cc-drift-v2.1.231-rc1` is skipped rather than ranked as `231` by `parseInt`
- `2.1.9 < 2.1.229` — comparison is component-wise, not lexicographic
- degenerate input (null, non-array, missing fields, unparseable branch) returns empty rather than throwing

Reuses the existing `isOlderThan` from `_drift-patch-helpers.mjs` rather than adding a second comparator.

Verified against the real shape: feeding the actual #966 / #973 listing with `bot/cc-drift-v2.1.232` emits exactly `966`.

CI-only — no runtime change, so no version bump, matching #964.

### DIFF
--- a/.github/workflows/cc-drift-watch.yml
+++ b/.github/workflows/cc-drift-watch.yml
@@ -284,6 +284,32 @@ jobs:
           # preserved for inspection.
           gh pr merge "$PR_URL" --auto --squash --delete-branch
 
+          # Close the drift PRs this one supersedes.
+          #
+          # The skip above only covers THIS branch. When CC moves again before
+          # the previous drift PR merges, that PR is left open carrying a LOWER
+          # maxTested and an OLDER package version -- merging it would walk
+          # master backwards. #966 sat open a full day that way, went
+          # conflicted, still collected an approving review, and had to be
+          # closed by hand.
+          #
+          # Runs AFTER the new PR is open and auto-merging, so a failure here
+          # can never leave the repo with nothing queued. Every call is
+          # `|| true` for the same reason: tidying is not worth failing a run
+          # whose real work already succeeded. The selection is unit-tested in
+          # test/close-superseded-drift-prs.mjs -- notably that it refuses to
+          # close a NEWER PR, which is the one mistake that would lose work.
+          gh pr list --state open --json number,headRefName \
+            | node scripts/close-superseded-drift-prs.mjs "$BRANCH" \
+            | while read -r OLD_PR; do
+                [ -n "$OLD_PR" ] || continue
+                echo "Closing #$OLD_PR -- superseded by $PR_URL"
+                gh pr comment "$OLD_PR" --body \
+                  "Superseded by $PR_URL, which carries a newer \`maxTested\`. This PR proposes an older value, so merging it would move \`master\` backwards. Closed automatically by \`cc-drift-watch.yml\`." \
+                  || true
+                gh pr close "$OLD_PR" --delete-branch || true
+              done || true
+
       - name: Close resolved drift issues
         if: steps.drift.outputs.exit_code == '0'
         env:
--- a/scripts/close-superseded-drift-prs.mjs
+++ b/scripts/close-superseded-drift-prs.mjs
@@ -1,0 +1,91 @@
+#!/usr/bin/env node
+// Pick out the cc-drift PRs that a newly-opened one supersedes.
+//
+// WHY THIS EXISTS. cc-drift-watch runs hourly. It already skips re-drafting
+// when a PR for *its own* branch is open, but it never looked at the older
+// sibling branches — so whenever CC moved again before the previous drift PR
+// merged, the old PR stayed open forever.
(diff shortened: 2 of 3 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped change with strong test coverage.

## What I checked
- Full diff (`gh pr diff 978`): workflow addition in `cc-drift-watch.yml`, new pure selection
  module `scripts/close-superseded-drift-prs.mjs`, and its test file.
- `isOlderThan` in `scripts/_drift-patch-helpers.mjs` (reused, not reimplemented) — confirmed it's
  a component-wise numeric comparator, not lexicographic, so `2.1.9 < 2.1.229` is handled correctly
  as the tests claim.
- Surrounding workflow context (`BRANCH`/`PR_URL` are set earlier in the same step, `EXISTING_PR`
  guard for the branch's own PR) to confirm the new block's placement and variable scope are correct.
- CI check-rollup: all 11 checks green (`build` × 3 node versions, `test`, `analyze`/CodeQL,
  `actionlint`, `docker-cap-drop-smoke`, `validate-package-json`, `triage`); `auto-merge` shows
  `skipping` which is its normal no-op state for a human-authored PR.

## Correctness
- `versionFromBranch` rejects suffixed/non-numeric version strings (`bot/cc-drift-v2.1.231-rc1` →
  `null`) rather than letting `parseInt` silently truncate to `231` and mis-rank — verified against
  `isOlderThan`'s `parseInt`-based parsing, and covered by a test.
- `selectSupersededPrs` only closes strictly-older same-prefix branches, never the caller's own
  branch, a same-version PR, or a newer PR — each guarded by a distinct test case.
- Workflow-level failure handling: `gh pr comment ... || true` and `gh pr close ... || true` inside
  the `while` loop, plus `|| true` on the whole piped block (`.github/workflows/cc-drift-watch.yml`
  new block), so a malformed `gh pr list` output or a `close-superseded-drift-prs.mjs` non-zero exit
  can't fail the step — consistent with the script's own parse-failure path (`process.exit(0)` on
  malformed JSON, emitting nothing).
- The new block runs after `gh pr merge "$PR_URL" --auto --squash --delete-branch`, so a failure in
  the tidy-up step can't block the real work (the new drift PR) from being queued.

## Reuse
- Reuses `isOlderThan` from `_drift-patch-helpers.mjs` instead of adding a second comparator, as the
  PR description states.

## Test coverage
- 21 assertions covering the #966 regression scenario, ordering (oldest-first), non-drift/human
  branches, degenerate inputs (null, non-array, missing fields), and the numeric-vs-lexicographic
  version comparison. This is thorough for a pure selection function.

## Minor
- The workflow block interpolates `$BRANCH` and `$PR_URL` into `gh pr comment`/`echo` without
  quoting concerns since both are generated by the script/`gh pr create` output rather than
  attacker-controlled input — not a blocking issue, just noting it was checked.

No changes requested.

PR #990, opened 2026-08-16
[#881 residue?] auto-rebake: template drift detected 2026-08-16

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-16T23:50:56+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.18` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

> [!CAUTION]
> ## ⚠ This rebake is dario#881 residue — do NOT treat it as a genuine CC prompt change
>
> The base system-prompt capture behind this PR matched the known [#881](https://github.com/askalf/dario/issues/881) signature: the extra `# Context management` paragraph beginning "When you have enough information to act, act.", which appears in roughly **1 local capture in 10** and had **never** reached CI before this run. See the `::warning::` annotation on the run for the exact clause that matched and the captured-vs-bundled lengths.
>
> **What to do:** re-dispatch `cc-drift-template-watch.yml`. If the next capture comes back clean, this PR is the anomaly — close it. Merging it bakes the residue into the shipped wire-shape contract and cuts a release for a prompt change that did not happen upstream.
>
> **Either way, comment the outcome on #881.** Whether the tripwire caught a real CI occurrence or a false positive, that is exactly the data the issue is open for. #881 defers the bake-time normalisation deliberately, so nothing here strips the paragraph — this is detection only.

### Summary

**Verdict:** ✅ Benign

- **system_prompt:** +279 chars net (text-content drift — see unified diff below)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.233) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.233, 24 tools, 6413 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6413 → 5030 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[...]

PR #992, opened 2026-08-17
drift-watch: gate auto-rebake PRs on a real content diff (#990)

askalf:
## Why

`cc-drift-template-watch.yml`'s auto-rebake step asks the right question — "did the bake actually change the bundle?" — of the wrong thing:

```bash
if git diff --quiet -- src/cc-template-data.json; then ... skip PR ... fi
```

`capture-and-bake.mjs` restamps `_captured` on **every** run, so the file is always textually dirty. That gate can never fire.

#990 is what that costs. Its `src/cc-template-data.json` is byte-identical to v5.5.17 apart from the `_captured` timestamp — verified key by key at both SHAs:

| | |
|---|---|
| `_captured` | **DIFF** (`2026-08-15T04:59:39.819Z` → `2026-08-16T23:50:32.289Z`) |
| `tools` (111682 B), `system_prompt` (4799 B), `system_prompt_variants` (30920 B), `anthropic_beta`, `header_order`, `header_values`, `body_field_order`, `tool_names`, `agent_identity`, `_version`, `_supportedMaxTested`, `_source`, `_schemaVersion` | identical |

Both files are exactly 168034 bytes. Yet the PR carries a patch bump to `v5.5.18` and a CHANGELOG entry announcing a "wire-fingerprint drift … now matches the current CC wire shape" — describing a change that is not in its own diff. Merging would cut an npm release for a wire-shape change that did not happen upstream.

The dario#881 tripwire did its job: it warned a human to look. Nothing *stopped* the content-empty PR from existing, and the tripwire only covers the one known residue signature — any transient capture reaches the ship gate the same way.

## What changed

- **`scripts/drift-report.mjs`** — new `meaningfulTemplateKeys(prev, now)` + `TRANSIENT_TEMPLATE_FIELDS`. Compares top-level keys by canonical JSON, ignoring only `_captured`. Deliberately narrow: `_version` / `_supportedMaxTested` / the user-agent token are **not** transient, because a label move is exactly the content of a label-sync PR and must still ship.
- **`scripts/check-rebake-content.mjs`** (new) — compares the freshly-baked working-tree bundle against `HEAD`. Exit 0 with the changed keys on stdout; exit 1 with a `::warning::` annotation when content-empty; exit 2 on git/parse error.
- **`.github/workflows/cc-drift-template-watch.yml`** — the auto-rebake step's `git diff --quiet` sanity check now calls that script, and logs the changed keys when it proceeds.

Only the *shape*-rebake path changes. The exit-3 label-sync path keeps its own `git diff --quiet` check, which is correct there: `label-sync.mjs` does not re-capture, so it never touches `_captured`.

## Test plan

`test/bake-drift-report.mjs`: **132 → 144 assertions, 0 fail** (`node test/bake-drift-report.mjs`). New case 42 covers the #990 shape (only `_captured` moved → empty), that genuine drift in each axis still surfaces (`system_prompt`, `tools`, `anthropic_beta`, `system_prompt_variants`), added/removed keys, multi-key sorting, nested key-order insensitivity, and that a `_version` move still reports so label-sync is unaffected.

End-to-end against the real artifacts, not fixtures:

- #990's actual head bundle in the working tree → `check-rebake-content.mjs` exits **1** and prints the content-empty annotation.
- same bundle plus one appended sentence in `system_prompt` → exits **0**, stdout `system_prompt`.

## Re-dispatch outcome

Re-dispatched `cc-drift-template-watch.yml` per #990's own instructions: run [31981172851](https://github.com/askalf/dario/actions/runs/31981172851). The #881 residue did **not** reproduce (scrubbed base 4802 vs 4751 bundled = +51; the #881 signature is +279 and its marker clause was absent). #990 is the anomaly and should be closed rather than merged.

That run did surface a *different* transient — the capture carried a trailing `<total_tokens>15000000 tokens left</total_tokens>` line after the `# Context management` paragraph. Unrelated to this gate, filed separately; it is a per-session budget counter, not wire shape. No new rebake PR was opened, because the de-dup saw #990 still open.

Source ticket: `00MSWH41FA60F7465256F288C9`. Refs #990, #881.

### DIFF
--- a/.github/workflows/cc-drift-template-watch.yml
+++ b/.github/workflows/cc-drift-template-watch.yml
@@ -221,12 +221,20 @@ jobs:
           fi
           echo "issue_881=$issue_881" >> "$GITHUB_OUTPUT"
 
-          # Sanity: did the bake actually change the bundle?
-          if git diff --quiet -- src/cc-template-data.json; then
-            echo "::warning::--check reported drift but bake produced no diff vs HEAD — likely a transient capture; skipping PR"
+          # Sanity: did the bake actually change the bundle's CONTENT?
+          #
+          # `git diff --quiet` cannot answer this (dario#990): capture-and-bake
+          # restamps `_captured` on every run, so the file is ALWAYS textually
+          # dirty and this gate never fired. PR #990 reached the ship gate with
+          # a bundle byte-identical to v5.5.17 apart from that timestamp,
+          # carrying a patch bump and a CHANGELOG entry for drift that was not
+          # in its diff. check-rebake-content.mjs compares key by key and
+          # ignores only the transient provenance stamp.
+          if ! changed_keys=$(node scripts/check-rebake-content.mjs); then
             echo "skipped=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
+          echo "bake changed template keys: $changed_keys"
 
           branch="bot/template-rebake-$(date -u +%Y%m%d-%H%M%S)"
           git config user.email "actions@github.com"
--- a/scripts/check-rebake-content.mjs
+++ b/scripts/check-rebake-content.mjs
@@ -1,0 +1,71 @@
+#!/usr/bin/env node
+/**
+ * Content-diff gate for an auto-rebake of the bundled CC template (dario#990).
+ *
+ * cc-drift-template-watch.yml runs this between capture-and-bake.mjs and
+ * rebake-release-prep.mjs, in place of the `git diff --quiet` sanity check
+ * that step used to rely on. That check asked the right question — "did the
+ * bake actually change the bundle?" — of the wrong thing: `_captured` is
+ * restamped by every bake, so the file is always textually dirty and the gate
+ * could never fire.
+ *
+ * PR #990 is what that costs. A transient `--check` capture (whose base
+ * matched the dario#881 residue signature) opened a rebake PR whose
+ * src/cc-template-data.json was byte-identical to the previous release apart
+ * from `_captured` — carrying a patch bump and a CHANGELOG entry announcing a
+ * "wire-fingerprint drift" that was not in the diff. The tripwire warned a
+ * human to check; nothing stopped the PR from existing.
+ *
+ * Compares the freshly-baked working-tree bundle against the committed one
+ * key by key, ignoring TRANSIENT_TEMPLATE_FIELDS.
+ *
+ * Exits: 0 the bake ships real content (changed keys listed on stdout)
+ *        1 content-empty — caller should skip the PR
+ *        2 usage / git error (also treated as skip-worthy by the caller)
+ */
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { meaningfulTemplateKeys, TRANSIENT_TEMPLATE_FIELDS } from './drift-report.mjs';
+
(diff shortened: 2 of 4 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues — approving. One minor note on test coverage/doc accuracy.

## What I checked
- `gh pr diff` for all 4 changed files (workflow, new `check-rebake-content.mjs`, `drift-report.mjs` addition, test file).
- CI check-rollup: all 12 checks green (`test`, `build (18/20/22)`, `CodeQL` x2, `actionlint`, `docker-cap-drop-smoke`, `validate-package-json`, `triage`, `auto-merge` skipped as expected for a non-bot PR).
- Cloned read-only to confirm `capture-and-bake.mjs` doesn't obviously construct `header_values`/nested objects with varying key order across runs (couldn't find a `header_values` builder in that script to fully confirm, see minor note below).

## Findings

**Minor — `scripts/drift-report.mjs:626` / `test/bake-drift-report.mjs` (test 42, "nested key order is not drift")**

The docstring says:

```js
 * Top-level keys whose value differs between two baked templates, ignoring
 * `TRANSIENT_TEMPLATE_FIELDS`. Compared by canonical JSON so key order inside
 * a nested object never reads as a change.
```

but the implementation is:

```js
if (JSON.stringify(prev?.[k]) !== JSON.stringify(now?.[k])) changed.push(k);
```

`JSON.stringify` is **not** order-canonicalizing — it serializes object keys in insertion order, so two objects with the same content but different key order produce different strings and would be (incorrectly) reported as "changed." The added test that's supposed to cover this:

```js
check(
    'nested key order is not drift',
    meaningfulTemplateKeys(
      { ...v5517, header_values: { a: '1', b: '2' } },
      { ...v5518, header_values: { a: '1', b: '2' } },
    ).length === 0,
  );
```

passes both sides the *same* literal `{ a: '1', b: '2' }` — identical key order on both sides — so it doesn't actually exercise reordering and would pass whether or not the implementation canonicalizes. This isn't a blocking bug: worst case is a false-positive "meaningful change" (an unnecessary rebake PR opens), which is the old, safe-but-noisy failure mode this PR is explicitly trying to move away from, not a reintroduction of the #990 danger (silently shipping empty rebakes). I couldn't confirm from the diff/context whether `capture-and-bake.mjs` ever produces non-deterministic key order in practice, so I'm not treating this as a live correctness bug — just flagging the doc claim and the test that doesn't back it up.

```suggestion
// either: sort keys before JSON.stringify to make the claim true, e.g.
function canonical(v) {
  if (Array.isArray(v)) return v.map(canonical);
  if (v && typeof v === 'object') {
    return Object.fromEntries(Object.keys(v).sort().map(k => [k, canonical(v[k])]));
  }
  return v;
}
// then: JSON.stringify(canonical(prev?.[k])) !== JSON.stringify(canonical(now?.[k]))
// or: fix the docstring to say "content equality, not key-order-insensitive" and
// give the test two genuinely different key orderings, e.g. { a:'1', b:'2' } vs { b:'2', a:'1' }
```

## What's good
- The fix directly targets the root cause described in the PR body (the old `git diff --quiet` gate could never fire because `_captured` always changes) with a narrow, well-justified transient-field allowlist — deliberately excluding `_version`/label fields so label-sync rebakes still ship.
- `check-rebake-content.mjs` has clear exit-code semantics (0/1/2) and the workflow caller treats both 1 and 2 as skip-worthy, which is correct per its own docstring.
- Test 42 is otherwise thorough: covers added/removed keys, multi-key sorting, `_version` still-reports, and the exact #990 shape as a regression case.
- PR description documents an actual re-dispatch validation against the real #990 artifacts, not just fixtures.

PR #996, opened 2026-08-17
ci(health-watch): stop spending codeload budget on every probe run

askalf:
## Problem

`cc-oauth-health.yml` ("dario serving health alert") failed 8 consecutive times on 2026-08-17 between 13:30Z and 14:50Z. The error was **not** in the probe — it was in `Prepare all required actions`:

```
Download action repository 'askalf/checkout-with-retry@115a640'
##[warning]Failed to download action 'https://codeload.github.com/askalf/checkout-with-retry/tar.gz/115a640'.
  Error: Response status code does not indicate success: 429 (Too Many Requests).
##[warning]Back off 14.109 seconds before retry.
##[warning]... 429 ... Back off 28.397 seconds before retry.
##[error]Failed to download archive '...' after 3 attempts.
```

Failing runs: 32039258026, 32038687325, 32038473167, 32038097823, 32037470540, 32036866452, 32036790725, 32036211103.

## Root cause

This workflow is the highest-frequency job on the box's self-hosted runner, and **every run** pulled an action tarball from `codeload.github.com` — solely to sparse-checkout `scripts/` for one dependency-free POSIX-sh file. Two independent clocks drive it:

- `workflow_dispatch` **every 10 min** from platform's `tools/watch-kick.sh` (systemd timer on the box)
- this file's own `schedule: */5`, which GitHub's free-tier scheduler actually delivered about once per 30 min (13:49, 14:16, 14:49, 15:14, 15:44Z)

~8 runs/hr of tarball downloads saturates the per-runner codeload budget. Nothing else on that runner pulls this action more often than `*/30`.

Critically, **"add backoff" is not an available fix**: the runner's own action-download retry is internal, already does 3 attempts with 14s/28s backoff, and was exhausted in every failing run. There is no knob. The only fix is to stop spending codeload budget here.

## Fix

**1. Remove the codeload dependency (primary).** Fetch `scripts/health-verdict.sh` with one authenticated `gh api repos/.../contents/...` raw GET. Different, far larger budget (5000/hr authenticated), and a raw file is a fraction of a tarball. Pinned to `?ref=${{ github.sha }}`, so the verdict logic can never drift from the workflow dispatching it. Our own 3-attempt linear backoff — and unlike the silent-healthy failure mode this watcher is built to avoid, a failure to load the verdict logic exits 1 loudly.

Side benefit: retires the sparse-checkout cone-mode footgun documented in the old step's comment (`fatal: '...' is not a directory`, run #31239237095) — no cone mode, so we fetch just the one file instead of the whole directory.

**2. Consolidate the duplicate clock (secondary).** `*/5` → `*/30`. That matches the cadence GitHub was actually delivering, so this removes load without reducing real coverage. The cron is **deliberately kept, not deleted**: `watch-kick.sh` is a single point of failure, and if the box's systemd timer dies this is the fleet's last-resort auth watcher — nothing else would notice.

Two bugs caught while writing the new step, either of which would have broken it in CI:
- A job-level `permissions:` block **replaces** the workflow-level one, so `contents: read` had to be restated at job level or the contents API 403s.
- `[ "$attempt" -lt 3 ] && sleep ...` as the loop's last statement is fatal under the runner's default `bash -e` — a false test on the final pass becomes the loop's exit status and kills the step *before* the `::error::` annotation prints. Uses `if/fi`.

## Test plan

Verified locally before pushing:

- **Real fetch:** `gh api repos/askalf/dario/contents/scripts/health-verdict.sh?ref=5c6d03e -H 'Accept: application/vnd.github.raw'` → 2878 bytes.
- **Verdict unchanged from the temp path** (the behaviour that matters — this watcher's failure mode is silence):
  - healthy body → `healthy / true / ok / 0 / <empty axis>` → no alert
  - `"oauth":"broken"` → axis `OAuth broken` → alerts
  - empty body (container down) → axis `OAuth unreachable` → alerts
- **Retry loop under `bash -e`** with a stubbed-failing `gh`: printed all three attempts, then the `::error::` annotation, exit 1. Confirms the `if/fi` fix.
- `test/health-verdict.mjs` is untouched and still covers the script itself; `actionlint` (required check) validates the workflow YAML.
- First green scheduled/kicked run on `master` post-merge is the real confirmation.

Source ticket: `01M0867F4H2WH1TVRX9YKY8QQ6`

### DIFF
--- a/.github/workflows/cc-oauth-health.yml
+++ b/.github/workflows/cc-oauth-health.yml
@@ -39,8 +39,21 @@ name: dario serving health alert (self-hosted)
 # box updates; it must never page because a field it wanted wasn't there.
 
 on:
+  # TWO clocks, deliberately, with the cron as the thin backstop.
+  #
+  # The real cadence comes from platform's tools/watch-kick.sh — a systemd
+  # timer on the box that workflow_dispatches this file every 10 min, because
+  # GH's free-tier scheduler throttles hard (this `*/5` was observed firing
+  # roughly once per 30 min, not 6x/hr).
+  #
+  # It was `*/5` here, which on top of the 10-min kick meant two independent
+  # clocks racing for the same job. Thinned to `*/30` — its actual observed
+  # rate anyway — so the cron stops adding load while still covering the case
+  # watch-kick.sh cannot: if the box's timer dies (or the box does), this is
+  # the fleet's last-resort auth watcher and nothing else would notice. Not
+  # deleted for exactly that reason; watch-kick.sh is a single point of failure.
   schedule:
-    - cron: '*/5 * * * *'
+    - cron: '*/30 * * * *'
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


sprayberry-redline (dismissed):

**Verdict:** No blocking issues — clean, well-reasoned infra fix. Approving.

## What I checked
- `gh pr diff 996` (`.github/workflows/cc-oauth-health.yml`, `CHANGELOG.md`).
- CI rollup (`gh pr checks 996`): `validate-package-json` shows failed, but its log
  (`gh run view 32043986779 --log-failed`) shows the exact same
  `codeload.github.com ... 429 Too Many Requests` fetching `askalf/checkout-with-retry`
  that this PR's own description documents and is fixing for `cc-oauth-health.yml`
  specifically. `validate-package-json` isn't touched by this diff (it uses
  `checkout-with-retry` too, per `grep` across `.github/workflows/`), so this is a
  pre-existing shared-infra flake, not a regression introduced here. `actionlint`,
  `CodeQL`, `analyze`, `triage` all pass.
- Cloned the repo to confirm the job-level `permissions:` replacement claim and the
  pre-PR step order (`cc-oauth-health.yml:53-70` before the change).

## Correctness
- The retry loop (new step, `cc-oauth-health.yml`) correctly avoids the `bash -e`
  trap it calls out: `if [ "$attempt" -lt 3 ]; then sleep $((attempt * 10)); fi` as
  the loop's last statement, rather than `&&`-chaining a command whose exit status
  would become the step's exit status on the final iteration. Traced attempt=3
  failure path: falls through the `if`, hits `echo "::error::..."` then `exit 1` —
  correct.
- `contents: read` is correctly restated at job level
  (`cc-oauth-health.yml` job `probe`): a job-level `permissions:` block replaces
  rather than merges with the workflow-level one, so without this the new
  `gh api repos/$REPO/contents/...` call would run with only the job's prior
  `issues: write` and no read scope. Matches the PR description's claim.
- `verdict=... sh "$RUNNER_TEMP/health-verdict.sh"` is updated consistently with the
  new fetch destination (`$RUNNER_TEMP/health-verdict.sh`); no stale
  `$GITHUB_WORKSPACE` reference left behind.

## Minor (non-blocking)
- `gh api "repos/$REPO/contents/scripts/health-verdict.sh?ref=$REF" ... > "$dest" 2>/dev/null`
  swallows the actual API error text (rate-limit message, 404, etc.) on each
  attempt — only the generic `attempt $attempt: could not fetch...` line is logged,
  so diagnosing *why* a future failure happened will require re-running by hand.
  Not blocking; the loud `exit 1` still fires correctly.

## What's good
- Root-caused with actual run IDs and log evidence rather than a guess, and the fix
  (drop the tarball-download action entirely, use a separate-budget REST call) is a
  correct and durable resolution rather than a bandaid retry/backoff (which the PR
  correctly notes doesn't apply — the runner's own backoff is what's exhausted).
- `*/5` → `*/30` change is justified against the schedule's actually-observed
  delivery rate, and the cron is kept (not deleted) with a clear reason
  (`watch-kick.sh` single point of failure) — good judgment call, not a driveby cut.
- CHANGELOG entry is accurate and matches the diff.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-documented infra fix. Approving.

## What I checked
- `gh pr diff` for the full changed hunks in `.github/workflows/cc-oauth-health.yml` and `CHANGELOG.md`.
- CI check-rollup (`statusCheckRollup`): all 9 checks green — `actionlint`, CodeQL `analyze`, `validate-package-json`, `build (18/20/22)`, `test`, `docker-cap-drop-smoke`. `auto-merge` shows SKIPPED (expected, not a gating check).

## Findings
None blocking. A few things I specifically verified as *not* bugs, since this is exactly the kind of shell-under-`bash -e` change that tends to hide off-by-ones:

- `set -uo pipefail` (new fetch step) does not clear the runner's default `bash -e {0}`; `-e` stays active, which is why the retry loop correctly uses `if [ "$attempt" -lt 3 ]; then sleep ...; fi` instead of `&&`-chaining the sleep (a false test on the last iteration under `-e` would otherwise have killed the step before the `::error::` annotation printed, per the PR's own note about the bug it caught).
- The job-level `permissions:` block now restates `contents: read` alongside `issues: write` — correct, since a job-level `permissions:` block replaces rather than merges with the workflow-level one, and the new `gh api .../contents/...` call needs it.
- `$RUNNER_TEMP/health-verdict.sh` is used consistently: written by the new fetch step, read by the later probe step (`sh "$RUNNER_TEMP/health-verdict.sh"`), and `RUNNER_TEMP` persists across steps within the same job, so no path mismatch.
- No injection risk in the interpolated `$REPO`/`$REF` — both come from `github.repository`/`github.sha`, not user input.

## What's good
- Root-caused the actual failure (codeload action-tarball 429s, not the probe itself) and picked a fix with a genuinely separate rate budget (contents API vs codeload) rather than just adding backoff to a retry that was already exhausted.
- The comments explain *why*, including two bugs caught while writing it (permissions replacement, `-e`/if-fi), which is exactly the kind of context that prevents someone re-introducing them later.
- `CHANGELOG.md` updated with a matching, accurate summary.
- Manual test plan (real fetch, verdict-parity across healthy/broken/empty bodies, retry-under-`bash -e` with a stubbed failing `gh`) covers the behavior that matters for a silent-failure-prone watcher.

Issue #993, opened 2026-08-17
dario HA mode


ramarro123:
in a kubernates env it will be great to have dario in HA mode, having 2 pod that share the same config.

what option we have to do that? (if it's possible at all)

what if i share a folder on a nas containing the config for dario, will two darioS be able to read or concurrency error will likely corrupt file? (appened to me all the time to be honest if the app isn't designed to lock or to concurrenctly access to that)

do you think it will be useful to think to some way to make dario HA?

thanks!

askalf:
Looked into this — a shared config file itself is fine (`config-file.ts` writes via temp-file + fsync + rename, so no torn writes), but the account pool's *runtime* state is all in-process memory, and that's what actually breaks with two replicas:

- **Rate-limit tracking is per-pod.** `pool.ts` keeps `accounts: Map<string, PoolAccount>` privately, updated only from responses that pod itself saw. Two pods sharing the same account files over NAS would each think an account has headroom and could both route to it at once, blowing past the real 5h/7d limit that neither pod can see the other consuming.
- **Session stickiness is per-pod too.** The stickyKey is a deterministic hash of the conversation's first message, so both pods compute the same key — but the *binding* of that key to an account lives in a private `sticky: Map` on whichever pod first saw it. Without sticky routing at the LB, turn 2 of a conversation can land on the other pod, which has no record of the binding and may pick a different account, breaking prompt-cache continuity mid-conversation.
- **The one that actually breaks accounts: OAuth refresh-token rotation.** Anthropic invalidates the previous `refresh_token` on every refresh. `accounts.ts` guards against a process racing *itself* with a per-alias single-flight map, but that map is explicitly scoped in-process — nothing stops two independent pods from both firing their own 15-min background refresh for the same shared account at the same time. Whichever refresh lands second uses an already-rotated-away token and the account needs full re-auth.

So a shared NAS mount alone gets you shared credentials, not HA — the failure mode is silent until two pods refresh the same account within the same window, which will happen eventually on a 15-min timer with 2 replicas.

Real HA here would need either active/passive (one pod live, one cold standby that only takes over on failure) or promoting the pool's in-memory state — rate limits, stickiness, and especially the refresh single-flight — to a shared store like Redis with a single elected refresher per account. Happy to think through the shared-store shape if there's interest, but wanted to flag the refresh-token race specifically since that's the one that actually locks an account out rather than just degrading routing.

ramarro123:
[...]

PR #1000, opened 2026-08-18
feat: distributed refresh lock via Cloudflare Durable Object (#993)

askalf:
## What

Closes the actual blocker behind #993 ("HA mode"). The pool's per-alias single-flight guard in `accounts.ts` only ever protected **one process** against itself. Two separate dario instances refreshing the same account could still race — and since Anthropic invalidates the previous `refresh_token` on every refresh, whichever instance's request landed second used an already-rotated-away token and needed full re-auth.

**A bare lock isn't enough by itself.** Two instances refreshing the same account *serialized but back-to-back* would still break — the second one's `refresh_token` was already burned by the first's refresh before it ever got a turn. So `/acquire` doesn't just block: a caller that loses the race gets the **winner's fresh credentials** back directly and adopts them, skipping its own (guaranteed-stale) refresh attempt.

New `cloudflare/refresh-lock/` — a small Worker + Durable Object, one instance per account alias (via `idFromName`), so state is strongly consistent regardless of pod count. Verified live end-to-end against the **real deployed Worker** (not mocked):
- acquire while held → wait
- release with credentials → cached
- a caller arriving *after* release → does **not** get `acquired:true`, gets the fresh credentials instead (this was the actual bug in my first draft — a bare "is the lock free" check misses the more common case of arriving just after the winner already finished)
- wrong-holder release → `409`
- bad token → `401`

## Safety

Optional and additive: unset `DARIO_REFRESH_LOCK_URL` and `doRefreshAccountTokenDistributed` is byte-identical to `doRefreshAccountToken` before this existed (test-verified: zero lock traffic, single fetch straight to the oauth endpoint). **Fails open** on any lock-service error — bad response, network error, timeout — so a Cloudflare outage falls back to today's in-process-only behavior rather than making dario's own refresh depend on a new external service.

## Test plan

- [x] 16 new tests (`test/account-refresh-distributed-lock.mjs`), same monkeypatched-fetch convention as the existing single-flight test — unset / acquired / adopt-fresher / fail-open / release-on-failure paths
- [x] Live-verified against the real deployed Worker (7 cases: acquire, contend, release, late-acquire-adopts-fresh, wrong-holder-409, bad-token-401)
- [x] `npm run build` clean
- [ ] ⚠️ **3 pre-existing failures, unrelated** — `template-config-scoped-tools.mjs` / `template-interactive-tools.mjs` / `tool-advertise-respects-client.mjs` fail identically under `--test-concurrency=8` **with or without this change** (confirmed by isolating: same 3 fail on a clean master rebuild, and again with this PR's test file physically removed from the tree). A real flake in the template subsystem, not introduced or worsened here — flagging rather than silently ignoring, but out of scope to fix in this PR.

## Deploy (when ready to activate)

```
cd cloudflare/refresh-lock && wrangler secret put LOCK_TOKEN && wrangler deploy
```
Then set `DARIO_REFRESH_LOCK_URL` + `DARIO_REFRESH_LOCK_TOKEN` on every dario instance sharing an account. Landing the code here does **not** activate it anywhere — that's a separate, deliberate step.


sprayberry-redline (approved):
**Automated review** (posted while the fleet's Code Reviewer is queued behind the Max-subscription cap — hand-driven this cycle, same identity/process either way).

This one deserves a closer read than usual since it changes credential-refresh behavior, so I traced the actual acquire/release logic rather than trusting the description.

- **The core correctness property is right, and it's the subtle one.** `doRefreshAccountTokenDistributed` doesn't just gate on "is the lock free" — every `acquire` call compares `currentExpiresAt` against the DO's cached credentials *before* even checking lock state (`cloudflare/refresh-lock/worker.mjs` lines ~42-55). That's what makes it correct for a caller arriving *after* the winner already released, not just one that arrives mid-refresh: without that check, a late arrival would see the lock free, acquire it, and redundantly refresh with an already-invalidated `refresh_token` — Anthropic invalidates the previous token on every refresh, so that path would silently reintroduce the exact bug this PR exists to fix. The PR body says this was caught and fixed during the author's own testing (first draft only checked cached creds while the lock was held) — verified the CURRENT code has the fix, not just the claim.
- **Safety invariant holds.** `refreshLockUrl()` returning `null` short-circuits straight to `doRefreshAccountToken(creds)` — confirmed via the "unset DARIO_REFRESH_LOCK_URL" test, which asserts exactly one fetch call and that it hits `/oauth/token`, not a lock endpoint. No behavior change for anyone not opting in.
- **Fails open, correctly.** Both the network-error path (`lockCall` catches and returns `null`) and the "gave up after 8 attempts" path fall back to a direct refresh rather than blocking. A Cloudflare outage degrades to today's behavior, doesn't add a new dependency dario's core function needs.
- **Failure-path release is right.** On a refresh throw, `release` is called *without* `credentials` — confirmed by the test asserting `!('credentials' in release.body)`. Caching a failure as if it were a fresh credential would be worse than not caching anything.
- One thing I'd flag, not blocking: the 3 template-tool test failures noted in the PR description are called out and the author verified them independently (same failures on a clean master rebuild, same failures with this PR's test file physically removed) — that's the right level of rigor for "is this pre-existing," and CI itself came back green regardless.

No blocking findings. Approving.

PR #1006, opened 2026-08-18
test: live dual-process integration proof for the refresh lock (#993)

askalf:
## What

The reviewer (this same reviewer, on #1000) correctly pushed back on treating #993 as solved off unit tests + isolated curl checks against the deployed Worker alone — nobody had actually run the failure scenario end-to-end: two genuinely separate dario processes racing to refresh the same account.

`test/integration/dual-instance-race.mjs` does that live:
- Two isolated `~/.dario` homes (per-process `HOME`/`USERPROFILE` override) — nothing shared between the two workers but the remote lock service.
- Real network calls to the **real deployed** `dario-refresh-lock` Cloudflare Worker.
- A local mock stands in **only** for Anthropic's token endpoint — hammering the real one with production credentials for adversarial testing risks actually burning the operator's live refresh token. Everything else is real, not simulated.
- The mock enforces the one behavior this whole fix exists to handle: a `refresh_token` is single-use, reusing an already-consumed one `400`s exactly like the real API (`invalid_grant`).

## Results

**Control run (`--no-lock`)** — proves the harness is meaningful before trusting a positive result: one worker gets a genuine `invalid_grant` rejection. That's the actual #993 bug, reproduced live, not assumed.

**With the real lock: 6/6 runs held.** Both workers succeed every time, exactly one real refresh reaches "Anthropic" each run, zero rejections.

One real bug in the harness itself along the way: the first 3 reruns showed `real refreshes = 0` and I nearly mis-read that as a failure — turned out the DO's own credential cache (intentional, 5-minute TTL) was correctly serving the *prior* run's cached credentials on a same-alias rerun, never touching the lock or the mock endpoint at all. Fixed by using a fresh random alias per run; documented in `test/integration/README.md` so it doesn't get re-discovered as a false alarm later.

## Test plan

- [x] Control run reproduces the real #993 bug (1/1)
- [x] Fixed run holds (6/6, after the alias-cache fix)
- [x] Not part of `npm test` (needs real network + secret) — same pattern as `e2e.mjs`/`compat.mjs`, own entry point: `npm run test:refresh-lock-race`
- [x] Full suite: 134/137 passing — the same 3 pre-existing `template-*`/`tool-advertise-respects-client` flakes already documented on #1000, unrelated to this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+- **Live integration test for the refresh lock (dario#993), and the finding that prompted it.** The unit tests + live curl checks against the deployed Worker proved the lock's API contract and dario's client code in isolation, but nobody had run two genuinely separate dario processes racing a real refresh end-to-end. `test/integration/dual-instance-race.mjs` does: two isolated `~/.dario` homes, real network calls to the real deployed Worker, a local mock standing in only for Anthropic's token endpoint (hammering the real one with production creds for adversarial testing risks burning the operator's live refresh token). Control run confirms the harness is meaningful — without the lock, one worker gets a genuine `invalid_grant`. With it: 6/6 runs held, both workers succeed, exactly one real refresh reaches "Anthropic" each time.
+
 ## [5.5.20] - 2026-08-18
 
 - **Template rebake** — re-captured `src/cc-template-data.json` after cc-drift-template-watch detected wire-fingerprint drift (+51 net chars, benign) against a live CC capture. Bundled fallback template now matches the current CC wire shape.
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "e2e": "node test/e2e.mjs",
     "stress": "node test/stress.mjs",
     "compat": "node test/compat.mjs",
+    "test:refresh-lock-race": "node test/integration/dual-instance-race.mjs",
     "lint:pkg": "node scripts/check-package-json.mjs",
     "drift:sdk": "node scripts/check-sdk-drift.mjs",
     "drift:wire": "node scripts/check-wire-drift.mjs",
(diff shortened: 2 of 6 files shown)


sprayberry-redline (approved):
**Automated review** (posted while the fleet's Code Reviewer is queued behind the Max-subscription cap — hand-driven this cycle, same identity/process either way).

This is the right response to the earlier gap flagged on #1000 — that PR's tests validated the lock's pieces in isolation, not the actual two-process failure scenario #993 describes. Read the harness itself, not just the PR description, since a test proving nothing would be worse than no test.

- **The control run is what makes this trustworthy.** `--no-lock` has to actually fail before the "it works" claim on the real run means anything — checked that the mock (`mock-anthropic-oauth.mjs`) genuinely enforces single-use refresh tokens (tracks the current valid token per alias, 400s `invalid_grant` on reuse) rather than trivially always succeeding. That's the one behavior the whole test depends on being real.
- **Isolation is real, not simulated.** `refresh-worker.mjs` runs as an actual separate `node` process via `spawn`, not two async calls sharing a process — the per-process `HOME`/`USERPROFILE` override means each worker's `~/.dario/accounts/` is genuinely independent. The only thing they share is the deployed Cloudflare Worker, which is the actual point.
- **The self-caught bug is a good sign, not a concern.** The PR body says the first 3 reruns misleadingly showed `real refreshes = 0` before the fix (unique alias per run) — verified this is exactly what you'd expect from the DO's own 5-minute credential cache correctly serving a same-alias rerun, not a flaw in the lock. Documented in the README so a future reader doesn't rediscover it as a false alarm.
- **Scope is right.** Correctly excluded from `npm test` (top-level `readdirSync` in `all.test.mjs` doesn't recurse into `test/integration/`) — confirmed this rather than trusting the claim, since a live-network test silently joining the default CI run would be its own bug.

No blocking findings. Approving.

PR #1005, opened 2026-08-18
auto-rebake: template drift detected 2026-08-18

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-18T14:17:20+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.19` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** ✅ Benign

- **system_prompt:** +51 chars net (text-content drift — see unified diff below)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.234) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.234, 24 tools, 6206 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6206 → 4802 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.233 captured 2026-08-15T04:59:39.819Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4802 → fable 9271 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4802 → opus-5 8137 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4802 → sonnet-5 13397 chars
[bake] check: drift detected — 1 differing slot (verdict: benign):
[bake] **Verdict:** ✅ Benign
[bake] 
[...]

PR #1001, opened 2026-08-18
auto-rebake: template drift detected 2026-08-18

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-18T05:05:20+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.19` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — per-model system-prompt variants only

- **system_prompt_variants.fable:** 9220 → 9271 chars (base system prompt, tools, and anthropic_beta unchanged)
- **system_prompt_variants.opus-5:** 8086 → 8137 chars (base system prompt, tools, and anthropic_beta unchanged)
- **system_prompt_variants.sonnet-5:** 13346 → 13397 chars (base system prompt, tools, and anthropic_beta unchanged)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.234) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.234, 24 tools, 6153 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6153 → 4751 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.233 captured 2026-08-15T04:59:39.819Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4751 → fable 9271 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4751 → opus-5 8137 chars
[...]

PR #1007, opened 2026-08-18
ci: content-labeler also assigns askalf when nothing is assigned yet

askalf:
Extends the content-labeler workflow (triage) to also auto-assign \`askalf\` when an issue or PR has no assignee yet, same trigger and same bot-skip guard as the labeling pass. Runs unconditionally relative to the label step (assignment happens regardless of whether any keyword label matched).

### DIFF
--- a/.github/workflows/content-labeler.yml
+++ b/.github/workflows/content-labeler.yml
@@ -1,15 +1,25 @@
-name: Content-label issues and PRs
+name: Triage issues and PRs
 
-# Keyword labels from title + body — the thing labeler.yml (path-based)
-# cannot do, since an issue has no changed files and a PR's *description*
-# of what it fixes doesn't necessarily overlap the paths it touches.
+# Two additive passes, on every issue and PR:
 #
-# Additive only: never removes a label, never touches one a human already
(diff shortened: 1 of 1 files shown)


sprayberry-redline (approved):
Reviewed the diff directly.

The restructure is sound: assignment now runs unconditionally (including for bot-authored items like the auto-rebake PRs from `cc-drift-template-watch[bot]`), while the bot-skip guard moves down to gate only the keyword-labeling pass — which is exactly right, since #1001 and #1005 both sat without an owner until someone went looking, and a rebake PR is precisely the kind of thing that most needs a human to notice and merge it.

Checked the two failure paths:
- `addAssignees` is wrapped in try/catch with a log-and-continue — a fork PR's read-only token or an invalid assignee won't block the labeling pass underneath it. Correct, matches the existing fork-guard pattern already in this file.
- The `target.assignees.length === 0` check is additive-only, as documented — never reassigns something a human already picked.

CI (Content-label issues and PRs, Label PRs, actionlint, CodeQL) is green on this branch. No concerns.

Approving.

PR #1008, opened 2026-08-18
feat: Redis-backed alternative to the Cloudflare refresh lock (#993)

askalf:
Second reference implementation of the exact same \`/lock/<alias>/acquire|release\` contract as \`cloudflare/refresh-lock\`, for operators who can't or don't want a Cloudflare dependency in the loop (airgapped environments, or already running Redis) — see the discussion on #993.

\`src/accounts.ts\` needs zero changes: it doesn't know which backend is behind \`DARIO_REFRESH_LOCK_URL\`, so this is a drop-in swap. Zero new runtime dependencies — \`redis-lock/resp-client.mjs\` is a minimal hand-rolled RESP2 client over \`node:net\`, not the \`redis\`/\`ioredis\` packages.

**Verified at the same rigor as the Cloudflare version:**
- 7-case contract test against a live Redis instance (acquire/release, contested acquire, late adoption of fresh credentials after release, wrong-holder release → 409, bad-token → 401) — all matched the Cloudflare Worker's behavior exactly.
- \`test/integration/dual-instance-race.mjs\` (the same live dual-process test used to validate the Cloudflare backend) pointed at this server instead: 6/6 clean runs, exactly one real refresh reaching the OAuth endpoint per run, the other worker adopting the winner's fresh credentials.

See \`redis-lock/README.md\` for deploy instructions.


sprayberry-redline (approved):
Reviewed the diff and the verification directly.

`redis-lock/` is a clean structural mirror of `cloudflare/refresh-lock`: same two-part correctness property (checked cache before checking lock state, unconditionally — this is the exact bug the Cloudflare version's first draft had, fixed here from the start by checking `creds:<alias>` before the lock regardless of whether it's held), same auth shape, same `/lock/<alias>/acquire|release` contract. `src/accounts.ts` genuinely needs zero changes to point at this instead — confirmed by reading `doRefreshAccountTokenDistributed()`, which only knows `DARIO_REFRESH_LOCK_URL`/`DARIO_REFRESH_LOCK_TOKEN`, not which backend answers them.

`resp-client.mjs`: hand-rolled RESP2 parser handles all the reply types this needs (simple string, error, integer, bulk string incl. nil, array incl. nested for EVAL) with correct incomplete-buffer detection (`parseReply` returns `null` and `_onData` waits for more bytes rather than throwing on a partial frame). FIFO queue matching is correct for RESP2's in-order-reply guarantee. No new runtime deps, consistent with dario's invariant.

`server.mjs`: the Lua release script (`GET`+conditional `DEL`) closes the same "don't delete someone else's lock after your lease already expired and got reassigned" hole the Cloudflare version handles — correct use of `EVAL` for atomicity here since a bare GET-then-DEL from JS would race.

Verification is the same bar as the Cloudflare PR: 7-case contract test against live Redis (matches Cloudflare's behavior exactly, including the late-adoption case), plus `dual-instance-race.mjs` — the same live dual-process harness, unmodified — pointed at this backend: 6/6 clean, one real refresh per run. Full `npm test` also clean (137/137).

Approving.

sprayberry-redline (dismissed):
Re-review after the branch-update merge commit (dismissed the prior approval per this repo's ruleset). Diffed `master...redis-refresh-lock`: the only new content since my last approval is #1007's unrelated `content-labeler.yml`/`CHANGELOG.md` changes pulled in by the merge — `redis-lock/` itself is byte-for-byte unchanged. Prior review stands. CI green.

Approving.

sprayberry-redline (approved):
Re-review after two fix commits. Both address real CodeQL findings, not false positives:

1. `redis-lock/server.mjs:45` (via the 500 handler) — the catch-all error handler was echoing the raw exception message straight back to the caller, which could leak internal details (Redis host/connection info, file paths) to anything hitting this endpoint. Now logs the full error server-side and returns a generic `{"error":"internal error"}` to the client.
2. `redis-lock/server.mjs:108` — the fix in (1) put `req.url` (attacker-controlled) into the format-string position of `console.error`, which CodeQL correctly flagged as an externally-controlled format string: a URL containing `%s`-style content could corrupt the logged output via `util.format`'s implicit substitution. Fixed by keeping the format string static and passing request data as a plain object argument instead.

Both fixes are error-path-only — no behavior change to any of the 7 verified contract cases, none of which exercise this path. CodeQL clean on the latest commit.

Approving.

PR #1013, opened 2026-08-19
docs: answer the suspension question directly in the README

askalf:
**Why:** "Will my account get suspended?" is the question that decides whether a visitor actually tries dario. The FAQ answered the adjacent *legal* framing ("does this violate Anthropic's terms?"), but not the *anxiety* framing — and the better, fuller answer was buried in [discussions/724](https://github.com/askalf/dario/discussions/724), where neither search nor an LLM summarizing the project is likely to weight it.

This promotes it to its own section in the trust cluster (right after "Honest about what this is"), with its own anchor so it can be linked and cited directly, and cross-links it from the existing FAQ entry so both entry points land on it.

**Framing is unchanged from the #724 answer** — deliberately: no promises, no reassurance I can't back, just a precise account of what dario does and doesn't do so the reader can weigh it themselves. It states plainly that letting non-CC tools use the subscription is the part that sits outside Anthropic's own client, and points at their terms and DISCLAIMER.md rather than hand-waving it.

Both external links verified live (200). Both in-page anchors verified — `#the-billing-split--a-contingency-dario-is-built-for` is already used elsewhere in the README, and `#trust--transparency` follows the same GitHub anchor rule.

Docs-only; no version bump.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,26 @@ dario uses your own subscription credentials, authenticates you as you, and impe
 
 ---
 
+## Will my account get suspended?
+
+The most common question about dario, and it deserves a straight answer: **I can't promise you won't be actioned, and I'd be skeptical of anyone who does.** Only Anthropic decides how it enforces its terms. What I can do is lay out exactly how dario works, so you can weigh the risk yourself instead of taking anyone's word for it.
+
+**What dario does:**
+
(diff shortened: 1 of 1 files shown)


sprayberry-redline (approved):

Issue #1014, opened 2026-08-19
compat goes green against a stale template after a CC upgrade — maxTested bumps land unvalidated


askalf:
Found while reviewing #1012. Two lags stack, and the result is that a `maxTested` bump can ship claiming validation that never happened — while CI is green the whole way.

## The two lags

**1. npm → runner.** `cc-drift-watch.yml` reads `@anthropic-ai/claude-code@latest` from the **npm registry** and drafts a `maxTested` bump for that version. The `compat` job runs on the self-hosted box against whatever CC is **installed there**. Nothing syncs them. In #1012 the watcher drafted a bump to `v2.1.235` while the box still had `v2.1.234`.

**2. binary → template cache.** Even after upgrading the box's CC, the *first* compat run still validates the **previous** version, because dario loads the cached capture into memory and refreshes it in the background. Its own log says so:

```
[dario] template: live capture, CC v2.1.234 (12h old), variants: fable+opus-5+sonnet-5
[dario] ⚠  template drift: cache is from CC v2.1.234 but installed CC is v2.1.235
        — background refresh will re-capture
```

That run **passed**. A second re-run, after the background re-capture had written `_version: "2.1.235"` to `~/.dario/cc-template.live.json`, was the first one to genuinely exercise the new version.

## Why it matters more than the patch delta suggests

`SUPPORTED_CC_RANGE.maxTested` is documented in `src/live-fingerprint.ts` as meaning we validated against that version, and its whole job is to *signal when to revalidate*:

> Update `maxTested` every time we validate against a new CC … Above this, dario is likely fine but it's explicitly untested, so users get a soft warn **and we get a signal to refresh the bundled snapshot + rerun e2e**.

Bumping it unvalidated tells users a version was tested when it wasn't, and silences the alarm that exists to prompt the testing. **A green compat that tested the wrong template is a worse signal than a red one** — it's the drift alarm reporting all-clear while looking at yesterday's data.

## Suggested fixes

- **Hard-fail compat on template drift** instead of warning. If the cache version and the installed binary disagree, the run is not a valid test of either — it should refuse rather than pass.
- **Or force a synchronous re-capture** before the compat suite starts, so the first run after an upgrade is already correct.
- **Gate the `maxTested` bump** on compat having actually observed that exact version, rather than on the npm registry's `latest`.
- Optionally, have `cc-drift-watch.yml` upgrade CC on the runner as an explicit pipeline step, so the two can't drift silently.

## Related timing note

[...]

PR #1015, opened 2026-08-19
chore(template): label refresh _version -> v2.1.235 (clears sdk-drift)

askalf:

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-19T03:30:08+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.235** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.235`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.235`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.5.22` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.235. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.


sprayberry-redline (approved):
Reviewed the diff.

Label-only refresh, and the diff confirms it — three fields move and nothing else:

```
_version:            2.1.234 → 2.1.235
user-agent:          claude-cli/2.1.234 → claude-cli/2.1.235
_supportedMaxTested: 2.1.234 → 2.1.235
```

`_captured` is correctly left alone, which is the right call: the changelog entry states `capture-and-bake --check` ran against live CC v2.1.235 and found zero shape drift vs the bundle, so this is a label catch-up, not a re-capture. Touching `_captured` would falsely imply a fresh capture happened.

Worth recording explicitly, because it retroactively validates #1012: **CC v2.1.235's wire shape is identical to v2.1.234's.** The `maxTested` bump that shipped in 5.5.21 was substantively safe — the process that let it merge unverified was the problem, not the value (see #1014, and #1016 for the fix).

Checked compat on this branch: green, and it ran against a 23s-old live capture of CC v2.1.235 — a genuinely fresh template, not the stale-cache case #1014 describes.

One cosmetic nit, not blocking: no blank line between the new `[5.5.22]` bullet and the `## [5.5.21]` heading beneath it. CommonMark still renders the heading correctly; it's just inconsistent with the spacing everywhere else in the file. Not worth a round-trip on an auto-generated PR.

Approving.

sprayberry-redline (approved):
Reviewed the diff.

Label-only refresh, and the diff confirms it — three fields move and nothing else:

```
_version:            2.1.234 → 2.1.235
user-agent:          claude-cli/2.1.234 → claude-cli/2.1.235
_supportedMaxTested: 2.1.234 → 2.1.235
```

`_captured` is correctly left alone, which is the right call: the changelog entry states `capture-and-bake --check` ran against live CC v2.1.235 and found zero shape drift vs the bundle, so this is a label catch-up, not a re-capture. Touching `_captured` would falsely imply a fresh capture happened.

Worth recording explicitly, because it retroactively validates #1012: **CC v2.1.235's wire shape is identical to v2.1.234's.** The `maxTested` bump that shipped in 5.5.21 was substantively safe — the process that let it merge unverified was the problem, not the value (see #1014, and #1016 for the fix).

Checked compat on this branch: green, and it ran against a 23s-old live capture of CC v2.1.235 — a genuinely fresh template, not the stale-cache case #1014 describes.

One cosmetic nit, not blocking: no blank line between the new `[5.5.22]` bullet and the `## [5.5.21]` heading beneath it. CommonMark still renders the heading correctly; it's just inconsistent with the spacing everywhere else in the file. Not worth a round-trip on an auto-generated PR.

Approving.

PR #1016, opened 2026-08-19
ci(compat): capture the live template before boot, and fail closed on drift

askalf:
Fixes the verification gap from #1014, found while reviewing #1012.

## The bug

`cc-template.ts` loads `~/.dario/cc-template.live.json` **synchronously at module init**, while the drift-triggered refresh is a **background** task whose result only lands for the *next* startup. So the first compat run after any CC upgrade on the runner reads the previous version's template, logs a warning, and **passes anyway**:

```
```

That run went green. It is how #1012 merged a `maxTested` bump to v2.1.235 whose only passing compat run had exercised **v2.1.234's** wire shape. The bump turned out to be fine — a later re-run confirmed v2.1.235 passes — but nothing in CI knew that at merge time.

## The fix — two layers

**1. Capture synchronously before boot.** `refreshLiveFingerprintAsync({ force: true })` runs as its own step, so the template on disk matches the installed binary before the proxy reads it. This makes the run **self-healing**: no manual re-run needed after a CC bump, and no silent-stale window. If the capture fails, the step exits 1 rather than proceeding on an unknown template.

**2. `--strict-template` as the backstop.** If the template is still drifted, or fell back to the bundled snapshot, the proxy now refuses to boot instead of serving a stale shape. The existing health-check failure path already `cat`s `proxy.log`, so the refusal message surfaces directly in the run.

The ordering matters: layer 1 makes the common case correct and automatic; layer 2 makes the failure case *loud* instead of green.

## Verified

- `--strict-template` is parsed (`src/cli.ts:449`) and threaded to `startProxy`.
- `refreshLiveFingerprintAsync` is exported from `dist/live-fingerprint.js`.
- Ran the exact capture invocation locally — captured cleanly, exit 0:
  ```
  [dario] live fingerprint refreshed from CC 2.1.235
  [compat] captured live template from CC 2.1.235
  ```
- The health-check failure branch dumps `proxy.log`, so a strict-template refusal is visible rather than surfacing only as a generic 30s timeout.

**This PR's own compat run is the real test** — it exercises the new capture step and the strict gate on the runner.

Refs #1014


sprayberry-redline (approved):
Reviewed the diff and the evidence.

This fixes the gap in #1014 — compat passing while exercising the previous CC version's template — and the diagnosis in the PR body matches what the code actually does: `cc-template.ts` reads the live cache synchronously at module init, while `refreshLiveFingerprintAsync` is a background task whose result only lands for the next startup. So the first run after a CC upgrade genuinely does test the old shape and pass.

The two layers are correctly ordered and neither is redundant:

- The forced synchronous capture makes the common case right *and* self-healing — no manual re-run after a CC bump. It exits 1 if the capture returns null, so a failed capture can't silently fall through to a stale template.
- `--strict-template` is a real backstop rather than decoration: it covers both drift and the bundled-snapshot fallback, which the capture step alone wouldn't catch if CC vanished from PATH.

Verified the mechanics rather than trusting the description:
- `--strict-template` is parsed at `src/cli.ts:449` and threaded through to `startProxy` — not a no-op flag.
- `refreshLiveFingerprintAsync` is exported from `dist/live-fingerprint.js`.
- The existing health-check failure branch already `cat`s `proxy.log`, so a strict-template refusal surfaces as its own message instead of an opaque 30s timeout. Good — that failure mode would otherwise be very confusing to debug.

**The proof is this PR's own compat run**, which exercised both layers on the runner:

```
[compat] captured live template from CC 2.1.235
[dario] template: live capture, CC v2.1.235 (9s old), variants: fable+opus-5+sonnet-5
```

9 seconds old, versus the 12-hours-stale capture that caused #1014. No `template drift` warning, and the proxy booted under `--strict-template`. All checks green: CI, CodeQL, compat, actionlint.

Approving.

sprayberry-redline (approved):
Reviewed the diff and the evidence.

This fixes the gap in #1014 — compat passing while exercising the previous CC version's template — and the diagnosis in the PR body matches what the code actually does: `cc-template.ts` reads the live cache synchronously at module init, while `refreshLiveFingerprintAsync` is a background task whose result only lands for the next startup. So the first run after a CC upgrade genuinely does test the old shape and pass.

The two layers are correctly ordered and neither is redundant:

- The forced synchronous capture makes the common case right *and* self-healing — no manual re-run after a CC bump. It exits 1 if the capture returns null, so a failed capture can't silently fall through to a stale template.
- `--strict-template` is a real backstop rather than decoration: it covers both drift and the bundled-snapshot fallback, which the capture step alone wouldn't catch if CC vanished from PATH.

Verified the mechanics rather than trusting the description:
- `--strict-template` is parsed at `src/cli.ts:449` and threaded through to `startProxy` — not a no-op flag.
- `refreshLiveFingerprintAsync` is exported from `dist/live-fingerprint.js`.
- The existing health-check failure branch already `cat`s `proxy.log`, so a strict-template refusal surfaces as its own message instead of an opaque 30s timeout. Good — that failure mode would otherwise be very confusing to debug.

**The proof is this PR's own compat run**, which exercised both layers on the runner:

```
[compat] captured live template from CC 2.1.235
[dario] template: live capture, CC v2.1.235 (9s old), variants: fable+opus-5+sonnet-5
```

9 seconds old, versus the 12-hours-stale capture that caused #1014. No `template drift` warning, and the proxy booted under `--strict-template`. All checks green: CI, CodeQL, compat, actionlint.

Approving.

PR #1022, opened 2026-08-19
chore(cc-drift): v5.5.23 — maxTested → v2.1.236

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.236 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.235` → `2.1.236` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.23`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.23] - 2026-08-19` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.236 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.235). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.235; npm latest is v2.1.236. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.236 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.236 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.236. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.236" step — no local run needed.)
- ✅ **Auto-merges** the moment the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`). `master` requires no review, so no human approval gates it.
[...]

PR #1024, opened 2026-08-19
ci(compat): fail closed when maxTested claims a CC the runner cannot back

### LINKED ISSUE #1014: compat goes green against a stale template after a CC upgrade — maxTested bumps land unvalidated

askalf:
Closes the **second** cause in #1014. #1016 fixed the stale-template lag; this fixes the version lag that survived it.

## The gap

`cc-drift-watch.yml` drafts a `maxTested` bump from the **npm registry's** `@latest`. `compat` runs against whatever CC is installed on the **self-hosted runner**. Nothing syncs them. When npm is ahead of the box, the PR claims support for a version nothing ever loaded — and compat goes green anyway.

This has now shipped twice:

| PR | claimed | compat actually validated |
|---|---|---|
| #1012 | v2.1.235 | v2.1.234 |
| #1022 | v2.1.236 | v2.1.235 |

Both were caught only because a human read the log line. That is not a control.

`#1016` was necessary but not sufficient — it guarantees the template is *fresh*, not that it is the *right version*. Worth being precise about that: on #1022 the capture step worked exactly as designed and still let an unvalidatable claim through, because freshness and correctness are different properties.

## The fix

After the forced capture, compare the captured CC version against `SUPPORTED_CC_RANGE.maxTested` and exit 1 if the runner is behind:

```
::error::compat cannot validate maxTested=2.1.240 — the runner captured CC 2.1.236.
Upgrade CC on the runner (npm i -g @anthropic-ai/claude-code@2.1.240) and re-run,
or lower maxTested.
```

The error names the exact remediation, so the next occurrence is a one-line fix rather than an investigation.

**Why fail closed:** a green compat that validated an *older* CC than the PR claims is worse than a red one. It is an affirmative "we tested this" attached to a version that was never exercised — precisely the signal `maxTested` exists to carry, inverted.

Reuses the already-exported `compareVersions` rather than reimplementing semver comparison in shell.

## Verified

Both paths, end to end, against the real capture pipeline:

```
# match (runner 2.1.236, claim 2.1.236)
[compat] captured live template from CC 2.1.236
[compat] maxTested gate ok: claimed 2.1.236 <= installed 2.1.236
exit=0

# mismatch (runner 2.1.236, claim 2.1.240)
::error::compat cannot validate maxTested=2.1.240 — the runner captured CC 2.1.236 …
exit=1
```

Truth table also checked: equal → pass, runner newer → pass, runner older → fail.

Closes #1014

### DIFF
--- a/.github/workflows/compat-test-self-hosted.yml
+++ b/.github/workflows/compat-test-self-hosted.yml
@@ -160,6 +160,30 @@ jobs:
               process.exit(1);
             }
             console.log('[compat] captured live template from CC ' + t._version);
+
+            // maxTested gate (dario#1014, second cause).
+            //
+            // The capture above fixes a STALE template. It does not fix the
+            // other lag: cc-drift-watch.yml drafts a maxTested bump from the
+            // npm registry's @latest, while this job runs against whatever CC
+            // is installed on the runner. When npm is ahead of the box, a PR
+            // claims support for a version nothing ever exercised — and compat
+            // goes green anyway. That shipped twice (#1012 -> v2.1.235,
+            // #1022 -> v2.1.236), both times caught only by a human reading
+            // the log.
+            //
+            // A green compat that validated an OLDER CC than the PR claims is
+            // worse than a red one: it is an affirmative 'we tested this'
+            // on a version that was never loaded. Fail closed instead.
+            const installed = t._version;
+            const claimed = m.SUPPORTED_CC_RANGE.maxTested;
+            if (m.compareVersions(installed, claimed) < 0) {
+              console.error('::error::compat cannot validate maxTested=' + claimed +
+                ' — the runner captured CC ' + installed + '. Upgrade CC on the runner' +
+                ' (npm i -g @anthropic-ai/claude-code@' + claimed + ') and re-run, or lower maxTested.');
+              process.exit(1);
+            }
+            console.log('[compat] maxTested gate ok: claimed ' + claimed + ' <= installed ' + installed);
           "
 
           # Background-start the proxy, capture PID for the always-run


sprayberry-redline (approved):
Reviewed the diff and the evidence.

This closes the half of #1014 that #1016 left open, and the PR is refreshingly direct about that rather than presenting #1016 as having been a complete fix. Worth restating because the distinction is the whole point: #1016 guarantees the template is **fresh**; it says nothing about whether it is the **right version**. Those are different properties, and #1022 demonstrated the gap within hours — its capture step worked exactly as designed and still let a `maxTested=2.1.236` claim through on a runner holding 2.1.235.

The failure mode being closed is the dangerous kind. A green compat that validated an *older* CC than the PR names is not a missing signal, it is an inverted one — an affirmative "we tested this" attached to a version that was never loaded. Failing closed is correct.

Implementation checks:

- Reuses the already-exported `compareVersions` instead of reimplementing semver comparison in shell, so the gate and `detectDrift` can't diverge on comparison semantics.
- Reads `SUPPORTED_CC_RANGE.maxTested` from the built `dist/`, i.e. the value this PR actually ships, not a re-parse of source.
- `compareVersions(installed, claimed) < 0` is the right predicate — runner newer or equal passes, runner behind fails. Confirmed against the truth table in the PR body.
- The error names the exact remediation (`npm i -g @anthropic-ai/claude-code@<claimed>`), which turns the next occurrence into a one-line fix rather than an investigation. That matters, because the two prior occurrences were caught only by a human reading a log line.
- Placed after the capture, so it gates on the freshly captured version rather than a stale cache — the two fixes compose in the right order.

**Verification is as good as it gets for a CI guard: it self-demonstrates on its own PR.** This run's compat logged

```
[compat] captured live template from CC 2.1.236
[compat] maxTested gate ok: claimed 2.1.236 <= installed 2.1.236
```

and the failure path was exercised locally against the real capture pipeline (runner 2.1.236, simulated claim 2.1.240 → exit 1 with the actionable message). Full CI green.

One note for later, not blocking: this makes the runner's CC version a hard dependency of every drift PR. That is the intended trade — correctness over throughput — but it means a CC release now requires upgrading the box before the drift PR can land. If that friction becomes a problem, the follow-up is to have `cc-drift-watch.yml` upgrade the runner as a pipeline step, which #1014 already suggests.

Approving.

PR #1027, opened 2026-08-20
chore(cc-drift): v5.5.25 — maxTested → v2.1.237

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.237 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.236` → `2.1.237` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.25`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.25] - 2026-08-20` and appends the drift-fix bullet

### Items in the drift report

- **scanner.layout** (high) — No scannable CC binary found for v2.1.237. Wrapper has no cli.js/cli.mjs and no @anthropic-ai/claude-code-linux-x64 entry in optionalDependencies was fetchable. Inspect the tarball layout and update jsCandidates or fetchNativeBinary in scripts/check-cc-drift.mjs.
- **compat.range** (medium) — CC v2.1.237 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.236). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.235; npm latest is v2.1.237. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.237 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.237 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.237. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.237" step — no local run needed.)
[...]

PR #1028, opened 2026-08-20
chore(template): label refresh _version -> v2.1.237 (clears sdk-drift)

askalf:

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-20T05:06:27+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.237** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.237`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.237`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.5.26` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.237. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.


sprayberry-redline (approved):

Bot-drafted by `cc-drift-template-watch.yml`, but the version fix on top was made by the same party approving this, so it is not independent review. Recording it rather than leaving it implicit.

**Read the history note below before merging** — this branch was force-pushed after an automated heal corrupted it.

## Scope

Three files: `src/cc-template-data.json`, `package.json`, `CHANGELOG.md`.

## Verified

- **The label bump is exactly what it claims.** `_version`, `_supportedMaxTested` and the `claude-cli/<version>` token in the `user-agent` header move `2.1.236` → `2.1.237`. `_captured` is unchanged, so this is a label refresh and not a re-capture — the contract `scripts/label-sync.mjs` states.
- **The wire shape is unchanged.** `cc-drift-template-watch` ran `capture-and-bake --check` against live CC v2.1.237 and found zero drift against the bundle, which is what makes a label-only refresh the right change rather than a rebake.
- **The version is now correct and distinct.** `package.json` 5.5.32 matches the `## [5.5.32]` heading and clears master's tip (5.5.31), so `cc-drift-auto-release` will ship on merge rather than fast-exiting its duplicate-tag gate. #1027 sits above it at 5.5.33, so the two can merge in either order without a version going backwards.
- Suite 139/139 on the merge.

## History note — this branch was force-pushed, and why

The automated heal from #1040 ran against this PR and got it wrong in two ways at once:

1. It renumbered **after** merging master. `resolve-release-conflicts.mjs` keeps the higher version, so by then `package.json` held *master's* 5.5.31 rather than this PR's 5.5.26 — and the renumberer renamed **master's `## [5.5.31]` section**, the notes for a release already tagged and on npm, while this PR's own entry sat orphaned at `## [5.5.26]`.
2. It renumbered this PR and #1027 onto the **same** version, because both read their floor from `origin/master`, which does not move during a run.

Merging that would have shipped 5.5.32 carrying 5.5.31's release notes and erased 5.5.31's own.

The branch was reset to its pre-heal commit and redone in the corrected order — renumber first, then merge — which renames this PR's own section and leaves master's untouched. Confirmed on the current head: `## [5.5.32]` carries the template-label entry, `## [5.5.31]` is present and still carries the drift-heal entry, and `_version` reads `2.1.237`.

Master was never damaged; the corruption existed only on this branch and #1027's.

Both workflow defects are fixed in #1042, with the ordering hazard pinned as a characterization test.

## Verdict

Approving. The change itself is a three-line label refresh backed by a live `--check`, and the version is now correct and distinct. The reason this sat for a day is not the diff — it is the collision loop #1040 and #1042 exist to end.

PR #1029, opened 2026-08-20
ci: the auto-release job must never fire from a fork

chaogebaba:
`cc-drift-auto-release` runs on an hourly cron and carries no repository guard, so every fork runs it on the same schedule. There are 54.

The gate asks whether a tag matching `package.json.version` already exists. A fork's tags are whatever it inherited when it diverged, so on any fork whose master has since moved the answer is no, and the job proceeds into the release path: tag, GitHub Release, `npm publish --access public --provenance`, GHCR push.

The publish step should fail on its own — npm's trusted publisher is bound to this repository, and the OIDC subject presented from a fork will not match. Two things sit in front of it though:

- the tag and the GitHub Release are created before the publish runs, so a fork produces those even when the publish is rejected;
- the duplicate-tag gate reads `npm view @askalf/dario`, so a fork is deciding whether to release by consulting *this* package's registry entry rather than anything about itself.

I found this on my own fork, where the job had been firing hourly for several days. Nothing shipped from it — but only because an unrelated breakage was killing every run before it reached the gate, not because anything was stopping it. When I fixed that breakage the runs started getting further, which is how it surfaced.

One line of condition, plus a comment saying why it is there so it does not get tidied away later. No behaviour change on this repository.

Worth saying explicitly since I cannot verify it from outside: if npm's trusted publisher is *not* in fact bound narrowly, or if any fork has an `NPM_TOKEN` secret of its own configured, the exposure is larger than the tag and the Release. The guard covers both cases either way.

### DIFF
--- a/.github/workflows/cc-drift-auto-release.yml
+++ b/.github/workflows/cc-drift-auto-release.yml
@@ -106,7 +106,21 @@ jobs:
     # For pull_request: require `merged == true` (rules out close-
     # without-merge). For schedule / workflow_dispatch: always proceed;
     # the gate step short-circuits when there's nothing to ship.
-    if: github.event_name != 'pull_request' || github.event.pull_request.merged == true
+    # Never from a fork. Everything below is triggered by nothing more than
+    # package.json.version moving on master, on an hourly cron -- and it tags,
+    # cuts a GitHub Release, runs `npm publish --access public --provenance`
+    # and pushes to GHCR. A fork inherits the workflow and the schedule but
+    # none of the authority, and its package.json is whatever it last merged,
+    # so the duplicate-tag gate finds no matching tag in the fork and the job
+    # proceeds. The publish itself should fail -- npm's trusted publisher is
+    # bound to this repository, and the OIDC subject from a fork will not
+    # match -- but the tag and the GitHub Release are created before that, and
+    # `npm view` against @askalf/dario is what the gate reads, so a fork also
+    # decides whether to "release" by looking at this package's registry
+    # entry rather than its own.
+    if: >-
+      github.repository == 'askalf/dario'
+      && (github.event_name != 'pull_request' || github.event.pull_request.merged == true)
     runs-on: ubuntu-latest
     # Serializes overlapping triggers so a pull_request:closed and a
     # schedule tick that fire within seconds of each other don't both


askalf:
This is a real hole and the guard is the right fix. Thanks for chasing it down from the fork side — that's not a vantage point I have.

Answering the part you flagged as unverifiable from outside, since you were right to flag it rather than assume:

**npm's trusted publisher is bound narrowly.** The job publishes via OIDC (`id-token: write`, `npm publish --access public --provenance`) and there is no `NPM_TOKEN` in the publish path at all — the workflow's own comment says as much: *"id-token: write is the npm AUTH now, not just provenance garnish… no NPM_TOKEN secret."* So the OIDC subject from a fork genuinely won't match, and your reasoning holds. The second case you raise — a fork configuring its own `NPM_TOKEN` — is outside anyone's control here, and as you say the guard covers it either way.

Which leaves exactly the exposure you described: **the tag and the GitHub Release get created before the publish is rejected**, and the duplicate-tag gate consults `npm view @askalf/dario` — so a fork decides whether to "release" by reading this package's registry entry rather than anything about itself. That second part is the one I'd have missed.

## One thing your PR made me check

`cc-drift-auto-release` isn't the only scheduled workflow here without a repository guard. It's **16 of 16**:

```
auto-merge-sweep                cc-oauth-health          scorecard
cc-billing-classifier-canary    cflite                   sdk-drift-watch
cc-drift-auto-release  <- this  codeql                   self-healing-supervisor
cc-drift-template-watch         dario-doctor-watch       stale
cc-drift-watch                  npm-drift-watch          wire-drift-self-hosted
cc-drift-watcher-liveness
```

Being precise about severity, because they are not equivalent and I don't want to inflate this:

- **`cc-drift-auto-release` is the only one that reaches outward.** It tags, releases, and attempts a publish. That's yours, and it's the one that mattered.
- The rest are **waste and noise, not exposure**. A fork's `GITHUB_TOKEN` is scoped to the fork, so the issue-filing watchers (`cc-drift-watch`, `sdk-drift-watch`, `npm-drift-watch`, the liveness and canary jobs) file into the fork, not here. `wire-drift-self-hosted` targets runners a fork doesn't have and just fails. `codeql` / `scorecard` / `cflite` burn fork minutes.

Still worth closing across the board — 54 forks running 16 hourly-to-daily jobs each is a lot of someone else's CI minutes spent on our behalf, and several of them read *this* package's state to make decisions, same as the gate you found. I'll open a follow-up for the other 15 rather than expand yours; a one-line security fix shouldn't have to carry a sweep.

Holding merge for @askalf, per how release-path changes go here. Nothing needed from you.

sprayberry-redline (approved):
**Automated review — approving.**

Unlike the rest of today's batch, this one is **independent**: @chaogebaba authored it, and neither the author nor anyone acting for them is approving it. No disclosure needed, and the approval means what an approval normally means.

## The finding is correct, and the part that mattered is the part I could not see

`cc-drift-auto-release` runs on an hourly cron with no repository guard, so every fork runs it on the same schedule. There are 54.

I can confirm the half the reporter flagged as unverifiable from outside: **npm's trusted publisher is bound narrowly.** The job publishes via OIDC (`id-token: write`, `npm publish --access public --provenance`) and there is no `NPM_TOKEN` anywhere in the publish path — the workflow's own comment says so: *"id-token: write is the npm AUTH now, not just provenance garnish… no NPM_TOKEN secret."* So a fork's OIDC subject genuinely will not match, and the publish genuinely should fail.

Which leaves exactly the exposure described, and this is the part worth crediting:

- **the tag and the GitHub Release are created before the publish is attempted**, so a fork produces both even when the publish is rejected;
- **the duplicate-tag gate reads `npm view @askalf/dario`**, so a fork decides whether to release by consulting *this* package's registry entry rather than anything about itself.

A fork whose master has diverged finds no matching tag, concludes there is something to ship, and proceeds. The reporter found this because it had been firing hourly on their own fork for days — and was explicit that nothing shipped only because an unrelated breakage was killing every run before it reached the gate, not because anything was stopping it. That distinction is the right one to draw and most reports would not have drawn it.

## The change

One condition plus a comment explaining why it is there:

```yaml
if: >-
  github.repository == 'askalf/dario'
  && (github.event_name != 'pull_request' || github.event.pull_request.merged == true)
```

Verified:

- **The existing predicate is preserved exactly.** The original `github.event_name != 'pull_request' || github.event.pull_request.merged == true` is parenthesised rather than rewritten, so the merged-only rule for the `pull_request` path is unchanged. Getting this wrong — dropping the parentheses and letting `&&` bind tighter than `||` — would have made the schedule path unreachable and silently stopped releases. It is right.
- **No behaviour change on this repository.** `github.repository` is `askalf/dario` here, so the added conjunct is constant-true and all three trigger paths (`pull_request` merged, `schedule`, `workflow_dispatch`) behave as before.
- **The comment survives a tidy-up.** It states what the job does — tag, Release, `npm publish`, GHCR push — so the next reader can see why the guard is load-bearing rather than decorative. That was asked for explicitly and it is the difference between a guard that lasts and one that gets removed as redundant.

`actionlint` and `validate-package-json` pass; the remaining checks are running. Fork-PR workflow runs needed maintainer release, which is why this sat without checks — all five are GitHub-hosted (`ubuntu-latest`), none touch the self-hosted runners, and the diff is fifteen lines in one workflow file.

## Scope, and what follows

This PR guards the one workflow that reaches outward. It is **16 of 16** scheduled workflows here that carry no repository guard, and that is tracked separately in #1039 rather than pushed into this PR — a one-line security fix should not have to carry a sweep.

To be precise about the rest, since inflating it would be easy: `cc-drift-auto-release` is the only one that tags, releases and publishes. The others are waste rather than exposure — a fork's `GITHUB_TOKEN` is scoped to the fork, so the issue-filing watchers file into the fork; `wire-drift-self-hosted` targets runners a fork does not have; `codeql` / `scorecard` / `cflite` burn fork minutes. #1039 also argues that the blanket `github.repository ==` form is *wrong* for several of them, because it would disable a forker's own push/PR security scanning to save cron minutes.

## Verdict

Approving, and thanks — this is the kind of report that is only findable from a vantage point the maintainer does not have, and it arrived with the reasoning, the mechanism, and an honest statement of what could not be verified from outside.

Issue #36, opened 2026-04-15
Report about 3.9.2 Hybrid mode.


boeingchoco:
First, I'm really thankyou about your feature request and making it by program code so fast!!

Here is what I've tested.

```
dario proxy --hybrid-tools --verbose
  Device identity: detected

  dario — http://localhost:3456

  Your Claude subscription is now an API.

  Usage:
    ANTHROPIC_BASE_URL=http://localhost:3456
    ANTHROPIC_API_KEY=dario

  OAuth: healthy (expires in 3h 28m)
  Model: passthrough (client decides)

[dario] #0 POST /v1/messages
[dario] #0 context-1m rejected (429) — retrying without it
[dario] #1 billing: five_hour (overage: ?)
[dario] #1 POST /v1/messages
[dario] #1 context-1m rejected (429) — retrying without it
[dario] #2 billing: five_hour (overage: ?)
[dario] #2 POST /v1/messages
[dario] #2 context-1m rejected (429) — retrying without it
[dario] #3 billing: five_hour (overage: ?)
[dario] #3 POST /v1/messages
[dario] #3 context-1m rejected (429) — retrying without it
[dario] #4 billing: five_hour (overage: ?)
[dario] #4 POST /v1/messages
[dario] #4 context-1m rejected (429) — retrying without it
[dario] #5 billing: five_hour (overage: ?)
[dario] #5 POST /v1/messages
[dario] #5 context-1m rejected (429) — retrying without it
[dario] #6 billing: five_hour (overage: ?)
[dario] #6 POST /v1/messages
[dario] #6 context-1m rejected (429) — retrying without it
[dario] #7 billing: five_hour (overage: ?)
[dario] #7 POST /v1/messages
[dario] #8 POST /v1/messages
[dario] #8 context-1m rejected (429) — retrying without it
[dario] #9 billing: five_hour (overage: ?)
[dario] #9 POST /v1/messages
[dario] #9 context-1m rejected (429) — retrying without it
[dario] #10 billing: five_hour (overage: ?)
[dario] #10 POST /v1/messages
[dario] #10 context-1m rejected (429) — retrying without it
[dario] #11 billing: five_hour (overage: ?)
[dario] #11 POST /v1/messages
[dario] #11 context-1m rejected (429) — retrying without it
[dario] #12 billing: five_hour (overage: ?)
[dario] #12 POST /v1/messages
[dario] #12 context-1m rejected (429) — retrying without it
[...]

PR #1034, opened 2026-08-20
fix: never leave a request ending on an assistant turn (#1033)

### LINKED ISSUE #1033: subagent returns 400
**What happened?**
Subagent completed but i have received err 400 in claude code
● API Error: 400 This model does not support assistant message prefill. The conversation must end with a user message.
**What did you expect?**
Successful subagent return
**Steps to reproduce:**
1. start subagent
2. wait for it to complete
3. get 400

**Environment:**
- OS: ubuntu24
- Node version: 24 
- dario version: 5.5.18
- Tool using dario (e.g. OpenClaw, Aider): claude code

askalf:
Fixes #1033.

## The report

@miklisanton hits this the moment a sub-agent returns, on dario 5.5.18 / CC / Ubuntu 24:

```
API Error: 400 This model does not support assistant message prefill.
The conversation must end with a user message.
```

## What is actually happening

CC emits standalone `<system-reminder>` / `<task_metadata>` user turns — notably right after a Task (sub-agent) result is folded back into the parent transcript. Both tags are in `ORCHESTRATION_TAG_NAMES`, so:

1. the orchestration-tag scrub empties that turn,
2. `sanitizeMessages`' drop-empty-messages filter (added for #744) removes it,
3. the request leaves dario ending on the **assistant** turn.

Anthropic reads a trailing assistant turn as a prefill ("continue from this text"), and Opus 4.6 under adaptive thinking + the `claude-code` beta refuses it. Nothing about the user's own conversation is malformed — dario makes it malformed on the way out.

Reproduced against `origin/master` (`298a89d`), before the fix:

```
A. trailing user turn = system-reminder only (array content)
  roles: user -> assistant          ends-on: assistant   <-- PREFILL 400
B. trailing user turn = system-reminder only (string content)
  roles: user -> assistant          ends-on: assistant   <-- PREFILL 400
C. control - trailing user turn has real text
  roles: user -> assistant -> user  ends-on: user   ok
```

A second, independent path produces the same shape — the trailing-empty pop in `buildCCRequest`:

```
D. client sends trailing user turn with content: []
  roles: user -> assistant          ends-on: assistant   <-- PREFILL 400
```

After the fix all four end on a user turn.

## The fix

One invariant, enforced at both sites: **dario must never turn a request that ended on a user turn into one that ends on an assistant turn.**

- **`sanitizeMessages` (`src/proxy.ts`)** restores the tail turn's *pre-scrub* content instead of dropping it, when dropping would expose an assistant turn. The orchestration tag survives in that one position only, which is the right trade in both directions: for a CC client the tag is CC's own injection, so keeping it is *more* wire-faithful, not less; for a non-CC client a lone tag as the final turn is the actual prompt, and forwarding it beats a hard 400. Every other position scrubs exactly as before — the guard is positional, and an emptied turn in the middle is still dropped.

- **The trailing-empty pop (`src/cc-template.ts`)** is restricted to **assistant** turns, which is the only shape it was written for (a thinking-only assistant turn emptied by the thinking strip above it). Popping an empty *user* turn exposed the assistant turn behind it and produced the very rejection the loop exists to prevent. An empty user turn is a malformed client request either way — leaving it in place lets the upstream name it accurately (`messages.N: content must contain at least one block`) instead of dario converting it into a misleading prefill error.

## What this must not regress, and doesn't

- #36 — a trailing **thinking-only assistant** turn is still dropped (covered by the existing `hybrid-tools.mjs` case, still green).
- #37 — a trailing assistant turn that still carries text / `tool_use` is still **not** popped. That was the v3.10.1 runaway loop (133 POSTs from one prompt); the guard here only ever *adds* a user turn back, never removes an assistant one.
- #744 / #54 — interior emptied turns and empty text blocks are dropped exactly as before, asserted explicitly.

## Tests

13 new cases — 11 in `test/sanitize-messages.mjs`, 2 in `test/hybrid-tools.mjs` — covering array and string content, `task_metadata` as well as `system-reminder`, the positional guard, and both no-regress controls.

`npm test`: **134 / 137 pass.** The 3 failures (`template-config-scoped-tools`, `template-interactive-tools`, `tool-advertise-respects-client`) are **pre-existing on clean `origin/master`** and are not caused by this change — they reproduce identically on an unmodified checkout, and all three pass when run against a clean `HOME`. They are an artifact of a local `~/.dario/cc-template.live.json`, which is a separate bug I am filing on its own.

## Release note

Bumps `5.5.25` → `5.5.26`. Heads-up: **#1028 also claims 5.5.26**, so whichever of the two lands second needs `gh pr update-branch` and a re-bump. Separately, **#1027 looks stale** — it bumps `5.5.24` → `5.5.25`, but master is already at 5.5.25 and `v5.5.25` is tagged and on npm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.26] - 2026-08-20
+
+- **Fixed: sub-agent turns rejected with `400 assistant message prefill`** (#1033) — CC emits standalone `<system-reminder>` / `<task_metadata>` user turns, notably right after a Task (sub-agent) result folds back into the transcript. Both tags are scrubbed by the orchestration-tag pass, so the turn emptied, the drop-empty-messages filter removed it, and the request left ending on the *assistant* turn. Anthropic reads that as a prefill and rejects it: *"This model does not support assistant message prefill. The conversation must end with a user message."* `sanitizeMessages` now restores that turn's pre-scrub content instead of dropping it, and the template build's trailing-empty pop is restricted to assistant turns — popping an empty *user* turn exposed the assistant turn behind it and caused the same rejection.
+
 ## [5.5.25] - 2026-08-20
 
 - **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.236` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.236 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.25",
+  "version": "5.5.26",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/cc-template.ts
+++ b/src/cc-template.ts
@@ -1845,10 +1845,19 @@ export function buildCCRequest(
   // appended its assistant reply locally, dario stripped it from the next
   // request, the model regenerated the same reply, dario stripped that, and
   // the loop never terminated (133 POSTs from a single user prompt).
+  //
+  // Restricted to ASSISTANT turns, which is the only shape this loop was
+  // written for (a thinking-only assistant turn emptied by the strip above).
+  // Popping an empty *user* turn is what the loop must not do: it exposes the
+  // assistant turn behind it and produces the very prefill rejection the loop
+  // exists to prevent (dario#1033). An empty user turn is a malformed client
+  // request either way — leaving it in place lets the upstream name it
+  // accurately ("messages.N: content must contain at least one block")
+  // instead of dario converting it into a misleading prefill error.
   while (messages.length > 0) {
     const last = messages[messages.length - 1];
     const contentEmpty = Array.isArray(last.content) && (last.content as unknown[]).length === 0;
-    if (contentEmpty) {
+    if (contentEmpty && last.role === 'assistant') {
       messages.pop();
       continue;
     }
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -675,6 +675,12 @@ export function sanitizeMessages(body: Record<string, unknown>, preserveTags?: S
   const messages = body.messages as Array<{ role: string; content: unknown }> | undefined;
   if (!messages) return;
   const patterns = orchestrationPatternsFor(preserveTags);
+  // Snapshot the tail turn before scrubbing. If the scrub empties it and the
+  // drop below would expose an assistant turn, we put the original content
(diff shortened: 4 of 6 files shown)


sprayberry-redline (approved):

This PR and the fix in it were authored by the same party approving it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

Adversarial pass below, including two behaviour changes that deserve to be named rather than absorbed silently.

## Scope

`src/proxy.ts` (the `sanitizeMessages` tail guard), `src/cc-template.ts` (the trailing-empty pop), 13 test cases across two files, plus the bump and CHANGELOG. Rebased onto master at 5.5.27; `src/cc-template.ts` auto-merged with no overlap against #1036, which moved the tool sets out of the top of that file.

## Verified

- **The reporter's failure is reproduced, not assumed.** Against unmodified master, a conversation ending in a `<system-reminder>`-only user turn leaves dario as `user -> assistant`, which is the prefill shape Anthropic rejects. The control — same turn carrying real text alongside the reminder — is unaffected.
- **The cause is specific to sub-agents for a real reason.** `system-reminder` and `task_metadata` are both in `ORCHESTRATION_TAG_NAMES`, and CC emits a standalone turn of exactly that kind when a Task result folds back into the parent transcript. That explains why the reporter sees it on sub-agent return and essentially never otherwise, which a generic "scrub is too aggressive" theory would not.
- **Two independent paths produced the shape, and both are closed.** The second (the trailing-empty pop in `buildCCRequest`) is reachable without the scrub at all, so fixing only `sanitizeMessages` would have left a live route to the same 400.
- **The regressions it must not cause, it does not.** A trailing thinking-only *assistant* turn is still dropped (#36), and a trailing assistant turn still carrying text or `tool_use` is still not popped (#37, the v3.10.1 runaway loop — 133 POSTs from one prompt). The guard only ever adds a user turn back; it never removes an assistant one, which is what makes it structurally incapable of reintroducing #37.
- **The guard is positional, and that is asserted.** An emptied turn in the middle is still dropped, so #744 and #54 behaviour is unchanged.
- Full suite 137/137 on the rebase.

## Findings against this PR

1. **For non-CC clients, an orchestration tag now reaches Anthropic where it previously did not.** The restore puts back *pre-scrub* content, so when an Aider/Cursor/OpenCode client's final turn is nothing but `<agent_persona>` or `<tool_call>`, that tag is now forwarded verbatim. The PR argues this is the right trade — a lone tag as the final turn *is* the actual prompt, and forwarding beats a hard 400 — and I agree, but it should be stated as what it is: a deliberate narrowing of the scrub's guarantee, not a pure bug fix. The scrub is a fingerprint-management feature, and this is the one position where it now yields.

2. **An empty trailing user turn now produces an upstream 400 where it previously produced a silent drop.** Restricting the pop to assistant turns means a client that sends `content: []` gets `messages.N: content must contain at least one block` from Anthropic instead of having the turn quietly removed. The PR's reasoning is sound — that error names the client's actual mistake, where the old behaviour converted it into a misleading prefill error — but a client that was previously "working" by accident will now see a hard failure. No known client does this, and CC does not, so the blast radius looks like zero; flagging it because "no known client" is an assumption, not a measurement.

3. **Minor: `tailHadContent` treats any non-null, non-string, non-array content as present.** A malformed `content: {}` would count as content worth restoring. Harmless in practice — such a body fails upstream validation regardless — but the check is looser than the three shapes it actually reasons about.

## Verdict

Approving. The diagnosis is precise, the fix is narrower than the obvious one (popping the trailing assistant, which is what caused #37), and the no-regress cases are covered by assertion rather than by argument.

Findings 1 and 2 are accepted trade-offs rather than defects, but they are behaviour changes on the request path and belong in the release notes, not only in the PR thread.

PR #1036, opened 2026-08-20
fix: apply the tool superset rule when LOADING a live template (#1035)

### LINKED ISSUE #1035: A live template capture silently drops interactive- and config-scoped tools, and rewrites history tool_use onto fallback slots
The superset rule that keeps `AskUserQuestion` / `EnterPlanMode` / `ExitPlanMode` / `TaskCreate` / `TaskGet` / `TaskList` / `TaskUpdate` in the tool bundle is enforced **only at bake time**. `loadTemplate()` applies no such rule to a user's *live* capture — so every dario install with CC present degrades itself within 24 hours of running.

CI never sees it, because CI has no `~/.dario/cc-template.live.json`.

## Mechanism

`refreshLiveFingerprintAsync` captures headlessly (`claude --print -p hi`) on a 24h TTL. CC v2.1.187+ does not advertise the interactive tools in `--print` mode, and CC's remote config decides the `Task*` four independently — the exact two causes `INTERACTIVE_ONLY_TOOLS` and `CONFIG_SCOPED_TOOLS` were introduced to compensate for.

`capture-and-bake.mjs` compensates (lines ~204 / ~225 / ~243, re-adding each set from the previous bundle). `loadTemplate` does not:

```ts
const cached = readLiveCache();
if (cached) {
  const age = Date.now() - new Date(cached._captured).getTime();
  if (age < LIVE_TTL_MS) {
    return withBundledVariants(cached);   // <-- tools taken wholesale from the capture
  }
```

`withBundledVariants` merges `system_prompt_variants` and nothing else. `TEMPLATE.tools` therefore becomes whatever the headless capture saw, and `CC_TOOL_DEFINITIONS_UNION` / `CC_NATIVE_NAMES_UNION` are derived from it.

## Observed on this machine

`~/.dario/cc-template.live.json`, `_source: live`, `_captured: 2026-08-19T22:57:28Z`, CC v2.1.236 — the same `_version` as the bundle:

| | bundled | live capture |
|---|---|---|
| tool count | 34 | **26** |
| `AskUserQuestion` | present | **missing** |
| `EnterPlanMode` | present | **missing** |
| `ExitPlanMode` | present | **missing** |
| `TaskCreate` / `TaskGet` / `TaskList` / `TaskUpdate` | present | **missing** |

## Impact

Two failures, and the second is the serious one. A client declaring five tools:

```
client declared : Read, Bash, TaskCreate, AskUserQuestion, ExitPlanMode
dario advertised: Bash, Read
```

1. **Tools are silently not advertised** even though the client declared them — the advertise-respects-client contract, i.e. the v4.8.93 regression, reached through the live-capture path instead of the bake.

2. **History `tool_use` blocks are rewritten onto a fallback slot with junk arguments.** `CC_NATIVE_NAMES_UNION` no longer contains `TaskCreate`, so it stops identity-mapping and falls into the unmapped round-robin. A prior `TaskCreate` call in the transcript goes out as:

```
history tool_use name on the wire: Grep
history tool_use input           : {"pattern":".","path":"."}
```

The model's own transcript is rewritten to claim it called `Grep` with garbage. This is precisely the failure mode the `CONFIG_SCOPED_TOOLS` comment warns about.

The same argument applies to `PLATFORM_ONLY_TOOLS`: a live capture on Linux carries no `PowerShell`, so a win32 client's `PowerShell` falls into the same round-robin — the v4.8.136 bug, likewise reachable again through the live cache.

## Reproduction

The repo's own tests already catch this — they just never run anywhere a live capture exists. On a machine with one:

```
$ npm test
not ok 120 - template-config-scoped-tools.mjs
not ok 122 - template-interactive-tools.mjs
not ok 124 - tool-advertise-respects-client.mjs
# tests 137 / pass 134 / fail 3
```

```
$ HOME=/tmp/clean node test/tool-advertise-respects-client.mjs   # no live cache
  ... all pass
```

Both against unmodified `origin/master` (`298a89d`).

## Suggested fix

Apply the superset rule where the template is *loaded*, not only where it is *baked* — and define it once so the two cannot drift apart again. `loadTemplate` should union any tool in `PLATFORM_ONLY_TOOLS ∪ INTERACTIVE_ONLY_TOOLS ∪ CONFIG_SCOPED_TOOLS` that the bundle has and the capture lacks, then re-derive `tool_names` to preserve the `tool_names === tools.map(name)` contract that the bake already learned to maintain.

I have this in progress and will open a PR.

askalf:
Fixes #1035.

## Summary

The superset rule that keeps `AskUserQuestion` / `EnterPlanMode` / `ExitPlanMode`, the `Task*` four, and other-platform tools like `PowerShell` in the tool set was enforced **only in `capture-and-bake.mjs`**. `loadTemplate` applied none of it to a user's **live** capture.

Captures are headless and refresh on a 24h TTL, and CC v2.1.187+ drops the interactive tools under `--print` — so every dario install with CC present degraded its own tool set within a day of running. CI never caught it because CI has no live cache.

## Evidence

A real cache on this machine (`_source: live`, CC v2.1.236 — the *same* `_version` as the bundle) carried 26 tools against the bundle's 34, missing all seven names above. With it present, against unmodified `origin/master`:

```
client declared : Read, Bash, TaskCreate, AskUserQuestion, ExitPlanMode
dario advertised: Bash, Read

history tool_use name on the wire: Grep
history tool_use input           : {"pattern":".","path":"."}
```

The second line is the serious one. `CC_NATIVE_NAMES_UNION` is derived from the loaded template, so a client that still declares `TaskCreate` stops identity-mapping, falls into the unmapped round-robin, and **has its own transcript rewritten to claim it called `Grep` with junk arguments** — the v4.8.93 failure mode, reached through the live-cache path instead of the bake.

After this change, same machine, same cache:

```
client declared : Read, Bash, TaskCreate, AskUserQuestion, ExitPlanMode
dario advertised: AskUserQuestion, Bash, ExitPlanMode, Read, TaskCreate

history tool_use name on the wire: TaskCreate
history tool_use input           : {"subagent_type":"Explore","prompt":"find X"}
```

## The change

`PLATFORM_ONLY_TOOLS`, `INTERACTIVE_ONLY_TOOLS` and `CONFIG_SCOPED_TOOLS`, plus one `mergePreservedTools()`, now live in `live-fingerprint.ts` as a **single definition**. `loadTemplate` is upstream of `cc-template.ts`, so the rule has to be expressible there; `cc-template.ts` re-exports all three so every existing import site (`capture-and-bake.mjs`, the invariant tests) is untouched.

`tool_names` is re-derived after the merge — the `tool_names === tools.map(name)` contract the bake already learned to maintain the hard way.

The merge is conservative by construction: it only ever adds a name the **bundle already carries** *and* the preserved sets name. It cannot invent a tool, and a stale entry is inert because advertise is an intersection with what the client declared. Asserted both ways in the tests.

## Tests

14 new assertions in `test/live-fingerprint.mjs`. They **fail on master** (7 failures) and pass here — verified by running this exact file against an unmodified `origin/master` build. They synthesize a degraded capture through `DARIO_LIVE_TEMPLATE_CACHE`, so CI exercises the path without needing CC installed.

`npm test`: **137 / 137**. The three suites that fail on master on any machine with a live cache — `template-config-scoped-tools`, `template-interactive-tools`, `tool-advertise-respects-client` — now pass with the real degraded cache in place.

## Deliberately not in this PR

`capture-and-bake.mjs` still has its own three merge loops rather than calling `mergePreservedTools()`. The *sets* are now single-sourced, so they cannot drift; only the loops are duplicated. I left them because the bake needs a live CC to run end-to-end and I could not exercise it here — switching it blind is the wrong risk to take quietly. Worth a follow-up.

## Release note

Bumps `5.5.25` → `5.5.27`, leaving `5.5.26` to #1034 / #1028. Whichever of these lands second will need `gh pr update-branch` and a re-bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.27] - 2026-08-20
+
+- **Fixed: a live template capture silently degraded the tool set** (#1035) — the superset rule that keeps `AskUserQuestion` / `EnterPlanMode` / `ExitPlanMode` / `TaskCreate` / `TaskGet` / `TaskList` / `TaskUpdate` (and other-platform tools such as `PowerShell`) in the bundle was enforced only at bake time. `loadTemplate` applied none of it to a user's live capture, so every install with CC present degraded itself within the 24h cache TTL: declared tools went unadvertised, and history `tool_use` blocks for the dropped tools were renamed onto fallback slots with junk arguments — the v4.8.93 failure mode, reached through the live-cache path. The rule is now defined once in `live-fingerprint.ts` and applied at load as well as at bake, with `tool_names` re-derived after the merge.
+
 ## [5.5.25] - 2026-08-20
 
 - **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.236` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.236 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.25",
+  "version": "5.5.27",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/cc-template.ts
+++ b/src/cc-template.ts
@@ -8,35 +8,23 @@
  * action required. See src/live-fingerprint.ts for the capture pipeline.
  */
 
-import { loadTemplate, promptVariantsOf, TemplateData, VARIANT_FAMILIES } from './live-fingerprint.js';
+import {
+  loadTemplate, promptVariantsOf, TemplateData, VARIANT_FAMILIES,
+  PLATFORM_ONLY_TOOLS, INTERACTIVE_ONLY_TOOLS, CONFIG_SCOPED_TOOLS,
+} from './live-fingerprint.js';
+
+// Re-exported so existing importers (scripts/capture-and-bake.mjs, the template
+// invariant tests) keep their import site. The definitions live in
+// live-fingerprint.ts because loadTemplate must apply the same superset rule to
+// a live capture that the bake applies to the bundle, and cc-template.ts is
+// downstream of that load (#1035).
+export { PLATFORM_ONLY_TOOLS, INTERACTIVE_ONLY_TOOLS, CONFIG_SCOPED_TOOLS };
 
 // Load template at module init — prefer live cache, fall back to bundled.
 const TEMPLATE: TemplateData = loadTemplate({ silent: true });
 
 /** The loaded template itself — source, version, capture age, all fields. Startup banners and drift checks read this directly. */
 export const CC_TEMPLATE: TemplateData = TEMPLATE;
-
-/**
- * Tools CC only ships on a specific platform. The bundled template is a
- * union capture (any platform the maintainer baked from), so we filter it
- * down to the running platform at module load. Real CC on the client side
- * only advertises the tools available to its host — forwarding a larger
- * set through dario would both leak a fingerprint (Anthropic sees tools
- * the client would never actually call) and risk tool_use round-trips
- * coming back for a tool the client has no handler for.
- *
- * PowerShell shipped in CC v2.1.116 on Windows; POSIX CC installs do not
- * advertise it. As of CC v2.1.162 the Glob/Grep tools are the same shape:
- * Windows CC advertises them, POSIX CC drops them and steers the agent to
- * shell `find`/`grep` instead (which PowerShell has no native equivalent
- * for). Registering them here filters them to win32 clients AND keeps a
- * POSIX auto-bake from dropping them out of the union — the v4.8.28
- * regression, where a Linux runner re-baked the bundle down to 28 tools.
- * Add new platform-scoped tools here as CC adds them.
- */
-export const PLATFORM_ONLY_TOOLS: Record<string, Set<string>> = {
-  win32: new Set(['PowerShell', 'Glob', 'Grep']),
-};
 
 /** Keep tool `t` unless its name is listed under a platform other than the current one. */
 export function filterToolsForPlatform<T extends { name: string }>(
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

This PR was authored by the same party approving it, and so was the issue it fixes (#1035). That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

What follows is an adversarial pass over the change, including two things it gets wrong.

## Scope

Five files. `live-fingerprint.ts` gains the three tool sets, a `mergePreservedTools()` helper and a `withPreservedTools()` wrapper; `cc-template.ts` loses its local copies of the sets and re-exports them instead; `test/live-fingerprint.mjs` gains 14 assertions; `package.json` / `CHANGELOG.md` carry the bump.

## Verified

- **The bug is real and reproduces on unmodified master.** A live cache on the reporting machine (`_source: live`, `_captured: 2026-08-19T22:57:28Z`, `_version: 2.1.236` — the *same* version as the bundle) carries 26 tools against the bundle's 34, missing `AskUserQuestion`, `EnterPlanMode`, `ExitPlanMode` and all four `Task*`.
- **The severe consequence is confirmed, not inferred.** With that cache present, a client declaring `[Read, Bash, TaskCreate, AskUserQuestion, ExitPlanMode]` gets `[Bash, Read]` advertised, and a history `tool_use` for `TaskCreate` goes out renamed to `Grep` with `{"pattern":".","path":"."}`. That is `CC_NATIVE_NAMES_UNION` shrinking and the unmapped round-robin claiming the name — the v4.8.93 failure mode, reached through the live-cache path rather than the bake.
- **The new tests are a genuine guard.** Run against an unmodified `origin/master` build they fail 7 assertions; against this branch they pass. They synthesize the degraded capture via `DARIO_LIVE_TEMPLATE_CACHE`, so CI exercises the path with no CC installed — which is the whole reason this was never caught.
- **The merge cannot invent a tool.** It only adds names the bundle already carries *and* `preservedToolReason()` names; `WebSearch` absent from a capture stays absent, asserted explicitly. Platform branch checked both ways: on `linux` it restores `PowerShell`/`Glob`/`Grep`, on `win32` it does not.
- **`tool_names` is re-derived after the merge**, preserving the `tool_names === tools.map(name)` contract the bake previously shipped artifacts in violation of.
- **Wire-shape direction is correct.** The advertised array is the intersection with client-declared tools, so this only ever restores a tool the client actually declared — moving the shape *toward* what a real CC would send, not away.
- Full suite 137/137 on a machine with the degraded cache; all checks green.

## Findings against this PR

1. **The bundle is now read and parsed twice at startup, and up to three times on the stale-cache path.** `loadBundledTemplate()` is not memoized — every call does `readFileSync` + `JSON.parse` on a ~110KB file. `withBundledVariants()` already called it; `withPreservedTools()` now calls it again. `loadTemplate` runs once at module init so the cost is a one-off and not worth blocking on, but it is a regression in a function that is easy to assume is cheap, and it becomes a real problem the moment anything calls `loadTemplate` per request. Memoizing `loadBundledTemplate` is a two-line fix and should follow.

2. **The repair is silent.** `withPreservedTools` logs which tools it restored, but `cc-template.ts` calls `loadTemplate({ silent: true })`, so in normal operation nobody ever sees it. That means both the degradation and its repair are invisible: a user whose capture is missing seven tools has no way to learn that from dario. `dario doctor` is the natural place to surface "live capture omitted N tools, restored from bundle", and it does not. Follow-up, not a blocker — the fix is still strictly better than the silent breakage it replaces.

## Accepted limitation, correctly disclosed

`capture-and-bake.mjs` keeps its own three merge loops rather than calling `mergePreservedTools()`. The PR says so and gives the reason — the bake needs a live CC to run end-to-end and could not be exercised locally. That is the right call: the sets are now single-sourced so they cannot drift, only the loops are duplicated, and a blind refactor of the script that writes the bundle is a worse risk than the duplication. Worth a follow-up once someone can run a real bake.

## Verdict

Approving. The change is well-evidenced, the tests fail-before/pass-after against real artifacts rather than only against fixtures, and the two findings above are follow-ups rather than defects in what ships.

Merging this ahead of #1034 / #1037 / #1038 is deliberate: it takes the suite from 134/137 to 137/137 on any machine with a live template cache, so every PR reviewed after it gets a clean signal instead of three failures that have to be explained away each time.

PR #1037, opened 2026-08-20
feat: report utilisation freshness on the account surfaces (#1032)

### LINKED ISSUE #1032: pool: util5h/util7d freeze at the moment of rejection and never recover
# pool: `util5h`/`util7d` freeze at the moment of rejection and never recover

## Summary

For an account the pool has parked (`status: rejected`), `util5h` and `util7d` in
`/accounts` stop updating entirely. The reported value is the rate-limit header
snapshot from the **last request that account served** — which, for a rejected
account, is precisely the request that hit the limit.

As the window rolls over, the real utilisation drops to zero, but dario keeps
reporting the old value indefinitely: no request goes through a parked account,
so no fresh rate-limit header ever arrives to correct it.

Net effect: **an account reports "5h window full" long after it is free again.**

## Environment

- dario **5.5.18**
- macOS (Darwin 25.4), Node 26.5.1
- `mode: pool`, three Claude OAuth accounts
- Observed via `GET /accounts` against the running proxy on `127.0.0.1:3456`

## Observation

Four samples of `GET /accounts`, same day, no restarts in between:

| time | `bls-…-max20` (rejected) | `bl-…-Pro5x` (rejected) | `bl-…-claude2-x20` (serving) |
|---|---|---|---|
| 10:50:54 | `util5h=1.01` | `util7d=1` | `util5h=0.94` |
| 10:51:54 | `util5h=1.01` | `util7d=1` | `util5h=0.94` |
| 10:52:54 | `util5h=1.01` | `util7d=1` | `util5h=0` ← window reset |
| 11:01:19 | `util5h=1.01` | `util7d=1` | `util5h=0.97` |

The account actively serving traffic behaves exactly as expected: it climbs,
resets to `0` when its 5-hour window rolls over, and climbs again.

In the same eleven minutes, the two parked accounts do not move by a single
digit. `bls-…-max20` sits at `1.01` throughout — a value **above 1.0**, which is
itself the fingerprint of a snapshot taken at the instant of rejection rather
than a current reading.

Full record at 11:01:19:

```json
{
  "mode": "pool",
  "accounts": [
    {"alias": "bls-…-max20",   "util5h": 1.01, "util7d": 0.47, "claim": "five_hour",  "status": "rejected",         "requestCount": 791},
    {"alias": "bl-…-Pro5x",    "util5h": 0,    "util7d": 1,    "claim": "seven_day",  "status": "rejected",         "requestCount": 675},
    {"alias": "bl-…-claude2",  "util5h": 0.97, "util7d": 0.26, "claim": "five_hour",  "status": "allowed_warning",  "requestCount": 1451}
  ],
  "healthy": 1, "exhausted": 2, "totalHeadroom": 1, "bestAccount": "bl-…-claude2"
}
```

## Steps to reproduce

1. Run `dario proxy` with a pool of ≥2 Claude OAuth accounts.
2. Drive traffic until one account is rejected on its 5-hour window.
3. Poll `GET /accounts` for longer than the 5-hour window.
4. The parked account's `util5h` never changes, while the serving account's
   value resets to `0` on schedule.

## Why this matters

1. **Every consumer shows a stale number as current.** The payload has no
   `lastSeenAt`, no age, no freshness flag, so the TUI, dashboards and scripts
   cannot distinguish a live reading from one that has been frozen for hours.
   There is no way to render it honestly.

2. **Pool metadata is derived from it.** `healthy`, `exhausted`,
   `totalHeadroom` and `bestAccount` all rest on these values, so both routing
   decisions and the operator's picture of the pool inherit the staleness.

3. **Recovery may be circular.** If a parked account is only reconsidered once
   its utilisation looks acceptable, it can never recover: the value cannot
   refresh without a request, and no request is sent while the value looks full.
   Whether dario re-probes parked accounts on a timer is not visible from the
   outside — if it does not, usable capacity stays locked out longer than
   necessary.

## Expected behaviour

Either of:

- **(a)** the reported utilisation reflects the current window, or
- **(b)** the payload states how old the reading is, so consumers can present
  it as a last-known value instead of a measurement.

## Suggested fixes

In increasing order of effort:

1. **Expose the reading's age.** Add `lastObservedAt` (or `utilAgeMs`) per
   account to `/accounts`, and to `perAccount` in `/analytics`. This is the
   smallest possible change and unblocks every downstream consumer immediately,
   even if the staleness itself stays.

2. **Expire the reading locally.** The 5h and 7d windows are known durations.
   Once `now - lastObservedAt` exceeds the window that produced a value, report
   it as unknown rather than continuing to serve the old number.

3. **Re-probe parked accounts.** After the window that caused the rejection
   would have elapsed, let one request through the parked account to refresh its
   headers, or optimistically return it to `allowed` and let the next 429 park
   it again.

Fix 1 alone would resolve the reporting side of this; 2 and 3 address the
routing side.

## Notes

- The naming already implies these semantics: `/analytics` exposes the same
  quantity as `lastUtil5h` / `lastUtil7d` — a *last* value, not a current one.
- `util5h: 1.01` (greater than 1) is consistent with a snapshot captured at the
  moment of rejection.
- Searched existing issues for `util5h` and for exhausted/recovery terms. The
  closest match is **#234** ("pool: selector doesn't fail over on auth errors
  (401/403), only on 429"), which concerns *when* to fail over rather than the
  freshness of the reported utilisation. No existing issue appears to cover this.

askalf:
Fixes #1032.

## The ask

@robincle filed this, then filed a correction against their own report — the recovery behaviour they suspected was circular turned out not to be, and they said so explicitly. That narrowed it to one clean ask, which is what this implements:

> **Suggested fix 1 — expose `lastObservedAt` (or `utilAgeMs`) per account** — is therefore still the ask, and on its own would resolve this.

## Why the reading freezes

`util5h` / `util7d` are a **snapshot of the last response an account served**. They do not tick on their own, and nothing refreshes them while an account is parked (rejected, or in auth cooldown) — so they stay pinned at whatever they read at the moment it was parked. Measured at exactly `1.01` across four samples over eleven minutes, while the account actively serving traffic moved normally and reset to `0` on schedule.

The pool *does* return parked accounts to service by itself, and the value corrects the moment it does. So this is a **reporting** problem, not a routing one: `/accounts` carried no timestamp and no freshness flag, so nothing downstream could tell a current reading from one frozen minutes ago. A dashboard reading it rendered "5-hour window full" for an account that had since reset and was free.

## The change

`updatedAt` was **already** on `RateLimitSnapshot` and already stamped by `parseRateLimits` on every response. It was simply never surfaced. Both account surfaces now carry:

```json
{ "alias": "bls-…-max20", "util5h": 1.01, "util7d": 0.62,
  "lastObservedAt": 1760000000000, "utilAgeMs": 660000, "status": "rejected" }
```

Derived by one exported `utilFreshness()` helper rather than inlined at each site — `/admin/accounts` documents itself as reporting *the same snapshot* `GET /accounts` exposes, so it must not be the one place a stale reading still looks current.

Two edges worth calling out, both asserted:

- `updatedAt === 0` is `EMPTY_SNAPSHOT`'s *never observed*. It reports as `null`, not as an age of ~56 years since epoch — otherwise an account that has simply never served a request would render as maximally stale, which is the same class of wrong reading this issue is about.
- Age floors at `0`, so clock skew cannot produce a negative.

This also removes the reporter's workaround. They had been deriving the reading's age from observed traffic, which by definition cannot work for a parked account — the exact accounts that need it.

## Deliberately not included

Suggested fix 2 — expiring the reading once its window has elapsed. It is a good idea, but `reset` and the util values are what `select()` routes on, so it wants its own change and its own reasoning about the failure modes. Fix 1 was described as resolving the issue on its own, so it ships alone. Happy to take fix 2 next if you want it.

## Tests

11 assertions in the new `test/util-freshness.mjs`, including the reported `1.01`-frozen-over-eleven-minutes shape, the live-reading control, never-observed, and clock skew. Auto-discovered by `all.test.mjs`, no registration needed.

`npm test`: **138 tests, 135 pass.** The 3 failures are `template-config-scoped-tools` / `template-interactive-tools` / `tool-advertise-respects-client`, which fail identically on unmodified `origin/master` on any machine with a live template cache — that is #1035, fixed by #1036, and unrelated to this change.

## Release note

Bumps `5.5.25` → `5.5.28`. There are now three fix PRs in flight each carrying a bump — #1034 (5.5.26), #1036 (5.5.27), this one (5.5.28) — plus bot PR #1028 also claiming 5.5.26. Whichever land after the first will need `gh pr update-branch` and a re-bump on the version line.


sprayberry-redline (dismissed):

This PR was authored by the same party approving it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

## Scope

`src/pool.ts` gains `utilFreshness()` and its `UtilFreshness` type; `src/proxy.ts` spreads it into both account surfaces; `src/admin-api.ts` widens `AdminAccountLive`; `test/util-freshness.mjs` is new. Rebased onto master at 5.5.27 with no source conflicts — #1036 touched `live-fingerprint.ts` / `cc-template.ts`, this touches the pool and proxy.

## Verified

- **The reported behaviour is real and correctly characterised.** `util5h` / `util7d` come from `parseRateLimits`, which only runs on a response. Nothing refreshes them while an account is parked, so they hold the value they had when it was parked — matching the reporter's `1.01` across four samples over eleven minutes, next to a serving account whose figures moved and reset normally.
- **It is a reporting fix, not a routing one, and the PR is honest about that.** The reporter filed a correction withdrawing the "recovery may be circular" hypothesis; that withdrawal is respected rather than quietly fixed anyway. Parked accounts do return to service on their own, and the value self-corrects when they do.
- **No new data collection was needed.** `updatedAt` was already on `RateLimitSnapshot` and already stamped by `parseRateLimits` on every response — the field existed and was simply never surfaced. That is the cheapest possible shape for this fix and it is the one taken.
- **Both surfaces move together.** `/admin/accounts` documents itself as reporting the same snapshot `GET /accounts` exposes, so updating only the public one would have left the admin surface as the single place a stale reading still looked current.
- **The edges that matter are asserted.** `updatedAt === 0` (`EMPTY_SNAPSHOT`, never observed) reports `null` rather than an age of ~56 years — which would be the same class of wrong reading this issue is about — and the age floors at 0 so clock skew cannot go negative. 11 assertions including the reported frozen-`1.01` shape; full suite 138/138.

## Findings against this PR

1. **The TUI does not render the new fields, which is where an operator is most likely to look.** The fix lands on `/accounts` and `/admin/accounts`, so a dashboard consuming the JSON is served — the reporter's case exactly. But someone watching `dario`'s own accounts view still sees a frozen `util5h` with nothing to indicate its age. That is the same defect in a different surface, and the PR does not say so. Follow-up.

2. **`AdminAccountLive` gains two required fields, which is a breaking change for any out-of-tree implementer of that interface.** It is internal today and every in-tree construction site is updated, so nothing breaks here. Worth knowing it is a typed contract, not just a response shape.

3. **Nit: `utilFreshness` returns an object that is spread at both call sites**, so the field names are fixed by the helper rather than by the surface. That is the right call for keeping the two consistent, but it means renaming a field is a two-surface change with no compiler assistance at the spread sites.

## Correctly scoped out

Suggested fix 2 from the issue — expiring the reading once its window has elapsed — is deliberately not here, and the reasoning holds: `reset` and the util values are what `select()` routes on, so changing what a lapsed window reads is a routing change wearing a reporting change's clothes. The reporter said fix 1 resolves the issue on its own. Shipping it alone, and asking them what a lapsed window *should* read as before touching it, is the right sequence.

## Verdict

Approving. Small, well-evidenced, uses data that already existed, and it removes a downstream workaround that could not have worked — the reporter had been deriving the reading's age from observed traffic, which by definition fails for exactly the parked accounts that need it.

sprayberry-redline (approved):

This PR was authored by the same party approving it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

## Scope

`src/pool.ts` gains `utilFreshness()` and its `UtilFreshness` type; `src/proxy.ts` spreads it into both account surfaces; `src/admin-api.ts` widens `AdminAccountLive`; `test/util-freshness.mjs` is new. Rebased onto master at 5.5.27 with no source conflicts — #1036 touched `live-fingerprint.ts` / `cc-template.ts`, this touches the pool and proxy.

## Verified

- **The reported behaviour is real and correctly characterised.** `util5h` / `util7d` come from `parseRateLimits`, which only runs on a response. Nothing refreshes them while an account is parked, so they hold the value they had when it was parked — matching the reporter's `1.01` across four samples over eleven minutes, next to a serving account whose figures moved and reset normally.
- **It is a reporting fix, not a routing one, and the PR is honest about that.** The reporter filed a correction withdrawing the "recovery may be circular" hypothesis; that withdrawal is respected rather than quietly fixed anyway. Parked accounts do return to service on their own, and the value self-corrects when they do.
- **No new data collection was needed.** `updatedAt` was already on `RateLimitSnapshot` and already stamped by `parseRateLimits` on every response — the field existed and was simply never surfaced. That is the cheapest possible shape for this fix and it is the one taken.
- **Both surfaces move together.** `/admin/accounts` documents itself as reporting the same snapshot `GET /accounts` exposes, so updating only the public one would have left the admin surface as the single place a stale reading still looked current.
- **The edges that matter are asserted.** `updatedAt === 0` (`EMPTY_SNAPSHOT`, never observed) reports `null` rather than an age of ~56 years — which would be the same class of wrong reading this issue is about — and the age floors at 0 so clock skew cannot go negative. 11 assertions including the reported frozen-`1.01` shape; full suite 138/138.

## Findings against this PR

1. **The TUI does not render the new fields, which is where an operator is most likely to look.** The fix lands on `/accounts` and `/admin/accounts`, so a dashboard consuming the JSON is served — the reporter's case exactly. But someone watching `dario`'s own accounts view still sees a frozen `util5h` with nothing to indicate its age. That is the same defect in a different surface, and the PR does not say so. Follow-up.

2. **`AdminAccountLive` gains two required fields, which is a breaking change for any out-of-tree implementer of that interface.** It is internal today and every in-tree construction site is updated, so nothing breaks here. Worth knowing it is a typed contract, not just a response shape.

3. **Nit: `utilFreshness` returns an object that is spread at both call sites**, so the field names are fixed by the helper rather than by the surface. That is the right call for keeping the two consistent, but it means renaming a field is a two-surface change with no compiler assistance at the spread sites.

## Correctly scoped out

Suggested fix 2 from the issue — expiring the reading once its window has elapsed — is deliberately not here, and the reasoning holds: `reset` and the util values are what `select()` routes on, so changing what a lapsed window reads is a routing change wearing a reporting change's clothes. The reporter said fix 1 resolves the issue on its own. Shipping it alone, and asking them what a lapsed window *should* read as before touching it, is the right sequence.

## Verdict

Approving. Small, well-evidenced, uses data that already existed, and it removes a downstream workaround that could not have worked — the reporter had been deriving the reading's age from observed traffic, which by definition fails for exactly the parked accounts that need it.

PR #1038, opened 2026-08-20
fix: /health must ask the same question select() asks (#1030)

### LINKED ISSUE #1030: /health reports healthy for a pool select() cannot serve from
`/health` and `select()` decide "can this pool serve" with different predicates, so a pool that fails every request reports `healthy` with `authenticated: true`.

`AccountPool.select()` (src/pool.ts:414) filters on three conditions:

```ts
const eligible = all.filter(a =>
  a.rateLimit.status !== 'rejected' &&
  a.expiresAt > now + 30_000 &&
  !isInAuthCooldown(a, now),
);
```

`derivePoolStatus` (src/health-response.ts) filters on one:

```ts
const usable = accounts.filter((a) => !a.inAuthCooldown);
```

`PoolAccountStatusLike` is `{ expiresAt, inAuthCooldown }` — so the expiry is already in the type, and is used only to render the "expires in" string, never for the usable decision. `rateLimit.status` is not in the type at all: src/proxy.ts:1974 maps the accounts down to those two fields before handing them over, so the surface cannot see the first of select()'s three conditions even if it wanted to.

## Reproduction

Against master, built with `npm run build`:

```js
import { AccountPool, isInAuthCooldown } from './dist/pool.js';
import { derivePoolStatus } from './dist/health-response.js';

const HOUR = 3_600_000;
const creds = (expiresAt) => ({ accessToken: 't', refreshToken: 'r', expiresAt, deviceId: 'd', accountUuid: 'u' });

function report(label, build) {
  const pool = new AccountPool();
  build(pool);
  const now = Date.now();
  // The exact mapping proxy.ts:1974 does before calling derivePoolStatus.
  const health = derivePoolStatus(
    pool.all().map((a) => ({ expiresAt: a.expiresAt, inAuthCooldown: isInAuthCooldown(a, now) })),
    now, false,
  );
  const picked = pool.select();
  const servable = picked
    ? (picked.rateLimit.status !== 'rejected' && picked.expiresAt > now + 30_000 && !isInAuthCooldown(picked, now))
    : false;
  console.log(`${label}: /health=${health.status} authenticated=${health.authenticated} select()=${picked?.alias ?? 'null'} servable=${servable}`);
}

report('all tokens expired', (p) => { p.add('a', creds(Date.now() - HOUR)); p.add('b', creds(Date.now() - 2 * HOUR)); });
report('all rate-limited', (p) => { p.add('a', creds(Date.now() + HOUR)); p.get('a').rateLimit.status = 'rejected'; });
report('healthy control', (p) => p.add('a', creds(Date.now() + HOUR)));
```

```
all tokens expired: /health=healthy authenticated=true select()=b servable=false
all rate-limited:   /health=healthy authenticated=true select()=a servable=false
healthy control:    /health=healthy authenticated=true select()=a servable=true
```

In the first two rows `select()` returns an account that fails its own eligibility test — the "all accounts exhausted" fallback below the filter — so the request goes out and 401s or 429s. `/health` says the proxy is fine throughout, which is the case it exists to catch: a monitor watching `/health` sees nothing, and `dario doctor` reads the same derivation.

The expired case is the one I would weight highest. Refresh runs on a 60s tick against a 45-minute margin, so a token that is expired *now* is one the refresh loop has already failed to roll — it is not about to fix itself, and it needs `dario login`.

## Note on the shape

The predicate is also inline at four sites in pool.ts (414, 478, 564, 621) rather than named once, which is presumably how the fifth reader came to implement a subset. Naming it — returning the reason, not just a boolean, since the surfaces want to say *why* — would let `/health` ask the same question the router asks, and would make the 503 body and doctor's routing line answer from it too.

I have a version of that running on a fork, but the fork has diverged far enough from master (pool.ts is roughly twice the size) that a patch from it would be importing an architecture rather than fixing this, so I am reporting rather than sending a PR. Happy to write one against master in whatever shape you prefer, or to leave it with you.

askalf:
Fixes #1030.

@chaogebaba filed this with the code references, a runnable reproduction, and the correct diagnosis. This implements the shape they proposed — naming the predicate and returning the reason rather than a boolean — which is what they offered to send a PR for. Taking it here so it isn't blocked on a round trip; the analysis is theirs.

## The bug

`select()` filters on three conditions. `derivePoolStatus` filtered on one:

```ts
// pool.ts select()
a.rateLimit.status !== 'rejected' && a.expiresAt > now + 30_000 && !isInAuthCooldown(a, now)

// health-response.ts derivePoolStatus()
accounts.filter((a) => !a.inAuthCooldown)
```

And it could not have done better: `PoolAccountStatusLike` was `{ expiresAt, inAuthCooldown }`, and `proxy.ts` mapped the accounts down to those two fields before handing them over — so the surface could not see `rateLimit.status` at all, and used `expiresAt` only to render the "expires in" string.

Result: a pool that fails every request reports `healthy` / `authenticated: true`. That is the exact case `/health` exists to catch — a monitor watching it sees nothing, and `dario doctor` reads the same derivation.

## Reproduction

The reporter's script, verbatim, before → after:

```
BEFORE (origin/master)
all tokens expired   /health=healthy  authenticated=true  select()=b  servable=false <-- LIES
all rate-limited     /health=healthy  authenticated=true  select()=a  servable=false <-- LIES
mixed failure        /health=healthy  authenticated=true  select()=b  servable=false <-- LIES
healthy control      /health=healthy  authenticated=true  select()=a  servable=true

AFTER
all tokens expired   /health=broken   authenticated=false select()=b  servable=false
                     reason: all tokens expired — run `dario login`
all rate-limited     /health=broken   authenticated=false select()=a  servable=false
                     reason: all accounts rate-limited
mixed failure        /health=broken   authenticated=false select()=b  servable=false
                     reason: no account can serve (rate-limited, token-expired)
healthy control      /health=healthy  authenticated=true  select()=a  servable=true
```

## The change

> The predicate is also inline at four sites in pool.ts (414, 478, 564, 621) rather than named once, which is presumably how the fifth reader came to implement a subset. Naming it — returning the reason, not just a boolean, since the surfaces want to say *why* — would let `/health` ask the same question the router asks.

That is what this does.

- `accountIneligibility(account, now)` returns `'rate-limited' | 'token-expired' | 'auth-cooldown' | null`, with `isAccountEligible()` as the boolean form. All four `pool.ts` sites now call it.
- `PoolAccountStatusLike.ineligible` is **required**, not optional. `health-response.ts` stays free of pool machinery — the reason is computed by the caller and passed in — and requiring it means a caller cannot silently reintroduce the subset.
- `broken` now names the cause instead of always claiming auth-cooldown. Mixed causes are listed rather than collapsed onto one, and an unrecognised reason falls through to a generic line rather than being treated as usable.
- `dario doctor` reads `/health`, so it inherits both the corrected verdict and the reason with no change of its own.

## Behaviour change — please read this bit

An expired-token pool now reports `broken` / 503 where it previously reported `healthy` / 200.

`test/health-response.mjs` explicitly asserted the old reading, under the heading *"expired-but-usable clamps to 0h 0m"* and the comment *"healthy (background refresh will roll it)"*. That reasoning does not hold, and the reporter is right about why:

> Refresh runs on a 60s tick against a 45-minute margin, so a token that is expired *now* is one the refresh loop has already failed to roll — it is not about to fix itself, and it needs `dario login`.

I have rewritten that test to assert the corrected behaviour and left the reasoning in place next to it, rather than deleting it. Flagging it prominently because it is the one place this PR changes what a passing test claimed, and if you disagree, this is the line to disagree with.

Uptime monitors keyed on the status code will now correctly alarm on an expired-token pool that previously looked fine. That is the intent, but it is a live-behaviour change on a monitored surface.

## Not included

The request-path 503 body could also answer from this predicate and currently does not. Different path, and it wants its own change.

## Tests

`test/health-response.mjs`: **111 pass, 0 fail**, including new coverage for all-rate-limited, mixed causes, one-eligible-account-keeps-it-healthy, and an unrecognised reason not being treated as usable.

`npm test`: 137 tests, 134 pass. The 3 failures are the #1035 template ones, identical on unmodified `origin/master`, fixed by #1036.

## Release note

Bumps to `5.5.29`. Four fix PRs are now in flight each carrying a bump — #1034 (5.5.26), #1036 (5.5.27), #1037 (5.5.28), this one (5.5.29) — plus bot PR #1028 also on 5.5.26. Whichever land after the first need `gh pr update-branch` and a re-bump on the version line.


sprayberry-redline (dismissed):

This PR was authored by the same party approving it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

The analysis behind the fix is @chaogebaba's, contributed on #1030 with a runnable reproduction and the exact code references; the implementation and this approval are not theirs.

## Scope

`src/pool.ts` (named predicate, four inline sites consolidated), `src/health-response.ts` (`derivePoolStatus` asks the router's question), `src/proxy.ts` (passes the reason through), `test/health-response.mjs`. Rebased onto master at 5.5.27; no source conflict with #1036.

## Verified

- **The bug is real and the reproduction runs.** Against unmodified master, a pool with every token expired — and separately one with every account rate-limited — reports `/health=healthy`, `authenticated=true` while `select()` returns an account that fails its own eligibility test. After the change both report `broken` with the cause named, and the healthy control is untouched.
- **The surface could not have got this right.** `PoolAccountStatusLike` was `{ expiresAt, inAuthCooldown }`, and `proxy.ts` mapped accounts down to those two fields before calling. `rateLimit.status` was not in the type at all, so the narrower predicate was not an oversight in `derivePoolStatus` so much as the only question it was given the data to ask.
- **Making the field required is the right lever.** Passing the reason in keeps `health-response.ts` free of pool imports — its header states that as a design property — while requiring it means a future caller cannot silently reintroduce the subset. The failure mode was a reader re-deriving a predicate; making the derivation unrepresentable at the boundary addresses that, where fixing only the instance would not.
- **Reason handling is not sloppy.** Mixed causes are listed rather than collapsed onto one, and an unrecognised reason falls through to a generic line instead of being treated as usable — asserted explicitly, which matters because "unknown means fine" is the exact shape of the original bug.
- **`dario doctor` inherits the fix for free** by reading `/health`, so the two do not diverge again.
- 137/137.

## The one thing to weigh before merging

**This changes what a monitored endpoint reports, and monitors will newly alarm.**

An expired-token pool now returns `broken` / 503 where it returned `healthy` / 200. `test/health-response.mjs` asserted the old reading under the heading *"expired-but-usable"* with the comment *"healthy (background refresh will roll it)"*. That test is rewritten here, which means this PR changes the meaning of a previously-passing assertion — the one thing in it that cannot be verified by the tests alone, because the tests are what changed.

The justification is the reporter's and it holds: refresh runs on a 60s tick against a 45-minute margin, so a token expired *now* is one the refresh loop has already failed to roll. It is not about to fix itself. `select()` has always refused to serve from it; only `/health` claimed otherwise.

So the change is correct, and the alarms it produces are true positives that were previously suppressed. But anything keyed on dario's `/health` status code — uptime monitoring, `depends_on: service_healthy`, container restart policies — may go red on pools that have been quietly broken for some time. That is the intended outcome and it should not be a surprise.

## Findings against this PR

1. **Fixed during review: a dead, duplicated summary table.** The PR originally added `INELIGIBILITY_SUMMARY` to `pool.ts` — exported, never imported, carrying the same three strings as `POOL_DEAD_SUMMARY` in `health-response.ts`. In a PR whose entire thesis is that a predicate duplicated across call sites drifts, that is the same mistake in miniature. Removed in `1d53370`; the local table in `health-response.ts` is the one that should exist, since importing from `pool.ts` would cost that module its no-pool-machinery property.

2. **The request-path 503 body still does not answer from this predicate.** The PR says so. It means an operator can now get an accurate reason from `/health` and a vaguer one from a failing request, which is a smaller version of the inconsistency being fixed. Follow-up, not a blocker.

## Verdict

Approving. The fix is the shape the reporter proposed, it consolidates four inline copies of the predicate rather than patching the fifth reader, and the required-field boundary makes the class of bug hard to reintroduce.

Merge deliberately, not casually — finding-free, but the `/health` status-code change is live-behaviour on a surface other systems watch.

sprayberry-redline (dismissed):

This PR was authored by the same party approving it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

The analysis behind the fix is @chaogebaba's, contributed on #1030 with a runnable reproduction and the exact code references; the implementation and this approval are not theirs.

## Scope

`src/pool.ts` (named predicate, four inline sites consolidated), `src/health-response.ts` (`derivePoolStatus` asks the router's question), `src/proxy.ts` (passes the reason through), `test/health-response.mjs`. Rebased onto master at 5.5.27; no source conflict with #1036.

## Verified

- **The bug is real and the reproduction runs.** Against unmodified master, a pool with every token expired — and separately one with every account rate-limited — reports `/health=healthy`, `authenticated=true` while `select()` returns an account that fails its own eligibility test. After the change both report `broken` with the cause named, and the healthy control is untouched.
- **The surface could not have got this right.** `PoolAccountStatusLike` was `{ expiresAt, inAuthCooldown }`, and `proxy.ts` mapped accounts down to those two fields before calling. `rateLimit.status` was not in the type at all, so the narrower predicate was not an oversight in `derivePoolStatus` so much as the only question it was given the data to ask.
- **Making the field required is the right lever.** Passing the reason in keeps `health-response.ts` free of pool imports — its header states that as a design property — while requiring it means a future caller cannot silently reintroduce the subset. The failure mode was a reader re-deriving a predicate; making the derivation unrepresentable at the boundary addresses that, where fixing only the instance would not.
- **Reason handling is not sloppy.** Mixed causes are listed rather than collapsed onto one, and an unrecognised reason falls through to a generic line instead of being treated as usable — asserted explicitly, which matters because "unknown means fine" is the exact shape of the original bug.
- **`dario doctor` inherits the fix for free** by reading `/health`, so the two do not diverge again.
- 137/137.

## The one thing to weigh before merging

**This changes what a monitored endpoint reports, and monitors will newly alarm.**

An expired-token pool now returns `broken` / 503 where it returned `healthy` / 200. `test/health-response.mjs` asserted the old reading under the heading *"expired-but-usable"* with the comment *"healthy (background refresh will roll it)"*. That test is rewritten here, which means this PR changes the meaning of a previously-passing assertion — the one thing in it that cannot be verified by the tests alone, because the tests are what changed.

The justification is the reporter's and it holds: refresh runs on a 60s tick against a 45-minute margin, so a token expired *now* is one the refresh loop has already failed to roll. It is not about to fix itself. `select()` has always refused to serve from it; only `/health` claimed otherwise.

So the change is correct, and the alarms it produces are true positives that were previously suppressed. But anything keyed on dario's `/health` status code — uptime monitoring, `depends_on: service_healthy`, container restart policies — may go red on pools that have been quietly broken for some time. That is the intended outcome and it should not be a surprise.

## Findings against this PR

1. **Fixed during review: a dead, duplicated summary table.** The PR originally added `INELIGIBILITY_SUMMARY` to `pool.ts` — exported, never imported, carrying the same three strings as `POOL_DEAD_SUMMARY` in `health-response.ts`. In a PR whose entire thesis is that a predicate duplicated across call sites drifts, that is the same mistake in miniature. Removed in `1d53370`; the local table in `health-response.ts` is the one that should exist, since importing from `pool.ts` would cost that module its no-pool-machinery property.

2. **The request-path 503 body still does not answer from this predicate.** The PR says so. It means an operator can now get an accurate reason from `/health` and a vaguer one from a failing request, which is a smaller version of the inconsistency being fixed. Follow-up, not a blocker.

## Verdict

Approving. The fix is the shape the reporter proposed, it consolidates four inline copies of the predicate rather than patching the fifth reader, and the required-field boundary makes the class of bug hard to reintroduce.

Merge deliberately, not casually — finding-free, but the `/health` status-code change is live-behaviour on a surface other systems watch.

sprayberry-redline (approved):

This PR was authored by the same party approving it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

The analysis behind the fix is @chaogebaba's, contributed on #1030 with a runnable reproduction and the exact code references; the implementation and this approval are not theirs.

## Scope

`src/pool.ts` (named predicate, four inline sites consolidated), `src/health-response.ts` (`derivePoolStatus` asks the router's question), `src/proxy.ts` (passes the reason through), `test/health-response.mjs`. Rebased onto master at 5.5.27; no source conflict with #1036.

## Verified

- **The bug is real and the reproduction runs.** Against unmodified master, a pool with every token expired — and separately one with every account rate-limited — reports `/health=healthy`, `authenticated=true` while `select()` returns an account that fails its own eligibility test. After the change both report `broken` with the cause named, and the healthy control is untouched.
- **The surface could not have got this right.** `PoolAccountStatusLike` was `{ expiresAt, inAuthCooldown }`, and `proxy.ts` mapped accounts down to those two fields before calling. `rateLimit.status` was not in the type at all, so the narrower predicate was not an oversight in `derivePoolStatus` so much as the only question it was given the data to ask.
- **Making the field required is the right lever.** Passing the reason in keeps `health-response.ts` free of pool imports — its header states that as a design property — while requiring it means a future caller cannot silently reintroduce the subset. The failure mode was a reader re-deriving a predicate; making the derivation unrepresentable at the boundary addresses that, where fixing only the instance would not.
- **Reason handling is not sloppy.** Mixed causes are listed rather than collapsed onto one, and an unrecognised reason falls through to a generic line instead of being treated as usable — asserted explicitly, which matters because "unknown means fine" is the exact shape of the original bug.
- **`dario doctor` inherits the fix for free** by reading `/health`, so the two do not diverge again.
- 137/137.

## The one thing to weigh before merging

**This changes what a monitored endpoint reports, and monitors will newly alarm.**

An expired-token pool now returns `broken` / 503 where it returned `healthy` / 200. `test/health-response.mjs` asserted the old reading under the heading *"expired-but-usable"* with the comment *"healthy (background refresh will roll it)"*. That test is rewritten here, which means this PR changes the meaning of a previously-passing assertion — the one thing in it that cannot be verified by the tests alone, because the tests are what changed.

The justification is the reporter's and it holds: refresh runs on a 60s tick against a 45-minute margin, so a token expired *now* is one the refresh loop has already failed to roll. It is not about to fix itself. `select()` has always refused to serve from it; only `/health` claimed otherwise.

So the change is correct, and the alarms it produces are true positives that were previously suppressed. But anything keyed on dario's `/health` status code — uptime monitoring, `depends_on: service_healthy`, container restart policies — may go red on pools that have been quietly broken for some time. That is the intended outcome and it should not be a surprise.

## Findings against this PR

1. **Fixed during review: a dead, duplicated summary table.** The PR originally added `INELIGIBILITY_SUMMARY` to `pool.ts` — exported, never imported, carrying the same three strings as `POOL_DEAD_SUMMARY` in `health-response.ts`. In a PR whose entire thesis is that a predicate duplicated across call sites drifts, that is the same mistake in miniature. Removed in `1d53370`; the local table in `health-response.ts` is the one that should exist, since importing from `pool.ts` would cost that module its no-pool-machinery property.

2. **The request-path 503 body still does not answer from this predicate.** The PR says so. It means an operator can now get an accurate reason from `/health` and a vaguer one from a failing request, which is a smaller version of the inconsistency being fixed. Follow-up, not a blocker.

## Verdict

Approving. The fix is the shape the reporter proposed, it consolidates four inline copies of the predicate rather than patching the fifth reader, and the required-field boundary makes the class of bug hard to reintroduce.

Merge deliberately, not casually — finding-free, but the `/health` status-code change is live-behaviour on a surface other systems watch.

Issue #1039, opened 2026-08-20
Scheduled workflows fire on all 54 forks — guard the cron, not the workflow


askalf:
Follow-up to #1029, which fixes the one instance that actually reaches outward. Opening this for the rest so that PR stays a one-line security fix rather than a sweep.

## State

**16 of 16** workflows carrying a `cron:` have no repository guard:

```
auto-merge-sweep                cc-oauth-health          scorecard
cc-billing-classifier-canary    cflite                   sdk-drift-watch
cc-drift-auto-release  <- #1029 codeql                   self-healing-supervisor
cc-drift-template-watch         dario-doctor-watch       stale
cc-drift-watch                  npm-drift-watch          wire-drift-self-hosted
cc-drift-watcher-liveness
```

@chaogebaba found this on their own fork, where `cc-drift-auto-release` had been firing hourly for days.

## Severity, stated honestly

These are not equivalent and I don't want to inflate the finding.

- **`cc-drift-auto-release` is the only one that reaches outward** — it tags, cuts a Release, and attempts `npm publish`. #1029 fixes it. The publish itself is bound by OIDC trusted publishing to this repository, but the tag and Release are created before that rejection, and the duplicate-tag gate reads `npm view @askalf/dario`, so a fork decides whether to release by consulting *this* package's registry entry.
- **The rest are waste and noise, not exposure.** A fork's `GITHUB_TOKEN` is scoped to the fork, so the issue-filing watchers file into the fork. `wire-drift-self-hosted` targets runners a fork doesn't have. `codeql` / `scorecard` / `cflite` burn fork CI minutes.

Still worth closing: 54 forks × ~16 jobs on hourly-to-daily crons is a lot of other people's CI minutes spent on our behalf, and several make decisions by reading *this* repo's published state, which is wrong on a fork by construction.

## Proposed shape — guard the trigger, not the workflow

The obvious fix is `if: github.repository == 'askalf/dario'` on every job, as #1029 does. That is right for the release job and **wrong as a blanket rule**: it would also disable `codeql`, `scorecard` and `cflite` on pushes and PRs in a fork, and a forker running security scanning on their own copy is a legitimate thing to do. We shouldn't break it to save cron minutes.

So guard the schedule specifically:

```yaml
if: github.event_name != 'schedule' || github.repository == 'askalf/dario'
```

Fork keeps its push/PR CI; the cron stops. Two groups:

[...]

PR #1040, opened 2026-08-20
ci: heal drift-bot PRs that lose a version race

askalf:
Follow-on to #1039. This is the loop that has been generating the PR backlog, not a symptom of it.

## The loop

Every drafter (`auto-draft-drift-fix`, `label-sync`, `rebake-release-prep`) computes its bump with `bumpPackageJsonPatch()` off the `package.json` in its own checkout — master **at draft time**. So:

- two PRs drafted before either merges claim the **same** version — #954 and #955 both took `5.5.10` off `5.5.9`;
- a PR drafted before an unrelated release lands claims one that is **already published** — #1027 bumped `5.5.24` → `5.5.25` with `v5.5.25` tagged and on npm.

`version-advances.sh` already catches this and refuses to arm the loser. That refusal is *correct*: merging it would either fast-exit `cc-drift-auto-release`'s duplicate-tag guard green having shipped nothing, or move master's version backwards below a published tag.

**But refusing is where the automation stopped.** Nothing renumbers the loser. It sits open, drifts BEHIND as master moves, and waits for a human — and since `master-protection` requires an approving review, that hand-fix is itself a push, which dismisses any approval the PR had collected and returns it to the review queue. Each collision costs a full round trip, and on a busy drift cycle collisions arrive faster than round trips complete.

That is the backlog, stated mechanically.

## What was already here

`scripts/resolve-release-conflicts.mjs` implements exactly the file-level rule this needs — *keep the higher version, keep both CHANGELOG sections* — is tested, and is **wired into no workflow at all**:

```
scripts that exist but are NEVER referenced by a workflow:
  resolve-release-conflicts.mjs      <-- this one
```

That is most of why the manual resolutions kept happening. The tool existed and never ran.

## The change

**`scripts/renumber-release-version.mjs`** (new, pure core + thin CLI). Above the base tip it is a no-op. At or below it, it sets the version to `bumpPatch(tip)` and renames the one CHANGELOG heading that carried the old version. It **refuses** when the CHANGELOG does not carry exactly one such heading — a renumbered `package.json` whose notes still announce the old version ships a release named after a different version, which is worse than an untouched PR.

**`.github/workflows/drift-pr-heal.yml`** (new). Hourly at `:47`, clear of the other four crons. For open drift-bot PRs in `DIRTY` or `BEHIND`: merge the base, resolve with the existing resolver, renumber, re-ask `version-advances.sh`, push. Any refusal resets the branch and leaves the PR exactly as found.

Four things I want a reviewer to actually check, because they are the load-bearing decisions:

1. **It refuses to run without `DARIO_DRIFT_BOT_PAT`.** A push made with `GITHUB_TOKEN` fires no `pull_request` event, so healing with it would update the branch and leave it with **no runs for the required checks** — stuck in a new and much less visible way. Skipping leaves it visibly stuck instead. That is deliberate, and it means this ships inert if the PAT is ever rotated away.
2. **Fork-guarded**, so this doesn't become the 17th unguarded scheduled workflow on the same day #1039 was filed.
3. **Serialized** by concurrency group. Two heals would compute the same floor from the same tip and renumber two PRs onto one version — the exact collision it exists to clear.
4. **Branch-prefix filter matches `auto-merge-sweep.yml`.** This job pushes to branches, so that filter is the only thing keeping it off human PRs. It needs to stay in step.

## Also: the bot has been telling maintainers not to look

The drift bot's own PR body said:

> `master` requires no review, so no human approval gates it.

That stopped being true when `master-protection` began requiring an approving review. It is printed on **every drift PR**, and it is the sentence that tells a maintainer they can ignore one. Corrected to say an approving review is required and is dismissed on any later push.

## Verification

30 assertions in `test/renumber-release-version.mjs` — both production collision shapes, the no-op path, five refusal paths, CRLF/LF round-tripping.

Verified end-to-end against the **real stuck PR #1027**, not just fixtures:

```
PR #1027 version : 5.5.25
master tip       : 5.5.27
[renumber] renumbered 5.5.25 -> 5.5.28
CHANGELOG        : ## [5.5.28] - 2026-08-20
re-run           : [renumber] unchanged 5.5.28 — already advances past 5.5.27
version-advances : true / "version advances 5.5.27 -> 5.5.28"
```

The gate that had been refusing #1027 now passes it. Both workflow `run:` blocks were extracted and checked with `bash -n`; the YAML parses and the guard/permissions are as intended.

Full suite **138/138**.

## Caveat

This PR bumps to `5.5.31` — the sixth claimant in a hand-maintained version ladder built over the last few hours (#1034 → 5.5.28, #1037 → 5.5.29, #1038 → 5.5.30, this → 5.5.31). I laddered those by hand precisely because this automation does not exist yet. If any of them merge in a different order, this one needs the same manual renumber it is written to eliminate — which is either the best argument for it or an unkind joke, depending on the hour.


askalf:
Closing and reopening to re-fire CI. The `synchronize` event for `c30f375` was partially delivered — `Auto-merge bot PRs` ran, and `CI` / `CodeQL` / `actionlint` / `Label` / `Triage` never got created. Nine minutes with no runs appearing, and githubstatus reports all systems normal, so this is dropped-webhook rather than queueing.

That is the exact failure mode `auto-merge-sweep.yml` was built for, and its header names the remedy: *"recovery meant manually re-running, pushing a commit, or updating the PR."* All three required workflows listen on `pull_request` with no `types:` filter, so `reopened` re-fires them without changing the head SHA.

Noting it here because it is a live instance of the reconciler-gap class, and worth a look at whether the sweep should cover *missing checks* on a bot PR and not just *unarmed* ones.

askalf:
Correction to the comment above: **that diagnosis was wrong.** It was not a dropped webhook, and the close/reopen was not what fixed it.

This PR had gone `DIRTY` when #1037 merged and moved master to 5.5.29. GitHub evaluates `pull_request` workflows against the computed merge commit, which does not exist while a PR is conflicting — so `CI` / `CodeQL` / `actionlint` stopped being created entirely, while `pull_request_target` workflows (`Spam removal`, `Auto-merge bot PRs`) kept firing because they run against the base. That asymmetry was the actual tell, and I read it as evidence of webhook loss instead of the ordinary consequence of a conflict.

What settled it: `pull_request` runs were being created normally for #1038 on the same repo at 17:13, minutes after this branch stopped getting them. A repo-wide webhook problem cannot be selective like that.

Resolving the conflict brought all six workflows back on the next push, with no intervention. The close/reopen did nothing except fire the two `pull_request_target` workflows a second time.

So there is no reconciler gap here, and the suggestion in the previous comment — that `auto-merge-sweep.yml` should also cover *missing checks* — should be disregarded. A bot PR with no checks is usually just a conflicting bot PR, and the fix for that is this PR, not another sweep.

Leaving both comments up rather than editing: the wrong reading is a plausible one and the distinguishing evidence is worth recording next to it.

sprayberry-redline (dismissed):

This PR was authored by the same party approving it, and so was the issue behind it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

Unusually for this batch, the change was exercised against live traffic before review: the same version-collision ladder it exists to eliminate produced three real conflicts today, and the tooling here resolved all three. Two defects surfaced that way. Both are fixed; both are described below rather than quietly folded in.

## Scope

`scripts/renumber-release-version.mjs` (new), `.github/workflows/drift-pr-heal.yml` (new), a one-line correction to the drift bot's PR body in `scripts/auto-draft-drift-fix.mjs`, 30 new assertions, plus the bump and CHANGELOG.

## The problem is real and already documented in-tree

`version-advances.sh` opens by describing it: *"Two bot PRs can independently bump package.json to the SAME version, because each is drafted off the current release without seeing the other (#954 and #955 both took 5.5.10 off 5.5.9, with different dates)."* That script refuses to arm the loser — correctly, since merging it either fast-exits the duplicate-tag guard green having shipped nothing, or moves master's version below a published tag.

Nothing then renumbers the loser. It sits open, drifts BEHIND, and waits. With `master-protection` requiring an approving review and dismissing it on push, the hand-fix costs a full round trip. That is the backlog mechanism, and it is the thing this PR removes.

## Verified

- **`resolve-release-conflicts.mjs` was wired into no workflow at all.** Confirmed by enumerating every `scripts/*.mjs` referenced from `.github/workflows/`. The file-level rule was written, tested, and never ran — which is most of why the manual resolutions kept recurring.
- **Verified against production data, not fixtures.** Against the real stuck #1027 (`5.5.25`, published; master at `5.5.27`) the renumberer produced `5.5.28`, moved the CHANGELOG heading with it, was idempotent on re-run, and flipped `version-advances.sh` from refusing to `"version advances 5.5.27 -> 5.5.28"`.
- **Exercised on three further live conflicts** during today's ladder — #1037 twice and #1038 — resolving `package.json` and `CHANGELOG.md` each time and correctly reporting the renumber as a no-op where the version already cleared the tip.
- **Refusal paths are asserted, not assumed**: zero or multiple CHANGELOG headings, a `package.json` with no version, a non-`X.Y.Z` floor, and a prerelease version all refuse. CRLF and LF both round-trip, which matters because a silent regex miss on CRLF would present as "nothing to rename" — safe, but permanently useless on a Windows checkout with no log line explaining it.
- Both workflow `run:` blocks extracted and checked with `bash -n`; YAML parses; guard and permissions are as intended. Full suite 138/138.

## Defects found by dogfooding, and fixed

1. **A source conflict would have killed the whole tick.** `resolve-release-conflicts.mjs` only touches the two version files, but a drift PR can also conflict in source — observed on #1038, where #1037 had added `utilFreshness` to the same pool import line #1038 was adding `accountIneligibility` to. The workflow would have resolved the version files, run `git add`, then failed on `git commit` with `src/proxy.ts` still unmerged. Under `set -e` that ends the job with every later PR in the batch unprocessed — one awkward PR silently starving the rest. Now the unmerged set is inspected before anything is resolved, and such a PR is skipped with the offending paths named.

2. **The first guard refused everything.** It joined the paths with `tr` and matched a `case` pattern; `tr` left a trailing space, so no valid input ever matched. It failed closed, which is the right direction, but it would have made the workflow a permanent no-op that looked like it was working. Replaced with `grep -vxE` over the path list — order-independent, separator-independent — and exercised against all six real shapes.

Worth stating plainly: both defects were in code already described as ready, and both were found by running it rather than reading it.

## Points a reviewer should weigh

- **It refuses to run without `DARIO_DRIFT_BOT_PAT`.** A push made with `GITHUB_TOKEN` fires no `pull_request` event, so healing with it would leave the PR with no runs for the required contexts — stuck in a newer and less visible way. Skipping keeps it visibly stuck. The consequence is that this ships **inert** if the PAT is ever rotated away, and the failure is a red job rather than a silent skip, which is the correct trade but should be a conscious one.
- **The branch-prefix filter is the only thing keeping this off human PRs**, because the job pushes commits. It must stay in step with `auto-merge-sweep.yml`; the two are now coupled by convention rather than by a shared constant.
- **Serialized by concurrency group.** Two heals would compute the same floor from the same tip and renumber two PRs onto one version — the collision this exists to clear.

## Separately: the bot has been telling maintainers not to look

The drift bot's PR body claimed *"`master` requires no review, so no human approval gates it."* That stopped being true when `master-protection` began requiring an approving review, and it is printed on every drift PR. Of everything in this PR, that sentence is the most likely single contributor to PRs sitting unreviewed, and it is a one-line fix.

## Verdict

Approving. The mechanism is understood rather than guessed at, the fix mostly connects machinery that already existed, and it has been run against four real conflicts including the one it was written for.

The `renumber-release-version.mjs` path resolution deserves a note for whoever touches it next: it resolves `package.json` / `CHANGELOG.md` from the script's own repo root, not from `cwd`. Correct inside the workflow, where they are the same checkout, and it silently read the wrong file when run across worktrees during testing. `--pkg` / `--changelog` exist for that case.

sprayberry-redline (approved):

This PR was authored by the same party approving it, and so was the issue behind it. That is not independent review, and the approval should not be read as implying it. Recording it rather than leaving it implicit.

Unusually for this batch, the change was exercised against live traffic before review: the same version-collision ladder it exists to eliminate produced three real conflicts today, and the tooling here resolved all three. Two defects surfaced that way. Both are fixed; both are described below rather than quietly folded in.

## Scope

`scripts/renumber-release-version.mjs` (new), `.github/workflows/drift-pr-heal.yml` (new), a one-line correction to the drift bot's PR body in `scripts/auto-draft-drift-fix.mjs`, 30 new assertions, plus the bump and CHANGELOG.

## The problem is real and already documented in-tree

`version-advances.sh` opens by describing it: *"Two bot PRs can independently bump package.json to the SAME version, because each is drafted off the current release without seeing the other (#954 and #955 both took 5.5.10 off 5.5.9, with different dates)."* That script refuses to arm the loser — correctly, since merging it either fast-exits the duplicate-tag guard green having shipped nothing, or moves master's version below a published tag.

Nothing then renumbers the loser. It sits open, drifts BEHIND, and waits. With `master-protection` requiring an approving review and dismissing it on push, the hand-fix costs a full round trip. That is the backlog mechanism, and it is the thing this PR removes.

## Verified

- **`resolve-release-conflicts.mjs` was wired into no workflow at all.** Confirmed by enumerating every `scripts/*.mjs` referenced from `.github/workflows/`. The file-level rule was written, tested, and never ran — which is most of why the manual resolutions kept recurring.
- **Verified against production data, not fixtures.** Against the real stuck #1027 (`5.5.25`, published; master at `5.5.27`) the renumberer produced `5.5.28`, moved the CHANGELOG heading with it, was idempotent on re-run, and flipped `version-advances.sh` from refusing to `"version advances 5.5.27 -> 5.5.28"`.
- **Exercised on three further live conflicts** during today's ladder — #1037 twice and #1038 — resolving `package.json` and `CHANGELOG.md` each time and correctly reporting the renumber as a no-op where the version already cleared the tip.
- **Refusal paths are asserted, not assumed**: zero or multiple CHANGELOG headings, a `package.json` with no version, a non-`X.Y.Z` floor, and a prerelease version all refuse. CRLF and LF both round-trip, which matters because a silent regex miss on CRLF would present as "nothing to rename" — safe, but permanently useless on a Windows checkout with no log line explaining it.
- Both workflow `run:` blocks extracted and checked with `bash -n`; YAML parses; guard and permissions are as intended. Full suite 138/138.

## Defects found by dogfooding, and fixed

1. **A source conflict would have killed the whole tick.** `resolve-release-conflicts.mjs` only touches the two version files, but a drift PR can also conflict in source — observed on #1038, where #1037 had added `utilFreshness` to the same pool import line #1038 was adding `accountIneligibility` to. The workflow would have resolved the version files, run `git add`, then failed on `git commit` with `src/proxy.ts` still unmerged. Under `set -e` that ends the job with every later PR in the batch unprocessed — one awkward PR silently starving the rest. Now the unmerged set is inspected before anything is resolved, and such a PR is skipped with the offending paths named.

2. **The first guard refused everything.** It joined the paths with `tr` and matched a `case` pattern; `tr` left a trailing space, so no valid input ever matched. It failed closed, which is the right direction, but it would have made the workflow a permanent no-op that looked like it was working. Replaced with `grep -vxE` over the path list — order-independent, separator-independent — and exercised against all six real shapes.

Worth stating plainly: both defects were in code already described as ready, and both were found by running it rather than reading it.

## Points a reviewer should weigh

- **It refuses to run without `DARIO_DRIFT_BOT_PAT`.** A push made with `GITHUB_TOKEN` fires no `pull_request` event, so healing with it would leave the PR with no runs for the required contexts — stuck in a newer and less visible way. Skipping keeps it visibly stuck. The consequence is that this ships **inert** if the PAT is ever rotated away, and the failure is a red job rather than a silent skip, which is the correct trade but should be a conscious one.
- **The branch-prefix filter is the only thing keeping this off human PRs**, because the job pushes commits. It must stay in step with `auto-merge-sweep.yml`; the two are now coupled by convention rather than by a shared constant.
- **Serialized by concurrency group.** Two heals would compute the same floor from the same tip and renumber two PRs onto one version — the collision this exists to clear.

## Separately: the bot has been telling maintainers not to look

The drift bot's PR body claimed *"`master` requires no review, so no human approval gates it."* That stopped being true when `master-protection` began requiring an approving review, and it is printed on every drift PR. Of everything in this PR, that sentence is the most likely single contributor to PRs sitting unreviewed, and it is a one-line fix.

## Verdict

Approving. The mechanism is understood rather than guessed at, the fix mostly connects machinery that already existed, and it has been run against four real conflicts including the one it was written for.

The `renumber-release-version.mjs` path resolution deserves a note for whoever touches it next: it resolves `package.json` / `CHANGELOG.md` from the script's own repo root, not from `cwd`. Correct inside the workflow, where they are the same checkout, and it silently read the wrong file when run across worktrees during testing. `--pkg` / `--changelog` exist for that case.

PR #1042, opened 2026-08-20
ci: fix two defects drift-pr-heal's first real run exposed

askalf:
#1040 shipped this workflow. Its first production run — dispatched manually against #1027 and #1028, because the `:47` cron missed its first tick — reported `healed=2 refused=0`, and **both results were wrong.**

## 1. It renumbered *after* merging, and renamed a shipped release's notes

`resolve-release-conflicts.mjs` keeps the **higher** version. So once a stale bot PR has merged master, its `package.json` holds *master's* version rather than its own. `renumber-release-version.mjs` replaces whichever CHANGELOG heading matches `package.json` — so it renamed master's `## [5.5.31]`, the notes for a release **already tagged and published to npm**, while the bot's own `## [5.5.26]` entry sat orphaned further down the file.

Merging that would have shipped 5.5.32 carrying 5.5.31's notes, and erased 5.5.31's.

**Nothing shipped was damaged.** Master's `[5.5.31]` is intact and `v5.5.31` is tagged; the corruption was confined to two unmerged branches, both since repaired by hand in the corrected order.

The fix is a step reorder — renumber before the merge, while `package.json` still holds the PR's own version and master's sections are not yet in the tree. Proven on #1028 from its pre-heal state:

```
pre-heal:        5.5.26, own section ## [5.5.26]
renumber first:  5.5.26 -> 5.5.32,  renames its OWN section
then merge:      ## [5.5.32]  <- the bot's template-label entry
                 ## [5.5.31]  <- master's shipped release, INTACT
```

## 2. Every PR in a batch took the same version

`tip` is read from `origin/master` per PR, and **master does not move during a run** — so #1027 and #1028 both computed floor `5.5.31` and both landed on `5.5.32`.

That is the collision this workflow exists to clear, reproduced by the workflow.

Two guards that look like they should have caught it, and structurally cannot:

- the **concurrency group** prevents two *runs* colliding — it says nothing about two PRs inside one run;
- **`version-advances.sh`** only ever compares against master, so a sibling in the same batch is invisible to it.

A running `floor` high-water mark now carries across the loop, claimed on every PR the run touches — including one it decided not to push, whose version must still not be handed out again. Comparison uses `sort -V`, so `5.5.9 < 5.5.10` rather than the lexical opposite.

Exercised:

```
tip=5.5.31 floor=(none)  -> 5.5.31   first PR in batch
tip=5.5.31 floor=5.5.32  -> 5.5.32   second PR: must clear its sibling
tip=5.5.40 floor=5.5.32  -> 5.5.40   master moved ahead of floor
tip=5.5.32 floor=5.5.32  -> 5.5.32   equal: no double-bump
tip=5.5.9  floor=5.5.10  -> 5.5.10   numeric, not lexical
```

## Tests

6 new assertions pinning the ordering hazard as a **characterization test**: the wrong caller renames the released section, the right one renames the PR's. `renumber()` is correct in both — the bug was always in what the caller handed it — so the test locks the *calling order* rather than the function, and a future reordering of the workflow has to break it to reintroduce this.

Both workflow `run:` blocks re-extracted and `bash -n`'d. Full suite **139/139**.

## Why the original dogfooding missed both

I exercised #1040 against three real conflicts before shipping it, and every one was a hand-authored PR whose version was **above** the tip — which is precisely the case that already worked. Neither defect is reachable from that shape: #1 needs a PR whose version is *below* the tip, and #2 needs *two* PRs in one batch. The first run with real bot PRs had both.

## Release note

Bumps to `5.5.34`, above #1028 (5.5.32) and #1027 (5.5.33).


sprayberry-redline (approved):

Authored by the same party approving it, and it fixes a workflow the same party shipped earlier today. Not independent review; recording it rather than leaving it implicit.

## What this fixes

#1040 shipped `drift-pr-heal.yml`. Its first production run — dispatched by hand, because the `:47` cron missed its first tick — reported `healed=2 refused=0` against #1027 and #1028, and both results were wrong.

**Defect 1 — it renamed a shipped release's notes.** `resolve-release-conflicts.mjs` keeps the *higher* version, so after merging master a stale bot PR holds master's version, not its own. `renumber-release-version.mjs` then renames whichever CHANGELOG heading matches `package.json` — which was master's `## [5.5.31]`, the notes for a release already tagged and on npm, while the bot's own entry sat orphaned. Merging would have shipped 5.5.32 carrying 5.5.31's notes and erased 5.5.31's.

**Defect 2 — both PRs took 5.5.32.** `tip` is read from `origin/master` per PR, and master does not move *during* a run.

## Verified

- **Defect 1's fix is a step reorder, and it was proven on real data before this PR existed.** #1028 was reset to its pre-heal commit and redone renumber-first: `5.5.26 -> 5.5.32` renaming its *own* section, then merge, leaving `## [5.5.31]` intact. Same for #1027 at 5.5.33. Both branches then merged cleanly as 5.5.32 and 5.5.33.
- **Defect 2's fix is exercised across the cases that matter**, including the one a lexical comparison gets backwards:

```
tip=5.5.31 floor=(none)  -> 5.5.31   first PR in batch
tip=5.5.31 floor=5.5.32  -> 5.5.32   second PR: must clear its sibling
tip=5.5.40 floor=5.5.32  -> 5.5.40   master moved ahead of floor
tip=5.5.32 floor=5.5.32  -> 5.5.32   equal: no double-bump
tip=5.5.9  floor=5.5.10  -> 5.5.10   numeric via sort -V, not lexical
```

- **The floor is claimed even when the run does not push.** A PR already holding a version must not have it handed to the next PR in the batch — easy to miss, since the obvious place to set it is beside the push.
- **The two guards that look sufficient and are not** are now documented in-place: the concurrency group prevents two *runs* colliding, and `version-advances.sh` only ever compares against master, so a sibling in the same batch is structurally invisible to it. Without that note, the next reader has two plausible reasons to delete the `floor` logic as redundant.
- **The new test is a characterization test, and that is the right shape.** `renumber()` is correct in both branches of it — the bug was always in what the caller handed it — so the test pins the *calling order*, and reordering the workflow has to break it to reintroduce this. 6 assertions; full suite 139/139.
- Both workflow `run:` blocks re-extracted and `bash -n`'d. Resolved against master at 5.5.33 with `## [5.5.31]`, `## [5.5.32]` and `## [5.5.33]` all intact.

## Independent confirmation of the diagnosis

After #1028 merged, the tick ran again against #1027 **on the unfixed workflow** and logged:

```
[renumber] unchanged 5.5.33 — 5.5.33 already advances past 5.5.32
HEALED — version advances 5.5.32 -> 5.5.33
```

Correct: no renumber commit, both master sections intact. That is the diagnosis holding up under test rather than being asserted — the defect is reachable *only* when the PR's version sits below the base tip. It also explains why #1040's dogfooding missed it: all three conflicts it was tried against were hand-authored PRs whose version was above the tip by construction, which is precisely the path that already worked.

## Findings against this PR

1. **It does not add a test for defect 2.** The floor logic lives in the workflow's shell, not in a script, so there is nothing importable to assert against — it was verified by extracting the selector and running it over five cases by hand, which is weaker than the characterization test covering defect 1. Lifting the selector into `scripts/` alongside `version-advances.sh` would make it testable and would put it next to the other rule the arming path already shares. Worth doing; not worth blocking a fix for two live defects.

2. **The workflow still only covers `bot/*` branches.** Correct — it pushes commits, and that filter is the only thing keeping it off human PRs — but it means the five hand-authored PRs merged today would not have been healed, and the manual rounds they cost are unchanged by this.

## Verdict

Approving. It fixes two real defects, one of which could have rewritten a published release's notes; the reorder was proven on the exact branches that were corrupted before the PR was written; and the diagnosis was independently confirmed by a later run of the unfixed code behaving correctly in the case the theory says it should.

PR #1043, opened 2026-08-20
ci: stop scheduled workflows firing on all 54 forks (#1039)

### LINKED ISSUE #1039: Scheduled workflows fire on all 54 forks — guard the cron, not the workflow

askalf:
Closes #1039. Follow-on to @chaogebaba's #1029, which fixed the one workflow that reached outward.

## The split is narrower than #1039 proposed

I argued in the issue for "guard the trigger, not the workflow", on the grounds that a blanket `github.repository ==` would break a forker's own push/PR security scanning. That worry was right in principle and **mostly wrong in practice** — I hadn't checked the trigger data. Having checked it:

| | count | other triggers besides `cron` | guard applied |
|---|---|---|---|
| `workflow_dispatch`-only | **13** | none | blanket `github.repository ==` |
| also `push` / `pull_request` | **2** | `codeql`, `scorecard` | schedule-scoped |

Thirteen of fifteen have no push/PR CI to break, so the blanket guard costs a fork nothing. Only `codeql` and `scorecard` needed the narrower form:

```yaml
if: github.event_name != 'schedule' || github.repository == 'askalf/dario'
```

## Severity, stated plainly

The 15 here are **waste, not exposure**. A fork's `GITHUB_TOKEN` is scoped to the fork, so the issue-filing watchers file into the fork rather than here. But 54 forks × 15 jobs on hourly-to-daily crons is a lot of other people's CI minutes spent on our behalf, and several decide whether to act by reading *this* package's registry entry or this repo's state — wrong on a fork by construction.

`cc-drift-auto-release`, the one that actually tagged, released and attempted an `npm publish`, was #1029.

## One that looked already-guarded and wasn't

`wire-drift-self-hosted` carried:

```yaml
if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```

That guards **PRs from forks**. On a fork, `github.repository` *is* the fork, so it's true for the fork's own PRs and does nothing to stop the workflow running there. It needed the repository guard too.

## Composition, not replacement

Two jobs had an existing job-level `if:` (`cc-drift-watch`'s `check`, and the above). Both are composed with the original wrapped in parentheses:

```yaml
if: >-
  github.repository == 'askalf/dario'
  && (needs.version_check.outputs.should_run == 'true')
```

Dropping those parens would let `&&` bind tighter than `||` and silently make a trigger path unreachable — the exact failure #1029 was careful to avoid, and the one thing here that could break this repo rather than a fork.

## Verification

**No behaviour change on this repository.** `github.repository` is `askalf/dario`, so every added conjunct is constant-true and all jobs evaluate exactly as before.

Verified by parsing every workflow file and asserting that each job of each cron-carrying workflow now has the guard: **18/18 jobs across 17 cron workflows, none missed** — including the two already covered (`cc-drift-auto-release` via #1029, `drift-pr-heal` by construction).

The risk that matters here is a wrong `if:` silently disabling something on *our* side, which fails quietly rather than loudly. `actionlint` is a required check and catches syntax; it does not catch semantics, so the parenthesisation above is the part worth a human eye.

### DIFF
--- a/.github/workflows/auto-merge-sweep.yml
+++ b/.github/workflows/auto-merge-sweep.yml
@@ -47,6 +47,9 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: github.repository == 'askalf/dario'
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
--- a/.github/workflows/cc-billing-classifier-canary.yml
+++ b/.github/workflows/cc-billing-classifier-canary.yml
@@ -68,6 +68,9 @@ jobs:
     permissions:
       contents: read
       issues: write
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: github.repository == 'askalf/dario'
     runs-on: [self-hosted, dario-drift]
     timeout-minutes: 5
     steps:
--- a/.github/workflows/cc-drift-template-watch.yml
+++ b/.github/workflows/cc-drift-template-watch.yml
@@ -62,6 +62,9 @@ jobs:
     # Runs only on the self-hosted runner labeled `dario-drift`. Github-hosted
     # runners don't have CC + OAuth, so this job effectively gates on whether
     # a maintainer has registered a runner with that label.
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: github.repository == 'askalf/dario'
     runs-on: [self-hosted, dario-drift]
     timeout-minutes: 8
     steps:
--- a/.github/workflows/cc-drift-watch.yml
+++ b/.github/workflows/cc-drift-watch.yml
@@ -39,6 +39,9 @@ jobs:
     # Gate job only probes npm + a cache sentinel — no GitHub API writes.
     permissions:
       contents: read
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: github.repository == 'askalf/dario'
     runs-on: ubuntu-latest
     outputs:
       version: ${{ steps.probe.outputs.version }}
@@ -107,7 +110,11 @@ jobs:
       issues: write
       pull-requests: write
     needs: version_check
-    if: needs.version_check.outputs.should_run == 'true'
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: >-
+      github.repository == 'askalf/dario'
+      && (needs.version_check.outputs.should_run == 'true')
     runs-on: ubuntu-latest
     steps:
       - uses: askalf/checkout-with-retry@115a6407547e9711edbc2e915838d495cad9583f  # v1.1.0
--- a/.github/workflows/cc-drift-watcher-liveness.yml
+++ b/.github/workflows/cc-drift-watcher-liveness.yml
@@ -43,6 +43,9 @@ jobs:
       # to an operator alert — the recovery path had never once fired
       # before #950 surfaced it.
       actions: write
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: github.repository == 'askalf/dario'
     runs-on: ubuntu-latest
     timeout-minutes: 3
     steps:
--- a/.github/workflows/cc-oauth-health.yml
+++ b/.github/workflows/cc-oauth-health.yml
@@ -71,6 +71,9 @@ jobs:
       # scripts/health-verdict.sh through the contents API with this token.
       contents: read
       issues: write
+      # Never from a fork. This job is meaningless anywhere but this
+      # repository, and 54 forks inherit the schedule. See #1039.
+    if: github.repository == 'askalf/dario'
     runs-on: [self-hosted, dario-drift]
     timeout-minutes: 4
     steps:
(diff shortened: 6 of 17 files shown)


sprayberry-redline (approved):

Authored by the same party approving it, and it closes an issue the same party filed. Not independent review; recording it rather than leaving it implicit.

## Scope

15 workflow files, plus the bump and CHANGELOG. No source changes. Follow-on to @chaogebaba's #1029, which covered the only one of the set that reached outward.

## The thing to actually check

This PR's risk is not that it fails to guard a fork. It is that a wrong `if:` **silently disables something on this repository** — a failure that produces no error, just a workflow that quietly stops running. `actionlint` is a required check and catches syntax; it does not catch semantics.

Two jobs had an existing job-level `if:` and are composed rather than replaced:

```yaml
if: >-
  github.repository == 'askalf/dario'
  && (needs.version_check.outputs.should_run == 'true')
```

```yaml
if: >-
  github.repository == 'askalf/dario'
  && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
```

Both wrap the original in parentheses. The second is the one that matters: without them, `&&` binds tighter than `||` and the `pull_request` path becomes unreachable — silently, on a self-hosted drift watcher nobody watches closely. That is the same trap #1029 was careful to avoid, and it is correct here.

## Verified

- **No behaviour change on this repository.** `github.repository` is `askalf/dario`, so every added conjunct is constant-true and every job evaluates exactly as before.
- **Coverage is complete and checked by parsing, not grepping.** Every workflow file was loaded as YAML and each job of each cron-carrying workflow asserted to carry the guard: 18/18 across 17 workflows, including the two already covered (`cc-drift-auto-release` via #1029, `drift-pr-heal` by construction).
- **The two forms are applied to the right files.** `codeql` and `scorecard` get the schedule-scoped form; the other thirteen get the blanket one. Worth checking explicitly, because the schedule-scoped string *contains* the blanket string as a substring — a naive "does it contain the guard" check passes for both and would not have caught a mix-up. Confirmed per-file.
- `actionlint` passes across all 15 changed files; all six required checks green.

## The correction this PR carries, and why it matters

#1039 argued for "guard the trigger, not the workflow" everywhere, on the grounds that a blanket guard would break a forker's own push/PR security scanning. That reasoning was sound and the conclusion was mostly wrong, because the trigger data was never checked before the issue was written. Thirteen of fifteen are `workflow_dispatch`-only besides cron and have no fork CI to break. The narrower form was needed for two, not for the six the issue named — `cflite` and `stale` were both listed there and are both dispatch-only.

The PR body and the issue both record that rather than quietly shipping the better split. Worth keeping visible: the original recommendation would have added conditional complexity to eleven files that did not need it.

## Finding

**`wire-drift-self-hosted` looked already-guarded and was not.** Its existing condition gates *PRs from forks* (`head.repo.full_name == github.repository`); on a fork, `github.repository` is the fork, so that is true for the fork's own PRs and never stopped the workflow running there. Anyone auditing this set by grepping for "fork" in the `if:` would have marked it done. Now genuinely guarded.

## Verdict

Approving. Mechanical change, complete coverage verified by parsing, no behaviour change here, and the one genuinely dangerous part — the two composed conditions — is parenthesised correctly.

Merging this closes #1039 and leaves `askalf/dario` with no unguarded scheduled workflow.

PR #1044, opened 2026-08-20
fix: OAuth watcher pages for a spent window, and calls a live container dead (#1041)

### LINKED ISSUE #1041: 🔴 dario OAuth unreachable — fleet LLM down
**dario OAuth is `unreachable`** — every fleet LLM call will 401 until this recovers.

| axis | reading |
|---|---|
| oauth | `unreachable` |
| serving probe | `not measured` |
| queue stalled | `0` ms |

- Reachable: **no** — container down or not answering
- Container: Up About a minute (healthy)
- Credential split check: NO — paths unified at inode 256799

> Full `/health?probe=1` body intentionally omitted (public repo). Read it on the box: `docker exec askalf-dario wget -qO- 'http://127.0.0.1:3456/health?probe=1'`

**Recover:** docs/recovery.md → *OAuth credential dead*. If the split check is YES, re-link both paths to the canonical and `docker restart askalf-dario`. If the refresh token itself is dead, transplant fresh creds (platform runbook).

_Auto-filed by `cc-oauth-health.yml` on the self-hosted runner._

askalf:
Fixes #1041 — which is a **true positive that should not have paged**, plus a second fault that made it point at the wrong runbook.

## 1. It paged for a throttle

5.5.30 (#1038) made `derivePoolStatus` answer the router's question rather than a narrower one. A pool whose accounts are all rate-limited therefore now reports `oauth: broken`, and `health-verdict.sh` raises an axis whenever `oauth != healthy` — so a spent usage window became a `[CRITICAL]` alert.

That contradicts a rule this repo had already settled, written **one branch below in the same file**:

```sh
elif [ "$probe_ok" = "false" ]; then
  # rate-limited / upstream-overloaded never reach here — dario reports those
  # as ok:true precisely so a watchdog does not restart on a throttle.
```

The oauth axis became a way around it. The credential is fine in that state and the window refills unattended.

`health-verdict.sh` now reads dario's `expiresIn` reason and suppresses the axis for a rate-limited pool **only**. The three reasons `accountIneligibility()` produces are treated distinctly:

| reason | pages? |
|---|---|
| `all accounts rate-limited` | **no** — refills on its own |
| `all tokens expired — run \`dario login\`` | yes |
| `all accounts in auth-cooldown` | yes |
| unreachable / empty body | yes |

## 2. It called a live container dead

The body was fetched with `wget -qO- || curl -s`. The container has **BusyBox** wget — which discards the body on a non-2xx and has no `--content-on-error` — and has **no curl at all**, so the fallback could never run:

```
$ docker exec askalf-dario wget --content-on-error -qO- .../health
wget: unrecognized option: content-on-error        # BusyBox v1.37.0
$ docker exec askalf-dario sh -c 'curl ...'
sh: curl: not found
```

So every legitimate 503 produced an empty body, `oauth` defaulted to `unreachable`, and the alert said *"Reachable: **no** — container down or not answering"* about a proxy that was up, healthy and answering accurately. That sends a reader straight to `docs/recovery.md → OAuth credential dead` for something that is not a credential problem.

Now fetched with **node** — the app's own runtime, guaranteed present — which keeps the body on an error status. "Reachable" now means *dario answered*, and the alert table carries the reason.

## Verified end-to-end against production

The box's pool is spent right now, so this was testable against the real thing rather than a fixture:

```
live body -> patched verdict:  oauth=broken  axis=[]                reason="all accounts rate-limited"
live body -> master  verdict:  oauth=broken  axis=[OAuth broken]    <- pages
```

The node fetch was also run against the live container and returned the full 503 body, which the old fetch could not.

## Tests

11 new assertions in `test/health-verdict.mjs`, using the exact body production returned tonight. They cover the suppression, both reasons that must still page, unreachable, a pre-5.5.30 body with no `expiresIn` (must not be mistaken for a throttle), and — the one most likely to bite later — a **healthy** pool, whose `expiresIn` is a *duration* (`"2h 0m"`) rather than a reason, so the match must not be loose. 43 passing.

## Note on ownership

The first fault is mine: #1038 shipped this morning and changed what `/health` reports on a monitored surface. I flagged in that review that *"uptime monitors will newly alarm; those are true positives that were previously suppressed."* That was right about credentials and wrong about throttles — a spent window is not an outage, and this repo had already decided that.

### DIFF
--- a/.github/workflows/cc-oauth-health.yml
+++ b/.github/workflows/cc-oauth-health.yml
@@ -152,7 +152,6 @@ jobs:
           # genuine loopback caller with no cf-ray is exactly what dario requires
           # before it will spend a token on this, so this path — and only this
           # path — gets the serving verdict.
-          URL='http://127.0.0.1:3456/health?probe=1'
           # Fetched with node, NOT wget/curl, because the body on a NON-2xx is
           # the whole point: /health answers 503 with a populated body whenever
           # the pool cannot serve, and that body says why.
--- a/.github/workflows/cc-oauth-health.yml
+++ b/.github/workflows/cc-oauth-health.yml
@@ -153,7 +153,19 @@ jobs:
           # before it will spend a token on this, so this path — and only this
           # path — gets the serving verdict.
           URL='http://127.0.0.1:3456/health?probe=1'
-          body="$(docker exec askalf-dario sh -lc "wget -qO- '$URL' 2>/dev/null || curl -s --max-time 20 '$URL'" 2>/dev/null || true)"
+          # Fetched with node, NOT wget/curl, because the body on a NON-2xx is
+          # the whole point: /health answers 503 with a populated body whenever
+          # the pool cannot serve, and that body says why.
+          #
+          # The container has BusyBox wget (no --content-on-error, it discards
+          # the body on an HTTP error) and NO curl at all — so the previous
+          # `wget -qO- || curl -s` produced an EMPTY body on every 503, the
+          # verdict script defaulted oauth to "unreachable", and the alert
+          # reported "container down or not answering" about a container that
+          # was up, healthy and answering truthfully (dario#1041). node is the
+          # app's own runtime, so it is always present.
+          FETCH='const http=require("http");const r=http.get({host:"127.0.0.1",port:3456,path:"/health?probe=1",timeout:20000},s=>{let b="";s.on("data",d=>b+=d);s.on("end",()=>process.stdout.write(b))});r.on("timeout",()=>{r.destroy();process.exit(0)});r.on("error",()=>process.exit(0));'
+          body="$(printf '%s' "$FETCH" | docker exec -i askalf-dario node - 2>/dev/null || true)"
           # `|| true`: grep exits 1 on no match, and this step runs under
           # the runner's default `bash -e {0}` plus pipefail — without it,
           # the empty-body case kills the job RIGHT HERE and the
@@ -170,6 +182,7 @@ jobs:
           probe_reason="$(printf '%s\n' "$verdict" | sed -n '3p')"
           stalled="$(printf '%s\n' "$verdict" | sed -n '4p')"
           axis="$(printf '%s\n' "$verdict" | sed -n '5p')"
+          reason="$(printf '%s\n' "$verdict" | sed -n '6p')"
 
           echo "dario /health -> oauth=$oauth probe=${probe_ok:-<not measured>}${probe_reason:+ ($probe_reason)} stalledForMs=$stalled"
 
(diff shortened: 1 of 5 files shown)


sprayberry-redline (approved):

Authored by the same party approving it, and it fixes a regression the same party shipped this morning. Not independent review; recording it rather than leaving it implicit.

## What this fixes

#1041 was a **true positive that should not have paged**, plus a second fault that pointed the reader at the wrong runbook.

**Fault 1 — paged for a throttle.** 5.5.30 (#1038) made `derivePoolStatus` answer the router's question, so an all-rate-limited pool now reports `oauth: broken`. `health-verdict.sh` raises an axis whenever `oauth != healthy`, so a spent usage window became `[CRITICAL]`.

The decisive detail is that the repo had **already settled this**, one branch below in the same file: dario reports a throttled probe as `ok:true` *"precisely so a watchdog does not restart on a throttle."* The oauth axis became a route around a rule that was already written down. That is what makes this a regression rather than a judgement call.

**Fault 2 — called a live container dead.** The fetch was `wget -qO- || curl -s`, and neither half can work here: BusyBox wget discards the body on a non-2xx and has no `--content-on-error`, and curl is not installed at all. Verified directly against the container rather than inferred:

```
wget: unrecognized option: content-on-error        # BusyBox v1.37.0
sh: curl: not found
```

So every legitimate 503 yielded an empty body, `oauth` defaulted to `unreachable`, and the alert reported *"container down or not answering"* about a proxy that was up and answering accurately.

## Verified

- **End-to-end against production, not a fixture.** The box's pool is spent right now, so the real body was run through both versions:

```
live body -> patched verdict:  oauth=broken  axis=[]              reason="all accounts rate-limited"
live body -> master  verdict:  oauth=broken  axis=[OAuth broken]  <- pages
```

- **The node fetch was run against the live container** and returned the full 503 body the old fetch could not. node is the app's own runtime, so availability is not an assumption.
- **Suppression is narrow and keyed on the reason, never on `oauth` alone.** Expired tokens page, auth-cooldown pages, unreachable pages, a pre-5.5.30 body with no `expiresIn` pages. Only `rate-limited` is suppressed.
- **The loose-match trap is covered.** A *healthy* pool sends `expiresIn` as a duration (`"2h 0m"`), not a reason — so a sloppier match could have suppressed real alerts on healthy-looking bodies or misread durations. Asserted explicitly.
- **The public-repo disclosure rule is untouched.** The raw body is still not pasted into the alert; only the reason string — which is one of three fixed operational phrases and carries no token countdown, request volume or session counts.
- `health-verdict.mjs` 43 passing; all six required checks green.

## Findings against this PR

1. **`actionlint` caught a dead variable I left behind** (`URL`, SC2034) — the leftover of the wget form. Fixed in `7e4a26f`. That is the **second** time today local `bash -n` passed something shellcheck rejected (the first was SC2001 on `drift-pr-heal.yml`). `bash -n` checks syntax, not style, and shellcheck is not installed on the dev box. Two occurrences in one day is a tooling gap, not bad luck — worth installing shellcheck locally or wiring a pre-push hook.

2. **Suppression is silent in the issue tracker.** When the pool is merely throttled the watcher now files nothing, which is the intent — but it also means a *prolonged* exhaustion looks identical to healthy from the tracker's point of view. The reason is still reported in the step log and on line 1/6 of the verdict, so it is observable; it is just no longer alarming. If sustained exhaustion is something worth knowing about, that wants a separate, lower-severity signal rather than re-arming this one.

## Verdict

Approving. It restores a rule the repo had already made, the diagnosis was verified against the live container rather than reasoned about, and the suppression is scoped tightly enough that every genuine failure mode still pages.

The ownership is worth stating plainly: fault 1 is a direct consequence of #1038, which I shipped this morning after writing in its own review that *"uptime monitors will newly alarm; those are true positives that were previously suppressed."* That was correct about credentials and wrong about throttles.

PR #1046, opened 2026-08-20
chore(cc-drift): v5.5.37 — maxTested → v2.1.238

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.238 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.237` → `2.1.238` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.37`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.37] - 2026-08-20` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.238 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.237). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.237; npm latest is v2.1.238. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.238 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.238 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.238. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.238" step — no local run needed.)
- ✅ **Auto-merges** once the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`) **and the PR has an approving review** — the ruleset on `master` requires one, and dismisses it on any later push. Until then this PR sits open regardless of how green it is.
- ✅ **Auto-releases** on merge: `cc-drift-auto-release.yml` tags `v5.5.37`, publishes `@askalf/dario@5.5.37` to npm (`--provenance`) + GHCR inline, and the box autodeploy timer picks it up within ~15 min.

**You only need to look if CI fails** — then auto-merge holds, this PR stays open with the failure visible, and the bot branch is preserved. Otherwise it is already on its way to npm.

### About this auto-draft

Only `compat.range` items are auto-patched by this script. Template re-capture is auto-handled separately by `cc-drift-template-watch.yml`. The remaining categories (scope rotations, URL / clientId / tokenUrl changes) require judgment and stay manual — the bot opens the plain drift-issue for those as before.

---

_Generated by `scripts/auto-draft-drift-fix.mjs`. Closes the detection-latency arc: [#112](https://github.com/askalf/dario/pull/112) (CF bypass), [#113](https://github.com/askalf/dario/pull/113) (hourly cadence), [#114](https://github.com/askalf/dario/pull/114) (auto-draft PR)._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.37] - 2026-08-20
+
+- **CC drift patch** — `SUPPORTED_CC_RANGE.maxTested` bumped `2.1.237` → `2.1.238` for CC v2.1.238. Auto-drafted by `cc-drift-watch.yml`. Template re-capture, if needed, is auto-handled by `cc-drift-template-watch.yml`.
 ## [5.5.36] - 2026-08-20
 
 - **The OAuth watcher no longer pages for a spent usage window, and no longer calls a live container dead** (#1041) — two faults, both surfaced by 5.5.30. (1) `derivePoolStatus` began answering the router's question, so a pool whose accounts are all rate-limited now reports `oauth: broken`; the watcher paged on it, contradicting the rule it already follows for a throttled probe (dario reports those as `ok:true` "so a watchdog does not restart on a throttle"). `health-verdict.sh` now reads dario's `expiresIn` reason and suppresses the alert for a rate-limited pool only — expired tokens, auth-cooldown, and an unreachable container all still page. (2) The body was fetched with BusyBox `wget -qO-`, which discards the body on a non-2xx, with a `curl` fallback that does not exist in the container — so every legitimate 503 produced an empty body, `oauth` defaulted to `unreachable`, and the alert reported "container down or not answering" about a container that was up and answering truthfully. Fetched via node now, and the alert carries the reason so a reader can tell "run `dario login`" from "wait for the window".
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ checklist.
 
 ## [Unreleased]
 
-## [5.5.37] - 2026-08-20
+## [5.5.38] - 2026-08-20
 
 - **CC drift patch** — `SUPPORTED_CC_RANGE.maxTested` bumped `2.1.237` → `2.1.238` for CC v2.1.238. Auto-drafted by `cc-drift-watch.yml`. Template re-capture, if needed, is auto-handled by `cc-drift-template-watch.yml`.
 ## [5.5.36] - 2026-08-20
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ checklist.
 
 ## [Unreleased]
 
-## [5.5.38] - 2026-08-20
+## [5.5.39] - 2026-08-20
 
 - **CC drift patch** — `SUPPORTED_CC_RANGE.maxTested` bumped `2.1.237` → `2.1.238` for CC v2.1.238. Auto-drafted by `cc-drift-watch.yml`. Template re-capture, if needed, is auto-handled by `cc-drift-template-watch.yml`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ checklist.
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


sprayberry-redline (approved):

Bot-drafted by `cc-drift-watch.yml`. The unblocking work on top — the heal that renumbered it and the CC upgrade that let compat validate it — was done by the approving party, so this is not independent review. Recording it rather than leaving it implicit.

## Scope

One field: `SUPPORTED_CC_RANGE.maxTested` moves `2.1.237` → `2.1.238`, declaring this dario release validated against CC v2.1.238. Plus the version bump and CHANGELOG.

## Verified — and the verification is the whole point of this PR

`compat` now passes on the self-hosted runner, and the gate line is the one that matters:

```
[compat] captured live template from CC 2.1.238
[compat] maxTested gate ok: claimed 2.1.238 <= installed 2.1.238
RESULTS: 10 passed, 0 failed, 0 warnings
```

That gate (#1024) is what makes this claim worth anything. It fails closed when `maxTested` names a CC the runner cannot actually exercise — which is exactly what it did here for several hours, correctly, because `cc-drift-watch` drafts from **npm** while the runner tracks the box's installed CC. It exists because #1012 once merged a `maxTested` bump to 2.1.235 whose only green compat run had exercised 2.1.234's template.

So a green compat here is affirmative evidence the runner really held 2.1.238, not merely that nothing objected. All seven required checks pass.

## What it took to get here, since none of it is visible in the diff

1. **The version collided twice.** Drafted as 5.5.37, then master moved to 5.5.39 while `drift-pr-heal` was mid-flight and renumbered it *to* 5.5.39 — a genuine race, not a logic error: the floor was correct when computed and master moved under it. The next heal tick resolved it cleanly:

   ```
   [renumber] renumbered 5.5.39 -> 5.5.40
   package.json: resolved   CHANGELOG.md: resolved
   HEALED — version advances 5.5.39 -> 5.5.40
   ```

   Worth noting the safety property held throughout: `version-advances.sh` would have refused a merge at 5.5.39 == 5.5.39, so the race could produce a stuck PR but never a bad release. And this is the first time the heal has run against a case it did not previously handle — renumber-before-merge (#1042) is what made it resolve rather than corrupt the CHANGELOG.

2. **compat needed CC 2.1.238 on the runner.** The box's own `cc-autoupdate.timer` was 5h25m away, so its unit was started early. Before doing that, the documented trap was re-checked: all of `claude-code-linux-x64`, `-musl` and `-arm64` are published at 2.1.238, so this was not the meta-published-native-missing case that turns an upgrade into a dead stub. Confirmed after: `2.1.238`, `Result=success`, and a real 338MB binary on disk rather than a stub.

## Findings

1. **This PR is only half of 2.1.238.** It moves `maxTested`; the bundled template still carries `_version: 2.1.237`. Now that the runner is on 2.1.238, `cc-drift-template-watch` should draft the label refresh on its next tick, and `sdk-drift-watch` closes #1031 only once *that* lands. Merging this alone does not clear #1045 or #1031.

2. **The two halves will collide on the version.** They always do — that is the loop #1040/#1042 exist for. It should now resolve unattended; this PR is the first evidence the heal does that correctly, so the next one is a reasonable test of whether it holds without a human.

## Verdict

Approving. One field, backed by a fail-closed gate that spent several hours correctly refusing it, on a version that now clears master.

The gate refusing this PR was never a fault in it — it was the system declining to certify a claim it could not check, which is precisely what it was built to do.

Issue #1018, opened 2026-08-19
strange cache behaviour


ramarro123:
i did a super dirty test

```
/home/bun/app # wget http://localhost:3456/v1/messages \
>   --header "x-api-key: write-key-here" \
>   --header "anthropic-version: 2023-06-01" \
>   --header "Content-Type: application/json" \
>   --post-data='{
>     "model": "claude-opus-5",
>     "max_tokens": 1024,
>     "messages": [
>       {
>         "role": "user",
>         "content": "ciao"
>       }
>     ]
>   }'
```
calling dario with an empty session (no litellm no other fancy stuff) and say "ciao"

Connecting to localhost:3456 (127.0.0.1:3456)
saving to 'messages'
messages             100% |************************************************************************************************************|  1513  0:00:00 ETA
'messages' saved
/home/bun/app # cat messages 
[...]

PR #1047, opened 2026-08-20
fix: Sonnet 5's $2/$10 is permanent — drop the cancelled cutover

askalf:
A dated bug with an eleven-day fuse.

## What breaks on 2026-09-01

Sonnet 5 launched at $2/$10, described as introductory pricing *"through 2026-08-31"*, after which it would rise to $3/$15. `analytics.ts` modeled that faithfully:

```js
'claude-sonnet-5': {
  input: 3, output: 15, cacheRead: 0.3, cacheCreate: 3.75,
  intro: { input: 2, output: 10, cacheRead: 0.2, cacheCreate: 2.5, until: '2026-08-31' },
},
```

**Anthropic has cancelled that increase and made $2/$10 the standard price.** The 2026-09-01 rise will not happen.

Left alone, `pricingRateFor` flips to $3/$15 on 2026-09-01 and every Sonnet 5 figure in `/analytics`, the TUI and doctor's cost lines reads 50% high — silently. No error, no warning, nothing in the output to suggest the number moved. Demonstrated against `master` vs this branch on identical records:

```
record dated        master     fixed
2026-08-20            $2         $2
2026-09-01            $3         $2     <- the bug
2027-01-01            $3         $2
```

A cost estimate that is quietly wrong is worse than one that is obviously missing, because it still gets quoted.

## The change

Sonnet 5 becomes a flat entry. The dated-`intro` mechanism **stays** in `pricingRateFor` — it is correct, and the next model to launch on a promotion will need it — but no entry carries one today, so that branch is now unexercised. The test says so explicitly rather than leaving a future reader to find out.

I deliberately did *not* keep an `intro` with a far-future `until`. A date nobody is watching is precisely what caused this.

## Tests

`test/pricing-intro-window.mjs` is rewritten around the new reality. The old cutover instant is now the **regression guard** rather than the assertion:

```js
check('first instant of 2026-09-01 -> STILL $2/$10', …)
check('far future (2028-01-01) -> STILL $2/$10', …)
```

18/18 there, **139/139** overall. Everything else in that file — Opus 5/4.8/4.7/4.6 rates, Fable 5's $10/$50, the `[1m]` tag stripping, the unknown-model fallback — is untouched and still passing.

## How this was found

I asserted the cutover as current fact in a public answer on #1018, having taken it from a *cached* price table instead of checking the live source. The maintainer caught it and asked me to re-verify every claim in that comment.

Six turned out to be wrong or overstated, and this is the one that would have actually mispriced something rather than merely misinformed a reader. The #1018 comment is being corrected separately.

Worth noting the shape of the failure for next time: the pricing entry was *correct when written* and had a test proving it. Both quietly became wrong when an external fact changed, and nothing in the repo could notice. Anything keyed to a future date deserves that same suspicion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.37] - 2026-08-20
+
+- **Fixed: `/analytics` would have over-stated every Sonnet 5 cost by 50% from 2026-09-01.** Sonnet 5 shipped at $2/$10 described as introductory "through 2026-08-31, then $3/$15", and `analytics.ts` modeled that as a dated cutover (`intro: { …, until: '2026-08-31' }`). Anthropic has since cancelled the increase and made $2/$10 the standard price. Left alone, `pricingRateFor` would have flipped to $3/$15 on 2026-09-01 — silently, with nothing in the output to indicate the number had moved. Sonnet 5 is now a flat rate. The dated-`intro` mechanism is retained for a model that genuinely has one; no entry uses it today.
+
 ## [5.5.36] - 2026-08-20
 
 - **The OAuth watcher no longer pages for a spent usage window, and no longer calls a live container dead** (#1041) — two faults, both surfaced by 5.5.30. (1) `derivePoolStatus` began answering the router's question, so a pool whose accounts are all rate-limited now reports `oauth: broken`; the watcher paged on it, contradicting the rule it already follows for a throttled probe (dario reports those as `ok:true` "so a watchdog does not restart on a throttle"). `health-verdict.sh` now reads dario's `expiresIn` reason and suppresses the alert for a rate-limited pool only — expired tokens, auth-cooldown, and an unreachable container all still page. (2) The body was fetched with BusyBox `wget -qO-`, which discards the body on a non-2xx, with a `curl` fallback that does not exist in the container — so every legitimate 503 produced an empty body, `oauth` defaulted to `unreachable`, and the alert reported "container down or not answering" about a container that was up and answering truthfully. Fetched via node now, and the alert carries the reason so a reader can tell "run `dario login`" from "wait for the window".
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.36",
+  "version": "5.5.37",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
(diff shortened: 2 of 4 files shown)


sprayberry-redline (approved):

Authored by the same party approving it. Not independent review; recording it rather than leaving it implicit.

## What this fixes

Sonnet 5 launched at $2/$10 described as introductory *"through 2026-08-31, then $3/$15"*, and `analytics.ts` modeled that as a dated cutover. Anthropic has since cancelled the increase and made $2/$10 permanent.

Left alone, `pricingRateFor` flips to $3/$15 on **2026-09-01 — eleven days out** — and every Sonnet 5 figure in `/analytics`, the TUI and doctor's cost lines reads 50% high, silently.

## Verified

- **The regression is demonstrated, not asserted.** Identical records priced against `master` and this branch:

```
record dated        master     fixed
2026-08-20            $2         $2
2026-09-01            $3         $2     <- the bug
2027-01-01            $3         $2
```

- **The external fact was checked against the live source**, which is the specific failure that caused this: the original entry was written from a price table that was correct at the time. Anthropic's current documentation states the $2/$10 introductory pricing "is now the standard price" and that the September 1 increase "will not occur."
- **Nothing else in the table moved.** Opus 5 / 4.8 / 4.7 / 4.6 at $5/$25, Fable 5 at $10/$50, Sonnet 4.6 at $3/$15, Haiku 4.5 at $0.80/$4, the `[1m]` tag stripping and the unknown-model fallback are all untouched and still asserted.
- **The `intro` mechanism is retained rather than deleted.** It is correct, and the next model to launch on a promotion will need it. Deleting it would mean rewriting it under time pressure later.
- 18/18 in `pricing-intro-window.mjs`, **139/139** overall, all six required checks green.

## The good judgement call

The obvious "fix" is to keep the `intro` block and push `until` far into the future. This PR explicitly refuses that, and the reasoning is right: **a date nobody is watching is exactly what caused this.** A far-future `until` would have re-armed the identical failure, just later and with less chance of anyone remembering why.

The test change is the other half of that. The old cutover instant is now the *regression guard* — `first instant of 2026-09-01 -> STILL $2/$10` — plus a 2028 check. So the moment that would have broken is now the moment that proves it is fixed.

## Findings

1. **The `intro` branch is now unexercised.** No entry carries one, so `pricingRateFor`'s intro path has no coverage. The PR documents this in the test header and tells the next person to add boundary coverage when they add a window, which is the honest handling — but it is still a live branch with no test. A tripwire asserting "no PRICING entry currently has an `intro`" would convert that into something CI notices; it needs `PRICING` exported, which is why it is not in this PR.

2. **The class of bug is unaddressed, and it is not unique to pricing.** This entry was correct when written and had a passing test. Both became wrong when an external fact changed, and nothing in the repo could detect it. `SUPPORTED_CC_RANGE`, the bundled template labels and the model catalog share that shape — the drift watchers exist precisely because someone already learned this for CC versions. There is no equivalent watcher for pricing. Worth a ticket; out of scope here.

## Verdict

Approving. Small, correct, demonstrated against the actual dates, and it resists the tempting fix that would have recreated the problem.

Worth recording how it surfaced: I stated the cutover as current fact in a public answer on #1018, having taken it from a cached table rather than the live source. The maintainer challenged it and asked for every claim in that comment to be re-verified. Six were wrong or overstated; this is the only one that would have mispriced something rather than merely misinformed a reader.

PR #1049, opened 2026-08-20
feat: pricing drift watcher, and fix Haiku 4.5's rates (#1048)

### LINKED ISSUE #1048: Pricing has no drift watcher — an entry that is correct today goes silently wrong when Anthropic changes it
Follow-up to #1047, which fixed one instance. The class is unaddressed.

## The shape

`PRICING` in `src/analytics.ts` encodes external facts — per-model input/output/cache rates. The Sonnet 5 entry was **correct when written and had a passing test**. Both quietly became wrong when Anthropic cancelled the scheduled increase and made $2/$10 permanent, and nothing in the repo could notice: no error, no failing test, no warning. It would simply have started reporting 50% high on 2026-09-01.

This repo already learned this lesson for a different external fact. CC versions get `cc-drift-watch`, template shape gets `cc-drift-template-watch`, the Agent SDK gets `sdk-drift-watch`, npm deps get `npm-drift-watch`. Every one exists because a value that tracks something outside the repo will drift, and a human will not notice.

**Pricing has no equivalent.** Neither does the `intro` mechanism's remaining branch, which now has no entry using it (see below).

## Why it matters more than it looks

The number is not decorative. `/analytics` `estimatedCost`, the TUI's cost column and `dario doctor`'s cost lines all derive from it, and users compare it against their own gateway's figure — dario#1018 is exactly that conversation. A cost estimate that is quietly wrong is worse than one that is obviously missing, because it still gets quoted.

## Possible shapes, roughly in order of effort

1. **A staleness assertion, cheapest.** A test that fails if `PRICING` has not been reviewed since some date — a `LAST_VERIFIED` constant plus a "this is older than N days" check. Crude, no network, catches the case where nobody has looked in a year.
2. **An `intro`-window tripwire.** `pricingRateFor`'s dated branch currently has **no entry exercising it**, so it is live code with no coverage. A test asserting "no `PRICING` entry carries an `intro`" would turn that into something CI reports the moment someone adds one — and prompt them to add boundary coverage. Needs `PRICING` exported.
3. **A real watcher**, matching the existing family: fetch the published pricing page on a schedule, diff the rates dario models, and open an issue on a delta. Most useful, most work, and it has the usual failure mode to design around — a page-format change must report "cannot parse" rather than "no drift".

(1) and (2) are small and would both have caught this one. (3) is the version that matches how the repo already handles CC drift.

## Related

Same class, worth checking whether they are covered: `SUPPORTED_CC_RANGE` (covered by `cc-drift-watch`), the bundled template labels (covered by `cc-drift-template-watch`), the model catalog and any model-capability assumptions (not obviously covered).

Anything keyed to a **future date** deserves particular suspicion — that is what made this one invisible rather than merely wrong.

askalf:
Closes #1048.

## Why

`PRICING` encodes external facts with no natural expiry. An entry is correct when written, has a passing test, and then goes silently wrong the moment Anthropic changes a price. No error, no failing test, nothing in the output to suggest the number moved — the first anyone hears is a wrong figure in front of a user.

It had already happened twice:

- **Sonnet 5** — the increase to $3/$15 was cancelled and $2/$10 made permanent, while the dated cutover stayed in the table. From 2026-09-01 every Sonnet 5 record would have priced 50% high (#1047).
- **Haiku 4.5** — carried Haiku 3.5's *entire row*. Fixed here.

The repo already watches every other external fact this way — `cc-drift-watch`, `cc-drift-template-watch`, `sdk-drift-watch`, `npm-drift-watch`. Pricing was the one feeding a user-visible number with nothing watching it.

## The watcher found a live bug on its first run

That is the strongest argument for it, so leading with it:

```json
{ "model": "claude-haiku-4-5", "field": "input",       "ours": 0.8,  "published": 1 }
{ "model": "claude-haiku-4-5", "field": "output",      "ours": 4,    "published": 5 }
{ "model": "claude-haiku-4-5", "field": "cacheRead",   "ours": 0.08, "published": 0.1 }
{ "model": "claude-haiku-4-5", "field": "cacheCreate", "ours": 1,    "published": 1.25 }
```

All four fields — Haiku **3.5**'s row, copied wholesale. Every Haiku 4.5 cost in `/analytics`, the TUI and `dario doctor` read about **20% low**, for an unknown period. After the fix the checker reports `"drift": [], "status": "clean"` against the live table.

## Failure mode is the design point

A watcher that quietly stops matching is worse than none: it reports "aligned" forever. So **every way of not knowing exits 2, never 0** — network failure, a missing table, a renamed column, HTML where markdown was expected, or a parse too thin to trust.

The workflow treats exit 2 as a skipped run and touches the issue in **neither** direction. Opening one would be a false alarm; closing one would silently retract a real finding because a page happened to be unreachable.

Two more things that could have produced a confident wrong answer:

- **Columns are matched by name, not position.** An upstream reorder would otherwise map cache-read prices onto cache-write fields and still report "aligned". Asserted with a fixture whose columns are deliberately shuffled.
- **The model table is located by its header content, not by ordinal.** The batch-pricing and fast-mode tables further down the same page are also `Model | $x / MTok` shaped, and a positional parser would happily read Sonnet 5's batch rate as its standard rate. Asserted.

It fetches the `.md` route (~43KB of `text/markdown`) rather than the human page (~1.1MB of app shell), and reports explicitly if it ever starts receiving HTML.

## Tests

34 assertions in `test/pricing-drift.mjs`, **most asserting the parser throws rather than succeeds**. Both real bugs are locked in as fixtures — the Sonnet 5 cutover and the Haiku 4.5 row — so a regression on either is a test failure rather than a discovery.

`npm test`: **140/140**. Live run against the published table: clean, exit 0.

## Notes for review

- `PRICING` is now exported. That is the only production-code change besides the Haiku rates.
- Daily at 09:23 UTC, off the hour and clear of the other watchers' `:00/:15/:27/:41/:47` cluster. Prices move rarely; noticing a day late costs a slightly stale estimate, not an outage.
- Fork-guarded per #1039.
- The workflow captures the checker's exit with `|| status=$?` rather than a bare call — under the runner's `bash -e` a bare call would kill the step on exit 1, which is the *drift* case this exists to report. (That is not hypothetical: `compat-test-self-hosted.yml` currently has the mirror-image of this bug — when its gate fails before the compat step runs, `exit "$COMPAT_EXIT_CODE"` dies with `numeric argument required` and the run reports a shell error instead of the real verdict. Visible on #1046 right now. Separate fix.)

### DIFF
--- a/.github/workflows/pricing-drift-watch.yml
+++ b/.github/workflows/pricing-drift-watch.yml
@@ -1,0 +1,150 @@
+name: Pricing drift watch
+
+# Diffs dario's PRICING table against Anthropic's published rates.
+#
+# WHY. PRICING encodes external facts with no natural expiry. An entry is
+# correct when written, has a passing test, and then goes silently wrong the
+# moment Anthropic changes a price. Twice already:
+#
+#   - Sonnet 5's increase to $3/$15 was cancelled and $2/$10 made permanent,
+#     while the dated cutover stayed in the table — from 2026-09-01 every
+#     Sonnet 5 record would have priced 50% high (#1047).
+#   - Haiku 4.5 carried Haiku 3.5's rates, ~20% low, for an unknown period —
+#     found by the first run of the checker this workflow drives (#1048).
+#
+# The repo already watches every other external fact this way (cc-drift-watch,
+# cc-drift-template-watch, sdk-drift-watch, npm-drift-watch). Pricing was the
+# one feeding a user-visible number with nothing watching it.
+#
+# CHEAP: one HTTPS GET of a ~43KB markdown page plus a build. No npm registry,
+# no CC download. Daily is plenty — prices move rarely, and the cost of
+# noticing a day late is a slightly stale cost estimate, not an outage.
+#
+# EXIT CODES (scripts/check-pricing-drift.mjs):
+#   0 aligned  -> close any open issue
+#   1 drift    -> open/refresh the issue
+#   2 could not determine -> WARN AND PASS, never treated as drift
+#
+# Exit 2 covers every way of not knowing: network failure, the page moving, a
+# renamed column, a parse too thin to trust. That distinction is the whole
+# design — a watcher that reports "aligned" because it could not read anything
+# is worse than no watcher, since the first sign of trouble is a wrong number
+# in front of a user.
+
+on:
+  schedule:
+    # 09:23 UTC daily. Off the hour and clear of the drift watchers, which
+    # cluster at :00/:15/:27/:41/:47.
+    - cron: '23 9 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: pricing-drift-watch
+  cancel-in-progress: false
+
+jobs:
+  pricing-drift-check:
+    name: Check published pricing
+    # Never from a fork. 54 forks inherit this schedule, and a fork filing
+    # pricing issues into its own tracker helps nobody. See #1039.
+    if: github.repository == 'askalf/dario'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
+    steps:
+      - uses: askalf/checkout-with-retry@115a6407547e9711edbc2e915838d495cad9583f  # v1.1.0
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020  # v4.4.0
+        with:
+          node-version: '22'
+
+      # The checker imports PRICING from dist/, so the build has to run. It is
+      # the only reason this job needs npm at all.
+      - name: Build
+        run: |
+          npm ci --no-audit --no-fund
+          npm run build
+
+      - name: Check pricing drift
+        id: check
+        run: |
+          # `|| status=$?`, NOT a bare call: this step runs under the runner's
+          # default `bash -e {0}`, so a non-zero exit would kill it before the
+          # status could be captured — and exit 1 is DRIFT, the case this exists
+          # to report. As an AND-OR list the failure is exempt from `set -e`.
+          status=0
+          node scripts/check-pricing-drift.mjs > pricing-drift.json 2>pricing-drift.err || status=$?
+          echo "status=$status" >> "$GITHUB_OUTPUT"
+          cat pricing-drift.json || true
+          cat pricing-drift.err >&2 || true
+          exit 0
+
+      - name: Open / refresh / close the pricing-drift issue
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          STATUS: ${{ steps.check.outputs.status }}
+        run: |
+          set -euo pipefail
+
+          # Recover the exit code even if the step's output was not set (the
+          # node process dying before writing $GITHUB_OUTPUT).
+          status="${STATUS:-2}"
+          echo "checker exit: $status"
+
+          gh label create pricing-drift --repo "$REPO" --color 1D76DB \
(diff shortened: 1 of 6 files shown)


sprayberry-redline (approved):

Authored by the same party approving it, and it closes an issue the same party filed. Not independent review; recording it rather than leaving it implicit.

## The finding that justifies the feature

The watcher caught a live bug on its **first run**, before this PR existed:

```json
{ "model": "claude-haiku-4-5", "field": "input",       "ours": 0.8,  "published": 1 }
{ "model": "claude-haiku-4-5", "field": "output",      "ours": 4,    "published": 5 }
{ "model": "claude-haiku-4-5", "field": "cacheRead",   "ours": 0.08, "published": 0.1 }
{ "model": "claude-haiku-4-5", "field": "cacheCreate", "ours": 1,    "published": 1.25 }
```

All four fields — Haiku **3.5**'s row, copied wholesale. Every Haiku 4.5 cost in `/analytics`, the TUI and `dario doctor` read ~20% low for an unknown period, and nothing in the repo could have told anyone. That is two silent pricing defects inside 24 hours (#1047 was the other), which is the case for the watcher made empirically rather than argued.

## Verified

- **Live run is clean after the fix.** `"drift": [], "status": "clean"`, exit 0 against the real published table — not a fixture.
- **The two hard-to-see wrong answers are both closed off**, and both are asserted:
  - *Columns matched by name, not position.* A reorder upstream would otherwise map cache-read prices onto cache-write fields and still report "aligned". Tested with a deliberately shuffled header.
  - *The table is located by header content, not ordinal.* The batch-pricing and fast-mode tables on the same page are also `Model | $x / MTok` shaped; a positional parser would read Sonnet 5's **batch** rate as its standard rate. Tested.
- **Every way of not knowing exits 2, never 0** — network failure, missing table, renamed column, HTML instead of markdown, implausibly thin parse. Verified by inspection and by five throwing test cases.
- **Exit 2 touches the issue in neither direction.** This is the subtle one and it is right: opening on exit 2 would be a false alarm, and *closing* on exit 2 would silently retract a real finding because a page was briefly unreachable.
- **Status capture survives `bash -e`.** `|| status=$?` rather than a bare call — a bare call would kill the step on exit 1, which is the drift case the workflow exists to report. Confirmed by running both the exit-1 and exit-2 shapes under `bash -e`.
- 34 parser assertions, `npm test` **140/140**, all six required checks green including `actionlint`.

## Scope judgement worth endorsing

Only models dario actually prices are compared. The published table lists fifteen; dario prices eight. Reporting the other seven as "drift" would produce a permanently-open issue that everyone learns to ignore — which is how watchers die. Conversely a model dario prices that has *vanished* upstream **is** reported, because a stale key falls through to the unknown-model fallback rate and prices silently wrong. That asymmetry is correct and non-obvious.

## Findings

1. **The `intro` branch remains unexercised.** `pricingRateFor`'s dated path still has no entry using it (noted in #1047 and unchanged here). This watcher does not cover it either — it compares flattened rates and drops `intro` in `comparable()`. So a future promotional window would be watched for its *standard* rate while its intro rate goes unchecked. Worth a line in the checker when the next `intro` lands.

2. **Cadence is a guess.** Daily at 09:23 is defensible — prices move rarely and a day's staleness costs a slightly wrong estimate, not an outage. But the two bugs it would have caught sat for an unknown period, so there is no evidence yet about how quickly divergence actually matters. Fine as a starting point; revisit if it fires.

3. **`MIN_PLAUSIBLE_ROWS = 8` is a magic number** that happens to equal the number of models dario prices today. If dario ever prices fewer than eight models the threshold silently becomes about the *published* table rather than a sanity floor. Harmless now, mildly confusing later.

## Verdict

Approving. It fixes a real mispricing, closes the class of bug rather than the instance, and its failure modes were designed first rather than discovered later — which is the specific thing that went wrong in both incidents it exists to prevent.

PR #1050, opened 2026-08-20
fix: compat reports its verdict, and the pricing watcher covers intro windows

askalf:
Two defects raised in review earlier today and deliberately left for a follow-up. This is the follow-up.

## 1. The compat job reported a shell error instead of its verdict

```yaml
exit "$COMPAT_EXIT_CODE"
```

`exit_code` is empty whenever the compat step never ran — and it does not run when an **earlier** step fails, which includes the `maxTested` gate refusing to validate a claim the runner cannot back. That is a legitimate, expected outcome, not a crash.

So the run died with:

```
line 3: exit: : numeric argument required
##[error]Process completed with exit code 2.
```

and surfaced *that* instead of `compat cannot validate maxTested=2.1.238`. **Visible on #1046 right now**: the gate did exactly its job, and a reader is shown a shell error rather than the reason.

Defaults to `1`. Reaching that step with no recorded exit code means something upstream already failed; the only wrong answer there is a green run.

## 2. The pricing watcher did not cover promotional windows

My own review of #1049 flagged that `comparable()` drops `intro`, so a model on a promotion would have its **standard** rate watched while the rate it actually bills at went unchecked.

**My first attempt at fixing it was wrong**, and wrong in the way this repo keeps re-learning: it compared the intro *rate* against the published standard. A promotional price differs from the standard by definition, so every entry carrying one would have reported drift **forever**, and the issue would have been learned-ignored inside a week. A watcher nobody reads is worse than no watcher — which is the exact argument I made in the review that prompted this.

What is genuinely checkable is the **date**:

```js
staleIntroWindows(pricing, nowMs)   // `until` in the past, or unparseable
```

That is precisely the Sonnet 5 shape (#1047). The bug there was never in the standard rate — it was a dated block that outlived the promotion it described. Dormant by design (no entry carries an `intro` today), so the next promotional window is covered on arrival rather than after the next incident.

Verified against the real case:

```
no intro block                                     -> 0 findings
promotion still running (until 2026-12-31)         -> 0 findings
last day of the window (until 2026-08-20, now 20th)-> 0 findings
lapsed (until 2026-08-19)                          -> lapsed
the Sonnet 5 shape (until 2026-08-31, now Sep 1)   -> lapsed
unparseable (until "soon")                         -> unparseable
```

Also renamed `MIN_PLAUSIBLE_ROWS` → `MIN_PUBLISHED_ROWS`. It is a sanity floor on the *published* table and happens to equal dario's model count today; tying the two together would silently turn it into "did we parse at least as many as we price", which is weaker and means something different.

## Tooling gap closed

Both of today's CI-only misses (`SC2001` on `drift-pr-heal.yml`, `SC2034` on `cc-oauth-health.yml`) got through because **`bash -n` checks syntax, not style**, and neither shellcheck nor actionlint was installed on the dev box.

`actionlint 1.7.1` — the same version this repo's CI pins — is now installed locally, and this PR was linted with it before pushing. Whole repo: clean.

Worth recording *why* shellcheck alone is not sufficient here. Extracting `run:` blocks and shellchecking them raw produces false positives, because shellcheck cannot know that `${{ github.repository }}` is an Actions expression rather than a shell expansion:

```
SC2296 (error): Parameter expansions can't start with {. Double check syntax.
```

actionlint substitutes those first and then delegates to shellcheck, which is why CI is clean on blocks that raw shellcheck rejects. The lesson is to run actionlint locally, not shellcheck.

## Tests

8 new assertions on `staleIntroWindows` — 42 in that file, **140/140** overall. `actionlint` clean across the repo.

### DIFF
--- a/.github/workflows/compat-test-self-hosted.yml
+++ b/.github/workflows/compat-test-self-hosted.yml
@@ -322,4 +322,20 @@ jobs:
         run: |
           # The job's pass/fail is the compat suite's exit code, regardless
           # of whether the comment step succeeded.
-          exit "$COMPAT_EXIT_CODE"
+          #
+          # DEFAULTED, because `exit_code` is empty whenever the compat step
+          # never ran — and it does not run when an EARLIER step fails, which
+          # includes the maxTested gate refusing to validate a claim the runner
+          # cannot back. In that case `exit ""` dies with
+          #
+          #   line 3: exit: : numeric argument required
+          #
+          # and the run surfaces a shell error (exit 2) instead of the real
+          # verdict. Seen on #1046: the gate correctly refused a
+          # maxTested=2.1.238 claim on a runner holding 2.1.237, and what a
+          # reader saw was the shell message, not "compat cannot validate".
+          #
+          # Failing closed on 1 is the right default — reaching this step with
+          # no recorded exit code means something upstream already failed, and
+          # the only wrong answer here is a green run.
+          exit "${COMPAT_EXIT_CODE:-1}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.39] - 2026-08-20
+
+- **Fixed: the compat job reported a shell error instead of its verdict.** `exit "$COMPAT_EXIT_CODE"` is empty whenever the compat step never ran — which is the case when an earlier step fails, including the `maxTested` gate refusing a claim the runner cannot back. The run then died with `exit: : numeric argument required` and surfaced that instead of "compat cannot validate maxTested=…". Visible on #1046. Defaults to 1 now: reaching that step with no recorded exit code means something upstream already failed, and the only wrong answer is a green run.
+- **The pricing watcher now covers promotional windows** (#1048 follow-up). It checks an `intro` block's `until` DATE rather than comparing its rate to the published standard — a promotional price differs from the standard by definition, so a rate comparison would report drift forever and the issue would be learned-ignored. A lapsed or unparseable `until` is reported instead, which is exactly the Sonnet 5 shape (#1047): the bug was never in the standard rate, it was a dated block that outlived its promotion. `MIN_PLAUSIBLE_ROWS` renamed `MIN_PUBLISHED_ROWS` — it is a sanity floor on the published table, deliberately not tied to how many models dario prices.
+
 ## [5.5.38] - 2026-08-20
 
 - **Fixed: Haiku 4.5 was priced at Haiku 3.5's rates** — `$0.80/$4` with cache `0.08/1` instead of `$1/$5` with cache `0.1/1.25`, so every Haiku 4.5 cost in `/analytics`, the TUI and `dario doctor` read about 20% low. The whole row had been copied from the wrong model.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.38",
+  "version": "5.5.39",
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

Authored by the same party approving it, closing two findings that same party raised in earlier reviews. Not independent review; recording it rather than leaving it implicit.

## 1. The compat exit-code default

`exit "$COMPAT_EXIT_CODE"` with an empty value dies as `exit: : numeric argument required`, so the run surfaces a shell error instead of the real verdict. The trigger is not exotic: the compat step does not run when an earlier step fails, and the `maxTested` gate failing is a **legitimate, expected** outcome, not a crash.

Verified this is live rather than theoretical — it is what #1046 is showing right now. The gate correctly refused a `maxTested=2.1.238` claim on a runner holding 2.1.237, and what a reader sees is `exit code 2 / numeric argument required`.

`${COMPAT_EXIT_CODE:-1}` is the right default, and the reasoning in the comment is the part worth keeping: reaching that step with no recorded exit code means something upstream already failed, so the only unacceptable outcome is a **green** run. Defaulting to 0 would have been the dangerous fix.

## 2. The intro-window coverage — note the first attempt

This is the interesting half, because the PR contains a rejected approach and says so.

The finding was that `comparable()` drops `intro`, leaving a promotional model's actual billing rate unwatched. The obvious fix — compare the intro *rate* against the published standard — is **wrong**, because a promotional price differs from the standard by definition. Every entry carrying one would report drift forever, the issue would be learned-ignored within a week, and the watcher would be dead while still appearing to work.

That is the same failure mode this PR's own sibling review warned about ("a permanently-open issue everyone learns to ignore"), reached by a different route. Catching it before merge rather than after is the right outcome; writing it down in the PR body so the next person does not re-derive the tempting version is better.

Checking the **date** instead is correct and is the actual Sonnet 5 shape (#1047): the bug was never in the standard rate, it was a dated block that outlived its promotion.

## Verified

- **Six behavioural cases, including both boundaries.** A running promotion reports nothing; the last day of a window still reports nothing; one day past reports `lapsed`; the exact Sonnet 5 case (`until 2026-08-31` evaluated 2026-09-01) reports `lapsed`; an unparseable `until` is reported rather than silently never expiring.
- **No false-positive risk today** — dormant by construction, since no entry carries an `intro`. The live checker still reports `"drift": [], "status": "clean"`.
- **`MIN_PUBLISHED_ROWS` rename is more than cosmetic.** The old name sat next to a value equal to dario's model count; the comment now states explicitly that tying the two together would turn a sanity floor into "did we parse at least as many as we price", which is a weaker and different check.
- 42 assertions in the file, **140/140** overall, all six required checks green.

## Tooling, which is the durable part

Two CI-only misses today (`SC2001`, `SC2034`) both got through because `bash -n` checks syntax and not style, and no linter was installed locally.

`actionlint 1.7.1` — the version CI pins — is now on the dev box and this PR was linted with it before pushing. The PR body also records **why shellcheck alone is not the right local tool**: raw-extracting `run:` blocks makes shellcheck reject `${{ github.repository }}` as `SC2296 Parameter expansions can't start with {`, because it cannot know that is an Actions expression. actionlint substitutes first and then delegates. Without that note the next person installs shellcheck, sees false errors, and concludes local linting is not worth it.

## Findings

1. **The tooling fix is not enforced.** actionlint is installed on one machine and used by convention. Nothing stops the next workflow edit from skipping it, and the failure mode is identical to today's — a red CI run twenty minutes later. A `lint:workflows` npm script, or a pre-push hook, would make it the default path rather than a remembered one. Small, and out of scope here.

2. **`staleIntroWindows` compares against `Date.now()` at call time**, so a window lapsing mid-run is a coin flip on which side of midnight the job lands. Irrelevant at daily cadence with day-granularity dates; noting it only because date-boundary logic is exactly what produced #1047.

## Verdict

Approving. Both findings closed properly rather than papered over, the rejected first attempt is documented instead of hidden, and the tooling gap that let two defects reach CI is closed at the source.

PR #1051, opened 2026-08-20
docs: how to run more than one dario against the same accounts (#993)

askalf:
Addresses #993.

## The gap

The refresh lock shipped across #1000 (Cloudflare Durable Object), #1006 (a real dual-process race test) and #1008 (a Redis backend, built specifically to answer @ramarro123's airgapped objection and his Redis suggestion).

None of it was discoverable:

```
mention of DARIO_REFRESH_LOCK_URL in docs/ or README.md:  none
```

Only the two backend subdirectory READMEs document it, and nobody finds those without already knowing they exist. We built the thing that was asked for, validated it against a real race, and a user could not learn it existed.

## What the page leads with, and why

**This is safe credential sharing, not HA.** Getting that wrong is expensive, so it is the first thing on the page rather than a footnote:

| | |
|---|---|
| OAuth refresh race | **solved** |
| rate-limit accounting | **not solved** — `pool.ts` holds the snapshot in process memory |
| sticky routing | **not solved** — the binding is per-process |

Anyone reading "you can share credentials" as "dario is HA" will scale a Deployment to 2 and blow through their 5h/7d windows wondering why. There is a table of what to actually run for each goal, and for most people the honest answer is still one replica — roughly where the issue itself landed.

## Details worth keeping

- **Why a plain mutex does not fix the race.** The loser waits, acquires the lock, then refreshes with an already-stale token — it just loses more slowly. The fix is that the loser adopts the *winner's* credentials.
- **Fail-open is a trade with a stated cost.** During a lock outage you are back to the plain race. Documented as such rather than presented as a guarantee.
- **The Redis backend is a coordination point, not consensus.** No leader election, not Raft. #993 asked about Raft and split-brain directly, so the page says plainly what this is not.
- **Verification is built around `--no-lock`** — reproduce the failure first, then see it prevented. Seeing the race happen is what makes the fix mean something.

## Note

Docs and README only; no source change. README line-count gate passes (+290, tolerance 1000).


sprayberry-redline (approved):

Authored by the same party approving it. Not independent review; recording it rather than leaving it implicit.

## What this closes

The refresh lock shipped across #1000, #1006 and #1008 — two backends, a real dual-process race test, and a Redis option built specifically to answer the airgapped objection raised on #993. It was undiscoverable: `DARIO_REFRESH_LOCK_URL` appeared nowhere in `docs/` or the README, only in two backend subdirectory READMEs.

A feature a user cannot find is a feature that was not delivered, regardless of how well it was tested.

## Verified

- **The scope claim is accurate against the source.** `pool.ts` keeps `accounts: Map` in process memory and updates it only from headers that instance saw; `computeStickyKey` binds per-process. So "rate-limit accounting and sticky routing are NOT solved" is a statement about the code, not a hedge.
- **The refresh-race description matches `src/accounts.ts`** — including the part that is easy to get wrong: a plain mutex does not fix it, because the loser then refreshes with an already-stale single-use token. The loser adopting the winner's credentials is the actual mechanism.
- **Fail-open is documented as a trade with its cost stated** (during a lock outage you are back to the plain race), matching the comment in `accounts.ts` that calls the lock "a resilience nice-to-have layered on top of dario's actual job".
- **Both backend READMEs exist and are linked**, so the page routes rather than duplicates.
- **The `--no-lock` flag is real** and the verification section is built around it correctly.
- Docs + README only, no source change. README line-count gate passes (+290, tolerance 1000).

## The judgement worth endorsing

Leading with "safe credential sharing, **not** HA" is the right call and the opposite of how this would normally be written. The tempting version announces the feature and buries the caveat; that version gets someone to scale a Deployment to 2 and silently overshoot their 5h/7d windows. Putting the non-solved half in the first table, and stating that one replica is still the honest answer for most people, costs nothing and prevents the actual failure.

Saying plainly that the Redis backend is "a coordination point, not a consensus system — not Raft, no leader election" is the same instinct. #993 asked about Raft and split-brain specifically, and the honest answer is that this does not do that.

## Findings

1. **The page is not linked from `docs/multi-account-pool.md`**, which is where someone reasoning about pool scaling would already be reading. The README entry is the only path in. One cross-link would close it.
2. **Nothing exercises the doc's setup instructions.** The Redis `docker run` block and the env-var contract are correct today by inspection; if the contract ever changes, this page drifts silently — the same class of problem the pricing watcher was built for. Low stakes, and not worth a watcher, but worth knowing it is prose rather than a tested path.

## Verdict

Approving. It documents something real, is accurate about what it does not do, and gives the reporter a way to prove it on his own cluster rather than taking our word for it.

PR #1053, opened 2026-08-21
fix: pin redis-lock's base image, and watch every Dockerfile with Dependabot

askalf:
Closes the open Scorecard alert on `redis-lock/Dockerfile:1`.

## The alert

```
tool:     Scorecard          rule: PinnedDependenciesID (medium)
file:     redis-lock/Dockerfile:1
message:  score is 9: containerImage not pinned by hash
created:  2026-08-18          ← when #1008 landed
```

Real, not a false positive. It was the **only** Dockerfile in the repo not pinned by digest:

| file | before |
|---|---|
| `Dockerfile` | `node:22-alpine@sha256:16e22a55…` ✅ |
| `.clusterfuzzlite/Dockerfile` | `base-builder-javascript@sha256:717f88cd…` ✅ |
| `redis-lock/Dockerfile` | `node:20-alpine` ❌ |

So an inconsistency rather than a policy question — and it was on the file users **copy into their own infrastructure** to run the refresh lock, which makes it the one most worth getting right.

Now pinned to the same `node:22-alpine` digest the root Dockerfile already uses and CI already exercises. That also drops an inconsistent Node major; `server.mjs` and `resp-client.mjs` import only `node:http` and `node:net`, so there is nothing version-specific to break.

## Pinning alone would have been the worse half of the fix

A pinned digest is **frozen**. Without something bumping it, the base stops receiving security patches and the pin quietly becomes the liability it was meant to prevent — while still satisfying the Scorecard check. That is precisely how this trap gets set.

`.github/dependabot.yml` covered `npm` (×3) and `github-actions`. It did **not** cover `docker`, so it now gains an entry per Dockerfile directory:

```
docker  /
docker  /redis-lock
docker  /.clusterfuzzlite
```

Verified all three are covered:

```
.clusterfuzzlite/Dockerfile   dir=/.clusterfuzzlite   covered
Dockerfile                    dir=/                   covered
redis-lock/Dockerfile         dir=/redis-lock         covered
```

Dependabot does not recurse, so it is one entry per directory — noted in a comment, because adding a Dockerfile elsewhere later means adding it here or it ships pinned and unwatched.

## The part the alert didn't say

**The root image was pinned-but-unwatched before this change too.** The alert pointed at `redis-lock` because that was the newest instance, but the gap was repo-wide: two digests, nothing bumping either. Fixing only what the scanner named would have left the main production image in the same state.

## Notes

- House style followed — weekly, Monday 09:00 UTC, `chore`-prefixed, matching the existing entries.
- No source change. `actionlint` clean; `dependabot.yml` parses.
- Expect a burst of Dependabot PRs on the first Monday tick as it catches up on three previously-unwatched images.


sprayberry-redline (dismissed):

**What I checked**
- `redis-lock/Dockerfile:6` — `FROM node:22-alpine@sha256:16e22a550f3863206a3f701448c45f7912c6896a62de43add43bb9c86130c3e2` matches the root `Dockerfile`'s pinned digest exactly (verified against `Dockerfile:3` on `master`), so this isn't just "a" pin, it's consistent with the image already built and exercised by CI.
- `.github/dependabot.yml` — adds `package-ecosystem: docker` entries for all three Dockerfile directories (`/`, `/redis-lock`, `/.clusterfuzzlite`), which correctly addresses the "pin without a bump mechanism" trap the PR description calls out. Dependabot doesn't recurse, and each entry maps to an actual Dockerfile location.
- `CHANGELOG.md` / `package.json` — version bump and changelog entry are consistent with house style and the rest of the diff.
- CI check-rollup: all checks green (CodeQL, build 18/20/22, test, docker-cap-drop-smoke, actionlint, compat).

No correctness, security, or test-coverage issues in the changed lines. This is a supply-chain hygiene fix (Dockerfile pin + Dependabot coverage) with no source/runtime code touched — scope matches the stated intent.

**Minor**
- The dependabot config comments are a bit verbose for a config file (explaining "why pin digests" rationale inline), but that's a documentation-density call, not a defect — leaving to author preference.

sprayberry-redline (approved):

## What I checked
- `gh pr diff 1053` (4 files, +61/-2) and CI check-rollup (`gh pr view --json statusCheckRollup`) — all 13 checks green (build 18/20/22, test, docker-cap-drop-smoke, CodeQL, actionlint, etc.).
- `redis-lock/Dockerfile:7` now reads:
  ```
  FROM node:22-alpine@sha256:16e22a550f3863206a3f701448c45f7912c6896a62de43add43bb9c86130c3e2
  ```
  I fetched the root `Dockerfile` from the repo directly and confirmed it pins the identical `node:22-alpine@sha256:16e22a550f3863206a3f701448c45f7912c6896a62de43add43bb9c86130c3e2` digest — so this isn't a fresh/unverified digest, it's reuse of one already exercised by the repo's own CI.
- `.github/dependabot.yml` adds three `package-ecosystem: docker` entries (`/`, `/redis-lock`, `/.clusterfuzzlite`), one per directory containing a Dockerfile, which is correct given Dependabot doesn't recurse — matches the PR's own claim and the file list.
- `package.json`/`CHANGELOG.md` version bump (5.5.42 → 5.5.43) is consistent between the two files.

## What's good
- The PR doesn't stop at satisfying the Scorecard rule (pinning the digest) — it also closes the gap that pinning alone creates (no bump path), and calls out that the root image had the same unwatched-pin gap even though the alert didn't name it. That's the right scope for this class of fix.
- Digest reuse from the existing root Dockerfile avoids introducing a second, unverified pin.

No correctness, security, or test-coverage issues found in the changed lines.

PR #1058, opened 2026-08-21
ci(canary): probe Opus 5, and stop an unreadable model field holding the alert open

### LINKED ISSUE #1003: CC canary: claim=seven_day served=<?> (2026-08-18)
## Billing classifier canary: 🟡 WARN

Generated by [`cc-billing-classifier-canary.yml`](.github/workflows/cc-billing-classifier-canary.yml) on 2026-08-18T06:16:16+00:00.

- **Captured `representative-claim`:** `seven_day` → bucket `subscription`
- **Requested model:** `claude-opus-4-8` → **served:** `<unparseable>` (`warn`)
- **fallback-percentage:** `0.5` · **5h-utilization:** `0.0`
- **Urgency:** investigate — a probe signal was missing or had an unrecognized value

### Likely causes

1. **Anthropic changed the classifier rules** — added a new signal dario doesn't replay, tightened an existing one, or flipped a threshold. Cross-check with the latest [`cc-drift-template-watch`](../../actions/workflows/cc-drift-template-watch.yml) report: if that's also flagging drift, the classifier change probably correlates with a wire-shape change in CC.
2. **Bundled template stale** — but in a way the drift watcher hasn't caught yet (e.g. a new signal slot CC doesn't even populate).
3. **OAuth credential is on a different account class** — pro/max account converted to api, or got disabled.

### How to investigate

1. Check recent [drift watcher runs](../../actions/workflows/cc-drift-template-watch.yml) — is there an open `cc-drift-template` issue or auto-rebake PR? Wire-shape may already be flagged.
2. Run the canary manually after any fix: `gh workflow run cc-billing-classifier-canary.yml --ref master`
3. If the cause is a new classifier signal, expect to need a code change in `src/cc-template.ts` to add the new slot to the canonical-rebuild path — and a corresponding template re-bake.

This alert auto-closes when the canary next returns a subscription bucket.

askalf:
## What

The daily billing-classifier canary (`cc-billing-classifier-canary.yml`) probes the flagship model to catch a silent server-side downgrade, and asserts the request bills against a subscription bucket. Two fixes:

1. **Probe Opus 5, not Opus 4.8.** The downgrade probe was pinned to `claude-opus-4-8`. `claude-opus-5` is now the flagship *and* the model the fleet actually runs (Developer/Alf), so the probe was guarding a stale target — a downgrade of the model we depend on wouldn't be what it watches. The probe model is now a single top-level `PROBE_MODEL` env var (`claude-opus-5`), inherited by both the probe and issue-management steps, so the next bump is one line.

2. **An unreadable served model no longer holds the alert open when billing is healthy.** When the probe can't parse the served `model` back from the response body, it recorded `model_verdict=warn`, which forced the combined status to `warn` and opened/held a `cc-billing-canary` issue. But an unreadable served model *while billing itself classifies as a subscription bucket* is a probe-quality note, not a billing incident.

## Why (concretely: #1003)

Issue **#1003** has been open since 2026-08-18, re-warning every single day. Investigation:

- The canary runs successfully daily and **auto-closes only on a clean run**. Every run since 08-18 came back 🟡 WARN, so it stayed open and just appended a comment.
- The WARN was **not** a billing problem — the billing claim is `seven_day` → `subscription` → **pass**. dario has been billing correctly the whole time.
- The WARN was purely the secondary model-downgrade check: `served_model` came back **empty**. The canary probes its own freshly-built dist on `:3457` (by design, to test the repo's code); that probe returns a body with no parseable `model` field, so `model_verdict=warn` → held the issue open.
- Reproduced the exact probe against production dario (`:3456`): **both** `claude-opus-4-8` and `claude-opus-5` return a clean, parseable `"model"` field. Production is healthy; the empty body is specific to the canary's own probe path.

So #1003 is a false positive that couldn't self-clear. This PR makes an empty served model a run-log note instead of an issue-holding WARN, so the canary recovers and closes it.

## What is preserved

A **confirmed** downgrade — the probe served a *real but cheaper* model — is still `model_verdict=fail` and opens an issue loudly. Only the *empty/unreadable* case is demoted, and only when billing is healthy. Billing `fail`/`warn` still escalate exactly as before.

## Verdict logic

```
model_verdict:
  ${PROBE_MODEL}*  -> pass         # served as requested
  ""               -> unreadable   # body had no model field (was: warn)
  *                -> fail         # a DIFFERENT, cheaper model — real downgrade

status:
  billing=fail OR model_verdict=fail   -> fail   (opens issue)
  billing=warn                         -> warn   (opens issue)
  else                                 -> pass   (closes issue)   <- unreadable lands here when billing is healthy
```

## Notes

- No version bump: a workflow change takes effect on merge to `master`; no release is needed.
- I could not reproduce the `:3457` empty-body condition directly (that instance is ephemeral, only up during a workflow run), so this fix is deliberately **robust to it**: even if the served model stays empty, a healthy billing bucket now yields `pass` and closes the issue. If the empty body reflects a real dist regression, `cc-drift-template-watch` is the watcher that would surface it.
- `actionlint` clean; YAML validates; CHANGELOG `[Unreleased]` entry added.

Closes #1003.

### DIFF
--- a/.github/workflows/cc-billing-classifier-canary.yml
+++ b/.github/workflows/cc-billing-classifier-canary.yml
@@ -27,7 +27,7 @@ name: CC billing classifier canary (self-hosted)
 #       - api:                   api                                     ✗ fail (classifier flip)
 #       - unknown:               anything else                           ⚠ warn (missing/new value)
 #
-# (B) MODEL DOWNGRADE — the probe requests the TOP model (claude-opus-4-8) and
+# (B) MODEL DOWNGRADE — the probe requests the TOP model (${PROBE_MODEL}) and
 #     asserts the `model` field in the RESPONSE BODY is still Opus. If Anthropic
 #     silently serves a cheaper model (Sonnet/Haiku) under load, served != asked
 #     and we fail. You cannot detect a downgrade by asking for the cheap model,
@@ -35,9 +35,12 @@ name: CC billing classifier canary (self-hosted)
 #     undocumented `fallback-percentage` + 5h-utilization headers are captured
 #     for the alert (context only — not asserted, meaning isn't documented).
 #
-# Combined verdict: any signal fail → fail; else any warn → warn; else pass.
+# Combined verdict: a billing flip or a CONFIRMED model downgrade → fail; an
+# unrecognized billing value → warn. An UNREADABLE served model (empty body
+# field) is a note only — it does NOT hold the alert open while billing is
+# healthy (that false alarm kept #1003 open for days). See the probe step.
 #
-# Cost: ~1 small subscription request per day (opus-4-8, ≤ 16 output tokens).
+# Cost: ~1 small subscription request per day (PROBE_MODEL, ≤ 16 output tokens).
 # Cron offset to 05:50 UTC. The old 06:30 slot sat exactly ON the drift
 # watcher's :00/:30 every-30-min grid (contrary to what this comment used
 # to claim) and inside the 06:30–06:37 single-runner pileup with
@@ -62,6 +65,14 @@ on:
 
 permissions:
   contents: read
+
+env:
+  # The model the downgrade-probe requests. It must be the current FLAGSHIP:
+  # the probe catches a silent server-side downgrade to a cheaper model, which
+  # you can only detect by asking for the most-capable one. Bump when a new top
+  # Opus ships (was claude-opus-4-8; claude-opus-5 is flagship and is what the
+  # fleet actually runs). One place to change, inherited by every step.
+  PROBE_MODEL: claude-opus-5
 
 jobs:
   canary:
@@ -152,7 +163,7 @@ jobs:
         run: |
           set -eo pipefail
 
-          # Probe with the TOP model (claude-opus-4-8), tiny output budget.
+          # Probe with the TOP model (${PROBE_MODEL}), tiny output budget.
           # Two signals from one request:
           #   1. BILLING — the `representative-claim` response header (which
           #      pool the request billed against).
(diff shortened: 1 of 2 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped workflow fix. Approving.

**What I checked**
- `gh pr diff` for both changed files (`.github/workflows/cc-billing-classifier-canary.yml`, `CHANGELOG.md`).
- CI check-rollup: all 13 checks green (CI build/test matrix, CodeQL, actionlint, triage/label bots), none pending or failing.
- Traced the verdict logic end-to-end: `model_verdict` case statement (now `unreadable` instead of `warn` for an empty served-model field) feeds the combined `status` calculation, which now only escalates to `warn` on `billing=warn`, and the new `if [ "$model_verdict" = "unreadable" ]` block only emits a `::warning::` log annotation — it does not touch `status`. This matches the PR's stated intent (an unreadable served model no longer holds the alert open when billing is healthy) and I don't see a path where `unreadable` still forces `warn`/`fail`.
- `PROBE_MODEL` is defined once at `env:` (top-level, workflow-wide) and consistently substituted into the curl payload, the case pattern (`"${PROBE_MODEL}"*`), the force-status override branch, and the two issue-body summaries — no leftover hardcoded `claude-opus-4-8` in the executable paths (only in trailing comments/history, which is fine).

**What's good**
- The env-var refactor for `PROBE_MODEL` removes 5+ duplicated hardcoded model strings, so a future flagship bump really is a one-line change as claimed.
- The distinction between "confirmed downgrade" (`fail`, still loud) and "unreadable body" (`unreadable`, now just a log note) is a sound fix for the described false-positive (#1003) without weakening the actual downgrade detector.
- CHANGELOG entry is accurate and matches the diff.

No correctness, security, or test-coverage concerns in the changed lines.

PR #1067, opened 2026-08-22
Never send or stamp empty text blocks — filter them in the rebuild, guard the breakpoint (v5.5.49)

### LINKED ISSUE #1066: cache_control stamped on empty text blocks kills long sessions with a 400
## Summary

`applyCcPromptCaching` places the rolling cache breakpoint on the last content block of a user turn without checking that the block is a non-empty text block. When that block is `{type:'text', text:''}`, the request upstream is rejected:

```
400 invalid_request_error
messages.10.content.0.text: cache_control cannot be set for empty text blocks
```

## Impact

This kills long agentic sessions rather than degrading them. The empty text block has to accumulate in the transcript, so requests succeed for several tool rounds and then every subsequent request from that turn onward fails with the same 400. In practice it caps how long any sub-agent can run, and because the failure surfaces through a proxy layer as a model-group error (`Available Model Group Fallbacks=None`) it reads as a routing or model problem rather than a request-shaping one.

## Root cause

`src/cc-template.ts` (v5.5.43, also present in 5.4.26):

```js
const blocks = msg.content;
blocks[blocks.length - 1] = { ...blocks[blocks.length - 1], cache_control: cacheControl };
```

The loop guards `msg.content.length === 0` but never inspects the block it stamps. Any trailing empty text block therefore receives a breakpoint.

## Reproduction

Unit, against the shipped build — no credentials or live proxy needed:

```js
import { applyCcPromptCaching } from '@askalf/dario/dist/cc-template.js';

const req = { messages: [
  { role: 'user',      content: [{ type: 'text', text: 'a' }] },
  { role: 'assistant', content: [{ type: 'text', text: 'b' }] },
  { role: 'user',      content: [{ type: 'text', text: 'q' }, { type: 'text', text: '' }] },
]};
applyCcPromptCaching(req, { type: 'ephemeral' });
// last block is now { type:'text', text:'', cache_control:{type:'ephemeral'} } -> upstream 400
```

End-to-end, a request carrying `cache_control` on an empty text block is forwarded to Anthropic unmodified and returns the 400 above, so there is no sanitizing pass downstream that would catch this.

## Suggested fix

Walk back past empty text blocks when choosing the breakpoint target, and skip the turn if no eligible block remains (it stays reachable via the next turn's anchor). PR to follow.

## Environment

- dario 5.5.43 (bug also present in 5.4.26)
- Claude Code -> LiteLLM -> dario -> Anthropic

LiveNathan:
Fixes #1066.

## Problem

`applyCcPromptCaching` stamps the rolling breakpoint onto the last content block of a user turn without inspecting it:

```js
blocks[blocks.length - 1] = { ...blocks[blocks.length - 1], cache_control: cacheControl };
```

The loop guards against an empty content *array* but not an empty *text block*. When the trailing block is `{type:'text', text:''}`, upstream rejects the entire request:

```
messages.10.content.0.text: cache_control cannot be set for empty text blocks
```

Because such a block only enters the transcript partway through a run, this doesn't fail fast — sessions complete several tool rounds and then every subsequent request dies. Downstream it surfaces as a model-group error, which sends people looking at routing rather than request shaping.

## Change

Walk back past empty text blocks to find an eligible target. If a turn has nothing legal to stamp, skip it rather than emit an illegal breakpoint — the turn stays reachable via the next request's rolling breakpoint, so this costs at most one anchor on a degenerate turn and never a 400.

The guard is deliberately narrow: it only matches `type === 'text'` blocks with a missing or empty string `text`. `tool_result` and other block types have no `text` field and remain eligible, so fan-out turns keep the anchor behavior added for #678.

## Tests

Six assertions added to `test/prompt-caching.mjs`, covering the trailing-empty case, the all-empty turn (fall through to an earlier turn), and `tool_result` eligibility as a regression guard on the guard's narrowness.

Per CONTRIBUTING:

- `npm run build` — clean.
- `npm test` — 140/140 pass.
- `npm audit --production --audit-level=high` — 0 vulnerabilities.
- No new runtime dependencies.

`cc-template.ts` is on the list requiring live-proxy verification. I ran the patched build against a live proxy: the reproducing request now succeeds, breakpoints are still placed on normal turns, and cache writes/reads are unchanged. I did not run `test/compat.mjs` in full, since it consumes real subscription usage on my own account — happy to if you'd like it before merge.


sprayberry-redline (dismissed):

## What I checked
- `gh pr diff` for the two changed files (`src/cc-template.ts`, `test/prompt-caching.mjs`).
- CI rollup: only `triage` (pass) and `auto-merge` (skipped, expected for a non-bot PR) run in this repo's workflow set — no unit-test CI gate exists here, so I relied on the diff plus the author's reported local run (140/140, `npm run build` clean, `npm audit` clean, live-proxy verification).

## Findings
None blocking.

## What's good
- `src/cc-template.ts:1548-1552` (`isEmptyTextBlock`) is a narrow, correctly-scoped guard: it only matches `type === 'text'` blocks with a missing/empty `text`, so `tool_result` and other block types remain eligible — verified by the added `tool_result block is still eligible for a breakpoint` test.
- The walk-back loop (`let bi = blocks.length - 1; while (bi >= 0 && isEmptyTextBlock(blocks[bi])) bi--; if (bi < 0) continue;`) correctly falls through to `continue` (skipping the `stamped++`) when a turn has nothing eligible, rather than emitting the illegal `cache_control` on an empty block that caused the original 400 in #1066.
- Test coverage directly exercises the three relevant cases: trailing-empty-block (falls back to the prior block in the same turn), all-empty-turn (falls through to an earlier turn's anchor), and the tool_result regression guard — these map 1:1 to the code paths touched.
- Change is minimal and localized; no unrelated refactoring.

askalf:
Pushed two commits onto this branch (thanks for the clean diagnosis and the guard, @LiveNathan — both survive).

**`21978d7` — the second half: filter the blocks out of the rebuilt request.** Live A/B against upstream showed the walk-back guard is necessary but not sufficient: Anthropic rejects an empty text block in the final user turn **outright**, stamped or not (`messages: text content blocks must be non-empty`), and treats whitespace-only text the same way. So on the guarded build, the reproducing shape still died — same 400, different message.

The rebuild now strips empty/whitespace-only text blocks from message content, and drops a user turn left with no blocks when it sits **mid-conversation** (the API combines the now-adjacent same-role turns — this is the shape that rides along in history). A **final** turn emptied this way is deliberately kept, per the #1033 decision: popping it would expose the assistant turn behind it and turn an honest error into a misleading prefill rejection. Your breakpoint guard stays as defense-in-depth.

Live matrix on the result (real upstream, request IDs logged):

| last-relevant shape | before | after |
|---|---|---|
| `["text", ""]` trailing empty | 200 | 200 |
| `[tool_result, ""]` | 200 | 200 |
| `[""]` mid-conversation | 200* | 200 |
| `[" "]` mid-conversation | — | 200 |
| `[""]` final turn | 400 (cache_control error) | 400 `user messages must have non-empty content` — accurate, names the message |
| `[" "]` final turn | 400 | 400 same, accurate |

\* mid-position empties turned out to pass upstream today even when stamped — the filter makes that robust rather than lucky, and keeps any such block off the wire entirely.

Also in: `test/empty-text-blocks.mjs` (12 assertions over the rebuilt wire body), full suite 141/141, and the v5.5.49 release bump so this ships on merge.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — the fix is well-scoped, and the two-layer defense (rebuild-side filter + breakpoint walk-back) is correctly ordered and tested.

**What I checked**
- `gh pr diff` for the full changed surface (`src/cc-template.ts`, `test/empty-text-blocks.mjs`, `test/prompt-caching.mjs`, version bumps).
- Surrounding context in `src/cc-template.ts` (the pre-existing "drop trailing empty turns" loop and the `dario#1033` final-turn rule it must not regress) and `src/proxy.ts:2912` for the `applyCcPromptCaching` call site, via a shallow clone of `master` + the PR ref.
- CI: `gh pr checks 1067` — all applicable checks pass (`build (18/20/22)`, `test`, `CodeQL`, `actionlint`, `docker-cap-drop-smoke`, `validate-package-json`); `compat`/`auto-merge`/`label`/`triage`/`wire-drift` show as skipped (conditional workflows, not failures).

**Findings**

None blocking. I traced the two behavioral changes closely:
- `src/cc-template.ts:1787-1798` (filter empty/whitespace text blocks from every message) runs *before* `src/cc-template.ts:1808-1812` (drop a resulting empty-content user turn, but only for `i < messages.length - 1`, i.e. never the final turn) — correctly preserves the `dario#1033` invariant that the final turn stays in place so a malformed request fails with the accurate upstream error instead of a misleading prefill rejection.
- The breakpoint walk-back in `applyCcPromptCaching` (`isEmptyTextBlock`, `src/cc-template.ts:1548-1553`) is defense-in-depth for callers that bypass `buildCCRequest`'s filter (as the standalone unit tests in `test/prompt-caching.mjs` do) — in the actual `proxy.ts:2912` call path the filter already runs first, so this is consistent with the stated intent rather than dead code.
- `test/empty-text-blocks.mjs` exercises the trailing-empty, whitespace-only, mid-conversation-drop, final-turn-kept, and `tool_result`-eligibility cases end-to-end through `buildCCRequest`, matching the scenarios described in the PR body and changelog.

I looked for a role-alternation issue where dropping a mid-conversation empty user turn between two assistant turns leaves adjacent same-role assistant messages (see the `test/empty-text-blocks.mjs` "session-killer" case, which asserts exactly that resulting shape). The PR body and changelog state this was live-verified against the upstream API across six shapes including this one, and I have no evidence in the diff contradicting that claim, so I'm not treating it as a blocking finding — flagging it here only as the one area I'd want CI/live-proxy confidence on if it ever regresses.

**What's good**
- Root-caused correctly: the walk-back guard from the original #1067 issue only changed *which* error upstream returned, not whether the request succeeded — this PR fixes the request shape itself, which is the right layer.
- Comments explain non-obvious invariants (why the final turn is deliberately kept, why only assistant trailing turns are popped) rather than restating the code.
- Version/changelog/lockfile bumps are consistent (5.5.49 across `package.json`, `package-lock.json`, `CHANGELOG.md`).

Issue #1077, opened 2026-08-23
#1067 doesn't cover genuine CC clients: empty-block filter sits below the isGenuineCCClient early return


LiveNathan:
### Summary

#1067 fixed the empty-text-block 400 for every client **except real Claude Code**. The `isGenuineCCClient` branch in `buildCcRequest` returns at `src/cc-template.ts:1740`, and the `── Drop empty text blocks from history ──` filter added by #1067 sits at `:1787` — 47 lines below the return, unreachable on that path.

Net effect: a genuine CC session that gets one empty text block into its transcript still dies with

```
400 {"type":"error","error":{"type":"invalid_request_error",
"message":"messages: text content blocks must be non-empty"}}
```

and, because the block stays in history, every subsequent request of that session dies too. This is the exact session-killer #1066 describes, on the one client that matters most.

### The gap

`src/cc-template.ts:1725-1731`, inside the genuine-CC branch:

```ts
for (const msg of messages) {
  if (Array.isArray(msg.content)) {
    for (const block of msg.content as Array<Record<string, unknown>>) {
      delete block.cache_control;
    }
  }
}
```

This strips the breakpoint but leaves the block. #1067's own comment at `:1787` states the principle this branch violates:

> Filtering here fixes the REQUEST, which no breakpoint guard can: skipping the stamp still leaves the illegal block on the wire.

`applyCcPromptCaching`'s `isEmptyTextBlock` walk-back has the same limitation by design — it declines to *stamp* the block, but nothing removes it.

I think the branch comment at `:1707-1710` is what produced this:

> Messages, thinking, effort, max_tokens, top-level key order: untouched — the client is the authority on its own wire shape.

That is right for `max_tokens` and key order. An empty text block isn't a wire-shape preference though — it's a hard upstream reject, so "forward verbatim" turns into "forward a request we know will 400."

### Observed impact

A `manage-issue` run with two sub-agents. Both sub-agents completed normally in 3 and 6 minutes (`stop_reason: end_turn`, full reports written, files on disk). Each `<task-notification>` was delivered to the parent within 1 second of the agent finishing. The parent's next API call 400'd on both, so **neither result was ever surfaced** — the run looked stalled for ten hours while the work had in fact been done.

Parent transcript, agent #1:

```
20:59:37  sub-agent finishes (end_turn)
20:59:38  queue-operation  <task-notification> ... <status>completed</status>
20:59:39  assistant  API Error: 400 ... "messages: text content blocks must be non-empty"
```

[...]

PR #1078, opened 2026-08-23
fix: hoist the empty-text-block filter above the genuine-CC early return (v5.5.52)

askalf:
## What

Hoists #1067's empty-text-block filter — and the mid-conversation empty-user-turn drop — above the `isGenuineCCClient` early return in `buildCCRequest`, so genuine Claude Code clients get the same protection as everyone else. Addresses #1077.

## Why

Exactly as @LiveNathan's report lays out (verified by inspection): the filter sat below the branch's `return`, unreachable for real CC — the one client whose messages are forwarded verbatim. An empty text block entering the transcript mid-run still 400'd every subsequent request of the session (`messages: text content blocks must be non-empty`, #1066), presenting as "sub-agents hang and never report" while the work sat finished on disk. #1067's tests passed with the bug present because none of their bodies satisfy `isGenuineCCClient`, and non-CC verification shows a false pass.

## How

- One shared implementation above the branch (the reporter's preferred shape — also closes the gap for any future early return). `messages` aliases `clientBody.messages`, so the mutations reach the genuine-CC body, which spreads `clientBody` — the same mechanism the branch's existing `cache_control` strip already relies on.
- The mid-conversation empty-turn drop travels with the filter (removing a block can empty a turn on any path). The final-turn/#1033 behavior is unchanged.
- The **thinking strip stays path-local** — genuine CC forwards its signed thinking verbatim; hoisting it would corrupt the path this fix protects.
- The branch comment's "client is the authority on its own wire shape" now states its one exception: a guaranteed upstream 400 is a defect, not a wire-shape preference.

## Verification

- New regression tests build a body `isGenuineCCClient` actually recognizes (billing header in `system[0]`, CC opener in `system[1]`) with empty/whitespace blocks in history: **4 checks fail against the pre-fix source, 19/19 pass with the fix**.
- Full suite: **141/141**.
- The reporter confirmed on a live proxy that their locally-patched equivalent restores sub-agent notification delivery (#1077).

## Release

In-PR bump to **v5.5.52** + CHANGELOG entry, per the release gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.52] - 2026-08-23
+
+- **Fixed: empty-text-block filter now reaches genuine Claude Code clients** (#1077). #1067's filter sat below the `isGenuineCCClient` early return, so real CC sessions — the one path forwarding messages verbatim — still died with `messages: text content blocks must be non-empty` on every request once an empty block entered the transcript (the #1066 session-killer; presented as "sub-agents hang and never report"). The filter and the mid-conversation empty-turn drop are hoisted above the branch so every path, present and future, shares one implementation; the thinking strip stays path-local (genuine CC forwards signed thinking verbatim). Regression-tested with a genuine-CC body. Reported with a full root-cause analysis by @LiveNathan.
+
 ## [5.5.51] - 2026-08-23
 
 - **CC drift patch** — `SUPPORTED_CC_RANGE.maxTested` bumped `2.1.240` → `2.1.241` for CC v2.1.241. Auto-drafted by `cc-drift-watch.yml`. Template re-capture, if needed, is auto-handled by `cc-drift-template-watch.yml`.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.49",
+  "version": "5.5.52",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "5.5.49",
+      "version": "5.5.52",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.51",
+  "version": "5.5.52",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped fix with regression coverage; approving.

## What I checked
- `gh pr diff` for `src/cc-template.ts`, `test/empty-text-blocks.mjs`, `CHANGELOG.md`, `package.json`/`package-lock.json`.
- Read surrounding source in a shallow clone to confirm two specific claims the PR makes:
  1. That the hoisted filter (`src/cc-template.ts:1708-1725`, the two loops iterating `messages`) actually reaches the genuine-CC branch. `messages` is assigned directly from `clientBody.messages` (same array reference), and the genuine-CC return builds `body = { ...clientBody, system }` (`src/cc-template.ts:1771-1779`) — since `messages`/`clientBody.messages` is the same array object, both the `.filter()` reassignment of `msg.content` and the `messages.splice(...)` mid-conversation drop are visible through that spread. Confirmed correct.
  2. That the thinking-strip stays path-local. `isGenuineCCClient` returns at line 1779, well before the "Strip thinking from history" loop at line 1813 — so genuine CC requests never hit it, matching the stated intent that CC's own signed thinking blocks aren't stripped.
- The mid-conversation-only drop loop starts at `messages.length - 2` (excludes the final turn index), matching the documented dario#1033 constraint; walked the new test's second fixture by hand (5 messages → empty user turn at index 2 removed → 4 messages) and it matches the asserted `body.messages.length === 4`.
- CI rollup (`gh pr checks 1078`): all checks pass (build, test, compat, CodeQL, wire-drift, etc.), nothing pending or failing.

## What's good
- The fix is a pure reordering/hoisting of existing, already-tested logic rather than new logic — low risk.
- New tests specifically target the previously-blind spot (`isGenuineCCClient`-recognized bodies with empty/whitespace text blocks), including the mid-conversation-empty-turn-drop interaction, which is exactly the gap #1067 missed.
- Changelog/version bump follow the repo's release-gate convention.

No correctness, reuse, or efficiency concerns found in the changed lines.

askalf:
Live-tested against the real upstream (both builds run locally in genuine-CC passthrough, upstream bodies captured at the `fetchImpl` seam):

| shape in genuine-CC history | v5.5.51 (pre-fix) | this PR |
|---|---|---|
| `''` beside real text (assistant + user turns) | 200 | 200 |
| whitespace-only block | 200 | 200 |
| `{type:'text', text:null}` | **400** `Input should be a valid string` | **200** |
| `{type:'text'}` (no text field) | **400** `Field required` | **200** |

Wire capture: pre-fix forwarded 2 malformed text blocks to api.anthropic.com; this PR forwarded 0, with the surrounding real blocks intact and all normal shapes answering identically.

Two findings worth recording:

1. The exact-`''` and whitespace shapes are neutralized on ALL paths in the live pipeline before `cc-template` ever runs — `sanitizeMessages` (proxy.ts, dario#54 lineage) trims every string text block and drops the ones that end up `''`. That is why @LiveNathan could not reproduce with hand-made minimal bodies: the residual live gap on the genuine-CC path is the **non-string-text** half of the filter's predicate (`text: null` / missing), which the scrub's `typeof === 'string'` guard skips and only the (previously unreachable) rebuild filter catches. A block that upstream rejects gets through pre-fix and is filtered by this PR — same class, same fix, confirmed live.
2. The compat run on this PR (live real CC through the patched build, self-hosted) is green, so normal genuine-CC traffic is unaffected.

PR #1079, opened 2026-08-23
fix: genuine-CC path inherits the trailing-empty-turn drop (v5.5.53)

askalf:
Stage 3 of #1077, which @LiveNathan called before v5.5.52 shipped (their follow-up crossed with the merge by 13 minutes): stage-1 filtering can empty a **trailing assistant** turn, and `content: []` at the end reads as a prefill upstream — the failure moves rather than disappears. Verified on the shipped v5.5.52 build: a trailing assistant turn holding a lone `text: null` block leaves `assistant[0]` on the wire.

## How

A copy of the trailing-empty-turn drop now runs above the `isGenuineCCClient` branch, next to stages 1–2. Constraints preserved exactly: **assistant turns only** — an empty final user turn is still forwarded (dario#1033's honest-error decision, covered by a new test), and a non-empty trailing assistant is never popped (the OpenClaw runaway, #37). The general-path original stays below, where it must run after the thinking strip; the loop is idempotent so the general path running both copies changes nothing.

## Verification

- @LiveNathan's three-stage repro, verbatim, asserting the exact expected shape `user[1] assistant[1] assistant[1] user[1]` — green.
- Trailing null-text assistant: fails 1/1 against v5.5.52's source, passes with this PR.
- #1033 final-user guarantee asserted on the genuine-CC path.
- Full suite 141/141.

## Release

In-PR bump to **v5.5.53** + CHANGELOG.

Addresses #1077

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.53] - 2026-08-23
+
+- **Fixed: genuine-CC path also inherits the trailing-empty-turn drop** (#1077 follow-up, stage 3 of 3). v5.5.52 hoisted the empty-block filter and the mid-conversation empty-user-turn drop above the `isGenuineCCClient` early return, but a trailing assistant turn the filter emptied (e.g. a lone `text: null` block) still left `content: []` at the end — which upstream reads as a prefill and refuses. A copy of the trailing drop (assistant turns only, per dario#1033/#37) now runs above the branch; the general-path original stays where it is, after the thinking strip it was written for. Predicted empirically by @LiveNathan ("the failure moves rather than disappears") before v5.5.52 shipped; their three-stage repro is now a verbatim regression test.
+
 ## [5.5.52] - 2026-08-23
 
 - **Fixed: empty-text-block filter now reaches genuine Claude Code clients** (#1077). #1067's filter sat below the `isGenuineCCClient` early return, so real CC sessions — the one path forwarding messages verbatim — still died with `messages: text content blocks must be non-empty` on every request once an empty block entered the transcript (the #1066 session-killer; presented as "sub-agents hang and never report"). The filter and the mid-conversation empty-turn drop are hoisted above the branch so every path, present and future, shares one implementation; the thinking strip stays path-local (genuine CC forwards signed thinking verbatim). Regression-tested with a genuine-CC body. Reported with a full root-cause analysis by @LiveNathan.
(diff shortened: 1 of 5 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-scoped fix.

## What I checked
- `gh pr diff 1079` (the changed lines in `src/cc-template.ts`, the new test cases in `test/empty-text-blocks.mjs`, and the version/changelog bumps).
- Shallow-cloned the branch to read the surrounding context in `src/cc-template.ts` around the new block (lines ~1725-1745) and the pre-existing "general path" trailing-assistant-drop loop it duplicates (lines ~1871-1880).
- CI check-rollup: all 16 checks green (`test`, `build (18/20/22)`, `compat`, `wire-drift`, `CodeQL`, `docker-cap-drop-smoke`, etc.) at head `db2a43d`.

## Analysis
The new loop at `src/cc-template.ts:1736-1744`:
```ts
while (messages.length > 0) {
  const last = messages[messages.length - 1];
  if (last.role === 'assistant' && Array.isArray(last.content) && (last.content as unknown[]).length === 0) {
    messages.pop();
    continue;
  }
  break;
}
```
is placed above the `isGenuineCCClient(clientBody)` branch (line ~1748), so the genuine-CC passthrough path now inherits the trailing-empty-assistant-turn drop that previously only ran in the general path (line ~1871-1880, the OpenClaw-runaway guard from #37). The condition is byte-identical in intent to the general-path original — `role === 'assistant'` only, never touching a trailing empty *user* turn — so the dario#1033 guarantee (final empty user turn forwarded, not popped) and the #37 guarantee (non-empty trailing assistant never popped) both hold on the new copy. Because the check is idempotent, the general path running both copies is a correct no-op, as the PR description claims and as I verified by comparing the two loop bodies directly.

## Test coverage
`test/empty-text-blocks.mjs` adds three cases exercising exactly the new code path: the reporter's full three-stage repro asserting the exact resulting shape, a direct trailing-null-text-assistant-turn case asserting the turn is dropped and `genuineCC === true`, and a case asserting the dario#1033 final-empty-user-turn guarantee still holds on the genuine-CC path. This maps 1:1 to the new logic's edge cases (assistant dropped, user kept).

## What's good
- The duplicated loop is small, correctly scoped, and the comment explains *why* it's a duplicate rather than a refactor into a shared helper (the general path's copy runs after a thinking-strip step this path doesn't have) — appropriate given the stated constraint.
- Version bump + CHANGELOG entry included in the same PR.
- All CI checks green, including the self-hosted wire-drift and compat suites.

No correctness, security, or coverage concerns in the diff.

Issue #1062, opened 2026-08-21
preservedToolReason misses 14 of the 24 tools a headless capture drops (#1035)


chaogebaba:
`mergePreservedTools` restores 10 of the 24 tools a headless capture drops, and loses the other 14.

#1035 makes `loadTemplate` apply the superset rule to a live capture, which was the right call — the bug it fixes is real and CI could never see it. But the rule it applies is `preservedToolReason`, and that is a hand-maintained allowlist. Measured against a headless capture on CC 2.1.236, the allowlist is missing more than half of what a headless capture drops.

### Measurement

Bundle: 34 tools. A `claude --print` capture on 2.1.236: 12 tools. So 24 are absent, and `preservedToolReason(name, 'linux')` classifies 10 of them:

```
preserved (10): AskUserQuestion, EnterPlanMode, ExitPlanMode, Glob, Grep,
                PowerShell, TaskCreate, TaskGet, TaskList, TaskUpdate

dropped   (14): CronCreate, CronDelete, CronList, DesignSync, EnterWorktree,
                ExitWorktree, Monitor, NotebookEdit, PushNotification,
                SendMessage, TaskOutput, TaskStop, WebFetch, WebSearch
```

`TaskOutput` and `TaskStop` are the tell. `CONFIG_SCOPED_TOOLS` registers the other four Task tools and deliberately omits those two, on the reasoning stated in its own docstring: "while TaskOutput/TaskStop, the rest of the task subsystem, stayed put." That was true of the capture the set was written against. It is not true of 2.1.236 — a headless capture there carries neither. The list rotted against the thing it tracks, in the specific place the docstring says it would not.

### Why this direction is the expensive one

Not my argument, yours — the `CONFIG_SCOPED_TOOLS` docstring already makes it:

> Dropping one is NOT inert either, and is worse: `CC_NATIVE_NAMES_UNION` is derived from this bundle, so a client that does declare `TaskCreate` stops identity-mapping, falls into the unmapped round-robin, and is renamed onto a fallback slot with junk args (the v4.8.93 regression).

That reasoning applies unchanged to the 14 above. A client declaring `WebSearch` against a dario whose live cache came from a headless capture now hits exactly that path. And because #1035 moved the rule onto the `loadTemplate` hot path, this reaches every install with CC present within the 24h cache TTL — the same blast radius #1035 was written to close, pointed the other way.

[...]

PR #1080, opened 2026-08-23
feat: refuse to bake or load through unclassified tool drops (v5.5.54)

askalf:
Addresses #1062 — with measurements first, code second.

## We measured instead of taking the report's numbers

Four real headless captures via dario's own `captureLiveTemplateAsync`:

| run | CC | platform | captured | absent vs bundle | unclassified |
|---|---|---|---|---|---|
| 1 | 2.1.241 | win32 | 28 (+`WaitForMcpServers`) | 7 | **0** |
| 2 | 2.1.241 | win32 | 27 | 7 | **0** |
| 3 | 2.1.241 | linux | 24 | 10 | **0** |
| 4 | **2.1.236** (the report's version, scratch-installed) | linux | 24 | 10 | **0** |

Every capture carried all 14 tools the issue says get dropped. `preservedToolReason` classified 100% of the genuine absences on both versions and both platforms. The report's 12-tool figure does not match what the binary sends.

## What the measurement DID prove

The roster is remote-config-shaped: `WaitForMcpServers` appeared in capture 1 and vanished in capture 2, minutes apart, same binary — on top of the documented 8/11→8/15 config-scoped flip. So the issue's *structural* concern deserves a tripwire, not a merge rewrite on an unverified premise.

## The guard

- **`unclassifiedToolDrops()`** names any bundle tool a capture omits that no preservation set classifies. Win32 subtlety pinned: a win32-only tool missing from a *win32* capture IS a drop.
- **`capture-and-bake` refuses to bake through one** (new exit 4). The dangerous path was drift-watch → auto-rebake → silently shrunken bundle shipped by auto-merge; now retirement is deliberate (`--allow-tool-drops` + evidence, prune the sets in the same change). `--check` mode names the would-be drops.
- **`loadTemplate` warns on the live path** with a pointer to #1062 — every install with CC present becomes a rot detector for the day the roster genuinely shifts.

Tests pin the semantics, including that `mergePreservedTools` and the guard **partition** the absent set — a name in neither is a hole. 142/142.

In-PR bump to **v5.5.54** + CHANGELOG per the release gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.54] - 2026-08-23
+
+- **Guard: a capture that drops unclassified tools can no longer bake or load silently** (#1062). The report's numbers did not reproduce — four real headless captures (win32 + linux, CC 2.1.236 and 2.1.241) each carried every tool the issue names, with `preservedToolReason` classifying 100% of the genuine absences — but the roster is provably remote-config-shaped (`WaitForMcpServers` appeared in one capture and vanished in the next, minutes apart), so the failure mode deserves a tripwire instead of a rewrite. New `unclassifiedToolDrops()` names any bundle tool a capture omits that no preservation set classifies; `capture-and-bake` now refuses to bake through one (exit 4, `--allow-tool-drops` + evidence to deliberately retire), `--check` mode warns, and `loadTemplate`'s live path logs it with a pointer to #1062 — turning every dario install with CC present into a rot detector for the day the roster shifts.
+
 ## [5.5.53] - 2026-08-23
 
 - **Fixed: genuine-CC path also inherits the trailing-empty-turn drop** (#1077 follow-up, stage 3 of 3). v5.5.52 hoisted the empty-block filter and the mid-conversation empty-user-turn drop above the `isGenuineCCClient` early return, but a trailing assistant turn the filter emptied (e.g. a lone `text: null` block) still left `content: []` at the end — which upstream reads as a prefill and refuses. A copy of the trailing drop (assistant turns only, per dario#1033/#37) now runs above the branch; the general-path original stays where it is, after the thinking strip it was written for. Predicted empirically by @LiveNathan ("the failure moves rather than disappears") before v5.5.52 shipped; their three-stage repro is now a verbatim regression test.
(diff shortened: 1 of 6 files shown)


sprayberry-redline (approved):

## What I checked
- `gh pr diff` (238 lines across 6 files: CHANGELOG, package.json/lock version bump, `scripts/capture-and-bake.mjs`, `src/live-fingerprint.ts`, new `test/unclassified-tool-drops.mjs`).
- CI check-rollup: 13/13 checks pass (`build`×3, `test`, `analyze`, `CodeQL`, `docker-cap-drop-smoke`, `compat`, `triage`×2, `actionlint`, `label`, `validate-package-json`); `auto-merge` is `skipping` (expected — gated, not a failure).

## Correctness — traced the new guard end to end
`unclassifiedToolDrops(captureTools, fallbackTools, platform)` at `src/live-fingerprint.ts:355-366` builds `have` from `captureTools` and returns fallback-tool names that are both absent from `have` and unclassified by `preservedToolReason`.

- **capture-and-bake.mjs:265** calls it as `unclassifiedToolDrops(scrubbed.tools, prev.tools || [], process.platform)` — `scrubbed.tools` at that point already has the platform/interactive/config-scoped preservation merges applied (the block sits right after the `preservedConfigScopedTools` merge), so a tool only surfaces here if it's still missing *and* unclassified after all legitimate restorations. That's the right point to gate the bake.
- **live-fingerprint.ts:479** calls it as `unclassifiedToolDrops(live.tools, bundled.tools, process.platform)` — raw capture vs. bundle, independent of the `mergePreservedTools` result computed just above it, so the warning reflects what the live capture actually omitted rather than the post-merge state. Consistent with the argument order in both call sites (`captureTools` first, `fallbackTools` second).
- The `--allow-tool-drops` branch in `capture-and-bake.mjs:256-268` correctly separates `--check` mode (warn only) from bake mode (exit 4 unless the flag is passed), and `prev.tools || []` guards a missing/empty fallback without throwing.

## Test coverage
`test/unclassified-tool-drops.mjs` is solid: it pins the win32-missing-on-win32 subtlety, verifies `mergePreservedTools` and the guard partition the absent set (no tool falls through both), and covers the identical-capture no-op case. Good use of a partition invariant test rather than just spot-checking a few names.

## Minor
- None worth listing — the module-level and inline comments in `live-fingerprint.ts` and `capture-and-bake.mjs` do a good job explaining the *why* (remote-config-shaped roster, v4.8.93 regression class) without restating the code.

Version bump (5.5.53 → 5.5.54) and CHANGELOG entry are present and match the release gate requirement.

PR #1085, opened 2026-08-24
auto-rebake: template drift detected 2026-08-24

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-24T09:27:43+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.56` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — per-model system-prompt variant only

- **system_prompt_variants.fable:** 9072 → 9220 chars (base system prompt, tools, and anthropic_beta unchanged)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.241) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.241, 24 tools, 6137 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6137 → 4751 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.241 captured 2026-08-23T22:53:22.462Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4751 → fable 9220 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4751 → opus-5 8086 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4751 → sonnet-5 13346 chars
[...]

PR #1081, opened 2026-08-23
auto-rebake: template drift detected 2026-08-23

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-23T22:53:47+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.55` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — per-model system-prompt variant only

- **system_prompt_variants.fable:** 9220 → 9072 chars (base system prompt, tools, and anthropic_beta unchanged)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.241) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.241, 24 tools, 6158 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6158 → 4751 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.241 captured 2026-08-19T16:59:03.395Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4751 → fable 9072 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4751 → opus-5 8086 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4751 → sonnet-5 13346 chars
[...]

PR #1088, opened 2026-08-24
test: derive FABLE_MARKER structurally instead of pinning a literal

### LINKED ISSUE #1087: test: replace FABLE_MARKER literal pin with a structural distinctness assertion (rotted twice)
`test/per-model-system-prompt.mjs` pins `FABLE_MARKER` to a literal prose clause from the Fable variant. That pin has now rotted **twice in two days**:

- #1081 (2026-08-23): CC 2.1.241 condensed Fable's `# Communicating with the user` section to one sentence and added `# Delivering work`. Marker repinned to `brief updates while you work help the user follow along`.
- #1085 (2026-08-24): the capture reverted Fable **byte-for-byte** to the pre-condensation 9220-char shape. The #1081 clause existed only in the condensed shape, so it vanished and CI went red again. Marker repinned to a clause verified present in both shapes.

Both captures report `_version: 2.1.241`. Anthropic is evidently A/B-ing or flip-flopping this variant at a fixed version, so a third rot should be assumed likely — and **any** literal is vulnerable to it, including the current one, which only survives the two shapes observed so far.

### Proposal

Replace the literal pin with a structural assertion that does not name any specific text:

> the Fable variant contains at least one `# ` section heading that appears in **none** of base / opus-5 / sonnet-5

That is the actual invariant the test cares about (Fable is distinct and gets routed correctly), it holds under both observed shapes, and it survives arbitrary editorial drift as long as Fable remains genuinely distinct. If Fable ever stops being distinct, the test *should* fail — that is a real regression, not rot.

`FABLE_IDENTITY` (`This iteration of Claude is Claude Fable 5`) can stay a literal; it is an identity string, not editorial prose, and has been stable across every capture.

Context: ticket `00MT71TKK68159718E534A0368`, analysis in https://github.com/askalf/dario/pull/1085#issuecomment-5393484094

askalf:
Fixes #1087.

The literal FABLE_MARKER pin rotted twice in two days (#1081 condensed shape, #1085 byte-for-byte reversion — same `_version` 2.1.241 both times: a remote-config A/B, not a version bump). Any prose literal is one flip from rot.

This replaces the pin with the invariant the suite actually cares about: **the Fable variant carries at least one top-level `# ` heading that appears in none of base / opus-5 / sonnet-5**, with the request-level checks using a heading derived from that set. The suite now survives editorial flip-flops and fails only if Fable genuinely stops being distinct — a real regression, not rot. `FABLE_IDENTITY` stays literal (identity string, byte-stable across every observed shape).

- 142/142 locally against the v5.5.56 template; the new structural check reports the full Fable heading list on failure for fast re-diagnosis.
- Bumps 5.5.56 → 5.5.57 (release gate) + CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.57] - 2026-08-24
+
+- **Test: FABLE_MARKER is now derived structurally, not pinned** (#1087). The literal prose pin rotted twice in two days as Anthropic A/B-served two Fable shapes at the same CC version. `per-model-system-prompt.mjs` now asserts the real invariant - the Fable variant carries at least one top-level `# ` heading absent from base/opus-5/sonnet-5 - and derives its marker from that set, so the suite survives editorial flip-flops and fails only if Fable genuinely stops being distinct. `FABLE_IDENTITY` stays literal.
+
 ## [5.5.56] - 2026-08-24
 
 - **Template rebake** — re-captured `src/cc-template-data.json` after cc-drift-template-watch detected wire-fingerprint drift against a live CC capture. Bundled fallback template now matches the current CC wire shape.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.56",
+  "version": "5.5.57",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/test/per-model-system-prompt.mjs
+++ b/test/per-model-system-prompt.mjs
@@ -13,21 +13,31 @@ function check(name, cond, detail) {
 }
 function header(n) { console.log(`\n=== ${n} ===`); }
 
-// FABLE_MARKER must be true of the Fable variant and FALSE of every other body
-// (base, opus-5, sonnet-5) — the negative assertions below depend on it.
+// The Fable marker is DERIVED, not pinned (#1087). A literal pin rotted twice
+// in two days: CC 2.1.241 first condensed Fable's '# Communicating with the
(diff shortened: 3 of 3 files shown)


sprayberry-redline (approved):

**Verdict:** Clean — no blocking issues found.

## What I checked
- `gh pr diff 1088` (86 lines across `CHANGELOG.md`, `package.json`, `test/per-model-system-prompt.mjs`)
- CI check-rollup via `gh pr checks 1088`: all checks passing (CodeQL, actionlint, analyze, build ×3, docker-cap-drop-smoke, test, triage ×2, validate-package-json); only `compat` still pending, non-blocking.

## Findings
None blocking. The change replaces a brittle literal-prose `FABLE_MARKER` pin (which had rotted twice in two days per the linked #1087) with a structurally-derived marker:

```js
const headings = (body) => new Set(body.match(/^# .+$/gm) ?? []);
const fableOnlyHeadings = [...headings(CC_SYSTEM_PROMPT_FABLE)].filter((h) =>
  !headings(CC_SYSTEM_PROMPT).has(h) &&
  !headings(CC_SYSTEM_PROMPT_OPUS5).has(h) &&
  !headings(CC_SYSTEM_PROMPT_SONNET5).has(h));
const FABLE_MARKER = fableOnlyHeadings[0];
```

I checked the edge case where `fableOnlyHeadings` is empty (`FABLE_MARKER` becomes `undefined`): the new explicit check `variant carries at least one Fable-only # heading ... fableOnlyHeadings.length >= 1` (test/per-model-system-prompt.mjs) fails first and prints the full heading list for diagnosis, so there's no silent false-pass through the later `.includes(FABLE_MARKER)` checks. `FABLE_IDENTITY` is correctly left as a literal since it's called out as byte-stable identity text, not editorial prose.

`package.json` version bump (5.5.56 → 5.5.57) and `CHANGELOG.md` entry are consistent with the PR description and each other.

## What's good
Well-scoped fix that directly addresses a real, previously-observed flakiness source (two pin rotations in two days per the PR body) by targeting the actual invariant instead of re-pinning prose. Good diagnostic output on failure (prints derived heading list).

PR #1093, opened 2026-08-24
fix: rewind an empty final user turn on the genuine-CC path (#1092)

### LINKED ISSUE #1092: genuine-CC: empty final user turn from stream-interruption retry 400s ("user messages must have non-empty content")
## What happened

A Claude Code session running a background subagent: the subagent finishes, CC streams the result back to the parent model, the stream drops ("Streaming response ended before any complete data was received"), CC retries without streaming, and the retry 400s:

```
API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.42: user messages must have non-empty content"},"request_id":"req_011CeNF2q7rBQckv2fkHVAx6"}
```

Every subsequent request in the session then fails the same way — the empty turn is stuck in history. (Fallback to `gemini-flash` also failed, with a separate "Invalid thought signature" error, but the primary 400 is the issue here.)

**Version**: 5.5.57 (npm latest).

## Why the existing filters miss it

`buildCCRequest` drops empty turns in three places, none of which cover this:

- empty `text` blocks → filtered, all paths (#1078 → #1077)
- empty **user** turn, *mid-conversation* → `for (let i = messages.length - 2; i >= 0; i--)` drops it (#1077)
- empty **assistant** turn, *trailing* → popped (#1079)

The surviving case is an **empty final user turn**: the `length - 2` loop deliberately stops short of the last message, and the trailing drop only pops `assistant`. So `messages[N]` with `role: "user", content: []` is forwarded verbatim, and Anthropic rejects it — the rule is "all messages must have non-empty content except for the optional final *assistant* message."

## The #1033 tension

The `cc-template.js` comment on the trailing drop explains why empty user turns are left in place: popping one would expose the assistant turn behind it and turn an honest "empty content" 400 into a misleading "prefill" rejection (#1033). That's right for a genuinely malformed third-party client — but here the empty user turn is produced by *Claude Code's own stream-interruption retry path*, not a bad client, so "let upstream name it accurately" is exactly the #1066 session-killer re-entering through the one turn the filter won't touch.

## One thing I couldn't capture

I'm inferring "final turn" from the code's `length - 2` skip — a mid-conversation empty user turn would already be dropped, so the final turn is the only surviving path. I don't have the raw request body at the moment of failure. If useful I can capture it with dario's debug/log output.

## Suggested direction

Leaving the fix shape to you, since it isn't a safe one-liner: dropping the empty final user turn exposes the preceding assistant turn (→ prefill rejection), so it'd need either (a) drop it *and* the preceding assistant turn, (b) substitute a minimal non-empty placeholder, or (c) recognize the CC retry artifact earlier and drop it before it enters history. Not something I want to guess at in a local patch.

askalf:
Fixes #1092.

## The bug

CC's stream-interruption retry can append a user turn with `content: []` as the final message. The three earlier empty-turn drops all deliberately skip the final turn:

- empty `text` blocks → filtered, all paths (#1078)
- empty **user** turn, *mid-conversation* → dropped (#1077)
- empty **assistant** turn, *trailing* → popped (#1079)

So an empty **final user** turn reaches the wire and 400s (`messages.N: user messages must have non-empty content`), then sticks in history and kills every later request — the #1066 session-killer re-entering through the one turn the filters won't touch. Reproduced deterministically (no live API needed): an empty final user turn survives `buildCCRequest` unchanged.

## The fix, and the #1033 tension it resolves

`#1033` deliberately leaves an empty final user turn in place so a **malformed third-party client** gets an honest upstream error instead of a mangled conversation. That reasoning is right for the general path and is **kept**. But on a **genuine CC** session the empty turn is CC's *own* retry artifact, and "let upstream name it" is exactly the session-killer — the honest 400 is still a 400, and it repeats forever.

The retry's intent is to regenerate the response whose stream was cut, so on the genuine-CC path the request is **rewound to it**: drop the empty user turn and the assistant turn behind it, landing back on the last real user turn for a clean regeneration.

- **Scoped to genuine CC** (billing header in `system[0]`, CC opener in `system[1]`). The general path keeps #1033 verbatim — asserted by a test.
- **Distinct from the #37 runaway loop**: #37 popped a *non-empty* trailing assistant with no empty user after it; this fires *only* on a trailing empty user turn and pops at most the one pair. Not a single-prompt runaway — each rewind corresponds to one real streaming failure and stops when streaming recovers.
- Also extends the mid-conversation empty-turn drop to **assistant** turns emptied by a null/non-string text block (the flavor the proxy-level string scrub, which is string-only, leaves behind).

## Tests

`test/empty-text-blocks.mjs` gains the #1092 rewind (both `content:[]` and emptied-block shapes), the general-path #1033 preservation, and the mid-assistant drop. 142/142 suite green. Verified by a deterministic 16-shape empty-content matrix through `sanitizeMessages` + `buildCCRequest`.

## Known gap, deliberately not guessed here

The same matrix shows an empty/null **text block nested inside a `tool_result` content array** still survives on both paths (requires a tool wrapper to emit a malformed text block — not the reported shape, and rare). Fixing it safely needs to know whether upstream accepts an emptied `tool_result` (`content: []` / `""`), which I can't confirm without a live capture. Flagging rather than shipping a transform that might itself 400. A captured body of that shape would let it be fixed correctly.

## Note on #1077

Fully resolved as shipped (5.5.52/53) — the reporter's three-stage repro passes verbatim in the matrix. Left open for the reporter to close per the earlier thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.59] - 2026-08-24
+
+- **Fixed: an empty final user turn no longer session-kills a genuine CC client** (#1092). CC's stream-interruption retry can append a `content: []` user turn as the final message; the three earlier empty-turn drops (#1077/#1078/#1079) all skip the final turn, so it reached the wire and 400'd (`messages.N: user messages must have non-empty content`), then stuck in history and killed every later request. On the genuine-CC path this is CC's own retry artifact, so the request is rewound to the interrupted response - the empty user turn and the assistant turn behind it are dropped, landing back on the last real user turn for a clean regeneration. The general path is unchanged: a third-party client's empty final user turn is still forwarded so upstream names it honestly (#1033). The rewind is bounded and distinct from the #37 runaway loop (fires only on a trailing empty user turn, pops at most one pair). The mid-conversation empty-turn drop now also covers assistant turns emptied by a null/non-string text block (the flavor the proxy-level string scrub leaves behind).
+
 ## [5.5.58] - 2026-08-24
 
 - **Template rebake** — re-captured `src/cc-template-data.json` after cc-drift-template-watch detected wire-fingerprint drift against a live CC capture. Bundled fallback template now matches the current CC wire shape.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.58",
+  "version": "5.5.59",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/cc-template.ts
+++ b/src/cc-template.ts
@@ -1711,17 +1711,21 @@ export function buildCCRequest(
       (b) => !(b.type === 'text' && (typeof b.text !== 'string' || (b.text as string).trim() === '')),
     );
   }
-  // A USER turn left with no blocks is dropped — but only MID-conversation,
-  // where the API combines the now-adjacent same-role turns and where the
-  // #1066 session-killer lives (the empty turn sits in history, so every
-  // later request carries it). The FINAL turn is deliberately left in place
-  // per the dario#1033 decision below: popping it would expose the assistant
-  // turn behind it and convert an honest "content must contain at least one
-  // block" into a misleading prefill rejection. Hoisted with the filter
-  // (dario#1077): removing a block can empty a turn on any path.
+  // A turn left with no blocks is dropped — but only MID-conversation, where
+  // the API combines the now-adjacent same-role turns and where the #1066
+  // session-killer lives (the empty turn sits in history, so every later
+  // request carries it). BOTH roles: a user turn whose only block was empty,
+  // and an assistant turn the filter emptied (a null/non-string text block the
+  // proxy-level string scrub leaves behind, dario#1092) — either one mid-stream
+  // is "messages.N: … content must contain at least one block" upstream. The
+  // FINAL turn is deliberately left in place per the dario#1033 decision below:
+  // popping it would expose the assistant turn behind it and convert an honest
+  // "content must contain at least one block" into a misleading prefill
+  // rejection. Hoisted with the filter (dario#1077): removing a block can empty
+  // a turn on any path.
   for (let i = messages.length - 2; i >= 0; i--) {
     const m = messages[i];
-    if (m.role === 'user' && Array.isArray(m.content) && (m.content as unknown[]).length === 0) {
+    if (Array.isArray(m.content) && (m.content as unknown[]).length === 0) {
       messages.splice(i, 1);
     }
   }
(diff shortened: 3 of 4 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean fix, well-tested, approving.

## What I checked
- `gh pr diff 1093` (full diff: `CHANGELOG.md`, `package.json`, `src/cc-template.ts`, `test/empty-text-blocks.mjs`)
- CI check-rollup: 14/14 green (`test`, `compat`, `analyze`, `CodeQL`, `wire-drift`, `docker-cap-drop-smoke`, builds on Node 18/20/22, etc.)
- Surrounding context in `src/cc-template.ts` around both changed loops (mid-conversation empty-turn drop, and the new genuine-CC final-turn rewind)

## Correctness review

**Mid-conversation drop loop** (`src/cc-template.ts`, widened from `m.role === 'user' && ...` to any role):
```
for (let i = messages.length - 2; i >= 0; i--) {
  const m = messages[i];
  if (Array.isArray(m.content) && (m.content as unknown[]).length === 0) {
    messages.splice(i, 1);
  }
}
```
Iterates high→low, so `splice(i, 1)` never disturbs indices not yet visited — safe. Widening from user-only to both roles matches the stated goal (an assistant turn emptied by a null/non-string text block, per #1092) and stays excluded from the final index, preserving the #1033 invariant.

**New genuine-CC final-turn rewind:**
```
const last = messages.length > 0 ? messages[messages.length - 1] : undefined;
if (last && last.role === 'user' && Array.isArray(last.content) && last.content.length === 0) {
  messages.pop();
  const prev = messages[messages.length - 1];
  if (prev && prev.role === 'assistant') messages.pop();
}
```
Traced the case where the assistant turn behind the empty final user turn was *itself* already emptied and removed by the mid-conversation loop above (since that assistant turn sits at `length-2`, not the excluded final index): after that removal, `last` is still the trailing empty user turn, and `prev` becomes the last real user turn (role `user`, not `assistant`), so the second `pop()` correctly does not fire — the request lands on the real user turn either way. No double-pop or wrong-turn scenario found.

Scoping to `isGenuineCCClient(clientBody)` is correct — the general path (#1033) is untouched, and the added test explicitly asserts the general path still forwards `content: []` verbatim.

## Tests
New tests in `test/empty-text-blocks.mjs` cover: the `content: []` rewind shape, the emptied-block (`[{text:''}]`) rewind shape, general-path #1033 preservation, and the new mid-conversation assistant-turn drop — matching the changed code paths. CI `test` job is green.

## What's good
Clear PR description tying the fix to the exact prior decisions it interacts with (#1033, #1077-1079, #37), and the code comments explain the *why* (bounded vs. the #37 runaway loop) rather than restating the diff.

sprayberry-redline:
## dario/live-test — PASS

Head SHA: `e7ea38d44a199a9a103838873f1cf847d7e9cacc`

| Stage | Result | Note |
|---|---|---|
| build (`npm run build`) | ✅ pass | `tsc` clean |
| test (`npm test`) | ✅ pass | 142/142 passed, 0 fail |
| doctor | ✅ pass | reports `[FAIL] OAuth not authenticated` as expected with no creds — did not crash, exit 1 is correct doctor behavior in a clean env |
| live probe | ⏭️ skipped | `DARIO_TEST_UPSTREAM_API_KEY` not set in tester env |

No stages attributable to this PR failed.

PR #1094, opened 2026-08-24
auto-rebake: template drift detected 2026-08-24

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-24T23:57:29+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.59` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — per-model system-prompt variant only

- **system_prompt_variants.opus-5:** 8086 → 8922 chars (base system prompt, tools, and anthropic_beta unchanged)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.241) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.241, 24 tools, 6135 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6135 → 4751 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.241 captured 2026-08-24T14:21:24.559Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4751 → fable 9072 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4751 → opus-5 8922 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4751 → sonnet-5 13346 chars
[...]

PR #1089, opened 2026-08-24
auto-rebake: template drift detected 2026-08-24

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-24T14:21:49+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v5.5.58` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — per-model system-prompt variant only

- **system_prompt_variants.fable:** 9220 → 9072 chars (base system prompt, tools, and anthropic_beta unchanged)

### Drift report

```
[bake] using CC at /usr/bin/claude (version 2.1.241) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.241, 24 tools, 6163 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6163 → 4751 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.241 captured 2026-08-24T09:27:18.099Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4751 → fable 9072 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4751 → opus-5 8086 chars
[bake] spawning CC (sonnet-5-variant: claude-sonnet-5) against loopback MITM to capture /v1/messages...
[bake] sonnet-5 system-prompt variant: base 4751 → sonnet-5 13346 chars
[...]

PR #1097, opened 2026-08-25
fix: remember A/B prompt arms — end the template-drift ping-pong (v5.5.61)

askalf:
**The "stuck in weird template drift" is a measured ping-pong.** Four auto-rebakes in ~25h (#1081 → #1085 → #1089 → #1094), with the `fable` variant flip-flopping between the **same two byte-stable shapes** (sha `8c7736e0…`/9072 chars ↔ `40c52c67…`/9220 chars), and `opus-5` now showing a second arm of its own (8086/8922). Anthropic **A/B-serves prompt arms per request** at the same CC version. The drift check compared each capture strictly (`!==`) against the single baked arm — so whichever arm the dice rolled read as drift, got rebaked onto, and armed the next flip. Forever. (The 5.5.57 `FABLE_MARKER` change taught the *test suite* this exact lesson; the watch itself never learned it.)

## The fix: a shape memory

- **`_variantShapeHashes`** on the template: every distinct shape ever observed per family (full sha256), **seeded** with the arms measured across the 08-19→08-24 bakes.
- **`classifyVariantShape()`** → `canonical` | `known-alt` | `new`. A re-served known arm is **not drift**.
- **Sticky canonical**: the bake keeps its baked arm when the capture is a known alt — the bundle moves only when CC serves a **never-seen** shape, which then grows the memory. Check mode logs known arms as not-drift; only genuinely new shapes open rebake PRs.
- **Retiring an arm is a deliberate manual edit** — same contract as `CONFIG_SCOPED_TOOLS`.

## Expected behavior after this lands
The hourly watch stops oscillating: captures of either fable arm (or either opus-5 arm) read clean; the next rebake happens only when CC actually ships something new. #1095 then closes on the first clean run.

Tests pin the classifier, the hash identity, the seeded bundle's invariants (every canonical present in its own memory; both measured fable arms remembered), and that the bake wires the classifier rather than a strict `!==`. **143/143.** In-PR bump to **v5.5.61** per the release gate.

Addresses #1095


sprayberry-redline (approved):

## What I checked
- `gh pr diff` (307 lines) across `capture-and-bake.mjs`, `live-fingerprint.ts`, `cc-template-data.json`, and the new `test/variant-shape-memory.mjs`.
- CI rollup (`gh pr checks 1097`): `build (18/20/22)`, `test`, `analyze`, `CodeQL`, `docker-cap-drop-smoke`, `actionlint`, `triage`, `label`, `validate-package-json` all `pass`. `compat` and `wire-drift` were still `pending` at review time — not blocking, but worth a glance once they land.

## Correctness
- `classifyVariantShape()` (`src/live-fingerprint.ts`) correctly prioritizes byte-equality to the current canonical before consulting the known-hash memory, and the new test file pins all three verdicts (`canonical`/`known-alt`/`new`) plus the no-canonical and no-memory edge cases — matches the described fix for the #1095 ping-pong (comparing a fresh capture strictly against a single baked arm).
- The shape-memory union in `capture-and-bake.mjs` (`const memory = {}` block) folds in the previous memory, the previous canonical, the freshly kept/baked variant, and the just-captured arm hash, so a `known-alt` capture still gets recorded even though it doesn't replace the canonical — this is what lets the set grow instead of only tracking the two arms currently in play.
- The five hashes seeded into `src/cc-template-data.json`'s `_variantShapeHashes` are all well-formed 64-char lowercase hex (verified locally), consistent with `variantShapeHash`'s `sha256().digest('hex')`.
- `--check` mode's `knownAltDiffs` filter reuses the same classifier against `prevVariants`/`newVariants`, so check and bake paths can't disagree on what counts as drift.

## What's good
- The new test file (`test/variant-shape-memory.mjs`) is thorough: it pins the hash identity, all three classifier verdicts including edge cases (no canonical, no memory), the shipped bundle's invariants (every canonical present in its own memory, sorted/dedup'd), and that the bake script actually wires the classifier rather than a strict `!==`. Directly targets the regression described in the PR body (four rebakes in 25h from A/B arm flip-flopping).
- Comments explain the non-obvious *why* (Anthropic's per-request A/B serving) rather than restating the code.
- Version bump (`package.json`/`package-lock.json` → 5.5.61) and CHANGELOG entry are consistent and match the release-gate convention referenced in the PR body.

No efficiency or reuse concerns — this is a small, well-scoped addition with matching test coverage.

sprayberry-redline:
## dario/live-test — PASS

Head SHA: `ef10a2cdee6cc71e5387bec3660a91cf46bcdc9d`

| stage | result | note |
|---|---|---|
| install | ✅ pass | `npm ci` clean |
| build | ✅ pass | `tsc` + asset copy clean (v5.5.61) |
| test | ✅ pass | 143/143 tests pass, 0 fail |
| cli | ✅ pass | `--help` exits 0 |
| doctor | ✅ pass | exits 1 as expected (no creds in sandbox) — reports `[FAIL] OAuth not authenticated`, `run dario login`, no crash |

Status: `dario/live-test` = success

PR #1106, opened 2026-08-25
fix(ci): compat must not touch the shared Claude credential (--no-claude-auth)

askalf:
**Re-running the compat job took the fleet's LLM down for ~3h today** (#1103, #1105). This makes that structurally impossible.

## What happened
The compat job forwards upstream with `ANTHROPIC_UPSTREAM_API_KEY` and never needs the Claude OAuth pool — but **dario loads that pool and arms its background refresh timer at startup regardless of routing target**. The credential it loads is `/root/.claude/.credentials.json`, *deliberately shared* with the platform `askalf-dario` container (per the existing header comment). Anthropic invalidates a refresh token the moment it's used, so this proxy's background refresh **rotated the token out from under the platform container**, which then looped `invalid_grant` and stopped serving until the credential was hand-recovered.

## The fix
`--no-claude-auth` (`proxy.ts:1648` — *"never touch the Claude token in OpenAI-only mode"*) is the flag built for exactly this case. Added to the proxy launch, with the reasoning inline.

The header comment already asserted *"the runner just READS the current token — no refresh attempted from this side, no rotation race."* That was true of the read path and **false of the background timer**; the flag is what makes the claim true. Sharing the credential file is only safe because this side now never authenticates with it.

Same failure class as the 2026-08-22 scratch-proxy incident — except that one was a hand-run proxy, and this one is **in CI**, so it would recur on every compat re-run.


sprayberry-redline:
## dario/live-test report — PASS

**Head SHA:** `b3810d9d3efb0b0eda31594c73462c5c494105ca`

| Stage | Result | Note |
|---|---|---|
| build (`npm run build`) | ✅ pass | tsc + asset copy, no errors |
| test (`npm test`) | ✅ pass | 143/143 tests passed, 0 fail, 47.1s |
| doctor | ✅ pass | ran and reported cleanly; expected `[FAIL] OAuth not authenticated` in a credential-less sandbox (not a crash) |

No stages attributable to this PR failed.

sprayberry-redline (changes requested):

**Verdict:** Request changes — the flag this PR adds breaks the compat job's own CI run (confirmed on this PR's head, run 32860260517).

## Findings

### 1. [Blocking] `--no-claude-auth` makes the compat job's readiness check fail every time
`.github/workflows/compat-test-self-hosted.yml` (proxy launch line, diff hunk 2):
```
node dist/cli.js proxy --passthrough --strict-template --no-claude-auth --port=3457 > proxy.log 2>&1 &
```
CI run for this exact PR head (32860260517, job `compat`) shows the proxy starting and printing its
full banner (`Claude auth: disabled (--no-claude-auth) ... Pool: empty ...`), then the workflow's own
readiness loop failing:
```
##[error]dario proxy never reached /health within 30s
```
Root cause, traced in `src/proxy.ts`: when `opts.noClaudeAuth` is true, the pool-population loop is
skipped entirely (`if (opts.noClaudeAuth) { console.error(...) } else { for (const acc of accountsList) pool.add(...) }`,
proxy.ts:~1604-1618) — even though `accountsList`/`ensureLoginCredentialsInPool()` already ran above
unconditionally. `pool.size` stays `0`. `/health`'s `currentStatus()` calls
`derivePoolStatus(pool.all()..., now, adminEnabled)` (health-response.ts:139), which for an empty pool
always returns `status: 'none'`, and `buildHealthResponse` treats `'none'` as `structurallyDead` → HTTP
503 (health-response.ts:214-224). `derivePoolStatus` only special-cases `adminEnabled` for the "empty
pool is fine" message — it has no awareness of api-key/`--no-claude-auth` mode, so it 503s regardless of
`ANTHROPIC_UPSTREAM_API_KEY` being set and working.
The workflow's own wait loop uses `curl -sf`, which treats a 503 as a failed check
(`.github/workflows/compat-test-self-hosted.yml:209`), so it spins for the full 30s and the job fails —
exactly what run 32860260517 shows, on this PR's own head commit.
```suggestion
Before merging, either:
- make /health treat api-key-mode / --no-claude-auth as a legitimately-healthy empty pool (mirror the
  `hasUpstreamApiKey`/`noClaudeAuth` awareness `requiresClaudeLogin` already has, in `derivePoolStatus`
  or its caller), or
- point the workflow's readiness probe at /livez (which just returns 200 unconditionally, per the
  code above /health) instead of /health for this job, since this job explicitly doesn't need OAuth
  readiness.
```
This isn't a flaky/unrelated CI failure — it's the direct, reproducible consequence of the one launch
flag this PR adds, checked via the CI check-rollup per review policy.

## What's good
- The comment rationale for *why* `--no-claude-auth` is load-bearing (not hygiene) is clear and
  correctly identifies the real prior incident (background refresh timer rotating a shared refresh
  token out from under the platform container). The intent of the fix is sound — it just needs to not
  break the job it's applied to.
- Doc-comment updates in the first hunk are consistent with the new behavior.

## What I checked
- `gh pr diff` (both hunks — comment updates + the flag addition).
- CI check-rollup: `compat` = FAILURE on this PR's head; all other checks (build/test/CodeQL/lint) green.
- Pulled the failing run's log directly and traced the failure into `src/proxy.ts` / `src/health-response.ts` via a shallow clone.

sprayberry-redline (dismissed):

**Verdict:** Approve — the compat-job regression from the prior review is fixed and CI confirms it ( `compat` check now passes on this head, 4m3s).

## What changed since my last CHANGES_REQUESTED (head e1c3b77 vs the reviewed head ce5dd28)
The prior head used `--no-claude-auth`, which broke the compat job's own traffic (that flag makes *all* Claude-bound requests return an unauthenticated error, and the compat suite is Claude-bound traffic forwarded via `ANTHROPIC_UPSTREAM_API_KEY`). This head drops `--no-claude-auth` entirely and instead runs the proxy under an isolated, empty `HOME`:

`.github/workflows/compat-test-self-hosted.yml`:
```
ISO_HOME="${RUNNER_TEMP:-/tmp}/dario-compat-home"
rm -rf "$ISO_HOME"
mkdir -p "$ISO_HOME/.dario"
cp "$HOME/.dario/cc-template.live.json" "$ISO_HOME/.dario/" 2>/dev/null || true
HOME="$ISO_HOME" node dist/cli.js proxy --passthrough --strict-template --port=3457 > proxy.log 2>&1 &
```
Since the proxy loads `/root/.claude/.credentials.json` via `homedir()` (`src/proxy.ts:183-185`), pointing `HOME` at an empty dir means it never reads or refreshes the shared credential — the actual root cause of the outage — while `upstreamApiKey` set means `requiresClaudeLogin()` (`src/proxy.ts:1680`, `requiresClaudeLogin` at `:875-885`) still returns `false` with an empty pool, so startup doesn't gate on a missing login and the background refresh interval no-ops on an empty pool. This is a materially better fix than the flag it replaces, and the `compat` CI check — the one that failed under the old approach — is green on this head.

`src/health-response.ts:216-231` / `src/proxy.ts:2137`: the isolation exposed a second, real bug — with OAuth state `'none'`, `/health` returned 503 even though the suite's own traffic was being served fine via `x-api-key`. Fixed by gating `structurallyDead` on the new `upstreamApiKeyMode` flag, while leaving the serving-probe path (`s.probe?.ok === false`) unconditional so a real failed round-trip still degrades in either mode. `test/health-response.mjs` adds matching coverage for both modes (api-key-mode 200s on `none`/`expired`/`broken`, still 503s on a failed probe; OAuth-mode verdicts unchanged).

## What's good
- The workflow comment accurately documents the outage mechanism and why `--no-claude-auth` was the wrong tool, matching the code.
- Test coverage for the new `upstreamApiKeyMode` branch is thorough and includes the "probe still wins" regression case.
- CHANGELOG entry accurately describes both fixes.

No blocking issues found in this diff.

sprayberry-redline:
## dario/live-test report — PASS

**Head SHA:** `73ec036cb7eaf8e13470bd229587ae6a46fb06e1`

| Stage | Result | Note |
|---|---|---|
| install (`npm ci`) | PASS | clean, 6 packages added |
| build (`npm run build`) | PASS | tsc + asset copy, no errors |
| test (`npm test`) | PASS | 143/143 tests passed, 0 failed |
| cli (`--help`) | PASS | exits 0, usage printed |
| doctor | PASS | reports `[FAIL] OAuth not authenticated` as expected with no credentials — did not crash |
| live probe | SKIPPED | `DARIO_TEST_UPSTREAM_API_KEY` not set in tester env |

sprayberry-redline (approved):

## What I checked
- `gh pr diff` for all 7 changed files, plus `src/proxy.ts` and `src/health-response.ts` context in a shallow clone of `fix/compat-no-claude-auth`.
- The PR's own CI check-rollup: all 14 checks green, including `compat` (SUCCESS, run 32863710501, completed 15:10:30Z) on this exact head commit `73ec036c`.
- A prior automated review on this PR (commit `ce5dd286`) requested changes because `--no-claude-auth` made the compat job's `/health` readiness probe 503 forever (empty pool → `structurallyDead`). That review is stale — the PR has since moved to a different fix (isolated `HOME` + `upstreamApiKeyMode` awareness in `buildHealthResponse`) at a new head, and the compat job now passes, so that blocking concern is resolved on the current diff.

## Findings
None blocking. Two observations, neither gating:

- **Correctness — verified, not a bug:** `src/health-response.ts:219-224` guards `structurallyDead` with `s.upstreamApiKeyMode !== true`, so existing OAuth-mode callers that never set the new optional field are unaffected (`undefined !== true` → `true`, old behavior preserved). The new `test/health-response.mjs` block explicitly covers both modes plus the "probe still wins" invariant (`H({ status: 'none', upstreamApiKeyMode: true, probe: { ok: false } }) === 503`), which is the right thing to pin down given dario#905's whole point is that a failed round-trip must stay authoritative. I traced this by hand and it holds.
- **Minor:** `.github/workflows/compat-test-self-hosted.yml` copies `cc-template.live.json` into the isolated `$ISO_HOME` with `|| true`, silently swallowing a copy failure; if the source file is ever missing, `--strict-template` will then fail with a less specific error than "template missing" would give. Not worth blocking on — the fallback is a refusal to boot either way, just with a slightly less direct error message.

## What's good
- The fix is well-targeted: giving the compat proxy its own empty `$HOME` (rather than the withdrawn `--no-claude-auth` flag) means it never loads the shared credential file at all, addressing the root cause instead of the symptom.
- `upstreamApiKeyMode` correctly threaded from `proxy.ts:2137` (`!!upstreamApiKey`) through to the health builder, with the incident context and reasoning captured inline where a future reader will actually see it.
- New tests are specific and cover the regression path (a compat-like caller with OAuth `'none'` but an API key configured) as well as confirming unchanged OAuth-mode behavior.
- CHANGELOG entry accurately describes both the credential-isolation fix and the health-response fix it exposed.

askalf:
Verification for review finding / ticket 01M0WSYADYQAAMV173R84VTGF8 (Developer):

The reported defect — `--no-claude-auth` leaves `pool.size=0`, `derivePoolStatus()` reports `none`, `buildHealthResponse` returns 503, and the workflow readiness probe `curl -sf .../health` therefore times out after 30s (seen on run 32860260517) — no longer applies to this branch head (73ec036). Confirmed against the head tree, not the original diff:

1. `--no-claude-auth` was **dropped** from the proxy launch. `.github/workflows/compat-test-self-hosted.yml:221-228` now isolates the credential with an empty `HOME` (`ISO_HOME`) instead, copying `cc-template.live.json` across so `--strict-template` still boots. The in-file comment at line 218 explicitly records that `--no-claude-auth` is the wrong tool because it makes Claude-bound requests fail.
2. The 503-in-API-key-mode bug is fixed at the right layer: `src/health-response.ts` adds `upstreamApiKeyMode` to `HealthStatusLike`, and `structurallyDead` is now gated on `s.upstreamApiKeyMode !== true`, so an empty/expired/broken OAuth pool is no longer treated as death when an upstream API key is what serves. `src/proxy.ts:2137` populates it with `upstreamApiKeyMode: !!upstreamApiKey`. The serving probe stays authoritative in both modes, so a real failed round-trip still 503s (dario#905 preserved).
3. `test/health-response.mjs` covers both directions: five api-key-mode cases (none/expired-unrefreshable/broken -> 200, failed probe -> 503, passing probe -> 200) plus four OAuth-mode regression cases asserting the old verdicts are unchanged, and `upstreamApiKeyMode: false` behaving like absent.
4. The readiness probe was left pointed at `/health` (the ticket offered `/livez` as the alternative); with fix (2) that is now the stronger of the two options, since it still fails the job on a genuinely dead proxy.

Runtime evidence: compat run 32863710501 passed in 4m03s, log shows `started dario proxy with PID 1040208` with no `dario proxy never reached /health within 30s` error. Full rollup is green (build 18/20/22, test, compat, wire-drift, CodeQL, actionlint, docker-cap-drop-smoke, dario/live-test) and `reviewDecision: APPROVED`.

Also clearing the stale blocker on intervention 00MT8RYCMSE9212AD6BE8F84AA ("Version bump needed before merge"): `package.json` is at 5.5.65 (from 5.5.63) with a matching `## [5.5.65] - 2026-08-25` CHANGELOG entry.

No code change filed by me — this PR already satisfies the requested change. Remaining step is the operator merge.

PR #1107, opened 2026-08-25
fix(doctor): report what SERVES, not the stale legacy credential (v5.5.63)

askalf:
`dario doctor` printed **`OAuth expired` while the proxy was answering 200s** on both Haiku and Sonnet — and `dario-doctor-watch` filed an issue for it on every run. #1105 was opened during exactly that healthy state.

## Cause
The OAuth row read **only** the legacy `credentials.json`. But dario#805 deliberately **keeps a newer pool token and refuses to overwrite that file**, so "legacy is stale" is an *expected steady state* — every recovery that restores a pool account leaves one behind. Meanwhile the separate **Pool** row was correctly reporting `1/1 healthy` two lines below. The doctor contradicted itself, and the watcher believed the wrong line.

## Fix
The row consults the **pool** first:
- live pool → `ok`, and it still **names the stale legacy file and cites #805** — disclosing rather than hiding, so the next reader knows why that file is left alone;
- nothing able to serve → unchanged `warn`/`fail`. Silencing a real outage would be worse than the false alarm it replaces.

Extracted as the pure **`oauthCheckRow()`**, mirroring the existing `checkIdentityDrift()` so the branch logic is unit-testable without touching the filesystem — 13 new assertions covering the #1105 case, pluralisation, both failure modes, and "an empty pool is not a live pool". **144/144.**

Verified against the box's real state (pool live + legacy expired): previously `WARN OAuth expired`, now `ok — pool credential live (1/1 account), the legacy credentials.json is expired and unused (dario#805)`.

PR #1108, opened 2026-08-25
fix(release): resolve lockfile conflicts, and preflight the artifacts (v5.5.67)

askalf:
## Why

Four version-bumping PRs were in flight at once on 2026-08-25. Every merge invalidated the others on the same two files, each had to be hand-resolved, and **three separate defects came out of that hand work — none of which any unit test could have failed.**

This closes the two mechanical holes that let them through.

## 1. The resolver never handled `package-lock.json`

`scripts/resolve-release-conflicts.mjs` already existed and already encoded the rule (keep the higher version, never resolve downward — a version below a published tag makes the duplicate-tag guard fast-exit green having shipped nothing). It covered `package.json` and `CHANGELOG.md`.

It did not cover the lockfile. Neither did the healer: `drift-pr-heal.yml` ran `git add package.json CHANGELOG.md`.

Those two gaps **compound**. The healer resolves a PR, leaves the lockfile carrying live `<<<<<<<` markers, and reports success. That file is then invalid JSON — and nothing downstream notices:

- `npm test` passes, because a warm `node_modules` means nothing ever parses the lockfile.
- `npm ci` is where it dies. Which is CI, and deploy.

This is not hypothetical — it happened on #1107, and the reviewer caught it only because it hypothesized exactly this cached-`node_modules` mechanism.

The lockfile carries the same version **twice** (root `.version` and `.packages[""].version`), each conflicting as its own lone-version hunk, so the existing rule applies verbatim — just more than once. `resolvePackageJson` was refactored into a shared `resolveVersionOnlyConflicts(text, label)` and registered for both files.

A genuine dependency-graph conflict still **throws**. Regenerating a lockfile is `npm install`'s job; a line-merger emitting a plausible tree nobody verified is worse than a hand-resolution.

## 2. `npm run preflight`

Four assertions, each one a defect that actually shipped:

| check | the defect it catches |
|---|---|
| no conflict markers in tracked files | markers committed live |
| every tracked `.json` parses | the invalid lockfile above |
| version agrees: `package.json` ↔ **both** lockfile slots ↔ CHANGELOG heading | fields drifting apart mid-collision |
| no release entry inside the CHANGELOG header comment | see below |

That last one is the subtle one. The header comment contains a literal `## [X.Y.Z] - YYYY-MM-DD` placeholder, so an inserter anchoring on the *first* `## [` writes the entry **inside the comment**, where it never renders. And stripping comments before parsing headings — which the version check *must* do — is precisely what blinds you to it: the buried entry vanishes from the parse and the placeholder survives, so every other check reads clean. So it is looked for explicitly.

Two deliberate design calls:

- **Marker detection is line-anchored**, so prose that merely *quotes* a marker does not trip it — `CHANGELOG.md` legitimately documents Cline's `<<<<<<< SEARCH` diff fence, and a checker that cries wolf on its own changelog gets turned off.
- **A file whose content is marker-shaped on purpose opts out** with a greppable `preflight:allow-conflict-markers` sentinel. The resolver's own test fixtures *are* conflicts. A path allowlist would have silently grown instead.

Wired into the existing `validate-package-json` job rather than a new one, so it **adds no required status check** to branch protection — same cheap, no-install class as the two checks already there.

## Verification

Not "it passes" — each check was run against a **reconstruction of the defect it exists to catch**, confirmed to fail, then confirmed to pass once repaired:

- lockfile with conflict markers → caught (marker + invalid-JSON, both)
- lockfile version drifted from `package.json` → caught
- a release heading buried in the header comment → caught

The third reconstruction is worth calling out: my first two attempts at it did **not** trip the checker, because the placeholder sits mid-line inside backticks and my injection never produced a line-start heading. That exposed a real gap in the check, which is why the buried-heading assertion exists at all rather than just the placeholder-intact one.

`npm test`: **144 pass, 0 fail** (10 new lockfile assertions). `npm run lint:pkg` OK. `npm run preflight` clean — it validated this PR's own version bump.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,15 @@ jobs:
       - name: README line-count claim still matches src/
         run: node scripts/check-readme-line-count.mjs
 
+      # Artifact-level defects a green `npm test` cannot catch: conflict
+      # markers left in a tracked file, a lockfile that stopped being valid
+      # JSON, versions drifting between package.json / lockfile / CHANGELOG,
+      # a release entry buried in the CHANGELOG's header comment. All four
+      # shipped for real during the 2026-08-25 release collision. Same cheap
+      # no-install class as the two above, so it lives in this job.
+      - name: preflight — release artifacts are internally consistent
+        run: node scripts/preflight.mjs
+
   build:
     needs: validate-package-json
     runs-on: ubuntu-latest
--- a/.github/workflows/drift-pr-heal.yml
+++ b/.github/workflows/drift-pr-heal.yml
@@ -176,7 +176,7 @@ jobs:
               continue
             fi
             if ! git diff --quiet -- package.json CHANGELOG.md; then
-              git add package.json CHANGELOG.md
+              git add package.json package-lock.json CHANGELOG.md
               git commit -q -m "chore: renumber release above ${effective_floor} (base moved under this PR)"
             fi
 
@@ -213,7 +213,7 @@ jobs:
                 refused=$((refused + 1))
                 continue
               fi
-              git add package.json CHANGELOG.md
+              git add package.json package-lock.json CHANGELOG.md
               # Belt and braces: if anything is still unmerged the commit would
               # fail and `set -e` would end the job, so verify before running it.
               if [ -n "$(git diff --name-only --diff-filter=U)" ]; then
--- a/.github/workflows/drift-pr-heal.yml
+++ b/.github/workflows/drift-pr-heal.yml
@@ -175,7 +175,13 @@ jobs:
               refused=$((refused + 1))
               continue
             fi
-            if ! git diff --quiet -- package.json CHANGELOG.md; then
+            # package-lock.json is in BOTH the test and the add: renumber now
+            # mirrors the new version into the lockfile's two version slots, so
+            # a renumber that changed only the lockfile must still commit.
+            # Testing just the other two would leave that write uncommitted and
+            # the lockfile would disagree with package.json — the exact drift
+            # `npm run preflight` fails on.
+            if ! git diff --quiet -- package.json package-lock.json CHANGELOG.md; then
               git add package.json package-lock.json CHANGELOG.md
               git commit -q -m "chore: renumber release above ${effective_floor} (base moved under this PR)"
             fi
@@ -183,7 +189,8 @@ jobs:
             # Conflicts are expected — that is the case being healed — so the
             # merge is allowed to fail and the resolver decides.
             if ! git merge "origin/$base" --no-edit >/dev/null 2>&1; then
-              # The resolver only ever touches package.json and CHANGELOG.md.
+              # The resolver only ever touches the three files that carry the
+              # version: package.json, package-lock.json and CHANGELOG.md.
               # A drift PR CAN also conflict in source — observed on #1038,
               # where #1037 had added utilFreshness to the same pool import
               # line this PR was adding accountIneligibility to. Resolving the
@@ -196,7 +203,7 @@ jobs:
               # how the separator lands, and the first draft of it refused
               # everything because `tr` left a trailing space.
               outside=$(git diff --name-only --diff-filter=U \
-                        | grep -vxE 'package\.json|CHANGELOG\.md' || true)
+                        | grep -vxE 'package\.json|package-lock\.json|CHANGELOG\.md' || true)
               if [ -n "$outside" ]; then
                 echo "  REFUSED — conflicts outside the version files:"
                 # read-loop rather than `sed 's/^/    /'`: shellcheck SC2001
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,28 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.67] - 2026-08-25
+
+### Fixed
+
+- Release-conflict resolver now handles `package-lock.json`. It carries the
+  same version twice (root `.version` and `.packages[""].version`), so a
+  release collision conflicts it exactly like `package.json` — but the
+  resolver had no handler for it and `drift-pr-heal` did not even `git add`
+  it. A healed PR could therefore carry a lockfile with live `<<<<<<<`
+  markers, i.e. invalid JSON. `npm test` passes anyway (a warm
+  `node_modules` means nothing parses the lockfile); `npm ci` is where it
+  dies, which is CI and deploy. A dependency-graph conflict still throws
+  rather than guessing — regenerating a lockfile is `npm install`'s job.
+
+### Added
+
+- `npm run preflight` (and a CI step) for artifact-level defects no unit test
+  can catch: conflict markers in tracked files, unparseable JSON, version
+  drift across `package.json` / both lockfile slots / the CHANGELOG heading,
+  and a release entry buried inside the CHANGELOG header comment. All four
+  shipped for real during the 2026-08-25 four-PR release collision.
+
 ## [5.5.66] - 2026-08-25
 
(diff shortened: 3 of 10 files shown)


sprayberry-redline:
## dario/live-test — PASS

Head: `bb59e66b84b71f79484a9549e7c0b71f03f44306`

| Stage | Result | Note |
|---|---|---|
| install | PASS | `npm ci` clean, 6 packages |
| build | PASS | `tsc` clean |
| test | PASS | 144/144 passed, 0 failed |
| doctor | PASS | exits 1 as expected with no credentials (`[FAIL] OAuth not authenticated`), no crash |
| live probe | SKIPPED | `DARIO_TEST_UPSTREAM_API_KEY` not set |

All stages clean.

sprayberry-redline (changes requested):

**Verdict:** 1 correctness issue that defeats the PR's stated purpose — requesting changes. Everything else (the resolver refactor, the preflight script, and the 93 new test assertions) reads well.

## Findings

### 🔴 Blocking — `drift-pr-heal.yml`'s "conflicts outside the version files" guard still excludes `package-lock.json`, so the new lockfile-resolution path is unreachable from the automated healer

`.github/workflows/drift-pr-heal.yml:29-30` and `:38-39` (changed lines):
```diff
-              git add package.json CHANGELOG.md
+              git add package.json package-lock.json CHANGELOG.md
```
These two hunks assume `resolve-release-conflicts.mjs` will have resolved a conflicted `package-lock.json` by the time this `git add` runs. But the merge-failure branch a few lines above each of those calls first filters out any conflicted path that isn't `package.json` or `CHANGELOG.md`, and this PR does not touch that filter — it's still, at the PR's head commit:
```
if [ -n "$outside" ]; then
  echo "  REFUSED — conflicts outside the version files:"
  ...
  git merge --abort || true
  refused=$((refused + 1))
  continue
fi
```
(unchanged in this diff — confirmed by fetching the file at the PR's head SHA).

**Failure scenario:** exactly the situation the PR's own description exists to fix — a release-bump PR whose merge conflicts `package.json`, `CHANGELOG.md`, *and* `package-lock.json` together (the normal shape, since all three carry the version). `git diff --name-only --diff-filter=U` now includes `package-lock.json`, which doesn't match `package\.json|CHANGELOG\.md`, so it lands in `$outside`, the guard fires, the merge is aborted, and the PR is REFUSED before `node scripts/resolve-release-conflicts.mjs` (and therefore the new `resolvePackageLock`) is ever invoked. The new resolver/`git add` code is correct but dead on this path — `drift-pr-heal` will now refuse essentially every real collision instead of healing it, which is a regression of the workflow's whole reason to exist ("This closes it: ... resolve the two-file conflict ... The PR then arms and releases through the normal path").

```suggestion
outside=$(git diff --name-only --diff-filter=U \
          | grep -vxE 'package\.json|package-lock\.json|CHANGELOG\.md' || true)
```

## What's good
- `resolveVersionOnlyConflicts` extraction and the new `resolvePackageLock` correctly reuse the existing "keep the higher version, throw on anything richer" rule, with the file name threaded into the error message.
- `scripts/preflight.mjs` is well-targeted at the exact three defects described (invalid JSON, version drift, buried-in-comment CHANGELOG heading), uses index-based comment detection to sidestep the `js/incomplete-multi-character-sanitization` CodeQL class of bug it calls out, and is wired into the existing `validate-package-json` job rather than adding a new required check.
- Strong test coverage for the new lockfile path: both version slots, a rejected dependency-graph conflict, and a no-op safety case.
- CI (`gh pr checks`) is all green, including `validate-package-json`, `build`, `test`, `CodeQL`, and `actionlint`.

## What I didn't verify
Did not run `preflight.mjs`'s buried-heading regex against edge cases beyond what the diff's own tests cover, and did not check whether other callers of `resolve-release-conflicts.mjs` (outside `drift-pr-heal.yml`) might have their own file-scope guard — only the one shown above was in the diff's blast radius.

askalf:
Good catch, and correct — fixed in c222803 (rebased onto #1109/#1110 too).

You were right that the new lockfile path was **dead code from the workflow's side**. I'd updated both `git add` sites but not the `outside=` filter that runs before either, so a real collision — which conflicts all three version-carrying files — would land `package-lock.json` in `$outside` and take the REFUSED branch without ever invoking the resolver.

### A second defect this surfaced

Chasing it down found a pre-existing one in the same path. `renumber-release-version.mjs` writes `package.json` and `CHANGELOG.md` but **not** the lockfile. Verified directly:

```
before: pkg=5.5.67 lock=5.5.67
[renumber] renumbered 5.5.67 -> 5.5.81
after:  pkg=5.5.81 lock=5.5.67
```

That drift was previously silent. With `preflight` landing in this PR it becomes a **red check** — so the healer would have started producing PRs that fail CI on a defect the healer itself introduced. Renumber now mirrors the version into both lockfile slots.

Worth noting this needs no refusal clause, unlike the CHANGELOG heading: those slots aren't independent information, they *mirror* `package.json`'s version by definition, so writing it is derivation, not a guess.

I also added the lockfile to the `git diff --quiet` test above the commit — otherwise a renumber touching only the lockfile would leave that write uncommitted.

### On your closing point

> none of those checks exercise `drift-pr-heal.yml`'s runtime logic, which is exactly why the finding above wasn't caught

That's the actual bug, so that's what I fixed. This is a **shape, not a one-off**: the workflow decides what's "ours" from a hardcoded list, the resolver decides from `HANDLERS`, and nothing tied them together — so teaching the resolver a new file silently made it unreachable.

`test/renumber-release-version.mjs` now reads the **actual grep pattern out of the workflow** and asserts every key of `HANDLERS` survives it, that real source paths are still refused (the #1038 shape must keep aborting), and that `-x` exact matching still rejects a lookalike like `packages/package.json`.

Confirmed it actually catches your finding by reverting the guard to its buggy form:

```
FAIL package-lock.json: resolver handles it AND the workflow lets it through
  :: workflow pattern /package\.json|CHANGELOG\.md/ would send package-lock.json to the REFUSED branch
53 pass, 1 fail
```

and back to `54 pass, 0 fail` once restored. Adding a fourth file to the resolver without updating the workflow now fails CI instead of shipping dead code.

144 tests pass (18 new), preflight clean, and end-to-end `renumber` moves `package.json`, both lockfile slots and the CHANGELOG heading together.

sprayberry-redline (approved):

**Verdict:** Prior blocking finding is fixed; approving.

## Re-review context

I previously requested changes on this PR (commit `bb59e66`) because `.github/workflows/drift-pr-heal.yml`'s "conflicts outside the version files" guard still excluded `package-lock.json`, making the new lockfile-resolution path unreachable. New commit `c222803` fixes exactly that.

## Verified fix

`.github/workflows/drift-pr-heal.yml` (changed lines):
```diff
-                        | grep -vxE 'package\.json|CHANGELOG\.md' || true)
+                        | grep -vxE 'package\.json|package-lock\.json|CHANGELOG\.md' || true)
```
This matches my suggested fix exactly, and both `git diff --quiet` checks / `git add` calls in the same file now include `package-lock.json` alongside `package.json` and `CHANGELOG.md`. The lockfile-resolution path added earlier in this PR is now reachable from the automated healer.

## What's good
- The commit also caught and fixed a second, related defect: `renumber-release-version.mjs` renumbered `package.json`/`CHANGELOG.md` but not the lockfile, which would have produced version drift that the PR's own new `preflight` check would then fail. The fix mirrors the version into the lockfile's two slots only when deriving from an already-known new version (no guessing), and leaves an absent lockfile alone.
- `test/renumber-release-version.mjs` now reads the actual `grep -vxE` pattern out of the workflow file and asserts every key of `resolve-release-conflicts.mjs`'s exported `HANDLERS` survives it, plus negative cases (`src/pool.ts`, `README.md`, a `packages/package.json` lookalike that `-x` must still reject). This directly guards against the class of bug just fixed — the workflow's allowlist silently drifting from the resolver's capability — recurring.
- CI is all green: `build`, `test`, `validate-package-json`, `CodeQL`, `actionlint`, `docker-cap-drop-smoke`.

## What I didn't verify
Did not re-verify the full lockfile-resolver logic from the original diff (already covered in my prior review); this pass focused on confirming the fix to the blocking finding and reviewing only the new commit's changes.

PR #1109, opened 2026-08-25
fix(canary): borrow the shared credential read-only, never rotate it (v5.5.68)

askalf:
## The hole this closes

Anthropic **invalidates the previous refresh_token on every refresh**. dario's own code already documents this — it's the entire reason `doRefreshAccountTokenDistributed` and the Cloudflare refresh-lock exist.

The consequence: a process that merely *reads* a shared credential store is harmless, but the moment it **refreshes**, every other holder of that token is silently logged out. **Copying the file first does not help** — the rotation happens upstream, so the copy invalidates the original too.

On 2026-08-25 a job on the self-hosted runner did exactly that. Production went down for ~3h on `invalid_grant`.

## Why isolation doesn't fix this one

#1106 fixed the compat suite by pointing `HOME` at a temp dir. That's the right fix when a job needs no real auth.

The billing-classifier canary is the **last workflow on that runner still starting a proxy against the shared store**, and neither escape works for it:

- **It can't use an API key.** An API key *is* the other billing bucket — it would test the wrong thing and always "pass."
- **An isolated credential was already tried, and failed.** With only sparse workflow fires it had no refresh authority between runs and repeatedly died of disuse. That history is recorded in the step's own comment.

It genuinely needs the shared subscription credential. So the fix isn't isolation — it's **removing the authority to rotate**.

## `DARIO_NO_TOKEN_REFRESH=1`

Borrow-don't-rotate: requests work for as long as the access token is valid, and a needed renewal **fails loudly** instead of quietly rotating a token production is holding.

Two deliberate placement calls:

- **Both paths are guarded.** dario has two credential stores — legacy `credentials.json` (oauth.ts) and the account pool (accounts.ts) — and which one loads depends on machine state. Guarding one is *worse* than guarding neither, because it reads as protected while leaving the hole open half the time.
- **The guard sits before the network exchange**, and before the single-flight bookkeeping. A guard placed after the exchange would still "throw" — having already invalidated production's token. That's the failure with an error message stapled on.

Strict parsing (`'1'` only; `'true'` and `'0'` do **not** enable it) so an operator typo yields normal behaviour rather than a surprise half-on state.

## A false alarm this would otherwise have introduced

Worth calling out, because the safety fix nearly created a reporting bug.

An expired borrowed token yields a 401 with no headers → `claim=""` → `bucket="unknown"` → **`status=warn`, a billing alert**. That would be a lie: nothing about the classifier changed, the canary simply couldn't run.

It's the same false-alarm shape that held **#1003** open for days while dario billed correctly the whole time — a precedent the workflow's own comments already warn about.

Now it fails as an *infrastructure* error and writes **no `status` output**, so the alert step (gated on `status != ''`) is skipped rather than opening a misleading issue. The error text says explicitly that it says nothing about billing, and what to do instead.

## Verification

- `npm test`: **145 pass, 0 fail** (15 new assertions).
- Both refresh paths asserted to throw the *disabled* error specifically — not incidentally failing on "no token available," which would pass a weaker test while proving nothing.
- **Default-unchanged is asserted directly**: with the flag unset, a failure must *not* be the disabled error. A guard that leaked into the default path would break every normal dario run.
- Workflow YAML parses; every `run:` block passes `bash -n`.
- Swept all workflows for others starting a proxy: only this one and `compat-test-self-hosted.yml`, which #1106 already fixed. This closes the last one.

### DIFF
--- a/.github/workflows/cc-billing-classifier-canary.yml
+++ b/.github/workflows/cc-billing-classifier-canary.yml
@@ -119,9 +119,25 @@ jobs:
 
       - name: Start dario proxy (canonical-rebuild — NOT passthrough)
         # OAuth credential: shares /root/.claude/.credentials.json with the
-        # platform dario, same rationale as compat-test-self-hosted.yml —
-        # the isolated /root/.claude-runner credential had no refresh
-        # authority between sparse workflow fires, repeatedly died of disuse.
+        # platform dario. An isolated credential was tried and does not work
+        # here — with only sparse workflow fires it had no refresh authority
+        # between runs and repeatedly died of disuse. And an API key is not an
+        # option either: an API key IS the other billing bucket, so it would
+        # test the wrong thing. This canary genuinely needs the shared
+        # subscription credential.
+        #
+        # BUT IT MUST NOT REFRESH IT. Anthropic invalidates the previous
+        # refresh_token on every refresh, so a rotation here logs out every
+        # other holder — production. On 2026-08-25 exactly that happened from
+        # a job on this runner and the fleet was down ~3h on `invalid_grant`.
+        # Copying the file first does not help; the rotation is upstream.
+        #
+        # DARIO_NO_TOKEN_REFRESH=1 borrows the credential read-only: requests
+        # work while the access token is valid, and a needed renewal fails
+        # LOUDLY instead of silently rotating prod's token away. A canary that
+        # cannot run is a fixable annoyance; a canary that takes the fleet down
+        # is not.
+        #
         # NO --passthrough flag: the canary specifically validates dario's
         # rebuild-from-canonical mode, the mode every non-CC dario user
         # runs in.
@@ -133,6 +149,8 @@ jobs:
         # using PLATFORM credentials, producing 401s that look like
         # classifier flips. Use :3457 so the workflow's own dist actually
         # runs.
+        env:
+          DARIO_NO_TOKEN_REFRESH: '1'
         run: |
           node dist/cli.js proxy --port=3457 > proxy.log 2>&1 &
           echo $! > proxy.pid
@@ -195,6 +213,25 @@ jobs:
           head -30 "$response_headers"
           echo ""
           echo "Captured: representative-claim='${claim}' served-model='${served_model}' fallback%='${fallback_pct}' 5h-util='${util_5h}'"
+
+          # --- Precondition, checked BEFORE any billing verdict is computed ---
+          # This canary borrows a shared subscription credential read-only
+          # (DARIO_NO_TOKEN_REFRESH=1, see the proxy step). If the access token
+          # expired, the probe gets a 401 with no headers — which would fall
+          # through to bucket="unknown" and raise a BILLING alert. That would be
+          # a lie: nothing about the classifier changed, the canary simply could
+          # not run. It is the same false-alarm shape that held #1003 open for
+          # days while dario billed correctly the whole time.
+          #
+          # So fail LOUDLY as an infrastructure problem and write no `status`
+          # output at all — the alert step is gated on `status != ''`, so it is
+          # skipped rather than opening a misleading billing issue.
+          if grep -q 'DARIO_NO_TOKEN_REFRESH' proxy.log 2>/dev/null; then
+            echo "::error::Canary could not run: the borrowed subscription credential needs refreshing, which this job is deliberately forbidden to do (refreshing would rotate the token out from under production). This says NOTHING about the billing classifier. Refresh the credential on the box, then re-run this workflow."
+            echo "=== proxy.log tail ==="
+            tail -40 proxy.log 2>/dev/null || true
+            exit 1
+          fi
 
           # Billing verdict
           case "$claim" in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,26 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.68] - 2026-08-25
+
+### Added
+
+- `DARIO_NO_TOKEN_REFRESH=1` — borrow a shared credential read-only. Anthropic
+  invalidates the previous refresh_token on every refresh, so any process that
+  refreshes a shared store logs out every other holder of it. Set this and both
+  refresh paths (legacy `credentials.json` and the account pool) refuse with an
+  explanatory error instead of rotating. Default behaviour is unchanged.
+
+### Fixed
+
+- The billing-classifier canary no longer risks the fleet. It runs on the
+  self-hosted runner against the shared subscription credential — it cannot use
+  an API key, because an API key IS the other billing bucket — and on
+  2026-08-25 a job of exactly this shape refreshed that credential and took
+  production down for ~3h with `invalid_grant`. It now borrows read-only, and
+  an expired token fails as an infrastructure error rather than falling through
+  to `bucket=unknown` and raising a billing alert that would be false.
+
 ## [5.5.67] - 2026-08-25
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.67",
+  "version": "5.5.68",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "5.5.67",
+      "version": "5.5.68",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.67",
+  "version": "5.5.68",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/accounts.ts
+++ b/src/accounts.ts
@@ -23,7 +23,7 @@ import { homedir } from 'node:os';
 import { randomUUID, randomBytes, createHash } from 'node:crypto';
 import { createServer, type IncomingMessage, type ServerResponse } from 'node:http';
 import { detectCCOAuthConfig } from './cc-oauth-detect.js';
-import { loadCredentials, saveCredentialsTokens, buildManualAuthorizeUrl, parseManualPaste, readLineFromStdin, enumerateKeychainCredentials, type KeychainEntry } from './oauth.js';
+import { loadCredentials, saveCredentialsTokens, buildManualAuthorizeUrl, parseManualPaste, readLineFromStdin, enumerateKeychainCredentials, tokenRefreshDisabled, refreshDisabledError, type KeychainEntry } from './oauth.js';
 import { openBrowser } from './open-browser.js';
 import { redactSecrets } from './redact.js';
 import { durableWriteFile } from './durable-write.js';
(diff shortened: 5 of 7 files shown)


sprayberry-redline (changes requested):

**Verdict:** 1 blocking correctness issue — the headline `drift-pr-heal` fix for `package-lock.json` conflicts can never actually run; the rest (the `DARIO_NO_TOKEN_REFRESH` guard, the canary hardening, `preflight.mjs`) reads well.

## Findings

### Blocking — `package-lock.json` conflicts are refused before the new resolver ever runs
`.github/workflows/drift-pr-heal.yml:198-199` (unchanged by this PR):
```
```
This PR's stated purpose (per the CHANGELOG entry for 5.5.67 and `scripts/resolve-release-conflicts.mjs`'s new `resolvePackageLock`) is that a release-collision merge conflicting `package-lock.json` should now be auto-healed instead of leaving invalid JSON in the tree. But the `outside` guard that runs immediately after `git merge` fails still only whitelists `package.json` and `CHANGELOG.md`. Since this PR's own version bump touches `package-lock.json` (see the diff to that file: `5.5.66` → `5.5.68`), any real drift PR healed by this workflow will have `package-lock.json` show up in `git diff --name-only --diff-filter=U`, `grep -vxE 'package\.json|CHANGELOG\.md'` will NOT filter it out, `outside` will be non-empty, and the job takes the `REFUSED — conflicts outside the version files` branch at line 200-206 — aborting the merge and never reaching `node scripts/resolve-release-conflicts.mjs` at all. The new `resolvePackageLock` handler is exercised by the unit tests but is unreachable from the workflow that was supposedly the whole point of adding it.

Even if that guard were fixed, the post-resolve staging at line 216 (`git add package.json CHANGELOG.md`) also omits `package-lock.json`, so the subsequent "unmerged paths remain" check at line 218-221 would still see it as unmerged and refuse anyway.

Concrete failure scenario: a future 2026-08-25-style release collision now conflicts `package-lock.json` (as this very PR's diff does) → `drift-pr-heal` merges, hits the conflict, sees `package-lock.json` in the unmerged set → treats it as "outside the version files" → aborts the merge and refuses the PR, exactly the pre-PR behavior this change claims to fix.

```suggestion
# .github/workflows/drift-pr-heal.yml
-              outside=$(git diff --name-only --diff-filter=U \
-                        | grep -vxE 'package\.json|CHANGELOG\.md' || true)
+              outside=$(git diff --name-only --diff-filter=U \
+                        | grep -vxE 'package\.json|package-lock\.json|CHANGELOG\.md' || true)
...
-              git add package.json CHANGELOG.md
+              git add package.json package-lock.json CHANGELOG.md
               # Belt and braces: if anything is still unmerged the commit would
```
(the second `git add` is the one at line 216, immediately after the resolver call — the line-179 one before the merge is fine as-is since `renumber-release-version.mjs` never touches the lockfile)

## What's good
- `DARIO_NO_TOKEN_REFRESH` is guarded on both credential paths (`oauth.ts` and `accounts.ts`), placed before any network exchange/single-flight bookkeeping, uses strict `'1'`-only parsing, and is covered by tests that assert the *specific* disabled-error message rather than an incidental failure — including a test explicitly designed to avoid mutating a real credential.
- The canary's false-alarm avoidance (writing no `status` output on infra failure so the alert step is skipped) is a real fix for a real precedent (#1003) and is easy to follow in the diff.
- `preflight.mjs`'s CHANGELOG-buried-heading check correctly avoids the CodeQL-flagged unsafe-comment-stripping pattern by indexing comment ranges instead of rewriting text.
- CI is fully green (build/test/CodeQL/actionlint/docker-cap-drop-smoke all pass) — but note none of those checks exercise `drift-pr-heal.yml`'s runtime logic, which is exactly why the finding above wasn't caught.

sprayberry-redline:
## dario/live-test — PASS

Head SHA: `7938206ee63c5ef83bc16aa972f2cd75fb4f01ce`

| Stage | Result | Note |
|---|---|---|
| install | ✅ pass | `npm ci` clean |
| build | ✅ pass | `tsc` clean, dist/cc-template-data.json copied |
| test | ✅ pass | 145/145 tests pass, 0 fail |
| cli | ✅ pass | `--help` exits 0 |
| doctor | ✅ pass | ran without crash; reports `[FAIL] OAuth not authenticated` as expected (no creds in scrubbed env) |
| live probe | ⏭️ skipped | `DARIO_TEST_UPSTREAM_API_KEY` not set — not a failure |

Status: https://github.com/askalf/dario/commit/7938206ee63c5ef83bc16aa972f2cd75fb4f01ce

askalf:
Fixed in 0298af4 — thanks, the guard was the real blocker.

`.github/workflows/drift-pr-heal.yml:199` now reads:

```
```

plus the stale comment above it (`# The resolver only ever touches package.json and CHANGELOG.md.`) updated to name the lockfile.

One correction to the review: the second half of the finding — `git add package.json CHANGELOG.md` at line 216 omitting the lockfile — was **already** fixed in this PR. Both `git add` lines (the pre-merge renumber one at 179 and the post-resolve one at 216) already read `git add package.json package-lock.json CHANGELOG.md` before this commit. Only the `outside` guard was left behind, so this commit is a one-line change plus the comment.

## Verification — end-to-end reproduction of the healer's merge path

Not a unit test; a real three-commit release collision in a scratch repo, run against `scripts/resolve-release-conflicts.mjs` as it stands on this branch. A realistic lockfile (40 dep entries) matters here: my first fixture was small enough that git collapsed both version slots into a single hunk spanning the dependency block, which the resolver correctly refuses as "not a lone version line". With normal padding the two slots conflict independently, exactly as described in the 5.5.67 CHANGELOG entry.

Unmerged set after `git merge master` fails: `CHANGELOG.md`, `package-lock.json`, `package.json`.

- **Pre-fix guard on that set** -> emits `package-lock.json` -> `outside` non-empty -> `REFUSED - conflicts outside the version files`, merge aborted, resolver never invoked. This is the dead-code claim, confirmed.
- **Post-fix guard on that set** -> empty -> falls through to the resolver, which prints `package.json: resolved` / `package-lock.json: resolved` / `CHANGELOG.md: resolved`.
- After `git add package.json package-lock.json CHANGELOG.md`: `git diff --name-only --diff-filter=U` is empty, so the belt-and-braces check passes and the commit runs.
- Resulting lockfile parses as JSON with root `.version` = 5.5.62 and `.packages[""].version` = 5.5.62, matching `package.json` = 5.5.62 — the higher side kept in all three slots, never resolved downward.

**Refusal behaviour preserved:** with `src/oauth.ts` in the unmerged set alongside `package-lock.json`, the new pattern still emits `src/oauth.ts`, so the #1038 source-conflict case still refuses cleanly instead of taking the job down under `set -e`.

Source ticket: 01M0X0T1RVBHNFFKM4K8NX32DR.

askalf:
Same finding as #1108 — fixed in `c222803`, which is now the base of this branch.

The `outside=` guard now allows `package-lock.json`, so the resolver is actually reachable; `renumber-release-version.mjs` also mirrors the version into the lockfile's two slots (it wrote only `package.json` + `CHANGELOG.md` before, which `preflight` would have turned into a red check on every healed PR).

The durable fix is a test that reads the **actual grep pattern out of `drift-pr-heal.yml`** and asserts every key of `HANDLERS` survives it — verified to fail when the guard is reverted to its buggy form. Full rationale in the [#1108 reply](https://github.com/askalf/dario/pull/1108#issuecomment-5414548539).

sprayberry-redline (dismissed):

**Verdict:** Prior blocking finding is fixed — approving.

## Re-review context

This is a re-review of the current head (`194b6a4`) after my earlier `CHANGES_REQUESTED` review (submitted `2026-08-25T17:20:39Z`) flagged that `drift-pr-heal.yml`'s `outside` guard and post-resolve `git add` still whitelisted only `package.json`/`CHANGELOG.md`, making the new `resolvePackageLock` handler unreachable from the workflow.

## What changed since my last review

Diffing the reviewed commit (`7938206`) against the current head (`194b6a4`):

- `.github/workflows/drift-pr-heal.yml`: the `outside` guard's `grep -vxE` pattern now includes `package-lock\.json`, and **both** `git add` staging lines (pre-merge renumber-commit and post-resolve) now include `package-lock.json`. This directly fixes the unreachable-resolver bug — a `package-lock.json` conflict no longer gets refused as "outside the version files."
- `scripts/renumber-release-version.mjs`: `renumber()` now takes an optional `lockJsonString`, mirrors the resolved version into the lockfile's `.version` and `.packages[""].version` slots, and `main()` reads/writes the lockfile only when present (`existsSync` guard) and only writes if the read succeeded — consistent with the existing "all writes or none" invariant the comment describes.
- `scripts/resolve-release-conflicts.mjs`: `HANDLERS` is now exported.
- `test/renumber-release-version.mjs` and a new binding test (in the resolve-release-conflicts suite) assert the workflow's actual `grep` allowlist regex is extracted and checked against every file `HANDLERS` claims to resolve, plus that non-resolvable files (`src/pool.ts`, `README.md`) and lookalikes (`packages/package.json`) are still correctly refused. This is exactly the kind of test that would have caught the original gap, and it's now a standing guard against the same two lists drifting apart again.

I traced the new `outside` regex and staging lines against the workflow's control flow: a `package-lock.json`-only conflict now reaches the resolver and gets staged, and the belt-and-braces "unmerged paths remain" check at line ~218 will see no leftover unmerged `package-lock.json` since it's now included in the `git add`. No new gap introduced by this fix.

## What's good
- The lockfile handling is optional/backward-compatible (`existsSync` check, `undefined` lock text skips the mirror) rather than assuming every checkout has one.
- The new test binds the workflow's regex directly rather than re-asserting a hardcoded copy of it, so it can't silently drift out of sync with a future workflow edit.
- CI checks relevant to this diff (`build`, `actionlint`, `validate-package-json`, `triage`) are green; no failures.

No blocking issues remain from this round.

sprayberry-redline:
| Stage | Result | Note |
|---|---|---|
| build (`npm run build`) | ✅ pass | tsc + asset copy, no errors |
| test (`npm test`) | ✅ pass | 145/145 passing, 0 fail |
| cli (`--help`) | ✅ pass | exits 0, help text renders |
| doctor | ✅ pass | reports `[FAIL] OAuth not authenticated` correctly (no creds provided) — does not crash |
| live probe | ⏭️ skipped | `DARIO_TEST_UPSTREAM_API_KEY` not set in tester env |

Head SHA: `8ba7dc4f7679fa4e2e67c1d7371b51052bc601f2`

**Verdict: PASS**

sprayberry-redline (approved):

## What I checked
- `gh pr diff 1109` (7 files, +224/-7): workflow change, CHANGELOG, version bump, `src/accounts.ts`, `src/oauth.ts`, new `test/no-token-refresh.mjs`.
- CI check-rollup (`gh pr checks 1109`): all 12 checks pass, including `test` and `compat`.
- A prior CHANGES_REQUESTED review on this PR (commit `194b6a49…`) was later dismissed and targeted a different file (`drift-pr-heal.yml`) not present in the current diff — not applicable to the current head (`8ba7dc4f`).

## Findings

None blocking. The core change is straightforward and correctly applied:

- `src/oauth.ts:912-913` — `refreshTokens()` checks `tokenRefreshDisabled()` and throws before setting `refreshInProgress`, so a refusal never poisons the in-flight promise.
- `src/accounts.ts:151-156` — `refreshAccountToken()` checks the same guard before touching `accountRefreshesInFlight`, matching the legacy-path placement for the same reason (both stores need the guard, since which one loads is machine-state-dependent).
- `.github/workflows/cc-billing-classifier-canary.yml:152-153,229-233` — `DARIO_NO_TOKEN_REFRESH=1` is set on the proxy step, and the probe step greps `proxy.log` for the marker text before computing a billing verdict, failing as an infra error (`exit 1`, no `status` output) rather than raising a false billing alert. I traced that `getAccessToken()` in `oauth.ts:1004-1008` logs `lastRefreshError` (which contains `DARIO_NO_TOKEN_REFRESH`, well within the 200-char truncation) via `console.error`, and the proxy step redirects both stdout/stderr into `proxy.log`, so the grep will actually see it in the real failure path.
- `test/no-token-refresh.mjs` exercises both refresh paths throwing the *disabled* error specifically (not a generic "no token" error), asserts the pool guard fires before any network call, and asserts the default-unchanged case without ever calling a live refresh — consistent with the CI `test` check passing.

## What's good
- Both credential stores are guarded symmetrically instead of just the one the PR's motivating incident hit — the comment on `accounts.ts:151` explicitly calls out why guarding only one path is worse than guarding neither.
- The false-alarm precondition in the canary avoids repeating the #1003 false-billing-alert failure mode.
- Version bump (`package.json`/`package-lock.json` 5.5.67→5.5.68) and CHANGELOG entry are consistent with the change.

PR #1110, opened 2026-08-25
feat(watch): alert when the box is not running what was released (v5.5.69)

askalf:
> Stacked on #1108 and #1109. Reviewable now; merge after those.

## The gap

`cc-drift-auto-release.yml`'s idempotency gate already proves the **artifacts** exist — the git tag, the npm version, the ghcr image.

Nothing proved the last leg: **that the box actually pulled them and is running them.** Publishing is not deploying.

That leg has failed silently before — dario's refresh-lock `lockId` code sat released-but-not-live on the Cloudflare Worker while every status surface read green. Each surface was reporting on a different link in the chain, and no single link *is* the deployment.

## Why the health watcher can't cover it

`cc-oauth-health.yml` is a genuinely good watcher — its axis 2 makes a real `max_tokens:1` request, which is why it caught today's outage in minutes (#1103, opened 11:35Z, auto-closed on recovery at 14:40Z).

But it probes the running container and asks *"are you serving?"* — and **a container running a stale version answers yes, truthfully.** Healthy and current are different questions, and only one of them was being asked.

## Design calls

- **90-minute grace window.** A release legitimately takes time to reach the box (multi-arch build → ghcr push → pull). Alerting the instant a tag appears would page on *every* release. The window is what makes this a deploy-**stuck** detector rather than a deploy-in-progress one.
- **Silent when the container is unreadable.** `cc-oauth-health.yml` owns container-down and files that alert with the right diagnostics. Two watchers filing separate issues for one outage doubles the noise on the worst possible day. Silence here is division of labour, not a blind spot.
- **`ahead` alerts regardless of grace.** Running something never released (hand-built image, aborted rollback) means the tag no longer describes what's running — that's the invariant.
- **The container's own `package.json` is the source of truth**, not the image tag. `:latest` is a moving label that says nothing about which build is resident.

## Verification

Verdict logic is a pure, unit-tested module rather than inline shell, because it only runs for real when a deploy is stuck — i.e. exactly when nobody is watching. **A watcher's failure mode isn't a crash; it's quietly reporting "aligned" forever.**

Driven through **the exact node invocation the workflow uses**, against the box's real state:

| scenario | result |
|---|---|
| real box: deployed `5.5.66`, released `v5.5.66` | `aligned`, no alert |
| behind, release 23m old | `deploying`, no alert |
| behind, release 300m old | `stuck`, **alert** |
| ahead of the latest release | `ahead`, **alert** |
| container unreadable | `unreadable`, no alert |

The aligned case also exercises the tag/package version split (`v5.5.66` vs `5.5.66`) — compared naively, that would make every correctly-deployed box look broken.

`npm test`: **146 pass, 0 fail** (29 new). `npm run preflight` clean.

### DIFF
--- a/.github/workflows/deployed-version-watch.yml
+++ b/.github/workflows/deployed-version-watch.yml
@@ -1,0 +1,159 @@
+name: Deployed version watch (self-hosted)
+
+# Is what's RUNNING what we last RELEASED?
+#
+# THE GAP. cc-drift-auto-release.yml's idempotency gate already proves the
+# artifacts exist — the git tag, the npm version, the ghcr image. Nothing
+# checked the last leg: that the box actually pulled the image and is running
+# it. Publishing is not deploying, and that leg has failed silently before
+# (dario's refresh-lock `lockId` code sat released-but-not-live on the CF
+# Worker while every status surface read green, because each was reporting on
+# a different link in the chain and no single link is the deployment).
+#
+# WHY THE HEALTH WATCHER CANNOT COVER THIS. cc-oauth-health.yml probes the
+# running container and asks "are you serving?". A container running a stale
+# version answers yes — truthfully. Healthy and current are different
+# questions and only one of them was being asked.
+#
+# DIVISION OF LABOUR. If the container is down or its version is unreadable,
+# this workflow stays SILENT: cc-oauth-health.yml owns container-down and
+# already files that alert with the right diagnostics. Two watchers filing
+# separate issues for one outage just doubles the noise on the worst day.
+#
+# The verdict itself lives in scripts/deploy-verdict.mjs and is unit-tested
+# (test/deploy-verdict.mjs), because this logic only runs for real when a
+# deploy is stuck — i.e. exactly when nobody is watching. A watcher's failure
+# mode is not a crash, it is quietly reporting "aligned" forever.
+
+on:
+  schedule:
+    # Hourly, offset off the :00 rush. A stuck deploy is not urgent to the
+    # minute — it is urgent that it is noticed at all rather than sitting for
+    # days.
+    - cron: '27 * * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: dario-deployed-version
+  cancel-in-progress: false
+
+jobs:
+  check:
+    permissions:
+      contents: read
+      issues: write
+    # Never from a fork: meaningless anywhere but this repository, and the
+    # forks inherit the schedule (see #1039).
+    if: github.repository == 'askalf/dario'
+    runs-on: [self-hosted, dario-drift]
+    timeout-minutes: 5
+    steps:
+      - uses: askalf/checkout-with-retry@115a6407547e9711edbc2e915838d495cad9583f  # v1.1.0
+
+      - name: Compare deployed version against the latest release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          # A release legitimately takes time to reach the box (multi-arch
+          # build, ghcr push, pull). Alerting the instant a tag appears would
+          # page on every single release, so this is what makes it a
+          # deploy-STUCK detector rather than a deploy-in-progress detector.
+          GRACE_MINUTES: '90'
+        run: |
+          set -uo pipefail
+
+          # The container's own package.json is the ground truth for what is
+          # RUNNING. Not the image tag: `:latest` is a moving label that says
+          # nothing about which build is actually resident.
+          deployed="$(docker exec askalf-dario node -e 'console.log(require("/app/package.json").version)' 2>/dev/null | tr -d '\r\n[:space:]' || true)"
+
+          released="$(gh release view --repo "$REPO" --json tagName --jq .tagName 2>/dev/null || true)"
+          published="$(gh release view --repo "$REPO" --json publishedAt --jq .publishedAt 2>/dev/null || true)"
+
+          age_ms=0
+          if [ -n "$published" ]; then
+            then_s="$(date -u -d "$published" +%s 2>/dev/null || echo 0)"
+            now_s="$(date -u +%s)"
+            if [ "$then_s" -gt 0 ]; then age_ms=$(( (now_s - then_s) * 1000 )); fi
+          fi
+
+          echo "deployed=${deployed:-<unreadable>} released=${released:-<unreadable>} release_age_min=$(( age_ms / 60000 ))"
+
+          verdict="$(DEPLOYED="$deployed" RELEASED="$released" AGE_MS="$age_ms" GRACE_MIN="$GRACE_MINUTES" \
+            node -e '
+              import("./scripts/deploy-verdict.mjs").then(({ deployVerdict }) => {
+                const r = deployVerdict({
+                  deployed: process.env.DEPLOYED,
+                  released: process.env.RELEASED,
+                  releaseAgeMs: Number(process.env.AGE_MS),
+                  graceMs: Number(process.env.GRACE_MIN) * 60000,
+                });
+                console.log(r.state);
+                console.log(r.alert ? "1" : "0");
+                console.log(r.summary);
+              }).catch((e) => { console.log("error"); console.log("0"); console.log(String(e && e.message)); });
+            ')"
+
+          state="$(printf '%s\n' "$verdict" | sed -n '1p')"
+          alert="$(printf '%s\n' "$verdict" | sed -n '2p')"
+          summary="$(printf '%s\n' "$verdict" | sed -n '3p')"
+          echo "verdict: $state — $summary"
+
+          gh label create dario-deploy-stuck --repo "$REPO" --color D93F0B \
+            --description "Released but not deployed on the box" --force >/dev/null 2>&1 || true
+          existing="$(gh issue list --repo "$REPO" --label dario-deploy-stuck --state open \
+            --json number --jq '.[0].number // empty' 2>/dev/null || true)"
+
+          if [ "$alert" != "1" ]; then
+            if [ -n "$existing" ]; then
+              gh issue close "$existing" --repo "$REPO" \
+                --comment "✅ Recovered — $summary ($(date -Iseconds))."
+              echo "closed #$existing (recovered)"
+            else
+              echo "OK — $state, no alert."
+            fi
+            exit 0
+          fi
+
+          if [ -n "$existing" ]; then
+            echo "$state — alert #$existing already open, leaving it."
+            exit 0
+          fi
+
+          image="$(docker inspect --format '{{.Config.Image}}' askalf-dario 2>/dev/null || echo '?')"
+          created="$(docker inspect --format '{{.Created}}' askalf-dario 2>/dev/null || echo '?')"
+
+          tmp="$(mktemp)"
+          {
+            echo "**The box is not running what was released.** $summary"
+            echo
+            echo "The release pipeline's own gate only proves the tag, the npm version and the ghcr image exist. It does not prove the box pulled them — and \`cc-oauth-health.yml\` cannot catch this either, because a container running a stale version still answers \"serving\" truthfully."
+            echo
+            echo "| field | value |"
+            echo "|---|---|"
+            echo "| state | \`$state\` |"
+            echo "| deployed | \`${deployed:-<unreadable>}\` |"
+            echo "| latest release | \`${released:-<unreadable>}\` |"
+            echo "| release age | $(( age_ms / 60000 )) min |"
+            echo "| image | \`$image\` |"
+            echo "| container created | \`$created\` |"
+            echo
+            echo "**Recover:** pull and recreate on the box —"
+            echo '```sh'
+            echo "docker pull $image"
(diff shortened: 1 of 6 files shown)


sprayberry-redline:
## dario/live-test — PASS

Head SHA: `8de96e0b659d41a1ad0f8fd96ede4ff933e7d263`

| Stage | Result | Note |
|---|---|---|
| install | ✅ PASS | `npm ci` clean, 6 packages added |
| build | ✅ PASS | `tsc && cp` clean |
| test | ✅ PASS | 146/146 tests passed, 0 failures |
| cli | ✅ PASS | `--help` exits 0 |
| doctor | ✅ PASS | exits 1 with `[FAIL] OAuth not authenticated` as expected (no credentials in sandbox) — reports cleanly, does not crash |
| live probe | ⏭️ SKIPPED | `DARIO_TEST_UPSTREAM_API_KEY` not set in tester environment |

Status: [`dario/live-test` success](https://github.com/askalf/dario/commit/8de96e0b659d41a1ad0f8fd96ede4ff933e7d263)

sprayberry-redline (changes requested):
_Automated review from the Sprayberry Labs fleet code reviewer._

**Verdict:** 1 correctness gap in the merge-conflict resolver's file allowlist; everything else (the new deploy-verdict watcher, preflight checks, and the token-refresh guard) reads solid and well-tested.

## Findings

### 1. [Correctness] `drift-pr-heal.yml`'s "outside the version files" filter was not updated for the new lockfile handling — `.github/workflows/drift-pr-heal.yml:198-199`

```yaml
```

This PR teaches `resolve-release-conflicts.mjs` to resolve `package-lock.json` conflicts too (new `resolvePackageLock`, registered in `HANDLERS` at `scripts/resolve-release-conflicts.mjs`), and both `git add` lines just above/below were updated to include `package-lock.json` (`drift-pr-heal.yml:179`, `:216`). But this `outside=` filter — which decides whether a merge conflict is "ours to resolve" before the resolver ever runs — was left matching only `package\.json|CHANGELOG\.md`.

Concrete scenario: a drift-heal run hits a merge where **only** `package-lock.json` is conflicted (the exact "release-collision, lockfile-only" case this PR was written to fix, per the CHANGELOG entry for 5.5.68/5.5.69). `git diff --name-only --diff-filter=U` reports `package-lock.json`; `grep -vxE 'package\.json|CHANGELOG\.md'` does **not** filter it out (regex doesn't anchor on `-x` exact-line-match against the literal `package.json`, and `package-lock.json` is a different string), so `outside` becomes non-empty, and the job takes the `REFUSED — conflicts outside the version files` branch and aborts the merge — even though the newly-added `resolvePackageLock` handler could have resolved it cleanly. The very capability this PR adds is unreachable from the auto-heal workflow that's supposed to use it.

```suggestion
              outside=$(git diff --name-only --diff-filter=U \
                        | grep -vxE 'package\.json|package-lock\.json|CHANGELOG\.md' || true)
```

## What's good
- `scripts/deploy-verdict.mjs` is a clean, pure, well-documented state machine with thorough unit coverage (`test/deploy-verdict.mjs`) including the tag/package version split.
- `scripts/preflight.mjs`'s CHANGELOG-comment handling explicitly avoids the incomplete-sanitization trap (index-based classification instead of stripping-then-rescanning) — good awareness of a real CodeQL-flagged footgun.
- `tokenRefreshDisabled`/`refreshDisabledError` guard is applied to both refresh paths (`src/oauth.ts`, `src/accounts.ts`), with a dedicated test (`test/no-token-refresh.mjs`) asserting both throw and that near-miss env values (`"true"`, `"0"`, empty) correctly leave refresh enabled.
- CI is fully green (14/14 checks, including the self-hosted compat job and CodeQL).

## What I checked
`gh pr view`/`gh pr diff` for the full diff, and a shallow clone to read `drift-pr-heal.yml` and `resolve-release-conflicts.mjs` in full context. Did not run the test suite locally — CI's own rollup is green (146 tests pass per the `dario/live-test` check).

askalf:
Addressed in 030c331.

`drift-pr-heal.yml`'s unmerged-path check now exempts the lockfile:

```
```

Without it a lockfile-only conflict — the exact case this PR adds `resolvePackageLock` for — landed in `$outside`, hit `REFUSED — conflicts outside the version files`, and aborted the merge before `resolve-release-conflicts.mjs` ever ran. The three exempt paths now match the resolver's `HANDLERS` keys exactly. Also rewrapped the stale comment at :186 that still named only two files.

askalf:
Same finding as #1108 — fixed in `c222803`, which is now the base of this branch.

The `outside=` guard now allows `package-lock.json`, so the resolver is actually reachable; `renumber-release-version.mjs` also mirrors the version into the lockfile's two slots (it wrote only `package.json` + `CHANGELOG.md` before, which `preflight` would have turned into a red check on every healed PR).

The durable fix is a test that reads the **actual grep pattern out of `drift-pr-heal.yml`** and asserts every key of `HANDLERS` survives it — verified to fail when the guard is reverted to its buggy form. Full rationale in the [#1108 reply](https://github.com/askalf/dario/pull/1108#issuecomment-5414548539).

sprayberry-redline (dismissed):
_Automated review from the Sprayberry Labs fleet code reviewer._

**Verdict:** Re-review at head `904be95` — the correctness gap from my prior CHANGES_REQUESTED is fixed exactly as suggested; no new blocking findings in the commits since. Approving.

## Prior finding — resolved

My previous review (on commit `8de96e0`) flagged that `.github/workflows/drift-pr-heal.yml`'s conflict allowlist regex wasn't updated to include `package-lock.json`, making the new `resolvePackageLock` handler unreachable from the auto-heal workflow. The fix landed exactly as suggested:

`.github/workflows/drift-pr-heal.yml:206`
```
```
The `git diff --quiet` gate and `git add` lines a few lines above were also extended to include `package-lock.json`, so the file is now committed, matched, and resolvable consistently.

## New commits since that review

The diff since `8de96e0` also adds lockfile-mirroring to `scripts/renumber-release-version.mjs` (`renumber()` now takes an optional `lockJsonString` and writes the derived version into both `.version` and `.packages[""].version`), with `test/renumber-release-version.mjs` covering: normal renumber, no-lockfile checkout, no-op case, and unrelated-field survival. This is straightforward derivation (not a guess) from `package.json`'s already-validated version, and is well covered.

Notably, `test/renumber-release-version.mjs` also adds a regression test that reads the actual `grep -vxE` pattern out of `drift-pr-heal.yml` and asserts every file in `resolve-release-conflicts.mjs`'s `HANDLERS` map passes it — directly binding the two lists so this exact class of bug (resolver knows a file, workflow allowlist doesn't) can't silently reoccur. Good defense-in-depth for a bug that had no other way to be caught (nothing in CI exercises `drift-pr-heal.yml` directly).

## What I checked
`gh pr view --json reviews` to find the prior CHANGES_REQUESTED commit (`8de96e0`) and current head (`904be95`); `git diff 8de96e0..904be95` for the full delta between reviews; read the fixed workflow section and the new lockfile-mirroring code and its tests in full. CI is green (12/12 relevant checks incl. CodeQL, test, build matrix). Did not re-run the test suite locally — trusting the green rollup.

sprayberry-redline (dismissed):
_Automated review from the Sprayberry Labs fleet code reviewer._

**Verdict:** No blocking findings — approving. Head `985a81b` has an identical file tree to `904be95caf1fd263224ab8d4a6860f6fe57aa62b` (verified via `git rev-parse <sha>^{tree}` on both — same tree hash), which this bot already reviewed and approved; this is a rebase, not new content.

## What I checked
- `gh pr diff` for the full unified diff (new `deployed-version-watch.yml`, `scripts/deploy-verdict.mjs` + tests, the `DARIO_NO_TOKEN_REFRESH` guard in `src/oauth.ts`/`src/accounts.ts`, and the canary hardening in `cc-billing-classifier-canary.yml`).
- `deployVerdict()`'s state machine (aligned/deploying/stuck/ahead/unreadable) against its own test suite — the grace-window boundary, the `v`-prefix/bare version comparison, and the "ahead" case all read correctly and are exercised by `test/deploy-verdict.mjs`.
- The refresh-guard (`tokenRefreshDisabled`/`refreshDisabledError`) is applied at the top of both `refreshTokens()` (`src/oauth.ts`) and `refreshAccountToken()` (`src/accounts.ts`) — the two paths a shared credential store can be refreshed from — so the borrow-don't-rotate invariant isn't reachable through only one of them.
- CI check-rollup: 14/14 green, including CodeQL, the self-hosted compat job, and the build matrix (node 18/20/22).
- Confirmed the current head's tree is byte-identical to a commit this bot already reviewed (see verdict above), so no re-diff of content was needed.

## What's good
- `scripts/deploy-verdict.mjs` is a small, pure, dependency-free state machine — easy to unit test and easy to trust precisely because it only matters when a deploy is silently stuck.
- The canary fix (`DARIO_NO_TOKEN_REFRESH=1`) directly addresses a real incident described in the PR body (~3h fleet outage from a refresh rotating a shared token) and fails loudly rather than silently producing a misleading billing alert.
- Division of labor between this new watcher and `cc-oauth-health.yml` (silent on unreadable/down container) is a deliberate, documented design choice, not an oversight.

No re-review of test execution was performed locally, per policy — trusting the green CI rollup.

sprayberry-redline:
## dario/live-test — PASS

Head: `985a81b9869c0290257f1449f97e0cf174378009`

| Stage | Result | Note |
|---|---|---|
| install | PASS | `npm ci` clean |
| build | PASS | `tsc && cp src/cc-template-data.json dist/` |
| test | PASS | 146/146 tests passed |
| doctor | PASS | exits 1 reporting `[FAIL] OAuth not authenticated` as expected with no credentials — did not crash |
| live probe | SKIPPED | `DARIO_TEST_UPSTREAM_API_KEY` not set in tester env |

No failures attributable to the PR.

sprayberry-redline (approved):
_Automated review from the Sprayberry Labs fleet code reviewer._

**Verdict:** No blocking findings — approving. (Note: a prior review on this PR was auto-dismissed by GitHub after new commits landed post-approval; this is a fresh review of the current head, `4b929fb`.)

## What I checked
- `gh pr diff` for the full change: new `.github/workflows/deployed-version-watch.yml`, `scripts/deploy-verdict.mjs` + `test/deploy-verdict.mjs`, and the version bump.
- `deployVerdict()`'s state machine (aligned/deploying/stuck/ahead/unreadable): the grace-window boundary (`releaseAgeMs < graceMs` vs `>=`) matches the "exact at window -> stuck" test; the `v`-prefix/bare-version comparison in `compareVersions` is exercised for both directions; `ahead` alerts regardless of grace as documented.
- Division of labor with `cc-oauth-health.yml`: `deploy-verdict.mjs` returns `alert:false` for `unreadable` on both empty `deployed` and empty `released`, so this workflow never double-files against the health watcher for a down/unreadable container.
- Duplicate-issue guard in the workflow: it looks up an existing open `dario-deploy-stuck` issue before filing (`gh issue list ... --json number --jq '.[0].number // empty'`) and closes it on recovery, so repeated hourly runs won't spam new issues — and `concurrency: { group: dario-deployed-version, cancel-in-progress: false }` prevents overlapping runs from racing on that check.
- CI check-rollup: 14/14 green (CodeQL, actionlint, build matrix 18/20/22, compat, test, docker-cap-drop-smoke, etc.), `auto-merge` correctly shows `skipping` (stacked PR, not yet mergeable).

## Minor (non-blocking)
`.github/workflows/deployed-version-watch.yml`, the `deployed=` line:
```
deployed="$(docker exec askalf-dario node -e 'console.log(require("/app/package.json").version)' 2>/dev/null | tr -d '\r\n[:space:]' || true)"
```
`tr -d` here isn't wrapped in an outer `[...]`, so `[:space:]` isn't treated as a POSIX character class — it's stripped as the literal characters `[`, `:`, `s`, `p`, `a`, `c`, `e`, `]`. In practice this is inert (npm version strings are digits and dots, so none of those literal characters ever appear, and `\r`/`\n` are still stripped correctly), but the intent ("strip whitespace") isn't what the code does. Not worth blocking on since it doesn't affect real version strings, but worth tightening to `tr -d '[:space:]'` (with the class properly bracketed) if this line is touched again.

## What's good
- `scripts/deploy-verdict.mjs` is a small, pure, dependency-free state machine — easy to unit test and exactly the kind of logic worth isolating since (per the PR's own framing) it only executes for real when a deploy is silently stuck.
- `test/deploy-verdict.mjs` covers the boundary case explicitly (`GRACE - 1` vs `GRACE`), the tag/bare-version split, and the "ahead" case regardless of freshness — good coverage for a component that's hard to exercise in production.
- The workflow is scoped to the base repo only (`if: github.repository == 'askalf/dario'`), preventing forks from inheriting the schedule.

No re-run of the test suite was performed locally, per policy — trusting the green CI rollup.

sprayberry-redline:
## dario/live-test — PASS

Head SHA: `4b929fb23b4eec11f00f2712db76f77f776e329f`

| Stage | Result | Note |
|---|---|---|
| install (`npm ci`) | ✅ pass | clean |
| build (`npm run build`) | ✅ pass | clean |
| test (`npm test`) | ✅ pass | 146/146 tests passed, 0 fail |
| cli (`--help`) | ✅ pass | exit 0 |
| doctor | ✅ pass | exit 1 as expected (no OAuth creds in test sandbox); reported `[FAIL] OAuth not authenticated`, did not crash |

No failures attributable to the PR.

PR #1111, opened 2026-08-25
fix(ci): pin the OAuth credential in wire-drift too (v5.5.70)

askalf:
## The gap

`wire-drift-self-hosted.yml` spawns `claude` on the box, and `claude` shares `/root/.claude/.credentials.json` with the platform dario container.

**Every other self-hosted workflow that runs CC re-pins that path to the container's canonical store first.** This one didn't:

| workflow | runs CC | re-pins first |
|---|---|---|
| `cc-billing-classifier-canary.yml` | ✅ | ✅ |
| `cc-drift-template-watch.yml` | ✅ | ✅ |
| `compat-test-self-hosted.yml` | ✅ | isolated `HOME` (#1106) |
| `wire-drift-self-hosted.yml` | ✅ | ❌ **this PR** |

A split credential family is the **2026-06-23 fleet outage**: two files carrying one refresh-token family rotate each other to death via Anthropic's reuse detection.

## It recurs

Not hypothetical, not a one-off. Found split again today — a **regular file at inode 290627** beside canonical **290583** — and `/root/.claude/` holds saved stubs from **2026-07-17, 07-18 and 08-02**. So roughly monthly, with the daily canary re-link as the only thing closing it.

## A detail that contradicts the runbook

The recovery notes assume CC's atomic refresh replaces the symlink with a real credential file. **It doesn't.** The clobbering file is an *emptied* credential:

```
accessTokenLen: 0,  refreshTokenLen: 0,  expiresAt: 0,  scopes: [...intact]
```

All four saved stubs share that exact shape. So something writes a **blank** credential on auth failure. No workflow in this repo writes that path, so the writer is still unidentified — worth a follow-up, but re-pinning before use is the cheap guard either way.

## Verification

- Workflow YAML parses; the new step passes `bash -n` (extracted and checked directly).
- `npm run preflight` clean, `npm run lint:pkg` OK, **146 tests pass**.
- The **live** split was repaired on the box separately: backed up first, and the script refuses to clobber if the standalone file is the fresher credential — it confirmed canonical was fresher (expiresAt), then relinked. Both paths now resolve to inode 290583.

### DIFF
--- a/.github/workflows/wire-drift-self-hosted.yml
+++ b/.github/workflows/wire-drift-self-hosted.yml
@@ -73,6 +73,35 @@ jobs:
           npm ci --no-audit --no-fund
           npm run build
 
+      - name: Pin OAuth credential to platform-dario canonical
+        # This job spawns `claude` on the box, and `claude` shares
+        # /root/.claude/.credentials.json with the platform dario container.
+        # Every OTHER self-hosted workflow that runs CC re-pins that path to the
+        # container's canonical store first; this one did not. Two files in one
+        # refresh-token family rotate each other to death via Anthropic's
+        # reuse detection — that is the 2026-06-23 fleet outage.
+        #
+        # The split is neither hypothetical nor a one-off: found again on
+        # 2026-08-25 (a REGULAR file at inode 290627 beside canonical 290583),
+        # alongside saved stubs from 2026-07-17, 07-18 and 08-02. Notably the
+        # clobbering file is an EMPTIED credential — accessToken "",
+        # refreshToken "", expiresAt 0, scopes intact — so something writes a
+        # blank credential on auth failure rather than CC refreshing over the
+        # link, which is what the runbook had assumed. Whatever the writer is,
+        # re-pinning before use is the cheap guard, and it is exactly what
+        # cc-billing-classifier-canary.yml and cc-drift-template-watch.yml
+        # already do.
+        run: |
+          CANON=/var/lib/askalf-dario/credentials.json
+          if [ -e "$CANON" ]; then
+            mkdir -p /root/.claude /root/.dario
+            ln -sfn "$CANON" /root/.claude/.credentials.json
+            ln -sfn "$CANON" /root/.dario/credentials.json
+            echo "pinned runner OAuth credential -> $CANON (inode $(stat -L -c %i "$CANON"))"
+          else
+            echo "::warning::canonical $CANON missing — platform dario down? proceeding as-is"
+          fi
+
       - name: Report claude version
         run: claude --version || echo "::warning::claude not on PATH — the wire-drift check will report a high-severity 'capture' finding"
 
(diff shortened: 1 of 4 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues — clean, well-scoped infra fix. Approving.

## What I checked
- `gh pr diff` for all four changed files (workflow YAML, CHANGELOG, package.json/lock).
- Compared the new step against the equivalent guard already shipped in `cc-billing-classifier-canary.yml` (cloned the repo to diff the two side-by-side).
- Confirmed `check-wire-drift.mjs` spawns `claude` with an isolated `HOME=cleanHome` tmpdir (scripts/check-wire-drift.mjs:63,111) — so the pinning step isn't needed for that spawn, but the added step runs *before* the separate `Report claude version` step, which does run `claude --version` against the real, unsandboxed `/root/.claude`, so the fix targets the right place.
- CI rollup (`statusCheckRollup`): all 15 checks SUCCESS/SKIPPED-as-expected — CI, CodeQL, compat, actionlint, build×3, test, docker-cap-drop-smoke, wire-drift itself. No red/pending checks.

## Findings
None blocking. The new step (`.github/workflows/wire-drift-self-hosted.yml:76-104`):
```yaml
run: |
  CANON=/var/lib/askalf-dario/credentials.json
  if [ -e "$CANON" ]; then
    mkdir -p /root/.claude /root/.dario
    ln -sfn "$CANON" /root/.claude/.credentials.json
    ln -sfn "$CANON" /root/.dario/credentials.json
    ...
  else
    echo "::warning::canonical $CANON missing — platform dario down? proceeding as-is"
  fi
```
matches the pattern already proven in `cc-billing-classifier-canary.yml` (same `ln -sfn` re-pin, same canonical path, same missing-canonical fallback), is idempotent, and fails soft (warns, doesn't block the job) if the canonical file is absent — reasonable given the platform-down case is out of this workflow's control.

## What's good
- Root-caused via inode comparison rather than assumption (290627 vs canonical 290583), and the CHANGELOG documents the correction to the runbook (blank credential on failure, not a refresh-over-symlink) — useful for the next person debugging this.
- Version bump is consistent across `package.json`, `package-lock.json`, and `CHANGELOG.md`.
- No test suite changes needed/expected for a self-hosted workflow YAML step; PR description notes `bash -n` validation on the extracted step, which is a reasonable substitute given this can't be unit tested.

sprayberry-redline:
## dario/live-test — PASS

Head SHA: `661a12f09238746ccfd424b9f187324aafea7d32`

| Stage | Result | Note |
|---|---|---|
| build (`npm run build`) | ✅ pass | tsc + asset copy clean |
| test (`npm test`) | ✅ pass | 146/146 passing, 0 fail/skipped |
| doctor | ✅ pass (expected FAIL row) | `[FAIL] OAuth not authenticated` is correct behavior with no creds in worktree HOME — doctor reported cleanly, did not crash |
| live probe | ⏭️ skipped | `DARIO_TEST_UPSTREAM_API_KEY` not set in this run |

No PR-attributable failures observed.

PR #1114, opened 2026-08-25
fix(bots): sync package-lock.json version when a drift bot bumps package.json

askalf:
## Problem

Every drift bot that patch-bumps `package.json` left `package-lock.json`'s two version slots behind. `scripts/preflight.mjs` — run by the **required** `validate-package-json` check — asserts they agree, so each bot PR failed a required check on a defect the bot itself introduced:

```
✗ package-lock.json: version 5.5.70 ≠ package.json 5.5.71
preflight: 1 finding — fix before pushing.
```

Seen on #1113 (`bot/cc-drift-v2.1.246`). The same gap exists in all three bump sites:

| script | used by |
|---|---|
| `scripts/auto-draft-drift-fix.mjs` | `cc-drift-watch.yml` (`bot/cc-drift-*`) |
| `scripts/rebake-release-prep.mjs` | `cc-drift-template-watch.yml` auto-rebake (`bot/template-rebake-*`) |
| `scripts/label-sync.mjs` | `cc-drift-template-watch.yml` label-sync (`bot/template-label-*`) |

## Change

- **`scripts/_drift-patch-helpers.mjs`** — new pure `syncLockfileVersion(lockJsonString, version)`: mirrors a version into the lockfile's root `.version` and `.packages[""].version`, returns canonical JSON text. Absent slots are left alone; bad input throws rather than silently no-op'ing.
- All three bump sites now call it immediately after `bumpPackageJsonPatch`.
- **`.github/workflows/cc-drift-template-watch.yml`** — both commit steps now `git add … package-lock.json …`. (`cc-drift-watch.yml` needs no workflow change: it stages from the drafter's `changedFiles`, which now includes the lockfile.)
- **`test/auto-draft-drift-fix.mjs`** — 13 new assertions covering both slots, dependency versions untouched, canonical formatting, guard rejections, absent-slot handling, and agreement with a `bumpPackageJsonPatch` result.

### Why not `npm install --package-lock-only`?

A version bump does not change the dependency graph — those two slots *mirror* `package.json` by definition. Writing the version into them is derivation, not a guess. This is the same reasoning `scripts/renumber-release-version.mjs` already spells out for its own lockfile write, and it keeps the bots free of an `npm install` step on the self-hosted runner.

## Test plan

- `node test/auto-draft-drift-fix.mjs` → **66 pass, 0 fail** (was 53).
[...]

PR #1113, opened 2026-08-25
chore(cc-drift): v5.5.71 — maxTested → v2.1.246

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.246 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.245` → `2.1.246` in `src/live-fingerprint.ts`
2. Bumps `package.json` version → `5.5.71`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.71] - 2026-08-25` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.246 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.245). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.245; npm latest is v2.1.246. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.246 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.246 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.246. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.246" step — no local run needed.)
- ✅ **Auto-merges** once the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`) **and the PR has an approving review** — the ruleset on `master` requires one, and dismisses it on any later push. Until then this PR sits open regardless of how green it is.
[...]

PR #1115, opened 2026-08-26
fix(compat): close the CC lag instead of reporting it (v5.5.72)

askalf:
Pairs with #1114. That one made drift bots sync `package-lock.json`; this one closes the *other* half of the same problem — and together they make a drift PR self-sufficient instead of needing a human twice.

## The loop this closes

`cc-drift-watch.yml` drafts a `maxTested` bump from the npm registry's `@latest`. The runner only moves to `@latest` once a day, when `cc-autoupdate.timer` fires (~04:58Z).

So every drift PR drafted between a version publishing and that tick claims a CC the runner cannot load. The `maxTested` gate correctly fails closed, and the PR sits red until someone upgrades the box by hand.

**That isn't a rare edge.** In one day it burned two operator approvals ("upgrade CC on the runner to 2.1.245", "…to 2.1.243") and blocked #1113 — and every resolution was the same one-line `npm i -g`. The gate was doing its job; nothing closed the loop it opened.

The step runs after build and before the capture, so the existing `force: true` capture re-reads the template from whatever CC ends up installed. That's the self-healing path #1014 already built, now with the claimed version actually present.

## Deliberate shape

**Upgrade only, never downgrade.** A PR that *lowers* `maxTested` must not drag this shared runner backwards — `cc-drift-template-watch` and the billing canary capture from the same binary, and pinning them to an older CC to satisfy one PR would corrupt their baselines. Verified numeric, not lexical (`2.1.9` vs `2.1.10`).

**The version is validated as a bare `X.Y.Z` before it reaches npm.** It comes from the PR's own source and this runner holds credentials. Fork PRs can't reach here — the job guard requires `head.repo.full_name == github.repository` — but that's one layer, not a reason to skip the second.

**Non-fatal by design.** If npm can't serve that version, it warns and continues; **the `maxTested` gate stays the authority** and still fails closed with its existing actionable message. This step can only ever *remove* a reason to fail, never invent one. That property is why the gate is untouched.

## Verification

| check | result |
|---|---|
| YAML parses | ✅ |
| run block `bash -n` | ✅ |
| upgrade/skip across 5 version pairs incl. lexical trap | ✅ |
| shape guard blocks `latest`, shell metacharacters, empty, trailing dot, prerelease | ✅ |
| node reader returns `maxTested` from a real dist build | ✅ |
| `npm test` | **146 pass** |
| `npm run preflight` / `lint:pkg` | clean |

Also worth noting: this branch hit a real three-file version conflict on rebase (`package.json`, `package-lock.json`, `CHANGELOG.md`) and `scripts/resolve-release-conflicts.mjs` from #1108 resolved all three cleanly — including the lockfile handler added there, which is the exact file that would have been left with live markers before. Both changelog entries survived and both lockfile slots agree.

### DIFF
--- a/.github/workflows/compat-test-self-hosted.yml
+++ b/.github/workflows/compat-test-self-hosted.yml
@@ -91,6 +91,61 @@ jobs:
           npm ci --no-audit --no-fund
           npm run build
 
+      - name: Bring the runner's CC up to the maxTested claim
+        # Close the lag INSTEAD of reporting it.
+        #
+        # cc-drift-watch.yml drafts a `maxTested` bump from the npm registry's
+        # @latest, but the runner only moves to @latest once a day when
+        # cc-autoupdate.timer fires (~04:58Z). Any drift PR drafted before that
+        # tick claims a CC version the runner cannot load, so the gate below
+        # fails closed and the PR sits red until a human upgrades the box by
+        # hand. That is not a rare edge: it burned two operator approvals and
+        # dario#1113 in a single day, and each time the resolution was to run
+        # the very command this step now runs.
+        #
+        # DELIBERATE SHAPE:
+        #   * UPGRADE ONLY, never downgrade. A PR that LOWERS maxTested must not
+        #     drag this shared runner backwards -- cc-drift-template-watch and
+        #     the billing canary capture from the same binary, and pinning them
+        #     to an older CC to satisfy one PR would corrupt their baselines.
+        #   * The version is validated as strict X.Y.Z before it reaches npm.
+        #     It comes from the PR's own source, and this runner is credentialed.
+        #     (Fork PRs cannot reach here -- the job-level guard requires
+        #     head.repo.full_name == github.repository -- but defence in depth.)
+        #   * NOT fatal on failure. If npm cannot serve that version, say so and
+        #     continue: the maxTested gate in the next step is still the
+        #     authority and will fail closed with its existing message. This
+        #     step only ever removes a reason to fail; it must never invent one.
+        #   * The capture in the next step is `force: true`, so it re-reads the
+        #     template from whatever CC is installed after this runs -- which is
+        #     exactly the self-healing path dario#1014 built.
+        run: |
+          set -uo pipefail
+          claimed="$(node -e 'import("./dist/live-fingerprint.js").then(m=>process.stdout.write(String(m.SUPPORTED_CC_RANGE.maxTested||"")))' 2>/dev/null || true)"
+          installed="$(claude --version 2>/dev/null | awk '{print $1}' || true)"
+          echo "claimed maxTested=${claimed:-<unreadable>} installed CC=${installed:-<none>}"
+
+          if [ -z "$claimed" ] || [ -z "$installed" ]; then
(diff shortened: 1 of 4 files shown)


sprayberry-redline:
## dario/live-test — PASS

Head SHA: `edeefed63ee863a6192c1504af3c89c010980649`

| Stage | Result | Note |
|---|---|---|
| install (`npm ci`) | ✅ PASS | clean, 6 packages added |
| build (`npm run build`) | ✅ PASS | tsc clean |
| test (`npm test`) | ✅ PASS | 146/146 pass, 0 fail |
| cli (`--help`) | ✅ PASS | exit 0 |
| doctor | ✅ PASS | no crash; correctly reports `[FAIL] OAuth not authenticated` with no creds present (expected) |
| live probe | ⏭️ SKIPPED | `DARIO_TEST_UPSTREAM_API_KEY` not set in tester environment |

Verdict: **PASS**

sprayberry-redline (approved):

## What I checked
- Full diff (`gh pr diff 1115`): the new `Bring the runner's CC up to the maxTested claim` step in `.github/workflows/compat-test-self-hosted.yml`, plus the mechanical `package.json`/`package-lock.json`/`CHANGELOG.md` version bump to 5.5.72.
- CI check-rollup: 14/14 checks pass (`build`, `test`, `compat`, `CodeQL`, `analyze`, `actionlint`, `docker-cap-drop-smoke`, etc.) — no failing or pending checks.
- Confirmed `SUPPORTED_CC_RANGE.maxTested` in `src/live-fingerprint.ts:1361` is a plain string literal, matching the diff's assumption that it's read from the built `dist/live-fingerprint.js`.
- Independently re-derived and tested the new shape-validation `case` statement (`*[!0-9.]*|""|.*|*.`) against `2.1.246`, `latest`, empty, `.2.3`, `2.3.`, `2.1.246-beta`, and a shell-metacharacter payload (`2;touch pwned`) — all the dangerous/malformed inputs are correctly rejected before the value ever reaches the quoted `npm i -g "@anthropic-ai/claude-code@${claimed}"` call.
- The job-level guard `github.event.pull_request.head.repo.full_name == github.repository` (pre-existing, `.github/workflows/compat-test-self-hosted.yml:72`) still gates fork PRs from reaching this self-hosted runner step at all, and the new step's own shape check is a second, independent layer on top of that.
- The upgrade-only comparison (`sort -V | tail -1` against `installed`) and the non-fatal fallback paths (`exit 0` on unreadable/malformed input, `::warning::` + continue on failed `npm i -g`) match the stated design: this step can only remove a reason for the existing `maxTested` gate to fail, never invent a new failure or silently swallow the gate's own authority.

## What's good
- Defense-in-depth reasoning is explicit and actually verified in code, not just asserted in the PR description (fork guard + input-shape guard + non-fatal fallback all independently checked out).
- `set -uo pipefail` (deliberately omitting `-e`) is the right choice given the step is designed to continue past soft failures.

## Minor (non-blocking)
- The bash step reimplements version comparison via `sort -V` rather than reusing the `compareVersions` helper already exported from the same `dist/live-fingerprint.js` this step already `node -e` imports from. Not a correctness issue (behavior is equivalent for well-formed `X.Y.Z` strings), just a small duplication of logic that already exists one file away.

PR #1118, opened 2026-08-26
fix(proxy): the #1033 tail-restore must deep-clone, not reference (v5.5.74)

### LINKED ISSUE #1117: subagent error 400
**What happened?**
 Error 400 when subagent completes
**Steps to reproduce:**
1. Ask claude code to launch a subagent and wait for its completion
2. Get 
" API Error: 400 messages.390: role 'system' must follow a 'user' message or an 'assistant' message ending in a server tool result;
  the directive-only form (content: [] with output_config) is accepted at any position"

**Environment:**
- OS: Ubuntu 24
- Node version: 24
- dario version: 5.5.72
- Tool using dario (e.g. OpenClaw, Aider): Claude code

askalf:
Fixes #1117.

## Reproduced live, twice — 400 on master, clean on this fix

Isolated proxy, isolated credential, a genuine background-subagent completion round-trip through real Anthropic traffic. Confirmed the exact failure on current master, then confirmed clean on this fix, identical scenario, same box:

| | |
|---|---|
| master | `claude exit=1` — `API Error: 400 messages.5.content.0.text: cache_control cannot be set for empty text blocks` |
| this fix | `claude exit=0` — "The subagent completed and returned exactly `ack`." |

## Root cause

`sanitizeMessages`' #1033 guard snapshots the tail message before scrubbing, so it can restore original content if the scrub would leave a request ending on an assistant turn (a prefill 400):

```ts
const tailContentBeforeScrub = tail ? tail.content : undefined;
```

`tail` is one of `messages`. The very next loop mutates every block's `.text` **in place**, tail included. Since `tailContentBeforeScrub` points at the *same* array and the *same* block objects, "restoring" it later restores nothing — it hands back the already-emptied blocks. The array's **length** survives unchanged (the guard's `tailHadContent` check is length-based, so it stays fooled), but the text is already gone, and any `cache_control` on that block rides along untouched.

A background-subagent completion turn whose entire text is an orchestration tag hits this exactly: the restored tail carries `{type:'text', text:'', cache_control:{type:'ephemeral'}}` — which Anthropic rejects. #1117's own report shows a *different* symptom from the identical defect ("role 'system' must follow a 'user' message…") — which error surfaces depends on what else sits in the corrupted position.

## Fix

```ts
const tailContentBeforeScrub = tail
  ? (Array.isArray(tail.content) ? structuredClone(tail.content) : tail.content)
  : undefined;
```

String content is untouched — `sanitizeContent` returns a new string rather than mutating in place, so it was never exposed to this. Node floor is `>=18`, so `structuredClone` needs no new dependency.

## Why the existing #1033 tests never caught it

They assert the message **survives** and the **role** is right. Neither checks that the restored **text** is non-empty. I proved this in-process before touching anything live — ran the exact #1033 fixture through the *unfixed* code and printed the restored block:

```
restored block text: ""
restored block cache_control: {"type":"ephemeral"}
```

Zero API cost, confirmed before any live traffic.

## Verification

- **9 new assertions** (`test/sanitize-messages.mjs`): the exact failing shape; a multi-block tail where only *one* block carries `cache_control` (proves the clone is per-block-faithful, not just array-shaped); the string path confirmed unaffected; an arbitrary extra field *and* a nested `cache_control` field both confirmed to survive the clone whole (proves `structuredClone` isn't narrowing to known fields).
- **Live, both directions** — see table above. Same repro script, same box, isolated `HOME` + isolated test credential throughout (never the shared credential).
- **146 tests pass** (9 new), `npm run preflight` clean, `npm run lint:pkg` OK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ checklist.
 
 ## [Unreleased]
 
-## [5.5.74] - 2026-08-27
+## [5.5.75] - 2026-08-27
 
 - **CC drift patch** — `SUPPORTED_CC_RANGE.maxTested` bumped `2.1.246` → `2.1.247` for CC v2.1.247. Auto-drafted by `cc-drift-watch.yml`. Template re-capture, if needed, is auto-handled by `cc-drift-template-watch.yml`.
 ## [5.5.73] - 2026-08-26
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.74",
+  "version": "5.5.75",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "5.5.74",
+      "version": "5.5.75",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "5.5.74",
+  "version": "5.5.75",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -679,8 +679,24 @@ export function sanitizeMessages(body: Record<string, unknown>, preserveTags?: S
   // drop below would expose an assistant turn, we put the original content
   // back rather than ship a prefill (dario#1033) — see the guard after the
   // filter for the reasoning.
+  //
+  // MUST be a deep clone when content is an array (dario#1117). `tail` is one
+  // of `messages`, and the per-block mutation loop just below rewrites each
+  // block's `.text` IN PLACE — so a bare `tail.content` reference here points
+  // at the SAME block objects the loop is about to scrub. "Restoring" it
+  // later then restores nothing: the array's length survives (the guard's
+  // `tailHadContent` check is length-based and stays fooled), but every
+  // block's text is already the post-scrub value, and any `cache_control` on
+  // that block rides along untouched. The restored tail then carries an empty
+  // text block WITH a live `cache_control` — a shape dario's own drop-filter
+  // exists to remove, and one Anthropic rejects outright with "cache_control
+  // cannot be set for empty text blocks", or (depending on what else sits in
+  // that position) a role-ordering 400 instead. Reproduced live against a
+  // background-subagent completion turn: dario#1117.
   const tail = messages.length > 0 ? messages[messages.length - 1] : undefined;
-  const tailContentBeforeScrub = tail ? tail.content : undefined;
+  const tailContentBeforeScrub = tail
+    ? (Array.isArray(tail.content) ? structuredClone(tail.content) : tail.content)
+    : undefined;
   for (const msg of messages) {
     if (typeof msg.content === 'string') {
       msg.content = sanitizeContent(msg.content, patterns);
--- a/test/sanitize-messages.mjs
+++ b/test/sanitize-messages.mjs
@@ -391,5 +391,107 @@ header('dario#1033 - the scrub must not leave a trailing assistant turn (prefill
   }
 }
 
+// -------------------------------------------------------------
+header('dario#1117 — the #1033 tail-restore must restore REAL text, not a shallow reference');
+{
+  // Reproduced live: a background-subagent completion round-trip, isolated
+  // proxy + isolated credential, real Anthropic traffic. Request #5 in that
+  // session ended:
+  //   [role=user, {type:'text', text:'', cache_control:{type:'ephemeral'}}]
+  // and Anthropic 400'd with "cache_control cannot be set for empty text
+  // blocks". A different transcript instead saw the *other* reported shape
+  // (dario#1117's own report): "role 'system' must follow a 'user' message
+  // …" — a second observable symptom of the identical defect, depending on
+  // which turn happens to sit in the corrupted position.
+  //
+  // Root cause: `tailContentBeforeScrub` was `tail.content` — a bare
+  // reference, not a copy. `tail` is one of `messages`, and the per-block
+  // mutation loop rewrites every block's `.text` IN PLACE, including tail's.
+  // So by the time the #1033 guard "restores" `tailContentBeforeScrub`, it is
+  // restoring the SAME objects the loop just emptied — the array's length
+  // survives (fooling the length-based `tailHadContent` check) but the text
+  // is already gone, and any `cache_control` rides along untouched. The
+  // restored tail ends up EXACTLY as invalid as the state #1033 was written
+  // to prevent, just via a different route.
+  const lastRole = (b) => b.messages[b.messages.length - 1]?.role;
+
+  {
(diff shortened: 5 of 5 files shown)


sprayberry-redline (dismissed):

**Verdict:** No blocking issues found — clean, well-targeted fix with solid regression coverage.

## What I checked
- `gh pr diff` for the full change (`src/proxy.ts`, `test/sanitize-messages.mjs`, `CHANGELOG.md`, version bumps).
- Read the pre-fix `sanitizeMessages` in `src/proxy.ts` (base branch) to confirm the described bug is real:
  ```ts
  const tail = messages.length > 0 ? messages[messages.length - 1] : undefined;
  const tailContentBeforeScrub = tail ? tail.content : undefined;
  for (const msg of messages) {
    ...
      for (const block of msg.content) {
        if (... 'text' in block ...) (block as {text:string}).text = sanitizeContent(...);
      }
  ```
  `tailContentBeforeScrub` is a bare reference to the same array/block objects the loop mutates in place, so the later `tail.content = tailContentBeforeScrub` restore (in the #1033 guard) hands back already-scrubbed blocks — length-based `tailHadContent` stays fooled, but text and any `cache_control` are the post-scrub values. This matches the PR's root-cause writeup exactly.
- The fix in `src/proxy.ts:687-688`:
  ```ts
  const tailContentBeforeScrub = tail
    ? (Array.isArray(tail.content) ? structuredClone(tail.content) : tail.content)
    : undefined;
  ```
  correctly breaks the aliasing only for the array-content case (string content is already immune since `sanitizeContent` returns a new string). `structuredClone` is available since Node 17, and `package.json` pins `engines.node >= 18.0.0`, so no new dependency or compat risk.
- New tests in `test/sanitize-messages.mjs` directly exercise the failure mode: single-block cache_control survival, per-block-faithful cloning in a multi-block tail, the unaffected string-content path, and preservation of nested/arbitrary fields (`ttl`, `custom_id`) through the clone — these are exactly the cases that would have caught the original bug.
- CI rollup (`gh pr checks 1118`): all checks pass (CodeQL, build ×3, compat, test, wire-drift, docker-cap-drop-smoke, etc.); `auto-merge` shows "skipping" (not a failure).

## What's good
- Root cause is precisely identified and the fix is minimal and scoped to exactly the aliasing bug described — no incidental refactoring.
- Test additions target the actual defect (text + cache_control pairing) rather than re-asserting the pre-existing #1033 guarantees, and include an adversarial multi-block case to prove the clone isn't just array-shaped.

No correctness, reuse, efficiency, or coverage concerns in the diff.

sprayberry-redline:
## dario/live-test — PASS

Head SHA: `9372a744fe14f8b97c32b9ea59cb3923e8b92a4d`

| Stage | Result | Note |
|---|---|---|
| install (`npm ci`) | PASS | clean |
| build (`npm run build`) | PASS | tsc clean |
| test (`npm test`) | PASS | 146/146 passing, 0 failures |
| cli (`--help`) | PASS | exit 0 |
| doctor | PASS | correctly reports `[FAIL] OAuth not authenticated` (no creds in test env — expected, not a crash) |
| live probe | PASS | proxy started, `/v1/messages` with `claude-haiku-4-5` returned exact `pong` response over passthrough |

Status: https://github.com/askalf/dario/commit/9372a744fe14f8b97c32b9ea59cb3923e8b92a4d/checks

sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean, well-targeted fix with solid regression coverage.

## What I checked
- `gh pr diff` for the full change (`src/proxy.ts`, `test/sanitize-messages.mjs`, version bumps).
- The surrounding `sanitizeMessages` function in the cloned branch to confirm how `tailContentBeforeScrub` is consumed at the restore site (`src/proxy.ts:756-769`).
- CI check-rollup: all 13 required checks green (`test`, `build (18/20/22)`, `compat`, `CodeQL`, `wire-drift`, etc.), `auto-merge` skipping (expected pre-approval).

## Analysis
The change at `src/proxy.ts:696-700`:
```ts
const tailContentBeforeScrub = tail
  ? (Array.isArray(tail.content) ? structuredClone(tail.content) : tail.content)
  : undefined;
```
correctly fixes the aliasing bug: previously `tailContentBeforeScrub` was a bare reference into the same `messages` array that the very next loop mutates block-by-block in place, so the "restore" at line 769 (`tail.content = tailContentBeforeScrub`) was restoring already-scrubbed objects. `structuredClone` on the array-content path breaks that aliasing while leaving the string-content path (already immutable via `sanitizeContent` returning a new string) untouched.

I traced the restore path at lines 756-769: `tailHadContent` is computed from the now-independent clone, and `tail.content = tailContentBeforeScrub` assigns the pre-scrub clone back, so cache_control and any extra per-block fields survive intact — consistent with the new tests.

## Test coverage
`test/sanitize-messages.mjs` adds 4 well-targeted cases: the exact failure shape (single block with cache_control), a multi-block tail (proving the clone is per-block, not just array-shaped), the string-content path (confirming no regression), and an arbitrary-extra-field + nested cache_control field case (confirming `structuredClone` isn't narrowing to known keys). This directly covers the gap the PR identifies in the original #1033 tests (which only checked message survival/role, not restored text content).

## What's good
- Root-cause explanation in the PR body and code comment both correctly identify the aliasing issue and are consistent with what the diff actually does.
- Fix is minimal and scoped — no unrelated refactoring.
- Live reproduction described (before/after) in addition to unit tests.

sprayberry-redline (approved):

**Verdict:** Clean fix — approving. No blocking issues found.

## What I checked
- `src/proxy.ts:679-684`: the actual fix — `tailContentBeforeScrub` changed from a bare reference (`tail.content`) to `Array.isArray(tail.content) ? structuredClone(tail.content) : tail.content`. This correctly breaks aliasing with the array the mutation loop below rewrites in place, so the #1033 restore path can no longer hand back already-scrubbed blocks. Traced the failure mode described (in-place `.text` mutation on the same array reference the snapshot pointed to, length-based `tailHadContent` check staying fooled while `cache_control` rides along on an emptied block) and it matches the diff exactly — the fix addresses it directly.
- `test/sanitize-messages.mjs`: new assertions cover the exact failing shape, a multi-block tail with `cache_control` on only one block (verifies per-block fidelity, not just array identity), the string-content path (confirms no regression there, since `sanitizeContent` already returns a new string), and a block with nested/extra fields (`ttl`, `custom_id`) surviving the clone — good coverage of the fix's actual boundary.
- `CHANGELOG.md`/`package.json`/`package-lock.json`: mechanical version bump (5.5.74 → 5.5.75), consistent across all three files.
- CI check-rollup: all checks pass (test, build 18/20/22, compat, CodeQL, docker-cap-drop-smoke, wire-drift, etc.) — 13/13 green, no pending checks.

## What's good
Root-cause writeup and in-code comment are unusually precise and match the diff exactly; the fix is minimal (one line, gated on `Array.isArray` so the string path is untouched) and the new tests target the specific gap (restored *text*, not just message survival/role) that let the original #1033 tests miss this.

No findings to raise — nothing to block on.

sprayberry-redline:
## dario/live-test — PASS

**Head SHA:** `561fcbc325b741223aecd9888040b3428ba29de4`

| Stage | Result | Note |
|---|---|---|
| build (`npm run build`) | ✅ pass | tsc clean |
| test (`npm test`) | ✅ pass | 146/146 tests passed, 0 failures |
| doctor | ✅ pass | reports `[FAIL] OAuth not authenticated` as expected (no creds in sandbox) — did not crash |
| live probe | ✅ PASS | `POST /v1/messages` (haiku-4-5) → HTTP 200, body contains `pong` |

**Verdict: PASS** — all stages clean, live probe confirmed working proxy round-trip against upstream.

PR #1124, opened 2026-08-28
chore(cc-drift): v5.5.77 — maxTested → v2.1.250

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.250 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.247` → `2.1.250` in `src/live-fingerprint.ts`
2. Bumps `package.json` + `package-lock.json` version slots → `5.5.77`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [5.5.77] - 2026-08-28` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.250 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.247). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.247; npm latest is v2.1.250. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.250 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.250 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.250. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.250" step — no local run needed.)
- ✅ **Auto-merges** once the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`) **and the PR has an approving review** — the ruleset on `master` requires one, and dismisses it on any later push. Until then this PR sits open regardless of how green it is.
[...]

PR #1121, opened 2026-08-27
chore(template): label refresh _version -> v2.1.247 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-08-27T09:54:25+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.247** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.247`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.247`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v5.5.74` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.247. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #1125, opened 2026-08-28
ci: post dario/live-test from a self-hosted workflow instead of the fleet agent

askalf:
## Why

`dario/live-test` is a required status on `master`. Until now it was posted by a fleet LLM agent (the **Dario PR Tester**): a host timer files a ticket per PR head, the agent runs ~30 tool-call iterations of install/build/test/cli/doctor/probe and posts the status. Every one of those stages is a shell command; the latency was the fleet's, not the work's.

Measured wait for the status on the last two merged PRs:

| PR | opened | `dario/live-test` posted | wait |
|---|---|---|---|
| #1121 | 09:54Z | 13:36Z | **3h 42m** |
| #1124 | 01:19Z | 02:21Z | **62m** (and only after a wedged ticket was cleared by hand) |

CI for the same PRs finished in ≤72s. The gate was ~50× slower than everything it gated, at ~$0.40 of Max subscription per PR.

## What

`.github/workflows/live-test.yml` — the agent's procedure verbatim (its `run()` scrubbed-env wrapper, stage order, timeouts, live-probe request and verdict rules), run deterministically on the `dario-drift` runner compat already uses, posting the same `dario/live-test` commit status plus one PR comment (updated in place, same marker pattern as compat). `pending` is posted at the start so the gate shows the run is live.

Boundaries carried over unchanged:
- **Never executes fork code** on the self-hosted runner — job `if` excludes cross-repo PRs, and the dispatch path re-checks `isCrossRepository` and refuses. Fork PRs get no status from this job, exactly as before.
- **Never touches the runner's credentials** — every stage runs under `env -i` with a throwaway `HOME` and `DARIO_IGNORE_CC_CREDENTIALS=1`; the live probe's key is passed in env only, never argv.
- Live probe uses `ANTHROPIC_COMPAT_API_KEY` (the secret compat already holds); absent → the stage is SKIPPED, not failed. A 401 from upstream is reported as **ERROR** (tooling) rather than FAIL (the PR), as the agent did.

Expected: status in ~2–3 minutes of a push, $0.

## Self-validating

This PR triggers the workflow on itself. If `dario/live-test: success` appears on this head, it was posted by the new job — that is the proof the required check works end to end.

## Follow-up (platform)

`tools/gh-ops-sweep.sh` item 8 still files `Live-test` tickets to the agent for every dario PR head. Once this is merged that becomes duplicate work; the follow-up gates item 8 on the head having no `dario/live-test` status yet, so the agent only ever acts as a fallback.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean replacement of the agent-driven `dario/live-test` gate with a deterministic self-hosted workflow.

## What I checked
- `gh pr diff 1125` — the full new file `.github/workflows/live-test.yml` (265 lines, single addition, no other files touched).
- `gh pr checks 1125` — all checks green, including `dario/live-test` itself (self-validating: this PR's own head shows `pass … All stages PASS: install/build/test(146/146)/cli/doctor/live-probe(pong)`), which is direct proof the new job runs end-to-end correctly.
- No prior review from this bot existed on this PR/head.

## Findings
None blocking.

**Minor (non-blocking):**
- `.github/workflows/live-test.yml:64-68` — fork-PR protection is implemented twice (the job `if:` condition and a re-check in the "Resolve the PR head" step via `isCrossRepository`), which is explicitly intentional ("Belt and braces for the dispatch path, which bypasses the `if` above") — not a defect, just noting it's deliberate defense-in-depth rather than redundancy to trim.
- Action pins use full 40-char commit SHAs with a trailing version comment (`askalf/checkout-with-retry@115a64...  # v1.1.0`, `actions/setup-node@820762...  # v7.0.0`) — good practice for supply-chain safety; I did not verify the comments match the actual tagged commit, but since the pin itself is an immutable SHA this has no security implication either way.

## What's good
- Careful credential isolation: every stage runs under `env -i` with a scrubbed `HOME`/`TMPDIR` and `DARIO_IGNORE_CC_CREDENTIALS=1`, with an inline comment explaining exactly why (avoids touching the runner's real OAuth credentials — a documented prior incident, dario#1081).
- The live-probe stage passes its test API key via `env` only, never `argv`, and uses a distinct dummy `x-api-key: dario` for the local proxy call vs. the real upstream key — matches the PR description's stated boundary.
- `doctor` verdict correctly treats an expected non-zero exit (no credentials) as PASS by inspecting output content rather than exit code alone, with a comment calling out that the old `| tail -30` agent behavior was masking this.
- Bash idiom `local out; out="$(...)"; local rc=$?` correctly avoids the classic "local swallows $?" pitfall by splitting declaration from assignment.
- `state=pending` posted up front so the required check doesn't look un-run while the job executes; final verdict step guarded by `if: always() && steps.pr.outputs.sha != ''` so partial failures still post a status instead of leaving the PR stuck pending.

sprayberry-redline:
## dario/live-test report — PR #1125 @ `4953b32`

| Stage | Result | Note |
|---|---|---|
| install (`npm ci`) | ✅ PASS | clean, 6 packages added |
| build (`npm run build`) | ✅ PASS | `tsc && cp src/cc-template-data.json dist/` clean |
| test (`npm test`) | ✅ PASS | 146/146 tests passed, 0 failed |
| cli (`--help`) | ✅ PASS | exit 0 |
| doctor | ✅ PASS | reported `[FAIL] OAuth not authenticated` as expected with no creds — no crash, exit 1 by design |
| live probe | ✅ PASS | `DARIO_TEST_UPSTREAM_API_KEY` was set; proxy started, HTTP 200, response text `pong` matched exactly |

Head SHA: `4953b321c1a76a810de2e1f2dd7046bceddc3146`

Verdict: **PASS**

PR #1126, opened 2026-08-28
ci(compat): pace live calls at 5s instead of 20s — the upstream here is an API key

askalf:
`test/compat.mjs` sleeps `PACE_MS` between each of its ~10 live calls. The default is **20s** when routed through dario (vs 500ms direct), which exists to keep a *subscription-backed* passthrough inside the 5h rate window.

The self-hosted compat job starts the proxy with `ANTHROPIC_UPSTREAM_API_KEY` — an API-key upstream with no such window — so it was paying that protection for nothing. Measured on run [33132511802](https://github.com/askalf/dario/actions/runs/33132511802) (2026-08-28): **215s in "Run compat tests", ~200s of it asleep**, holding the only `dario-drift` runner — and everything else queued on it (live-test, template drift) — for the duration.

Sets `DARIO_COMPAT_PACE_MS: '5000'` on the job. Calls stay paced; the runner hold drops from ~4.3 min to ~1 min. No change to what is asserted.

Not a required check, so this is about runner contention rather than merge latency. This PR touches the workflow file, which is in the job's `paths` filter, so it exercises the new pacing on itself.


sprayberry-redline (approved):

**Verdict:** No blocking issues — approving. One informational note on the added comment's rationale.

## What I checked
- `gh pr diff` — single-file, 9-line addition to `.github/workflows/compat-test-self-hosted.yml` (a new job-level env var `DARIO_COMPAT_PACE_MS: '5000'` plus an explanatory comment). No other files touched.
- `test/compat.mjs` in the target repo for how `DARIO_COMPAT_PACE_MS` is consumed (`PACE_MS = parseInt(process.env.DARIO_COMPAT_PACE_MS ?? (VIA_API_KEY ? '500' : '20000'), 10)`), confirming this is an existing, documented override point ("Overridable via DARIO_COMPAT_PACE_MS ... against a higher-tier key"), not a new mechanism.
- CI check-rollup: all checks green, including `compat` (SUCCESS) — and since this workflow file is in the job's own `paths` filter, the compat job actually ran *with* the new 5s pacing on this PR's own head commit, which is direct empirical evidence the change works end-to-end.

## Findings
None blocking.

**Minor / informational** — `.github/workflows/compat-test-self-hosted.yml:81-89` (new comment):
```
+      # test/compat.mjs sleeps PACE_MS between each of its ~10 live calls. Its
+      # default of 20s exists to keep a SUBSCRIPTION-backed passthrough under
+      # the 5h rate window; here the proxy is started with an API-key upstream
+      # (ANTHROPIC_UPSTREAM_API_KEY below), which has no such window.
```
`test/compat.mjs`'s own comments describe the 20s default differently: it exists because this run routes through the **per-token API pool** (the same `ANTHROPIC_UPSTREAM_API_KEY`/`ANTHROPIC_COMPAT_API_KEY` used here), which "has real per-minute / per-token rate limits; the 20s default keeps the 10-test burst comfortably under them (and is gentle on a freshly minted key)" — and explicitly disclaims the "~3/min Max cap" framing as a misdiagnosis. So the new comment's "no such window" framing conflates the Max/OAuth 5h subscription window (genuinely irrelevant here) with the per-token pool's own rate limits (which *do* apply to this exact upstream and aren't addressed by the new comment). This doesn't affect correctness of the shipped config — the override point is the same one `compat.mjs` already documents, and CI passed with 5s pacing on this run — but a future reader tightening `DARIO_COMPAT_PACE_MS` further based on this comment alone could reintroduce the exact per-token throttling `compat.mjs` warns about. Worth a follow-up comment tweak, not a blocker.

## What's good
- Minimal, single-purpose change (env var + comment only), no logic touched.
- Uses the pacing override `test/compat.mjs` already documents rather than inventing a new one.
- Self-validating: this PR's own diff falls inside the workflow's `paths` filter, so the compat job exercised the new pacing on itself and passed.

sprayberry-redline:
## dario/live-test — PASS

**Head SHA:** `ed73a8b4166fbced76af88f0e50d80fe1d739294`

| Stage | Result | Note |
|---|---|---|
| install | PASS | `npm ci` clean (6 packages added, 1 informational allow-scripts warning) |
| build | PASS | `tsc && cp` clean |
| test | PASS | 146/146 tests passed, 0 failed |
| doctor | PASS | reports `[FAIL] OAuth not authenticated` as expected with no credentials — did not crash |
| live probe | PASS | proxy up, POST `/v1/messages` → HTTP 200, response text `pong` |

No failures. Full run: install → build → test → cli → doctor → live probe, all green.

sprayberry-redline:
**dario/live-test** — head `200970d117dc9ac8f27b7a86dc185a9e41473672`

| Stage | Result | Note |
|---|---|---|
| install | ✅ PASS | `npm ci` clean, 6 packages added |
| build | ✅ PASS | `tsc` clean |
| test | ✅ PASS | 146/146 passing (55.0s) |
| cli | ✅ PASS | `--help` OK |
| doctor | ✅ PASS | reports `[FAIL] OAuth not authenticated` as expected (no creds provided), no crash |
| live probe | ✅ PASS | proxy up, `/v1/messages` → 200, response text `pong` |

Verdict: **PASS**

askalf:
Release gate: this PR has no `package.json` version bump, so `cc-drift-auto-release.yml`'s `gate` step would find the existing `v5.5.78` tag and emit `proceed=false` — merging as-is ships nothing (no build, no npm publish, no GHCR push).

Opened #1128 against this branch with the bump (5.5.78 → 5.5.79 + package-lock + CHANGELOG entry, no source/workflow changes). Merge #1128 into `ci/compat-pace` first, then merge this PR to master and the release chain will run.

Source ticket: 00MTCVI3VRD1CE6CADEC80D7E3

PR #1128, opened 2026-08-28
chore(release): v5.5.79 — add the version bump #1126 needs to actually ship

askalf:
Adds the release-gate data field missing from #1126.

## Why

#1126 changes `.github/workflows/compat-test-self-hosted.yml` only. `cc-drift-auto-release.yml`'s `gate` step short-circuits when a `v<version>` tag already exists — with `package.json` still at `5.5.78` and `v5.5.78` already tagged, merging #1126 would emit `proceed=false` and skip build / npm publish / GHCR push. The merge would be a no-op for release purposes. CI on #1126 is green; this is a missing data field, not a test failure.

## What

- `package.json`: `5.5.78` → `5.5.79`
- `package-lock.json`: root + self-entry version to match (same two-line shape as prior release commits, e.g. 9fc6708)
- `CHANGELOG.md`: new `## [5.5.79] - 2026-08-28` section under `[Unreleased]`, describing the compat-pace change from #1126

No source or workflow files touched — this is purely the release trigger.

## Base

Targets `ci/compat-pace` (not `master`) so the bump becomes part of #1126 rather than a competing release. Merge this first, then merge #1126.

## Test plan

- CI on this branch should be green (no source change).
- After merge into `ci/compat-pace`, #1126's diff vs master should show `package.json` at `5.5.79`.
- On merge of #1126 to master, `Auto release on version bump` should report `proceed=true` (no `v5.5.79` tag exists) and run the build → tag → release → npm publish → GHCR push chain.

Source ticket: 00MTCVI3VRD1CE6CADEC80D7E3

PR #1129, opened 2026-08-28
feat(proxy): forward a genuine Claude Code client's conversation untouched (v5.5.80)

askalf:
## What

`sanitizeMessages` — the orchestration-tag scrub — no longer runs on requests from a genuine Claude Code client. Their conversation is forwarded exactly as it arrived.

## Why

The scrub exists to make **other** harnesses (Aider, Cursor, OpenClaw, …) look like CC by removing their orchestration wrappers. CC's own `<system-reminder>`, `<task-notification>` and `<env>` blocks *are* the CC wire shape. Stripping them meant every request through dario was less faithful than the same request sent direct — and the empty blocks and emptied turns the scrub leaves behind were the origin of a whole class of defects: #54, #744, #1033, #1092, #1117. Each needed its own guard, and each guard patched the previous one's hole (see #1118, and the companion PR for #1117's second report).

The genuine-CC branch of `buildCCRequest` already declares "messages: untouched — the client is the authority on its own wire shape". This makes that true from the first byte.

## How

One early return at the top of `sanitizeMessages`, using the detection `buildCCRequest` already uses for its byte-faithful branch: billing header in `system[0]`, CC opener in `system[1]`. `proxy.ts` already imports from `cc-template.ts` (not the reverse), so no cycle.

What still runs for genuine CC, unchanged: the empty-text filter and the #1092 retry-artifact rewind in `buildCCRequest` — those handle CC's *own* malformed shapes, not scrub leftovers.

What is unchanged: non-CC clients (scrubbed exactly as before, #1033 tail-restore included) and `--preserve-orchestration-tags`.

## Tests

`test/sanitize-messages.mjs` +7: genuine CC → messages byte-identical (mid-conversation reminder survives, tag-only tail keeps its `cache_control`); the same conversation from a non-CC client → scrubbed and tail-restored as before; billing header without a CC opener → still scrubbed; explicit preserve set → still honoured. 66/66 in the file, 146/146 suite.

## Release

Bumps to **v5.5.80** (master took 5.5.79 in #1128 while this was open). The companion #1117 fix, #1130, is rebased on top of this branch as 5.5.81 and merges after it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,27 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.80] - 2026-08-28
+
+### Changed
+
+- **A genuine Claude Code client's conversation is forwarded untouched.**
+  `sanitizeMessages` — the orchestration-tag scrub that makes other harnesses
+  (Aider, Cursor, OpenClaw, ...) look like CC — no longer runs on requests
+  from Claude Code itself. CC's own `<system-reminder>`, `<task-notification>`
+  and `<env>` blocks are the CC wire shape; stripping them made every request
+  through dario less faithful than the same request sent direct, and the empty
+  blocks and emptied turns the scrub left behind were the origin of #54, #744,
+  #1033, #1092 and #1117, each needing its own guard. Detection is the check
+  `buildCCRequest` already uses for its byte-faithful branch (billing header
+  in `system[0]`, CC opener in `system[1]`); the empty-content filters that
+  genuine CC still needs (#1092's retry artifact) live there and keep running.
+  Non-CC clients and `--preserve-orchestration-tags` are unchanged.
+
+
+
+- **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.250` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.250 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
+
 ## [5.5.79] - 2026-08-28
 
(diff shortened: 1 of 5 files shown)


sprayberry-redline (approved):

## What I checked
- `gh pr diff 1129` (147 lines across CHANGELOG.md, package.json/package-lock.json, src/proxy.ts, test/sanitize-messages.mjs).
- Read `isGenuineCCClient` (src/cc-template.ts:1667-1676) and its existing use in `buildCCRequest` (src/cc-template.ts:1774) via a shallow clone, to confirm the detector `sanitizeMessages` now calls is the same one `buildCCRequest` already trusted for its "messages: untouched" branch — it is, so this closes a real gap rather than introducing a new trust boundary: previously `sanitizeMessages` scrubbed a genuine-CC request's messages *before* `buildCCRequest` ran its own genuine-CC branch, silently contradicting that branch's "untouched" comment. This change makes the two paths agree.
- CI (`gh pr checks 1129`): still running at review time (build/test/compat/live-test/wire-drift pending; validate-package-json, actionlint, label, triage already passing). `mergeable_state` currently reports `dirty`/`CONFLICTING` against `master` (PR is 2 commits behind, 1 ahead) — this is a git-state/rebase note, not a defect in the diff, but will need a rebase before merge; flagging so it isn't missed.

## Correctness
`src/proxy.ts:677-693` — the added early return:
```
if (isGenuineCCClient(body)) return;
```
placed right after the `if (!messages) return;` guard, before the scrub/tail-snapshot logic. Traced the call site (`src/proxy.ts:2831`, `sanitizeMessages(parsed, opts.preserveOrchestrationTags)`) — `parsed` is the client's raw request body, matching what `isGenuineCCClient` expects (checks `body.system[0]`/`[1]`). No mutation of `messages` happens before the new check, so genuine-CC requests now skip scrubbing entirely and non-CC requests fall through to the unchanged path below. Consistent with the PR description and no logic error found.

## Test coverage
`test/sanitize-messages.mjs` adds cases for: byte-identical genuine-CC messages (including a mid-conversation reminder and a tag-only tail's `cache_control`), the same conversation scrubbed for a non-CC client, a billing-header-only body (still scrubbed, confirming system[1] is required), and `--preserve-orchestration-tags` still working for non-CC clients. Good coverage of the new branch and its boundary (billing header alone isn't sufficient).

## What's good
Comment at proxy.ts:677 clearly explains the detection reuse and cross-references the defect chain (#54, #744, #1033, #1092, #1117) this consolidates. Version bump and CHANGELOG are consistent with the code change.

## Not blocking, for awareness
- CI was still mid-run when this review was submitted (build/test/compat jobs pending). Treating the already-green checks (validate-package-json, actionlint, triage) plus the diff read as sufficient given the review budget; a failing test/build job surfacing after this review would need a follow-up look.
- Branch currently shows `mergeable_state: dirty` vs `master` — will need a rebase, unrelated to code correctness.

PR #1130, opened 2026-08-28
fix(proxy): an empty final user turn must reach the #1092 rewind, not become a prefill 400 (v5.5.81)

askalf:
Fixes #1117 — the **second** report (Monitor "stream ended" → `400 This model does not support assistant message prefill`). The first report was fixed in #1118.

## Root cause — found in code, reproduced in-process

CC's stream-interruption retry appends a user turn with `content: []` as the final message. That is exactly the shape the #1092 rewind (#1093) recovers on the genuine-CC path — but the request reaches `buildCCRequest` **after** `sanitizeMessages` (`proxy.ts`, the call just after the body parse), and there:

1. the message-level drop (#744) removes the empty final user turn;
2. the #1033 tail-restore skips it — `tailHadContent` was false, so "only restore a tail that had content" let it go.

`buildCCRequest` therefore saw a request already ending on the assistant's real reply, the #1092 rewind had nothing to rewind, and upstream answered with the prefill rejection. The #1092 tests call `buildCCRequest` directly and could not see the order.

Timeline: the drop + guard landed in #1034 (08-20); the rewind in #1093 (08-25) was written for a turn the pipeline never delivered to it. Not a regression from #1118.

## Fix

Remove the `tailHadContent` condition from the #1033 guard: an empty trailing user turn is restored too. Restoring an empty tail is never worse than dropping it —

- genuine CC: `buildCCRequest` rewinds the pair as designed (#1092);
- general path: upstream names the malformed turn honestly ("content must contain at least one block") — the #1033 decision, unchanged — instead of dario converting it into a misleading prefill 400.

## Verification

New chained test in `test/empty-text-blocks.mjs` — `sanitizeMessages` → `buildCCRequest` on the Monitor shape (tool_use → tool_result "stream ended" → assistant reply → `content: []`):

| | |
|---|---|
| master | **3/5 fail** — scrub drops the tail, request ends on the assistant turn, no rewind |
| this fix | 5/5 — tail kept, genuine CC recognised, request rewinds to the Monitor `tool_result`, no empty turn survives |

Existing suites unchanged: `sanitize-messages` 59/59, `empty-text-blocks` 39/39, 146/146 overall.

## Release

Bumps to **v5.5.81**, stacked on #1129 (v5.5.80) so the two CHANGELOG entries don't collide — merge #1129 first and this becomes a one-commit diff. Both fixes stand on their own: #1129 skips the scrub for genuine CC, this one fixes the guard for every other client.

PR #1131, opened 2026-08-28
ci(live-test): drop the hand-posted dario/live-test status — the job's check-run is the gate

askalf:
## What

`live-test.yml` stops hand-posting the `dario/live-test` commit status. The job's own `live-test` check-run is the gate.

## Why

Every PR showed the self-hosted live test **twice** on the checks list: once as the check-run GitHub creates for the job, once as the `dario/live-test` status the job posted via the API at start (`pending`) and end (verdict). The status only existed because the master ruleset required that exact context — inherited from the fleet-agent era (#1125 kept it so the ruleset would not need editing).

The job already ends with `[ "$state" = "success" ]`, so its check-run fails on any non-PASS verdict — the status carried no information the check-run didn't.

## Done alongside

The master ruleset (`master-protection`) now requires the `live-test` check-run (GitHub Actions) instead of the `dario/live-test` status — switched **before** this PR so nothing is ever gated on a status no longer posted. #1129 and #1130 already carry a passing `live-test` check-run.

## Diff

- header comment: what the ruleset requires now, and why the status went
- the `Mark dario/live-test pending` step: removed
- the verdict step: status POST removed; PR comment and the exit-status line unchanged
- no `statuses:` permission was ever declared (the token came from `gh`), nothing else to drop

No version bump: workflow-only, same as #1125.

### DIFF
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -1,6 +1,8 @@
 name: Live test (self-hosted)
 
-# Posts the `dario/live-test` commit status the master ruleset requires.
+# The `live-test` check-run this job produces is what the master ruleset
+# requires — the job's exit status IS the verdict (see the last line of the
+# verdict step). It also posts one PR comment with the stage report.
 #
 # Until 2026-08-28 that status came from a fleet LLM agent (the "Dario PR
 # Tester"): a host timer filed a ticket per PR head, an agent picked it up,
@@ -10,10 +12,13 @@ name: Live test (self-hosted)
 # and 62m on #1124 (the latter only after a stall was cleared by hand), for
 # ~$0.40 of subscription per PR. The stages below are that agent's procedure
 # verbatim, run deterministically on the same runner compat already uses,
-# posting the same status and one PR comment, in minutes and for nothing.
+# posting one PR comment, in minutes and for nothing.
 #
-# The status is set to `pending` first so the merge gate shows the run is
-# live rather than looking un-run for the whole job.
+# Until 2026-08-28 (later that day) the job ALSO hand-posted a
+# `dario/live-test` commit status — the agent's context, kept so the ruleset
+# would not need editing. That put the same run on the checks list twice
+# (its own check-run and the status). The ruleset now requires the
+# check-run, and the status is gone.
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues — clean workflow simplification, approving.

**What I checked**
- Full diff (`.github/workflows/live-test.yml`, +11/-18, single file) — removes the `Mark dario/live-test pending` step and the `dario/live-test` status POST in the verdict step, replacing the header comment to explain why.
- `gh search code "dario/live-test" --repo askalf/dario` — the only remaining references are the header comment describing history and the (still-present in default branch, pre-merge) step names; no other workflow or script in the repo depends on the `dario/live-test` status context, consistent with the PR body's claim that the master ruleset was switched to require the `live-test` check-run instead.
- CI check-rollup (`gh pr checks 1131`): all 14 checks pass, including `live-test` itself (1m15s) — the job runs clean on its own new logic.

**Findings:** none. The job's exit relies on `[ "$state" = "success" ]` failing the step (and thus the check-run) on any non-PASS verdict (`live-test.yml` verdict step, unchanged logic aside from the removed POST), so the check-run genuinely carries the same information the status used to. The removed lines are a clean subtraction with no dangling references.

**What's good:** the header comment is rewritten to explain the historical context and why the status is gone rather than left stale — useful for the next reader. Small, single-purpose diff scoped tightly to the stated goal.

sprayberry-redline:
## dario/live-test report

**Head SHA:** `438a888e641d2575b40fde96a0301b13d75d9979`

| Stage | Result | Note |
|---|---|---|
| install (`npm ci`) | PASS | clean, 6 packages added |
| build (`npm run build`) | PASS | tsc + asset copy clean |
| test (`npm test`) | PASS | 146/146 passed, 0 failed |
| cli (`--help`) | PASS | exits 0 |
| doctor | PASS | reports `[FAIL] OAuth not authenticated` as expected (no creds provided), does not crash |
| live probe | PASS | proxy started, `POST /v1/messages` → HTTP 200, response text `pong` |

**Verdict: PASS**

askalf:
⚠️ **Merging this PR as-is will ship nothing.** Opened #1132 to fix it.

`package.json` on this branch is v5.5.81 — identical to `master`. `cc-drift-auto-release.yml`'s `gate` step short-circuits when a `v<version>` tag already exists (`v5.5.81` was cut by #1130), so the release pipeline skips and the workflow change never reaches npm/GHCR. Exactly what happened to #1125 on 2026-08-28, which needed a separate follow-up (#1128) to release.

#1132 bumps to **v5.5.82** and is based on **this branch**, so the bump becomes part of this PR's diff rather than consuming the next version slot ahead of it.

**Merge order:**
1. #1132 → `ci/live-test-drop-duplicate-status`
2. this PR → `master` (the merge commit then carries v5.5.82 and the release fires)

Source ticket: 01M14NV6GESR7KHCX7CTAAXV2V

PR #1010, opened 2026-08-18
feat: Codex/ChatGPT-subscription account scaffolding — the "altman" engine (#1009)

askalf:
Scoping/scaffolding for #1009 — not a finished feature, and not wired into request routing.

`src/codex-oauth.ts` + `src/codex-accounts.ts` add a fully isolated OAuth account pool for Codex CLI's ChatGPT Plus/Pro subscription flow (`dario codex add/list/remove`), structurally parallel to the Claude engine but deliberately **not** a shared abstraction with it — two providers isn't enough to know what a good generalization looks like yet, and this shouldn't touch the mature, battle-tested Claude path.

**This is gated on one open question**, same root cause as everything #993 built for Claude: does OpenAI invalidate the previous `refresh_token` on every refresh? No public source documents this — every existing third-party OAuth client for this flow (including the one this code's constants come from) is single-instance, single-user, and has never had a reason to find out. If OpenAI rotates like Anthropic does, this needs the same pool/lock architecture #993/#1008 built. If it's more forgiving, it may need much less.

`test/manual/codex-refresh-race.mjs` answers that directly — not part of `npm test`, needs a real ChatGPT Plus/Pro login. Fires two concurrent refreshes against the same token and reports which of the two outcomes happened.

40 new unit tests (`test/codex-oauth.mjs`, `test/codex-accounts.mjs`) cover everything that doesn't need live credentials: PKCE, authorize-URL construction, token-response parsing/error handling (monkeypatched fetch, no real network), and storage round-trip (isolated temp `$HOME`).

Holding this open rather than merging immediately — next step is running the live race test once a subscription is available, which may change the shape of what comes after this.

PR #1137, opened 2026-08-29
feat(codex): wire the ChatGPT-subscription engine into request routing

askalf:
Stacked on #1010 (`codex-altman-scaffold`) — **base is `codex-altman-scaffold`, not `main`**, so #1010 stays reviewable on its own. Both stay drafts; the operator un-drafts and merges bottom-up.

Ticket: DEV-2ab264f0. This PR wires the scaffolded ChatGPT-subscription ("altman") engine into request routing: a `/v1/chat/completions` request naming a model that account may use is served from the stored ChatGPT Plus/Pro subscription instead of an API key.

## What routes where

Routing lives in `provider-adapter.ts` — the existing seam, extended with a third `ProviderId`. No new decision point.

| Request | Provider |
|---|---|
| OpenAI-shape path, model is a slug the Codex backend **lists** for the stored account | `codex` |
| OpenAI-shape path, model carries `codex:` / `chatgpt:` prefix (stripped before forwarding) | `codex` |
| OpenAI-shape path, `isOpenAIModel(model)` and a backend is configured | `openai` (unchanged) |
| Anything else, incl. all `/v1/messages` | `claude` (unchanged) |

`codexAdapter` has priority 200, above `openaiAdapter` (100), so a listed slug reaches the subscription even when an API-key backend is also configured. Anything the backend does **not** list — `gpt-4o` and friends — is untouched and still lands on that backend. An explicit `claude:` / `openai:` prefix wins over codex.

Anthropic-shape `/v1/messages` is deliberately never claimed by codex: the ChatGPT backend speaks Responses and the translation here is written against the OpenAI request shape. A messages⇄Responses translation would be a second, separate piece of work.

## Model discovery (no hardcoded slugs)

The first cut hardcoded `gpt-5-codex` / `codex-*` patterns. Live testing showed **every** such name 400s with `The '<model>' model is not supported when using Codex with a <subscription> account.` — the routable set is per-account and moves. So:

- `getCodexModelSlugs(creds)` does `GET ${CODEX_BACKEND_BASE_URL}/models?client_version=<v>` with the account's own headers and keeps only `visibility=list` slugs, cached 10 min per account (`clearCodexModelCache()` for tests).
- **`client_version` is a REQUIRED backend query param** — the request is rejected without it. It is the `CODEX_CLIENT_VERSION` constant (currently `0.152.0`, overridable via `DARIO_CODEX_CLIENT_VERSION`) precisely so bumping it stays a one-line change when the backend starts gating on a newer CLI.
- Discovery never throws; on failure the slug list is empty and routing simply falls through to the existing providers.
- Those slugs are appended to dario's `/v1/models` so an OpenAI-compatible client's model picker can discover them. `/v1/models` keeps its "always answers" property.

## Translation

`codex-backend.ts` owns chat/completions ⇄ Responses in both directions:
`chatCompletionsToResponses()` (messages incl. `system`, `tools` → Responses function tools, assistant `tool_calls` → `function_call`, `tool` role → `function_call_output`) and `createResponsesTranslator()` (Responses SSE → `chat.completion.chunk` deltas incl. tool-call argument deltas and `finish_reason`, or a single `chat.completion` with `usage` when non-streaming).

Auth per `buildCodexHeaders()`: `Authorization: Bearer <access_token>` + `ChatGPT-Account-ID: <id>`, the id read from the `https://api.openai.com/auth`.`chatgpt_account_id` claim of the stored `id_token`. Endpoint, headers and claim were read out of the `openai/codex` CLI source rather than guessed (file references are in the `codex-backend.ts` header comment).

## Refresh

Deliberately **not** the pool/lock/lease architecture of #993/#1008. `test/manual/codex-refresh-race.mjs` was run twice against a live ChatGPT Plus account: two concurrent refreshes of the same `refresh_token` both succeed and each returns a *different* new `refresh_token` — OpenAI does not invalidate the previous one. So this needs only: refresh when `expiresAt` is inside the buffer, persist whatever comes back, and collapse concurrent refreshes **within one process** with an in-memory promise per alias. Nothing cross-process.

## `dario codex add`

Now accepts the full redirect URL pasted straight from the browser address bar (`http://localhost:1455/auth/callback?code=…&state=…`) as well as bare `code` / `code#state`. The manual-paste flow has nothing listening on 1455, so the address bar is the only place a user can get the code — the first live login failed exactly here. Same parse applied in `test/manual/codex-refresh-race.mjs`. Prompt and CLI help updated to say so.

## Live verification (operator, dev box, real Plus subscription)

Not runnable in CI — no subscription credentials there. All passed:

- Auth accepted at `CODEX_BACKEND_BASE_URL/responses`; no User-Agent / originator gate observed.
- Non-streaming `chat.completion` with `usage`.
- Streaming `chat.completion.chunk` SSE.
- Tool call → `finish_reason=tool_calls` with well-formed `arguments`.
- Tool-result round trip (`function_call_output`) → final answer.
- Slugs this account lists today: `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`. `gpt-5.6-sol` and `gpt-5.6-terra` both answered.

To re-run after merge, with an account added via `dario codex add <name>` and the proxy on 3456:

```
curl -s localhost:3456/v1/models | jq -r '.data[].id'
curl -s localhost:3456/v1/chat/completions -H 'content-type: application/json' \
  -d '{"model":"gpt-5.6-sol","messages":[{"role":"user","content":"say ok"}]}'
curl -Ns localhost:3456/v1/chat/completions -H 'content-type: application/json' \
  -d '{"model":"gpt-5.6-sol","stream":true,"messages":[{"role":"user","content":"count to 3"}]}'
```

## Tests

Unit tests use an injected/monkeypatched `fetch` — no network, same pattern as `test/codex-oauth.mjs`. No live model-loop tests in `npm test`.

## Unknowns / left for follow-up

- **`/v1/responses` inbound is not implemented.** dario's inbound remains chat/completions only. A passthrough would be cheap for the non-streaming case but the streaming and error paths need their own handling, so it is deliberately out of scope here.
- **`client_version` gating is a guess about the future.** `0.152.0` is accepted today; there is no signal about when or whether the backend starts requiring a newer one. The constant + env override is the mitigation, not a fix.
- **Slug cache TTL is 10 min, picked not measured.** If the backend rotates slugs faster, a request can 400 until the cache expires.
- **Rate-limit / quota headers from the ChatGPT backend are not surfaced** in dario's usage reporting (`dario doctor --usage` knows nothing about codex accounts). Separate piece of work.
- **Single stored account only.** `selectCodexAccount()` picks one; there is no rotation or failover across multiple codex accounts, by design for this PR.


askalf:
Branch complete — all four scope items are pushed (8 commits): backend + routing + slug discovery + `/v1/models`, the redirect-URL paste in `dario codex add`, `test/codex-backend.mjs`, and the README/CHANGELOG section.

**CI note — `npm test` does not run on this PR, by design of the workflow triggers, not by anything this branch does.** `ci.yml` is `on: pull_request: branches: [master]`, and this PR targets `codex-altman-scaffold`. What did run and pass here: `compat` (self-hosted), `wire-drift`, `actionlint`, `label`, `triage`. `live-test` and `auto-merge` skip on a draft. So the new `test/codex-backend.mjs` assertions are **not yet machine-verified** — they will run automatically once #1010 lands on master and this PR retargets, or locally with `npm run build && node test/codex-backend.mjs`. It is picked up by `test/all.test.mjs` with no registration needed (the driver auto-discovers `test/*.mjs`), and imports from `dist/` like its sibling `test/codex-oauth.mjs`, so it needs a build first.

Both the test file and the manual harness are syntax-checked (`node --check`) but not executed in this run — no `dist/` in the agent container.

`grep -ri cursor` over every file this branch created or modified returns nothing. The remaining hits elsewhere in `README.md` / `CHANGELOG.md` are pre-existing release history and the dario#190 BYOK notes in `src/proxy.ts`, none of which this branch touches; rewriting them would be editing unrelated shipped history.

One editorial change outside the new text: the previous `[Unreleased]` entry said the scaffolding was "NOT wired into request routing yet", which this PR falsifies. Reworded to "when it landed, not yet wired" so the section isn't self-contradictory when both entries ship in the same release.

askalf:
### Follow-up push: tsc narrowing fix + the tests for the routing change

Head is now `4f72dd6`. Three commits on top of the scaffold: `fc46bc8` (the fix), `30825eb` (tsc + tests), `4f72dd6` (test hermeticity).

**The routing fix was live-found and live-verified, not inferred.** The operator built the branch on the dev box against a real ChatGPT Plus account and ran `dario proxy --port=18458 --passthrough` with **no `--no-claude-auth` and no Claude login at all**:

| probe | result |
| --- | --- |
| bare slug `gpt-5.6-sol` | 200, `PIPE OK` |
| prefixed `chatgpt:gpt-5.5` | 200 |
| streaming `gpt-5.6-terra` | `chat.completion.chunk` frames |
| `claude-sonnet-5` | clean 503 `No account configured` |
| `GET /v1/models` | codex slugs tagged `owned_by: "openai"` |

Before `fc46bc8` all three chat probes returned 503 — `pool.select()` and the empty-pool 503 at `src/proxy.ts` ~L2612 ran *before* the provider-decision block at ~L2771, so a user whose only credential is a ChatGPT subscription died at a guard for a credential they do not have. The unit tests could not see it: routing in `provider-adapter.ts` and translation in `codex-backend.ts` were both already correct in isolation; only the *ordering inside the request handler* was wrong. That is why the new test drives a real proxy rather than the pure helpers.

**What changed since `fc46bc8`:**

1. **`src/proxy.ts:2620` — TS2339 `Property 'alias' does not exist on type 'never'`.** `let poolAccount: PoolAccount | null = null;` is only ever assigned inside the `selectPoolAccount()` closure, so control-flow analysis narrowed the annotated declaration to `null` for the entire rest of the handler and `poolAccount?.alias` (L2997 and ~15 other sites) stopped compiling. Now declared `let poolAccount = null as PoolAccount | null;`, which gives the declared type without seeding a narrowing. `accessToken` is a plain `string`, not a union, so it does not narrow and is unchanged.

2. **`test/codex-route-empty-pool.mjs` (new)** — the regression lock the ticket asked for. Hermetic: `HOME` at a `mkdtemp`'d dir holding one stored codex account, `DARIO_CODEX_BASE_URL` pointed at a local stub serving both `/models` discovery and the `/responses` SSE stream, and `fetchImpl` wired to a thrower so any request that reaches the *Claude* upstream fails the test loudly rather than touching the network. Asserts: empty pool + stored account + listed slug → 200 from codex with the stored token (not the 503); `chatgpt:` prefix routes and is stripped before forwarding; streaming returns translated deltas and `[DONE]`; a `claude-*` model **still** gets the empty-pool 503 (the guard moved, it did not disappear); `/v1/models` tags codex slugs `openai` and Claude models `anthropic`.

3. **`test/no-claude-auth-flag.mjs`** — `requiresClaudeLogin` cases for the new `hasCodexAccount` parameter, including that omitting the argument keeps the pre-#1137 answer.

**Open item for review, stated rather than chased:** this PR's base is `codex-altman-scaffold`, and `.github/workflows/ci.yml` triggers on `pull_request: branches: [master]` — so `validate-package-json`/`build`/`test` do not run here. The green rollup on this PR is actionlint + compat + wire-drift + label + triage only. **Neither the tsc fix nor the two test files have been executed by CI on this branch**; they will first be exercised when the stack retargets `master`. Locally there is no `node_modules` and no `dist/`, and the test imports `../dist/proxy.js`, so it cannot be run in the agent container either. `node scripts/check-readme-line-count.mjs` does pass (`~26k vs actual 26,798`, +798 of 1000 tolerance) — worth knowing since that job gates `build`/`test` and a trip there shows up as SKIPPED, not FAILURE.

Next probe if anything is off: retarget to `master` (or push the stack) and read the `test` job log for `ok N - codex-route-empty-pool.mjs`.

Source ticket: DEV-7a98e650.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — this is a well-scoped, well-tested change; approving.

## What I checked
- Full diff (`gh pr diff 1137`) across all 14 changed files: routing seam (`provider-adapter.ts`), new translation module (`codex-backend.ts`), credential storage/refresh (`codex-accounts.ts`), and the `proxy.ts` request-handler wiring (deferred pool-account selection).
- Surrounding context via a shallow clone of `codex-altman-routing` for `proxy.ts` and `codex-accounts.ts` to confirm control flow around the new `selectPoolAccount()` closure and the codex-credential-refresh block.
- CI rollup (`gh pr view --json statusCheckRollup`): all 13 checks green (build ×3, test, CodeQL, compat, wire-drift, actionlint, etc.); `live-test` and `auto-merge` show `SKIPPED` (expected — no live subscription credentials in CI, matches the PR body's own note).

## What's good
- The core bug this PR exists to fix — an empty Claude pool 503-ing before the codex adapter ever got a chance to claim the request — is fixed at the right layer (`selectPoolAccount()` deferred past the routing block in `proxy.ts:2870`) and locked down with a dedicated hermetic integration test (`test/codex-route-empty-pool.mjs`) that fails loudly (`noNetwork` fetch stub) if the regression reappears, rather than silently passing.
- Model routing is discovery-based rather than hardcoded slug patterns (`getCodexModelSlugs`/`isCodexModel`), with a documented, tested fallback (empty slugs → falls through to existing providers) when the backend is unreachable — avoids hard-coding a moving target.
- `chatCompletionsToResponses`/`createResponsesTranslator` (`src/codex-backend.ts`) have solid unit coverage including the tool-call round trip, per-request-instance isolation (explicitly guards against the interleaved-stream state bug referenced from #642), and junk/malformed-event handling.
- Secrets handling looks correct: `buildCodexHeaders` never puts `refreshToken` on the wire (asserted in `test/codex-backend.mjs`), and `extractChatGPTAccountId` only decodes the JWT payload for a routing header — no signature-trust claim is made anywhere.

## Minor (non-blocking)
- `src/provider-adapter.ts:83-84` — `codexAdapter.claimsPrimary` returns `ctx.forcedProvider === 'codex' || isCodexModel(...)`, gated on `ctx.hasCodexAccount`. If a client sends an explicit `codex:`/`chatgpt:` prefix but no codex account is actually configured (`hasCodexAccount` false), the codex adapter declines and — because `openaiAdapter` also declines on `forcedProvider === 'codex'` (`provider-adapter.ts:100`) — the request silently falls through to the Claude adapter with a `codex:`-stripped model name (e.g. `gpt-5.6-sol`) that Claude doesn't understand, rather than a clear "no codex account configured" error. Low impact (only hit by a user who explicitly opts into a provider they haven't set up), and not something this PR needs to block on.

No correctness bugs, security holes, or missing test coverage found in the changed lines.

Issue #1138, opened 2026-08-29
codex accounts added at runtime are not seen until the proxy restarts


askalf:
Shipped in v5.5.84. Live-found on the Hetzner deploy 2026-08-29.

**Symptom.** The proxy started 14:47Z with zero codex accounts. `dario codex add fleet` stored the account at 14:49Z. Until `docker restart`, the running proxy advertised no gpt slugs on `/v1/models` and routed `gpt-5.6-sol` down the Claude path, where it 404s as an unknown model.

**Cause.** `src/proxy.ts` computes `hasCodexAccount` once at startup (~L1549). The per-request codex block (~L2822) only ever flips it FALSE — when the accounts disappear — never TRUE when they appear. The Claude pool has the same startup-only shape, but `dario login` restarts the proxy by convention; `dario codex add` does not, and its help text does not say to.

**Fix.** Re-check `listCodexAccountAliases()` per request instead of trusting a startup flag. It is a readdir of a directory holding a handful of files; cache the NEGATIVE answer ~30s so an idle proxy with no codex account is not stat-ing per request, and leave the positive answer uncached so `codex remove` still takes effect immediately. Same for the `/v1/models` path. Plus a line in `dario codex add` output saying a running proxy picks the account up on its own.

askalf:
Fixed and released in **v5.5.85**; the box is on 6.0.3, so it has been running this for a day.

Implemented as prescribed: routing and `/v1/models` both re-ask `hasAnyCodexAccount()` per request, only the NEGATIVE answer is cached (30s) so an idle proxy with no codex account isn't stat-ing per request, and the positive answer is uncached so `codex remove` still takes effect immediately. `saveCodexAccount` also clears the negative cache, so an account stored in-process is never hidden by an entry taken moments earlier — that one wasn't in the report and closes the last gap. The startup probe stays, now only deciding what to print and whether a Claude login is required to boot. `dario codex add` ends with "no restart needed (dario#1138)."

`test/codex-runtime-account-detect.mjs` is the reported symptom end to end, on one process:

```
booted with no codex account → no gpt slugs advertised
  the codex slug is not advertised
  the codex backend was not consulted
account added mid-run → next request routes to codex, no restart
  the slug is now advertised on the SAME process
  the request is served, not 404/503 off the Claude path
  it reached the codex backend
```

Plus the cache semantics directly: added out-of-process stays absent inside the TTL, present past it with no restart, removed goes false immediately.

Closing.

PR #1139, opened 2026-08-29
fix(codex): detect accounts added while the proxy is running

askalf:
Addresses #1138.

## What

`src/proxy.ts` resolved "is a codex account available" once at startup and only ever flipped it FALSE. An account stored by `dario codex add` against an already-running proxy therefore stayed invisible until a restart — `/v1/models` advertised no gpt slugs and a listed slug fell through to the Claude path as an unknown model. Live-found on the Hetzner deploy 2026-08-29 (proxy up 14:47Z, account added 14:49Z, only fixed by `docker restart`).

Routing and `/v1/models` now ask a new `hasAnyCodexAccount()` in `codex-accounts.ts` per request:

- it is a `readdir` of a directory holding a handful of files;
- the **absent** answer is cached ~30s, so an idle proxy with no codex account does not touch the filesystem per request;
- the **present** answer is not cached, so `codex remove` still takes effect on the next request (this preserves the behaviour of the `else { hasCodexAccount = false }` branch that is removed);
- an in-process `saveCodexAccount` clears the negative entry, so a save cannot be hidden by a cache taken moments earlier.

The startup probe stays, but only to print the banner and to answer `requiresClaudeLogin()` — the boot-time question of whether a Claude login is required is genuinely a startup question.

`dario codex add` now prints that a running proxy picks the account up on its own, which is the thing its help text never said.

## Test plan

New `test/codex-runtime-account-detect.mjs` (auto-discovered by `test/all.test.mjs`), hermetic — HOME in a mkdtemp, `DARIO_CODEX_BASE_URL` at a local stub, `noClaudeAuth`, `fetchImpl` wired to a thrower so any leak to the Claude upstream fails loudly:

1. `hasAnyCodexAccount` unit behaviour on an injected clock — absent, account appears out-of-process, still absent inside the 30s TTL, present after it, absent again immediately after `removeCodexAccount`, present again after an in-process `saveCodexAccount`.
2. The regression itself: a real loopback proxy booted with **zero** codex accounts advertises no gpt slug and does not consult the backend; the account file then appears mid-run and the very next `/v1/models` advertises the slug (`owned_by: openai`) while `/v1/chat/completions` is served 200 by the codex stub — same process, no restart.

Existing `test/codex-route-empty-pool.mjs` covers the startup-with-an-account path and is unchanged.

Source ticket: DEV-1b3997c2.


sprayberry-redline (approved):

**Verdict:** No blocking issues found -- clean fix, well-tested, approving.

## What I checked
- Full diff (gh pr diff 1139, 326 lines across src/cli.ts, src/codex-accounts.ts, src/proxy.ts, test/codex-runtime-account-detect.mjs).
- Cloned the PR branch and confirmed there are no leftover references to the old module-level 'let hasCodexAccount' -- every call site in src/proxy.ts now goes through the new hasAnyCodexAccount() (lines 1553, 2505, 2823) or the one-time startupCodexAliases array used only for the boot banner and requiresClaudeLogin() (lines 1553, 1745).
- CI rollup: all checks green (build, test, compat, live-test, wire-drift, CodeQL, actionlint, etc.) as of the latest commit 0521181.

## Correctness trace
- src/codex-accounts.ts:64-89 (hasAnyCodexAccount / saveCodexAccount): the negative cache (codexAbsentUntil) is only set when present === false, never cached when true, and saveCodexAccount unconditionally resets it -- so an in-process add can't be hidden behind a stale negative entry, matching the PR's stated invariant. Traced through the new test's t0/t0+5000/t0+31000 sequence and it holds.
- src/proxy.ts:2820-2831: replacing 'hasCodexAccount && isOpenAI' with 'isOpenAI && await hasAnyCodexAccount()' and dropping the 'else { hasCodexAccount = false }' re-arm doesn't reintroduce a stale-true window -- an emptied directory is picked up by the very next hasAnyCodexAccount() call (which arms the 30s negative cache itself), so behavior is equivalent to the old re-arm but no longer requires the mutable flag.
- requiresClaudeLogin(...) at line 1745 still receives a startup-only snapshot (startupCodexAliases.length > 0), which is correct -- that's a boot-time gate, not a per-request routing decision, and the PR's comment explains this distinction directly.

## What's good
- New test (test/codex-runtime-account-detect.mjs) is hermetic (mkdtemp HOME, stub backend, fetchImpl thrower to catch any accidental Claude-path leak) and covers both the unit-level cache semantics and the actual regression (proxy boots with zero accounts, account appears mid-run, next request routes without restart) -- good coverage for the exact bug being fixed.
- Comments clearly state the invariants (why absence is cached and presence isn't), which will help the next person touching this code.

askalf:
Added the missing version bump: **5.5.84 -> 5.5.85** (`dc6220e`).

Without it the release gate filters this PR out of the auto-release and the fix ships to zero devices — flagged by gh-ops-sweep, tracked on ticket 00MTEKQKOE3D98076A9F550E18.

- `package.json` + `package-lock.json` bumped via `npm version --no-git-tag-version`, same shape as prior release commits (e.g. 085ecb3).
- `CHANGELOG.md` gets a `## [5.5.85] - 2026-08-29` section under `[Unreleased]`, per the release convention in the file header.

No source or test changes — the codex runtime-detection commits are untouched. Re-running CI on the new head.

askalf:
CI on `dc6220e` (the bump commit): **CI**, **actionlint**, **Wire drift (self-hosted)**, **CodeQL**, **Triage/Label** all green; **Live test** and **Compat test (self-hosted)** are the live-network drift jobs and were still running at time of writing — same posture as the pre-bump head. Review decision is APPROVED, mergeable.

The version-bump blocker (intervention 00MTEKQF0910836474CD73100C) is cleared by this commit. The merge intervention 00MTEKLVSZC1060163F1457296 is already operator-approved. Merge is the operator step; a dario release/deploy is a further step after that.

Issue #1009, opened 2026-08-18
Altman?


zytegalaxy:
**What would you like?**

Can you add "altman" mode that does the same with codex subscription?

`dario add altman`

**Why?**

For those who use codex and want to avoid subscription fee

**Alternatives you've considered:**

askalf:
Status update, since this has been quiet for a couple of days and you asked for it.

**It is being built.** #1010 has the scaffolding — 754 lines across `src/codex-accounts.ts`, `src/codex-oauth.ts` and the CLI wiring, plus three test files including a refresh-race harness.

**It is on a temporary hold, and not for a reason to do with the code.** The OAuth and account-pool paths need a real Codex/ChatGPT subscription to develop and test against, and there isn't one on the box yet. Reviewing or merging it now would mean shipping an auth path nobody has exercised end-to-end — which for a credential path is the one place that is not worth guessing at.

The subscription is expected shortly and work resumes within about a day of it landing. I have marked #1010 as a draft and labelled it `blocked` so it is not repeatedly queued for review while it waits, and so the stale bot does not close it out from under the work.

One thing worth flagging honestly rather than leaving you to find out later: #1010 was opened on 2026-08-18 against 5.5.25, and `master` is now at 5.5.35 — nine releases in a single day, several touching `live-fingerprint.ts`, `pool.ts` and `proxy.ts`. So expect a rebase before it lands, not just a merge.

I will post here when it comes off hold. Thanks for the request — it is a good one, and `dario add altman` is the right shape for it.

sprayberry-redline:
Scope update on this one.

The hold on #1010 was for a ChatGPT subscription to test the OAuth refresh behaviour against. That is being sorted now, so the branch comes off hold shortly: first step is a rebase onto current master (it was opened against 5.5.25; master is 5.5.83 and the branch conflicts), then the live refresh-race test, then wiring the engine into request routing so a stored account actually serves requests.

One thing has changed since this was filed: Cursor's direct OpenAI access is reported to end on November 12 following its acquisition. That makes this the same need for a much larger group than Codex CLI users — anyone on a ChatGPT plan who wants to keep using it from Cursor via a local OpenAI-compatible endpoint. So the target is no longer "Codex accounts in the pool" but "ChatGPT-subscription accounts served on the OpenAI-compatible path, usable from Cursor or Codex". Same engine, wider front door.

Will update here as the pieces land.

askalf:
[...]

PR #1140, opened 2026-08-29
fix(codex): open a streamed reply with the assistant role so SDK harnesses can assemble it (v5.5.86)

askalf:
Continues the altman engine (#1009). Found while de-risking the public claim — *"your ChatGPT plan in any OpenAI-compatible harness"* — with a strict wire-contract check against the live route on the box.

## The defect

`createResponsesTranslator` forwarded the first text delta as the **first `chat.completion.chunk` a client ever sees**:

```
data: {"id":"chatcmpl-…","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"content":"Hi"},"finish_reason":null}]}
```

The reference OpenAI stream opens with a role-only frame first:

```
data: {…,"choices":[{"index":0,"delta":{"role":"assistant","content":""},"finish_reason":null}]}
```

Accumulators built on that contract — openai-node's `ChatCompletionStream`, and the harnesses layered on it — use that frame to **open** the assistant message. Without it the assembled message carries no role, which reads fine in a terminal and then fails the moment the harness sends that turn back as history. This is precisely the class of bug that surfaces after launch as "it doesn't work in <harness>", because curl never notices and a typed client always does.

## The fix

The opener is emitted at `response.created` — the "a message is starting" signal, and the point where the chunk id is already resolved. A `roleSent` latch covers every other branch, so a stream that begins with a text delta, a tool call, or an empty completion still cannot put anything on the wire ahead of the role, and the role is never announced twice.

## Evidence

The contract check on the box (`gpt-5.6-sol`, live `fleet` ChatGPT account) — **42 of 43 assertions passed before this change; this was the one**:

```
/v1/models              7/7   list shape, target slug listed, created/owned_by typed
non-stream completion  11/11  id/object/created/model/choices/message/finish_reason/usage
stream                 13/14  SSE framing, [DONE], per-chunk shape … role opener FAILED
tool calls              8/8   tool_call id/type/name, arguments as a JSON *string*, finish_reason
error shape             3/3   4xx + .error.message (SDKs surface this)
```

## Tests

`test/codex-backend.mjs` now carries the contract explicitly: the role opener is the first frame, has no `finish_reason`, is announced exactly once, and appears on tool-call streams too — plus the two paths where it could have been skipped (a stream with no `response.created`, and a response that produced no content at all).

`node test/codex-backend.mjs` → 102 passed / 0 failed (was 94). `npm test` → **151/151**. `tsc --noEmit` clean.

v5.5.86 bumped in-PR (package + lock + CHANGELOG) so the release gate ships it.


sprayberry-redline (approved):

**Verdict:** No blocking issues found — the fix is correct, well-tested, and CI is fully green.

## What I checked

- `src/codex-backend.ts`: the `roleSent` latch (`let roleSent = false;`) and the `opened()` helper wrapping every emission path.
- Every branch that can write a frame to the wire — `response.created`, `response.output_text.delta`, `response.output_item.added`, the tool-call argument delta, and `response.completed`/`response.incomplete` — now routes through `opened(...)`, so a role-only frame (`delta: {role:"assistant", content:""}`) is guaranteed to precede the first content/tool-call frame, and the latch (`if (roleSent) return rest; roleSent = true;`) prevents it from ever firing twice.
- `test/codex-backend.mjs`: traced the updated frame-index assertions (e.g. `frames[1].choices[0].delta.content === 'Hel'`, `frames[4].choices[0].finish_reason === 'tool_calls'`) against the new opener-first ordering — indices line up correctly with the shift caused by the new leading frame.
- The two new defensive cases — a stream that starts at `response.output_text.delta` with no `response.created`, and a response with `response.created` + `response.completed` and no content at all — both correctly assert the role opener appears first and exactly once.
- CI check-rollup: all checks (`CI` build/test matrix, `Compat test`, `Live test`, `CodeQL`, `actionlint`) report `SUCCESS`; `auto-merge` is `SKIPPED` (expected — non-bot PR).

I did not find a code path where the latch can be bypassed or double-fired.

## What's good

- The version bump (5.5.85 → 5.5.86) is bundled correctly across `package.json`/`package-lock.json`/`CHANGELOG.md`, keeping the release atomic.
- Test coverage is thorough: happy path, tool-call streams, and both edge cases where the opener could theoretically be skipped.
- The PR body documents the wire-contract evidence (42/43 → 43/43 assertions) that motivated the fix.

## Minor (non-blocking)

- `CHANGELOG.md` diff introduces a double blank line before the `## [5.5.85]` heading (cosmetic only).

PR #1141, opened 2026-08-29
feat(codex): serve Anthropic-shape /v1/messages from a ChatGPT subscription (v5.5.87)

askalf:
Stacked on #1140 (base is `fix/codex-stream-role-opener`, so the diff here is just this change; it retargets to `master` when #1140 merges).

Continues the altman engine (#1009). This is the half that lets a **Claude-shaped harness run on a ChatGPT plan**.

## The defect, and how we found it

`codexAdapter` refused anything that wasn't `/v1/chat/completions` — the source said why: *"An Anthropic-shape /v1/messages request would need a second translation that doesn't exist, so it stays with Claude."* So any Anthropic-SDK client naming a subscription slug fell through to Claude and got:

```
404 {"type":"not_found_error","message":"model: gpt-5.6-sol"}
```

That is exactly how askalf's own fleet found it. A forge agent configured on `gpt-5.6-sol` (`ANTHROPIC_BASE_URL=http://dario:3456`, Anthropic SDK → `/v1/messages`) was dispatched for the first time and 404'd. The feature had been deployed and declared live for hours; nothing had exercised it.

## The change

- **`src/anthropic-responses-translate.ts`** — the missing half: Messages → Responses requests, Responses → Messages responses, and the Responses event stream → Anthropic SSE. Ported from the 5.1 work that lived outside the repo and was live-probe validated there. Translator logic is **byte-identical from its first exported function**; only the header changed and the types it needed were inlined so the module is self-contained (no `openai-translate.ts` dependency, and none of its chat/completions code comes along as dead weight).
- **`forwardToCodex` takes a `shape`** so both wire formats share ONE transport — same headers, timeout, upstream POST, read loop, error paths — with only the two translation ends swapped:

```
openai     chat/completions ─┐                    ┌─ chat.completion(.chunk)
                             ├→ Responses (codex) ┤
anthropic  messages ─────────┘                    └─ Anthropic SSE / Message
```

  `stream: true` is forced upstream for both shapes (the backend is always streamed and collapsed here — the non-streaming Anthropic body is rebuilt from the `response` carried by the terminal `response.completed`), and an Anthropic-shape client now gets **Anthropic-shape errors** `{type:'error',error:{type,message}}`, because an SDK reads its own error shape.
- **`codexAdapter` drops the path guard.** The claim is now purely MODEL-driven — a `codex:`/`chatgpt:` prefix, or a slug the backend itself listed for the stored account.

## Why dropping the guard is safe

A Claude model is never in `codexModels`, so Claude traffic is untouched on either path. That's the whole risk of this change, so it's pinned in the truth table rather than argued:

```
claude model on the anthropic path is untouched by a codex account   → claude
anthropic path, discovered slug, but no codex account                → claude
forced claude beats a discovered slug on the anthropic path          → claude
undiscovered gpt model on the anthropic path                         → claude
discovered slug on the anthropic path                                → codex
forced codex on the anthropic path                                   → codex
```

The `openai` (API-key backend) adapter **keeps** its OpenAI-path guard — it has no Messages translation. That's separate work.

## Tests

`forwardToCodex` is now driven end to end in **both** shapes against an injected upstream and a fake `ServerResponse` — request translation (`input[]` not `messages[]`, forced `stream`, `system`→`instructions`, `input_schema`→`parameters`), the non-streaming collapse (a real Anthropic `Message` with `input_tokens`/`output_tokens`), streaming (Anthropic event names, every `data:` line valid JSON, no OpenAI `[DONE]` on the Anthropic wire), the unchanged OpenAI shape, and the error body. That end-to-end coverage is the point: a mistranslation here is invisible until a real client tries to read the stream.

`npm test` → **153/153**. `tsc --noEmit` clean. v5.5.87 bumped in-PR so the release gate ships it.

## Follow-up (not in this PR)

Once released and deployed, `rev_code_reviewer_gpt` should work unchanged — that's the fleet dogfooding this unblocks. Its trigger is paused meanwhile.


askalf:
**Reviewer scoping note** — the previous review run hit its 40-turn cap while still opening files, so here is where the risk actually is. This PR is ~1,400 lines but only ~90 of them are new logic.

**Skip: `src/anthropic-responses-translate.ts` (1,171 lines).** It is a *verbatim port*, not new code. Its logic is byte-identical to the source from its first exported function onward — verifiable mechanically:

```
diff <(sed -n '/^export function anthropicToResponsesRequest/,$p' <source>) \
     <(sed -n '/^export function anthropicToResponsesRequest/,$p' src/anthropic-responses-translate.ts)
```

Only the file header changed: the types it used to import were inlined so the module is self-contained, and an unused `estimateTokenCount` re-export was dropped. It arrives with its own 28 tests (`test/anthropic-responses-*.test.mjs`), which pass unchanged from the source repo.

**Review these three, in this order:**

1. **`src/provider-adapter.ts`** — the only behavioural risk in the PR. `codexAdapter` loses its `isOpenAIPath` guard, so it can now claim `/v1/messages`. The question to answer: *can this steal Claude traffic?* The claim is model-driven (`isCodexModel` / an explicit `codex:` prefix) and a Claude model is never in `codexModels`, but that is the thing worth checking rather than taking on trust. The six new truth-table rows in `test/codex-backend.mjs` are the assertions for it.

2. **`src/codex-backend.ts`** — `forwardToCodex` gains a `shape` parameter. Worth checking: the OpenAI branch is untouched (same line-based loop), `stream: true` is forced upstream for both shapes, and the non-streaming Anthropic body is rebuilt from the terminal `response.completed` rather than invented.

3. **`src/proxy.ts`** — 2 lines. The codex credential lookup is no longer gated on `isOpenAI`, and the shape is passed through.

**Not a regression risk:** `README.md` (line-count gate) and the version/CHANGELOG bump.

sprayberry-redline (changes requested):

**Verdict:** one correctness bug worth fixing before merge (silently-swallowed upstream failure on the non-streaming Anthropic path); the rest of the translation is careful and well-tested. CI is all-green (18 checks, including live-test and wire-drift).

## Findings

### 1. [Medium] `src/codex-backend.ts:537` — non-streaming Anthropic responses drop `response.failed`, returning a fake empty success instead of an error

```js
        if (t === 'response.completed' || t === 'response.incomplete') {
          const r = (ev as { response?: ResponsesResponse }).response;
          if (r) terminalResponse = r;
        }
```

`ResponsesResponse.status` is documented (`src/anthropic-responses-translate.ts:363`) as `completed | failed | in_progress | cancelled | queued | incomplete`, and the translator's own terminal-event switch treats all three terminal types uniformly:

```js
        type === 'response.completed' ||
        type === 'response.incomplete' ||
        type === 'response.failed'
```
(`src/anthropic-responses-translate.ts:1085-1087`)

But the `emitAnthropic` capture in `forwardToCodex` (used only to feed the non-streaming JSON collapse) checks just two of the three. If the upstream Responses stream terminates with `response.failed`, `terminalResponse` is never set, and the collapse falls through to:

```js
res.end(JSON.stringify(responsesToAnthropicResponse(terminalResponse ?? {}, model)));
```
(`src/codex-backend.ts:601`)

`responsesToAnthropicResponse({}, model)` has no `output` to walk and no `resp.error` handling, so it returns a well-formed `type:'message'` with `content: []`, `stop_reason: 'end_turn'`, `usage: {input_tokens:0, output_tokens:0}` — a non-streaming Anthropic-SDK client (e.g. Claude Code on a `codex:` slug) sees what looks like a normal, empty, successful turn instead of any indication the backend failed. There is no test exercising `response.failed` anywhere in `test/anthropic-responses-stream.test.mjs` or `test/codex-backend.mjs` to catch this.

```suggestion
        if (t === 'response.completed' || t === 'response.incomplete' || t === 'response.failed') {
          const r = (ev as { response?: ResponsesResponse }).response;
          if (r) terminalResponse = r;
        }
```
(and ideally have `responsesToAnthropicResponse` surface `resp.error`/a failed status as a visible text block or a distinct stop_reason, rather than silently returning empty content.)

## What's good
- `forwardToCodex`'s shared-transport design (one function, `shape` picks translation ends) is a clean way to avoid duplicating headers/timeout/error-path logic between the two wire formats.
- The new translator module is thorough and well-documented, with real end-to-end coverage of both streaming and non-streaming Anthropic-shape traffic against an injected upstream (`test/codex-backend.mjs`), plus a dedicated routing truth table in the same file confirming Claude traffic is untouched now that the `isOpenAIPath` guard is dropped in `provider-adapter.ts`/`proxy.ts`.
- Defensive JSON parsing (`safeParseArguments`, `stringifyToolInput`) avoids throwing mid-response on malformed upstream data.
- CI is fully green (build/test on Node 18/20/22, live-test, wire-drift, CodeQL).

## What I checked
Read the full diff for `src/anthropic-responses-translate.ts` (new, 1278 lines), `src/codex-backend.ts`, `src/provider-adapter.ts`, `src/proxy.ts`, and skimmed the three test files for coverage shape. Did not run the suite locally (per policy) — relied on the PR's CI check-rollup, which is all green.

askalf:
Verified `812dc21` against the review finding (ticket 00MTF1GL1QB21215CA1F0EB859) — it resolves it, and goes further than the finding did.

The finding was scoped to the Anthropic non-streaming collapse. `812dc21` fixes that (`isTerminalResponsesEvent` now admits `response.failed`, and the collapse returns 502 with a shape-appropriate error body instead of `responsesToAnthropicResponse({})`), and additionally fixes the same gap on the chat path, which the finding did not name: `createResponsesTranslator` was also two-of-three terminal, so a failed chat stream ended with no finish frame and no `[DONE]`. Sharing `isTerminalResponsesEvent`/`isFailedResponse` between both paths is the right shape — that is what stops them drifting apart again, which is how this arose.

Two notes, neither blocking:

1. **Uncovered edge, correctly handled but unpinned.** A bare `{ "type": "response.failed" }` with no `response` object: both paths get it right (`anthropicFailed` is set on the event type, and the translator ORs `type === "response.failed"` before consulting `isFailedResponse`), but no test asserts it, so a future refactor that keyed only off `isFailedResponse(e.response)` would regress silently to an empty 200. One case in the `response.failed` block would pin it:
   ```js
   await forwardToCodex({}, res, body, CREDS, "*", {}, 5000, false, "anthropic",
     fakeUpstream([{ type: "response.created", response: { id: "resp_b" } }, { type: "response.failed" }], {}));
   check("bare response.failed still fails closed", res.statusCode === 502);
   ```

2. **`isFailedResponse` treats any non-null `error` as a failure, including on a `completed` payload.** That is the safe direction (fail closed) and the success-path guard covers the normal case, but it does mean a `response.completed` that carries a non-fatal `error` annotation alongside real output would 502 and discard that output. If the Responses API never does that, it is a non-issue; worth knowing the behaviour is deliberate rather than incidental.

No PR from me — `812dc21` is a superset of what I had staged, so a second PR would only conflict. CI 15/15 green.

sprayberry-redline (approved):

**Verdict:** re-review at head `812dc21` — the fix commit correctly closes the finding from my prior CHANGES_REQUESTED review. Approving.

## What changed since the last review

Commit `812dc21` ("fix(codex): treat response.failed as terminal on BOTH paths; never collapse a failure into an empty success") directly targets the bug I flagged: the non-streaming Anthropic collapse only checked `response.completed`/`response.incomplete`, so a stream ending in `response.failed` silently returned a well-formed empty success.

Verified in `src/codex-backend.ts`:
- New shared predicates `isTerminalResponsesEvent`, `isFailedResponse`, `failedResponseMessage` are now used on both the Anthropic collapse path (`emitAnthropic`, `codex-backend.ts:568-577`) and the OpenAI/chat path (`createResponsesTranslator`'s terminal branch, now via `isTerminalResponsesEvent`), so the two paths can't drift again — the fix also caught and closed the same pre-existing gap on the chat path, which I hadn't flagged but is a legitimate related bug.
- `upstreamFailed` is computed after the read loop and, for a non-streaming client, short-circuits to a 502 with a shape-appropriate error body (`codex-backend.ts:625-641`) **before** reaching the code that would have called `responsesToAnthropicResponse(terminalResponse ?? {}, model)` — confirmed by reading the surrounding control flow: the early `return` sits ahead of both the `isAnthropic` and OpenAI-shape success-response blocks.
- Streaming clients are correctly left alone (the terminal event is already on the wire; `emitAnthropic` and the chat-path `chunk()` translator already surface it during the read loop).
- Tests added in `test/codex-backend.mjs` cover the predicates, chat-path stream termination on failure (finish frame + `[DONE]`), non-streaming 502 + error-shape assertions for both Anthropic and OpenAI shapes, the streaming path staying 200/clean, and a false-positive guard on a successful response.

I did not re-run the suite locally (per policy) — CI's `test` and `live-test` checks are green, along with all other checks in the rollup (`build` x3, `compat`, `docker-cap-drop-smoke`, `wire-drift`, `CodeQL`, etc.).

No new correctness issues found in this delta.

## What's good
- Consolidating the terminal-event check into shared predicates instead of patching the two call sites independently is the right fix — it's exactly the kind of drift the original bug was caused by.
- The added tests exercise both success and failure on both wire shapes, not just the happy path.

PR #1142, opened 2026-08-30
fix(codex): never send max_output_tokens to the ChatGPT backend — it 400s the request (v5.5.88)

askalf:
Found by running the **first live end-to-end test of the Anthropic path** on the box after v5.5.87 deployed. Every `/v1/messages` request to a gpt slug came back 400.

## The bug

The Codex backend rejects the parameter outright:

```
400 {"detail":"Unsupported parameter: max_output_tokens"}
```

Both builders set it from the client's `max_tokens` / `max_completion_tokens`, so **both wire shapes** fail whenever a client asks for an output cap. Reproduced directly against the live backend — the *chat*-derived body 400s identically, so this is a pre-existing bug on that path too, not something the Anthropic work introduced.

It hid for as long as it did because every smoke test so far happened to omit `max_tokens`. It then became a **100% failure** the moment v5.5.87 went live, because the Messages API *requires* `max_tokens` and so always produced it.

## Verified live

Same body, same account, one field different:

| | |
|---|---|
| with `max_output_tokens` | `400 Unsupported parameter` |
| without | `200`, streams `response.output_text.delta` normally |

## The fix

Stripped in `forwardToCodex`, not in either translator — it is a property of *this backend*, not of either wire format. The same builders stay correct against an API-key Responses endpoint, which does support the parameter.

Tests assert it never reaches the upstream body for either shape, and for `max_completion_tokens` too. `npm test` 153/153, `tsc` clean. v5.5.88 bumped in-PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.88] - 2026-08-30
+
+- **A request that asks for an output cap no longer 400s against a ChatGPT subscription (dario#1142).** The Codex backend rejects the parameter outright — `400 {"detail":"Unsupported parameter: max_output_tokens"}` — and both request builders set it from the client's `max_tokens` / `max_completion_tokens`, so both wire shapes failed whenever a client asked for one. It hid for as long as it did because every smoke test happened to omit `max_tokens`; it then surfaced as a 100% failure the moment the Anthropic path went live, because the Messages API *requires* `max_tokens` and so always produced it. Verified against the live backend: the identical body 400s with the field and streams a normal reply without it. The strip lives in `forwardToCodex`, not in either translator, because it is a property of this backend rather than of either wire format — the same builders remain correct against an API-key Responses endpoint, which does support the parameter.
+
+
 ## [5.5.87] - 2026-08-29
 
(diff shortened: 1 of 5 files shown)


sprayberry-redline (approved):

## What I checked
- `gh pr diff` (5 files, +49/-3: CHANGELOG, package.json/lock version bump, `src/codex-backend.ts`, `test/codex-backend.mjs`).
- CI check-rollup: all 14 checks passing (`test`, `compat`, `live-test`, `CodeQL`, `build` x3, etc.), `auto-merge` correctly skipping.
- No standing `sprayberry-reviewer` review exists yet for this head SHA — this is the first review.

## Findings
None blocking.

The fix itself, `src/codex-backend.ts:545` — `delete (upstreamBody as Record<string, unknown>).max_output_tokens;` — is placed after `upstreamBody` is constructed from *either* branch (`anthropicToResponsesRequest` or `chatCompletionsToResponses`), so it correctly strips the rejected field for both wire shapes unconditionally, matching the PR's stated reasoning (this is a property of the backend, not of either translator). `delete` on a possibly-absent key is a safe no-op, so callers that never set the field are unaffected.

Test coverage in `test/codex-backend.mjs:645-676` directly exercises the claim: strips `max_output_tokens` for the Anthropic-shape path (from client `max_tokens`), the OpenAI/chat-shape path (from client `max_tokens`), and the `max_completion_tokens` variant, plus a sanity check that the rest of the body (`input`, `stream: true`) still arrives. This lines up with the two "both wire shapes fail" scenarios described in the PR body.

## What's good
- Root-caused correctly: the bug is scoped to the backend/transport layer (`forwardToCodex`), not duplicated into both translators.
- The comment at `src/codex-backend.ts:537-546` explains the non-obvious *why* (backend-specific rejection, hidden by smoke tests omitting `max_tokens`) rather than restating the code.
- Regression tests target the exact failure mode described, across both client wire shapes.
- CHANGELOG and version bump are consistent with the change.

PR #1143, opened 2026-08-30
fix(codex): fold the non-streaming Anthropic body from the stream, not the empty terminal event (v5.5.89)

askalf:
Third and last defect from the live end-to-end run of the Anthropic path on the box. After v5.5.88 the proof went **17/18** — streaming returned text, the Claude path was untouched, and non-streaming returned a well-formed message with `content: []`.

## The bug

The collapse read content off the terminal `response.completed` event. Correct against the standard Responses API; wrong against **this** backend. Captured live:

```
response.completed keys: ["type","response","sequence_number"]
.response.output: []
translated content: []
```

The content only ever exists in the delta events. So a non-streaming client got a perfectly well-formed Anthropic message with empty content — **a silent empty answer**, indistinguishable from a model that chose to say nothing. That is the same failure shape as the two bugs before it, which is why it is worth fixing properly rather than special-casing.

## The fix

`createAnthropicMessageAssembler()` folds the Anthropic events this module already emits back into one `Message`, and the collapse uses it. So the streamed body and the collapsed body now come from **one** translation instead of two that can disagree.

This is the discipline the chat path always had — `createResponsesTranslator.complete()` accumulates from deltas for exactly this reason, which is why the OpenAI shape never showed the bug.

## Test

Pins the real backend shape rather than my earlier assumption: a terminal event carrying `output: []` must still yield the delta text, with usage and stop_reason intact.

`npm test` 153/153, `tsc` clean, README gate green. v5.5.89 bumped in-PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ checklist.
 
 ## [Unreleased]
 
+## [5.5.89] - 2026-08-30
+
+- **A non-streaming Anthropic reply from a ChatGPT subscription is no longer empty (dario#1143).** The collapse read its content off the terminal `response.completed` event, which is correct against the standard Responses API and wrong against this backend: the ChatGPT Codex backend sends `response.completed` with `output: []`, so the content — which only ever exists in the delta events — was dropped and the client received a perfectly well-formed message with `content: []`. A silent empty answer, indistinguishable from a model that chose to say nothing. Found by running the first live end-to-end test of the path against a real subscription: streaming returned the text, non-streaming returned nothing. The body is now FOLDED FROM THE STREAM by `createAnthropicMessageAssembler`, which is the discipline the chat path always had (`createResponsesTranslator.complete()` accumulates from deltas for exactly this reason) — so both the streamed and collapsed bodies now come from ONE translation instead of two that can disagree. The regression test pins the real shape: a terminal event carrying `output: []` must still yield the delta text.
+
+
 ## [5.5.88] - 2026-08-30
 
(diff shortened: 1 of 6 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found - clean fix with solid regression coverage.

**What I checked**
- `src/anthropic-responses-translate.ts`: the new `createAnthropicMessageAssembler()` folds `message_start` -> `content_block_start`/`content_block_delta` -> `message_delta` events into one `Message`, mirroring the existing accumulate-from-deltas discipline the OpenAI chat path already used. Traced the text, thinking, and tool_use (`partialJson` + `safeParseArguments`) accumulation paths and the `usage`/`stop_reason` fold in `message_delta` - all correctly propagate to the final message.
- `src/codex-backend.ts`: `antAssembler` is only fed (`antAssembler!.push(produced)`) when `!clientWantsStream`, so streaming clients pay no extra cost; the non-streaming collapse now calls `antAssembler!.push(antTranslator!.end())` then `antAssembler!.message(model)` instead of the old `responsesToAnthropicResponse(terminalResponse ?? {}, model)`. Confirmed `antTranslator.end()` is a no-op once the translator already finalized (e.g. via `response.completed`), so there is no double-counting of `message_delta`/`message_stop`.
- `test/codex-backend.mjs`: the new test reproduces the exact live-observed shape (`response.completed` with `output: []`, content only in `response.output_text.delta` events) and asserts the assembled text, usage, and stop_reason all survive the collapse - this is a real regression test for the bug described, not a rewrite of the assumption that caused it.
- `gh pr checks` - all checks green, including `live-test`, `build (18/20/22)`, `compat`, `test`.

**What's good**
- The fix unifies the streamed and collapsed body construction into one translation path, closing off the class of "two translations that can disagree" bugs (this is the third in that series per the PR description) rather than special-casing this backend.
- Test fixture is grounded in an actual captured live response shape, not a guess.

No correctness, security, or test-coverage concerns in the diff. Approving.

PR #1144, opened 2026-08-30
fix(codex): scrub the upstream body against an allowlist — temperature and friends 400 this backend (v5.5.90)

askalf:
Found the moment the GPT code reviewer was un-paused: it 400'd on its very first real dispatch, *after* the Anthropic path was proven working 18/18 by hand.

## Why my proof missed it

My live tests never sent a `temperature`. forge does — it sets one from the agent's `provider_config`. So the reviewer failed on every dispatch while hand-built requests passed.

## What this backend actually accepts

Probed field-by-field against a live subscription:

```
REJECTED: temperature, top_p, max_output_tokens, presence_penalty,
          frequency_penalty, seed, metadata, top_logprobs, truncation, service_tier
ACCEPTED: tools, tool_choice, parallel_tool_calls, reasoning, instructions
```

**This was breaking the OpenAI path too**, not just the new Anthropic one — `chatCompletionsToResponses` passes `temperature` and `top_p` straight through, and most OpenAI clients send a temperature by default.

## Allowlist, not a denylist

v5.5.88 deleted `max_output_tokens` specifically. That treated the symptom — this is the third parameter in the same class to bite. The transport now scrubs against an allowlist of what the backend accepts.

Deliberately an allowlist and not the ten known-bad names: the backend is undocumented and restrictive, so the failure mode we must not have is *"we started sending a new field and every request 400s"*. Dropping an unknown field degrades one request; sending one breaks all of them.

The scrub lives at the transport, so both request builders stay correct against an API-key Responses endpoint, which does accept these.

## Tests

Each rejected name asserted stripped; every accepted field asserted preserved; an unknown field asserted dropped (the fail-safe property); plus end-to-end through `forwardToCodex` in both shapes. codex-backend 174/174, full suite green, `tsc` clean. v5.5.90 bumped in-PR.


sprayberry-redline (approved):

## What I checked
- `gh pr diff` (225 lines) covering `src/codex-backend.ts`, `test/codex-backend.mjs`, `CHANGELOG.md`, `package.json`/`package-lock.json`.
- Cross-referenced `CODEX_SUPPORTED_FIELDS` (src/codex-backend.ts) against both request builders that feed `forwardToCodex`: `chatCompletionsToResponses` (src/codex-backend.ts:204) and `anthropicToResponsesRequest` (src/anthropic-responses-translate.ts:606) to confirm the allowlist doesn't silently drop a field either builder legitimately sets.
- CI check-rollup: all 14 checks green (build ×3, test, live-test, compat, CodeQL, actionlint, docker-cap-drop-smoke, etc.) — `gh pr checks 1144`.

## Correctness
The allowlist (`model, input, stream, store, instructions, tools, tool_choice, parallel_tool_calls, reasoning`) is an exact match for the field set both builders can emit, minus the three the backend rejects (`temperature`, `top_p`, `max_output_tokens`). I traced each builder's output fields by hand and found no field that's produced by either translator but missing from the allowlist — so no working parameter (e.g. `tools`, `reasoning`, nested `reasoning.summary`, `parallel_tool_calls`) gets silently stripped by this change.

`toCodexSupportedBody` (src/codex-backend.ts new code) is a straightforward, side-effect-free filter (`if (body[k] !== undefined) out[k] = body[k];`), correctly preserving falsy-but-defined values like `store: false` and `stream: true` rather than dropping them (a common `if (body[k])` truthiness bug this avoids).

## Test coverage
Good: `test/codex-backend.mjs` asserts each rejected field name is stripped, each accepted field survives, an unknown/future field is dropped (the fail-safe property the PR body calls out), no duplicates in the allowlist, and two end-to-end assertions through `forwardToCodex` covering both the Anthropic and OpenAI-chat request shapes.

## Minor (non-blocking)
- CHANGELOG.md has a doubled blank line before the `## [5.5.89]` entry (cosmetic).

## What's good
Replacing the prior single-field deny-and-delete (`max_output_tokens`) with an allowlist is the right fix for the failure class described — an undocumented, restrictive backend where a future unlisted field breaking every request is worse than one degraded request. The doc comment on `CODEX_SUPPORTED_FIELDS` explains the tradeoff clearly, and the scrub is correctly placed at the transport layer (`forwardToCodex`) rather than in either translator, keeping both builders correct for the public API-key Responses endpoint.

PR #1145, opened 2026-08-30
v6.0.0 — route any client shape to any subscription, and fail over between them

askalf:
**dario now routes any client shape to any of your subscriptions, and fails over between them.**

dario has been a Claude proxy that recently learned to reach a ChatGPT subscription on one path. The 5.5.8x line quietly finished the harder half of that — the Codex backend serves Anthropic-shape requests too — and this release takes the consequences seriously. Two consumer subscriptions, no API keys, either one able to carry the whole deployment. That is a different product than 5.x described, so it gets a major.

Everything here is opt-in, and a single-entry `--pool-fallback` behaves exactly as it did before, so existing configs are untouched.

---

### Failover can target a subscription, on both wire shapes

`--pool-fallback` could previously only re-point a request at an api-key backend, and only on `/v1/chat/completions`. That made it **inert** for anyone whose second provider is a ChatGPT plan rather than an API key — which is how askalf's own fleet runs, and why it went fully dark on Claude 429s twice on 2026-08-29 with an idle ChatGPT subscription sitting beside it.

Failover now picks whichever provider can actually serve the nominated model, prefers the subscription (no per-token cost), and covers Anthropic-shape clients — Claude Code, the Anthropic SDKs, agent runtimes — which previously had nowhere to go and simply failed. The api-key path keeps its OpenAI-only guard: there is still no Messages translation on that route.

### Failover is symmetric, via chains

```bash
dario proxy --pool-fallback=gpt-5.6-sol,claude-sonnet-5
```

Read left to right, each provider taking the first entry it can serve. Drained Claude pool → the gpt slug. Rate-limited or failing subscription → back to the Claude pool. Neither subscription hitting its ceiling can take a deployment down on its own.

Only a **429 or 5xx** defers. A **400 still surfaces** — a bad request that fails over just reproduces itself on the other provider and buries the real cause. That distinction is tested explicitly, because it is exactly the shape of the bug that shipped in v5.5.90.

### `dario add altman`

The command asked for in #1009 — plus `amodei`, `chatgpt`, `claude`. Naming whose plan you are attaching is the distinction that matters once a Codex account is a peer of the Claude pool rather than a side door. It dispatches into the existing implementations; `dario codex add` and `dario login` are unchanged and undeprecated.

### `dario doctor` reports failover readiness

Distinguishes off / armed-but-**INERT** / one-way / symmetric, and names a concrete fix. The inert state is the point: armed with nothing to fall back to is green on every other check and incapable of doing anything. Nothing reported it because nothing asked "armed" and "has somewhere to go" as a single question.

`failoverReadiness()` is pure so all five branches are tested — a live host can only ever exercise the branch matching its own credentials, which is precisely how this went unnoticed.

### Shadow compare — `x-dario-compare: <model>`

Runs the same prompt past a second model family beside the real answer and writes both to `~/.dario/compare/`, in the client's own wire shape.

Once either subscription can serve either shape, the useful question stops being *can I reach GPT* and becomes *which of these is better at my work* — which your own traffic answers far better than a benchmark does.

The client is never affected: the tee only observes bytes already on their way out, the request is never held open for the comparison, and a comparison that fails, times out, or has nowhere to go is dropped with the record still written. Both sides are stored as raw payloads rather than extracted text, since extraction is where a bug would quietly make two answers look more alike than they are.

---

### Verification

| | |
|---|---|
| `tsc --noEmit` | clean |
| `npm test` | 154 files, 0 fail |
| `test/codex-backend.mjs` | 195 pass |
| `test/compare.mjs` | 26 pass (new) |
| `test/doctor-formatter.mjs` | 48 pass |
| README line-count gate | ok (28,949 / 58 files) |

72 new assertions. Every stale doc site that asserted the removed constraint — `src/cli.ts` help, `src/config-file.ts`, `src/proxy.ts`, `docs/commands.md` (2 rows), `docs/multi-account-pool.md`, README — was swept; `grep` for the old claims returns nothing.

### Not covered by tests

`runCompare()` reads real credentials and would make a network call on any machine with a Codex account, so it is exercised only through its pure parts. The codex→Claude fall-through is wired and typechecked but, like every wire path in this line of work, deserves one live request before it is trusted — the last five bugs here all passed CI and read as successes.


sprayberry-redline (commented):

2 correctness issues to address in the runtime paths reviewed.

## Findings

### High — transport failures do not enter symmetric failover (`src/codex-backend.ts:744-755`)

> ```ts
>   } catch (err) {
>     const detail = err instanceof Error ? err.message : String(err);
>     if (verbose) console.error(`[dario] codex backend (${creds.alias}) error: ${detail}`);
>     if (!res.headersSent) {
>       res.writeHead(502, { 'Content-Type': 'application/json', ...securityHeaders });
>       res.end(errBody('Upstream Codex backend error', { account: creds.alias }));
>     } else {
>       try { res.end(); } catch { /* already closed */ }
>     }
>     return true;
> ```

`deferOnUnavailable` is honored for HTTP 429/5xx responses above, but not here. If `fetchImpl` rejects before headers (DNS failure, connection refusal/reset, or timeout abort), this writes a 502 and returns `true`; the caller's `if (served) return` then prevents the nominated Claude fallback. No response bytes had been committed, so this is a failover-eligible Codex outage, but the request fails instead.

```suggestion
    if (deferOnUnavailable && !res.headersSent) {
      console.log(`[dario] codex account ${creds.alias} unavailable (${detail}) — deferring to the next provider`);
      return false;
    }
    if (!res.headersSent) {
      res.writeHead(502, { 'Content-Type': 'application/json', ...securityHeaders });
      res.end(errBody('Upstream Codex backend error', { account: creds.alias }));
    } else {
      try { res.end(); } catch { /* already closed */ }
    }
    return true;
```

Please also test an injected `fetchImpl` rejection with `deferOnUnavailable` enabled.

### Medium — same-millisecond comparisons overwrite each other (`src/compare.ts:207-213`)

> ```ts
>     const stamp = record.ts.replace(/[:.]/g, '-');
>     const safeModel = record.comparedModel.replace(/[^\w.-]/g, '_').replace(/\.{2,}/g, '.');
>     const path = join(dir, `${stamp}-${safeModel}.json`);
>     writeFileSync(path, JSON.stringify(record, null, 2), 'utf-8');
> ```

The filename contains only a millisecond timestamp and model, while `writeFileSync` replaces existing files. Two concurrent requests comparing the same model can finish in one millisecond and derive the same path, so the second silently destroys the first record. Comparison work is explicitly concurrent, making this reachable under normal parallel traffic.

```suggestion
    const safeModel = record.comparedModel.replace(/[^\w.-]/g, '_').replace(/\.{2,}/g, '.');
    const path = join(dir, `${stamp}-${safeModel}-${randomUUID()}.json`);
    writeFileSync(path, JSON.stringify(record, null, 2), { encoding: 'utf-8', flag: 'wx' });
```

Import `randomUUID`, or use another unique request/monotonic identifier; retain exclusive creation to prevent silent replacement.

## What's good

The provider selection is split into pure, focused helpers; 400 vs 429/5xx behavior is explicit; and clearing `parsedBody` after the Codex→Claude swap avoids a subtle stale-model bug.

Scope note: this PR is 1,094 changed lines. I prioritized runtime routing/failover, Codex transport, provider adapter, shadow compare, and their targeted tests; I did not line-review every documentation/package metadata change.

askalf:
Addressed the review finding on `src/codex-backend.ts:744-755` (ticket `00MTFQH6SI328CEC6AA5CB0463`).

**What changed** — the `catch` around the upstream POST always answered the client 502, so failover covered the 429/5xx case but not the case where the subscription is simply unreachable: a rejecting `fetch` (DNS, refused connection, reset socket) or our own `AbortController` firing on `upstreamTimeoutMs` never produces a status to read, so it never reached the `deferOnUnavailable` branch at line 622. A gpt-bound request then died 502 with an idle Claude pool sitting beside it — exactly the shape v6.0.0 exists to fix.

Now, when `deferOnUnavailable` is armed **and** `!res.headersSent`, the catch returns `false` instead. The `headersSent` guard preserves the documented contract: once a stream is in flight it is too late to hand the request to anyone else, so a half-sent stream is still finished on this provider. With failover unarmed the 502 is unchanged.

**Tests** (`test/codex-backend.mjs`, in the existing symmetric-failover block):
- a rejecting fetch (`TypeError: fetch failed`) returns `false` and writes nothing — `headersSent === false && ended === false && chunks.length === 0`
- an `AbortError` (our own upstream timeout, which lands in the same catch) also defers
- with `deferOnUnavailable = false` the transport failure still 502s — the default is unchanged

Commits: cf06a52 (fix), 3165587 (tests), 57af28b (doc contract).

**Not verified in-run:** the test file imports from `dist/`, so I did not execute it locally under the build wall — CI on 57af28b is the authoritative signal. Untouched by design: `src/proxy.ts:2977` already treats any `served === false` as the cue to swap the body to `claudeTarget`, so the new deferral path needs no caller change.

sprayberry-redline (commented):

2 correctness issues to address in the reviewed core failover/compare paths.

## Findings

### Medium — `src/codex-backend.ts:169-170` — an unknown model is assumed to be Claude-servable

> `return models.find(m => !isCodexModel(m, slugs)) ?? null;`

This tests only that a model is absent from the selected Codex account; it does not establish that the Claude pool can serve it. For example, with Codex slugs `["gpt-5.6-sol"]` and `--pool-fallback=gpt-5.6-sol,gpt-4o` (or a typo in the second slug), a 429 from Codex selects `gpt-4o`, `canDefer` becomes true, and `proxy.ts:2984-3001` re-points the request to the Claude pool. The original retryable Codex error is consequently replaced by an invalid-model response from Claude, even though `gpt-4o` may actually belong to the configured OpenAI backend. The same unsupported chain is reported as “symmetric” by `doctor.ts:89-93` solely because it has two entries.

Determine capability positively rather than defining “Claude model” as “not listed by this Codex account”; if Claude capabilities cannot be enumerated, require an explicit provider prefix for reverse-failover entries and make doctor validate the same rule.

```suggestion
export function pickClaudeFallback(models: readonly string[]): string | null {
  return models.find((model) => model.startsWith("claude:"))?.slice("claude:".length) ?? null;
}
```

### Low — `src/proxy.ts:2873-2881` — the response claims a comparison happened before knowing it can run

> `const compareTarget = readCompareTarget(req.headers as unknown as Record<string, unknown>);`
> `if (compareTarget) {`
> `  const compareShape = isOpenAI ? ("openai" as const) : ("anthropic" as const);`
> `  const tee = teeResponse(res);`
> `  res.setHeader(COMPARE_RESULT_HEADER, compareTarget);`
> `  ...`
> `  const running = runCompare({`

`runCompare()` can legitimately return `skipped` for no account, an unsupported target, invalid JSON, credential failure, or an upstream failure, but every such response still carries `x-dario-compared-with: <target>`. A caller using that header as the advertised confirmation that shadow comparison ran records a false positive, while the actual skip exists only in a server-side file/log.

Expose attempt state honestly—e.g. name this header `x-dario-compare-requested`—or perform the capability checks before headers are sent and set `x-dario-compared-with` only when dispatch actually starts. If asynchronous completion must remain invisible to the primary response, the response header should not claim successful comparison.

```suggestion
res.setHeader("x-dario-compare-requested", compareTarget);
```

## What’s good

The failover response contract distinguishes retryable 429/5xx/transport failures from client-side 4xx errors, and the changed backend tests exercise that boundary. The compare implementation also keeps its failures off the primary request path and sanitizes model-derived filenames.

Review scope: this PR is 1,139 changed lines, so I reviewed the core changed implementation in `src/proxy.ts`, `src/codex-backend.ts`, `src/provider-adapter.ts`, `src/compare.ts`, `src/doctor.ts`, and their focused tests. I did not line-review the documentation/versioning-only files.

sprayberry-redline (commented):

Verdict: 1 high-severity correctness issue makes the new Claude→Codex subscription failover unreachable in key advertised configurations.

## Findings

### High — `src/proxy.ts:3024-3034` — Codex fallback runs only after the old selector has already rejected the request

> `if (!selectPoolAccount()) return;`
>
> `if (!upstreamApiKey && !poolAccount && await tryCodexPoolFallback(`
> `  req, res, body, poolFallbackModels, isOpenAI ? 'openai' : 'anthropic', 'pool exhausted',`
> `)) {`
> `  return;`
> `}`

The newly added Codex fallback is reached only if `selectPoolAccount()` returns `true`. However, that selector still considers a null pool selection viable only when `openaiBackend !== null && isOpenAI`; it writes a 503 and returns `false` otherwise. Concrete failures:

* With a drained Claude pool, a configured Codex account, no API-key OpenAI backend, and `--pool-fallback=gpt-5.6-sol`, an OpenAI-shape Claude request gets the selector's 503 instead of reaching `tryCodexPoolFallback()`.
* Even if an API-key backend exists, the advertised Anthropic-shape failover gets the same early 503 because `isOpenAI` is false.

Thus the subscription-only setup and the Anthropic wire shape highlighted by this PR cannot use selection-time Claude→Codex failover. Make Codex viability part of selection, or attempt the provider-independent fallback before the selector can answer.

```suggestion
      // Try subscription failover before pool selection can terminate an
      // exhausted request. An empty pool can still retain the existing setup
      // error behavior inside tryCodexPoolFallback/selectPoolAccount as desired.
      if (!upstreamApiKey && pool.size > 0 && await tryCodexPoolFallback(
        req, res, body, poolFallbackModels,
        isOpenAI ? 'openai' : 'anthropic', 'pool exhausted',
      )) {
        return;
      }
      if (!selectPoolAccount()) return;
```

A regression test should exercise a drained non-empty Claude pool with (1) only a Codex subscription and (2) an Anthropic `/v1/messages` request; the existing provider-adapter tests do not traverse this selector.

## What's good

The failover status classification is careful: 429/5xx and pre-header transport failures defer, while client-caused 4xx responses remain visible. The pure chain pickers and byte-free-decline tests are also useful boundaries.

Scope note: this PR is 1,217 changed lines, so I prioritized the runtime behavior in `src/proxy.ts`, `src/codex-backend.ts`, `src/provider-adapter.ts`, and `src/compare.ts`, plus their new tests; I did not line-review every documentation and CLI/doctor change.

sprayberry-redline (dismissed):

**Verdict:** No blocking issues found — clean implementation with good self-correction visible in the fixup history.

## What I checked
Read the full diff (`gh pr diff 1145`, ~1150 net lines across src/codex-backend.ts, src/proxy.ts, src/provider-adapter.ts, src/doctor.ts, and the new src/compare.ts), plus the fixup commits appended to the same PR, and cross-checked the final state of the changed files in a shallow clone (`src/codex-backend.ts`, `src/proxy.ts`). CI check-rollup is all-green (CodeQL, build×3, compat, docker-cap-drop-smoke, live-test, test, wire-drift, triage all `pass`; `auto-merge` shows `skipping`, not a failure).

## Notable: the PR fixed its own bug mid-review
The diff includes a fixup commit renaming `pickNonCodexFallback` → `pickClaudeFallback` and changing its body from `!isCodexModel(m, slugs)` to `!isCodexModel(m, slugs) && /^claude/i.test(m)` (`src/codex-backend.ts:169-183` in the final state). The original version would fail over to any model that merely *wasn't* a Codex slug — a typo or third-party model name in the chain would get forwarded to the Claude pool, trading a recoverable 429 for an unrecoverable 404. The fix requires the fallback candidate to look like a Claude model and fails closed otherwise. I verified this is the code that actually lands (`grep` confirms no remaining references to the old `pickNonCodexFallback` name in `src/` or `test/`), and it's covered by new assertions in `test/codex-backend.mjs` (e.g. `pickClaudeFallback(['gpt-5.6-sol', 'llama-3.1-70b'], SLUGS) === null`).

## What's good
- `forwardToCodex`'s new `deferOnUnavailable` return-boolean contract is careful about not double-responding: it only returns `false` (defer) before any byte is written (on the initial non-ok upstream response), and every other exit path returns `true`.
- The `parsedBody = null` reset at `src/proxy.ts:3001` after swapping `body` to the Claude-bound payload is correctly threaded through — the downstream parse at `src/proxy.ts:3096` (`parsedBody ?? JSON.parse(body...)`) will re-parse the swapped body rather than serve a stale parse, which the PR's own comment flags as the exact bug class it was hunting.
- `writeCompareRecord`'s filename sanitization (`comparedModel.replace(/[^\w.-]/g, '_').replace(/\.{2,}/g, '.')`) closes path-traversal via both separator-stripping and dot-run collapsing.
- Test coverage is thorough for the new pure functions (`pickCodexFallback`/`pickClaudeFallback`, `failoverReadiness`) with explicit adversarial cases (typos, wrong-family models, pluralization edges).

## Minor
- `runCompare` in `src/compare.ts` always compares against a Codex account and never against the Claude pool, by design per its own doc comment — worth confirming that's the intended permanent scope rather than a v1 gap, but not a blocking concern since it's explicitly documented as deliberate.

I did not re-run the test suite locally (trusting the green CI rollup per policy) and did not deep-trace the streaming/SSE translation paths in `forwardToCodex` beyond the return-value contract change, since those are pre-existing and untouched by this diff.

sprayberry-redline (commented):

**Verdict:** 2 correctness issues to address in the new failover/compare paths.

## Findings

### High — `src/proxy.ts:2202-2207` — Codex fallback failures never continue to the API-key fallback

> `await forwardToCodex(`
> `  req, res, fallbackBody, creds, corsOrigin,`
> `  { ...SECURITY_HEADERS, 'x-dario-pool-fallback': fallbackModel },`
> `  upstreamTimeoutMs, verbose, shape,`
> `);`
> `return true;`

`tryCodexPoolFallback()` is documented to decline so its caller can continue to the API-key backend, but this call leaves `deferOnUnavailable` at its default `false` and then unconditionally returns `true`. Concrete failure: configure both a Codex account and an OpenAI-compatible backend, exhaust the Claude pool, and have the nominated Codex model return 429 (or 5xx/transport failure). `forwardToCodex` writes that error to the client, this helper returns `true`, and the callers return without trying the healthy OpenAI backend at lines 3067/3962. Thus preferring the subscription turns into making it terminal when unavailable.

```suggestion
    return await forwardToCodex(
      req, res, fallbackBody, creds, corsOrigin,
      { ...SECURITY_HEADERS, 'x-dario-pool-fallback': fallbackModel },
      upstreamTimeoutMs, verbose, shape, fetch, true,
    );
```

Please also cover this through the helper/caller path (Codex 429 plus a configured API-key backend), rather than only testing `forwardToCodex`'s pure decline behavior.

### Medium — `src/proxy.ts:2912-2924` — the recorded “verbatim” request is read from a later-mutated body

> `res.on('finish', () => {`
> `  void running.then((result) => {`
> `    let request: unknown = null;`
> `    try { request = JSON.parse(body.toString()); } catch { /* recorded as null */ }`
> `    const asObj = (request ?? {}) as Record<string, unknown>;`
> `    const written = writeCompareRecord({`
> `      ts: new Date().toISOString(),`
> `      path: urlPath,`
> `      shape: compareShape,`
> `      streaming: asObj.stream === true,`
> `      primaryModel: typeof asObj.model === 'string' ? asObj.model : '(unknown)',`
> `      comparedModel: compareTarget,`
> `      request,`

The finish callback closes over mutable `body`, while the new Codex→Claude failover later assigns `body = Buffer.from(swapped)` at line 3017. Concrete failure: send a GPT request with `x-dario-compare`, have Codex defer on 429, and fall back to `claude-sonnet-5`. When the primary response finishes, the record parses the swapped Claude body, so `request` and `primaryModel` say Claude even though the client sent the GPT model. That breaks the stated replay/audit contract and can mislabel the comparison. Snapshot the body used to start the comparison before routing can mutate it (or, if “verbatim” literally means wire input, snapshot before the earlier prefix/alias rewrites).

```suggestion
        const compareRequestBody = Buffer.from(body);
        const running = runCompare({
          body: compareRequestBody,
          // ...
        });
        // In the finish callback, parse compareRequestBody rather than body.
```

## What's good

The provider-selection helpers and unavailable-status matrix make the intended failover rules unusually explicit, and the exclusive-create compare-record test covers a real cross-process race. I reviewed every changed `src/` hunk and the targeted failover/compare tests; because this PR is over 800 changed lines, I did not line-review the documentation and lockfile changes.

sprayberry-redline (approved):

**Verdict: no blocking issues found — approving.** This is a large, well-tested PR (16 files, 72 new assertions per the changelog) and CI is fully green (14/14 checks). Scope of this review: `src/proxy.ts` (dispatch/selection changes), `src/codex-backend.ts` (`forwardToCodex` defer contract, `pickCodexFallback`/`pickClaudeFallback`), `src/provider-adapter.ts` (`poolFallbackOutcome`/`route`), `src/compare.ts` (new file, shadow compare), `src/doctor.ts` (`failoverReadiness`), and `src/cli.ts` (`dario add`). I did not re-derive line-by-line coverage of `test/*.mjs` beyond the new `header(...)` blocks, and did not exercise the pool's `select()` interaction with `canDefer` under a live rate-limit scenario — noting that as unverified rather than claiming it's clean.

**What's good**
- `forwardToCodex`'s decline contract is careful: it only returns `false` (signalling the caller can fail over) before any byte is written — `src/codex-backend.ts` gates the transport-failure defer on `!res.headersSent`, and the `!upstream.ok` branch defers before `res.writeHead` is called. That avoids the "abandon a stream half-sent" failure class the PR explicitly calls out.
- `pickClaudeFallback` fails closed on anything that doesn't look like `claude-*` (`src/codex-backend.ts`), with tests for a typo (`cluade-sonnet-5`) and a third-provider name (`llama-3.1-70b`) both correctly resolving to `null` rather than being swapped in and 404ing upstream.
- `writeCompareRecord` uses `wx` (exclusive create) with a numeric suffix retry loop to avoid same-millisecond record collisions across processes — a real race the PR's own changelog says was caught in review, and the fix (flag `wx` + loop) is a correct way to close it.
- The `selectPoolAccount()` gate fix (removing the `openaiBackend !== null && isOpenAI` requirement before deferring, and moving the `isOpenAI` guard down to the api-key-specific branch only) is exactly the right shape for the bug described: it stops 503ing Anthropic-shape/Codex-only deployments before the dispatcher ever runs, while `src/proxy.ts:3062` (`poolFallbackModel && openaiBackend && isOpenAI`) still correctly keeps the api-key path OpenAI-only since that route has no Messages translation.

**Minor / non-blocking**
- In the codex-primary path (`src/proxy.ts` ~2999-3023), after a codex account declines (429/5xx) and the request falls through to "Claude's turn" with `body` swapped to `claudeTarget`, `selectPoolAccount()` is called again further down. If the pool has seats (`pool.size > 0`, satisfying `canDefer`) but none are currently *eligible* (all cooling), `poolAccount` comes back null and the code re-enters `tryCodexPoolFallback` with the *same* `poolFallbackModels` chain — potentially re-hitting the same just-declined codex account within the same request before finally 503ing. Not a correctness bug (no loop, bounded to one extra round-trip) and plausibly rare, but worth a look if failover latency under simultaneous exhaustion ever becomes a complaint.
- `dario add <name>` mutates the shared `args` array in place (`args[1] = 'add'`, `args[2] = args[2] ?? who`) before delegating to `codex()`/`login()`. Works correctly for the current call sites since nothing reads `args` afterward, but it's a slightly fragile pattern for a codebase that otherwise treats `args` as an input rather than a scratch variable.

I did not run the test suite locally per policy; the CI check-rollup (all 14 checks passing, including `test`, `compat`, `live-test`, `wire-drift`) is the test signal here.

PR #1146, opened 2026-08-30
fix(compare): unique record names + exclusive create so concurrent compares cannot clobber each other

askalf:
Addresses the Medium review comment on #1145 (`src/compare.ts:207-213`): the compare record filename was `${ts}-${model}.json`, derived entirely from values that are **not unique across concurrent requests**. Two same-model comparisons that land in the same millisecond produce the same path, and `writeFileSync` with the default `w` flag silently truncates the earlier one. The sample that gets lost is precisely the concurrent traffic — the most interesting comparisons are the ones under load.

## Change
- Append a random 8-hex component (`randomUUID().slice(0, 8)`) to the filename.
- Create with `flag: 'wx'` (`O_EXCL`) so an existing name is an error, never an overwrite.
- On `EEXIST`, retry under a fresh name up to 5 times, then return `null`. Any other errno (ENOSPC, EACCES) rethrows immediately to the existing catch — it would fail identically under a different name, so retrying it is just latency on a served request's success path.

The function keeps its contract: never throws, returns the written path or `null`.

## Test plan
- `node test/compare.mjs` — existing assertions still hold; the filename assertions check `includes(model)` and `!includes(':')`, both unaffected by the added suffix, and the traversal assertion still passes since the suffix is hex.
- Concurrency check: N `writeCompareRecord()` calls with an identical `ts` and `comparedModel` now yield N distinct paths and N files on disk (previously 1 file).

Source ticket: 01M198XWNZA2B1FS01FDYQDNYH
Finding: 00MTFQH7X4A12CB83C2D6F0441 (Code Reviewer (GPT), review of #1145)


askalf:
Verification notes.

**Reproduced the defect first.** Old name is `${ts}-${safeModel}.json` with `writeFileSync` default flag `w`: 20 records sharing one `ts` + `comparedModel` collapse to **1 file on disk**, the last writer winning silently and `writeCompareRecord` returning a path for all 20 — the caller has no way to know 19 were lost.

**After the change**, the same 20 calls give 20 distinct paths, 20 files, and each file holds its own body (checked by tagging each record with a sequence number and reading them all back). Also re-checked the pre-existing guarantees the new suffix could plausibly have broken: the model stays greppable in the name, the traversal case `evil/../../name` still cannot escape the directory, and an unwritable directory still returns `null` rather than throwing. Confirmed against an extracted copy of the new function under `node` (repo deps aren't installed in my container); the same assertions are now in `test/compare.mjs` and CI's `live-test` job is SUCCESS on this branch, which is the authoritative run since it compiles `dist/` first.

**What I did not verify, and the exact next probe if a reviewer wants it:** the retry loop's EEXIST branch is unreachable in practice — a 32-bit random suffix makes a genuine collision vanishingly rare, so the tests exercise the happy path only. To cover it deliberately, stub `randomUUID` to a constant and assert that call N+1 with a fixed suffix returns `null` after 5 attempts rather than overwriting. I left that out because stubbing a `node:crypto` import means restructuring the module for injection, which is a larger change than the bug warrants.

Source ticket: 01M198XWNZA2B1FS01FDYQDNYH

sprayberry-redline (commented):

**Verdict:** No blocking issues found in the scoped collision fix.

I reviewed the PR-requested changes in `src/compare.ts:201-239` and the regression coverage in `test/compare.mjs:133-168`. The new random suffix removes timestamp/model aliasing, `flag: 'wx'` makes creation atomic rather than overwriting, and the bounded `EEXIST`-only retry preserves the function's best-effort/non-throwing contract. The test verifies distinct paths, file count, and per-record payload preservation for repeated same-timestamp/same-model writes.

**What's good:** The fix combines uniqueness with exclusive creation instead of relying on randomness alone, and avoids retrying deterministic filesystem failures such as `EACCES` or `ENOSPC`.

Review scope note: this PR reports 1,199 changed lines and includes the broader v6.0 feature branch; per the large-PR calibration budget, I focused on the compare-record overwrite fix named in the PR body and its tests, rather than re-reviewing every broader feature hunk.

askalf:
Superseded by #1145, which merged an equivalent fix: `writeFileSync` with `wx` (exclusive create) and an EEXIST retry, so same-millisecond compare records get distinct files across processes. This PR reaches the same guarantee with a `randomUUID` suffix rather than a counter — no functional gap either way, so closing in favour of what's already on master rather than reworking a conflicting branch.

The finding behind it was real and was fixed. Thank you.

PR #1147, opened 2026-08-30
fix(codex): reverse failover must positively validate the Claude target, not infer it from absence

askalf:
Fixes the finding on #1145 (`src/codex-backend.ts:169-170`). Source ticket: `01M198XWPBXARFZW72ZCQEFYDV`.

**Based on `feat/subscription-failover`, not `master`** — the code being fixed only exists on the PR branch.

## The bug

```ts
export function pickNonCodexFallback(models, slugs) {
  return models.find(m => !isCodexModel(m, slugs)) ?? null;
}
```

That is elimination, not a capability test. "Absent from the Codex slugs" also describes:

- **an api-key model** — `--pool-fallback=gpt-5.6-sol,gpt-4o` re-points a rate-limited subscription request at *Anthropic* as `gpt-4o`, turning a 429 into a 400 from a provider nobody named;
- **a typo** — `gtp-5.6-sol`, `claude-sonnett-5`. A misspelling is silently "not a slug", so it becomes the Claude target;
- **every entry in the chain, when discovery is down.** `getCodexModelSlugs` deliberately degrades to an EMPTY slug set on an unreachable backend. Under elimination an empty set means *nothing is a Codex model*, so the whole chain reads as Claude-servable — the failure mode is total exactly when the subscription is already unhealthy, which is when failover runs.

The last one is why this is worth fixing rather than documenting: the wrong answer is unconditional in the state that triggers the code path.

## The fix

`pickNonCodexFallback` → `pickClaudeFallback`, which requires a **positive** claim from the Claude side. New `src/claude-model.ts` owns the predicate:

- an explicit `claude:` / `anthropic:` provider prefix (the operator naming the provider outranks any name-shape guess);
- a `claude-` family id (`claude-opus-5`, `claude-sonnet-5[1m]`);
- a family shorthand the **catalog resolves** (`opus`, `sonnet1m`) — the same `resolveAliasAgainst` the Claude request path already uses, so the picker and the forwarder agree on what resolves.

A recognized non-Claude prefix is a definite **no**, not a fall-through: `openai:claude-opus-5` is pointing somewhere else on purpose.

An entry neither provider can serve is now **skipped**, so `claudeTarget` stays null, `canDefer` is false, and the subscription's real 429/5xx reaches the client instead of a fabricated 400 from Anthropic.

### Deliberate non-changes

- **A model listed by BOTH providers still stays with the subscription.** `!isCodexModel` is kept as the first conjunct, so a single-entry `--pool-fallback=<slug>` remains one-way exactly as pre-6.0.
- **An unreleased `claude-` id is still forwarded** (`claude-sonnet-6` → forwarded). The family prefix is the contract; only Anthropic can adjudicate a version, and a client asking for a model dario has not heard of yet must not be blocked by dario's catalog freshness.
- **A cold catalog degrades narrower, not wrong.** With no bases, explicit `claude-` ids and prefixes still qualify; only shorthands stop resolving.

## Changes

| file | |
|---|---|
| `src/claude-model.ts` | new — `isClaudeServableModel`, pure over an injected base set |
| `src/codex-backend.ts` | `pickNonCodexFallback` → `pickClaudeFallback(models, slugs, bases)` |
| `src/proxy.ts` | call site passes `getCachedBases()`; import + comment updated |
| `test/codex-backend.mjs` | +18 assertions |

`pickNonCodexFallback` is removed rather than deprecated — the only two references were this PR's own call site and its test.

## Verification

`npm run build` was not run locally (per operator policy — CI compiles). The pure rule was extracted with its call-site logic into a standalone file and run under `node --experimental-strip-types`: **23/23 pass**, covering every assertion added here.

CI is the authoritative build/test signal for `tsc --noEmit` and the full suite.

### Not verified

No live request. The picker is pure and fully covered; what remains unexercised is the same wire path #1145 already flags — one real 429 from the subscription, confirming the deferred request reaches the Claude pool as the selected model.


askalf:
### CI coverage on this PR — read before trusting the green

`.github/workflows/ci.yml` triggers on `pull_request: branches: [master]`. **This PR targets `feat/subscription-failover`, so the CI job — and with it `npm test` and the `test/codex-backend.mjs` assertions I added — did not run here.** Nothing is failing; the job was never scheduled. Worth stating plainly so the checkmarks are not read as more than they are.

What the green checks *do* prove:

- **`compat` — pass (1m41s).** Its `Install + build` step runs `npm ci && npm run build`, i.e. the real `tsc`. So the new `src/claude-model.ts`, the `pickClaudeFallback` signature, and both `src/proxy.ts` call-site edits **compile clean under strict TS**. That was the main thing I could not check locally (build wall).
- `actionlint`, `label`, `triage` — metadata only, unrelated to this change.

What is **not** covered here:

- the 18 new + 5 amended assertions in `test/codex-backend.mjs`. I ran the rule and the picker logic standalone under `node --experimental-strip-types` — **23/23 pass** — but that is my extraction of the logic, not the committed test file executing against `dist/`.

Both close automatically the moment this lands on `feat/subscription-failover`: #1145's own CI run then executes the committed suite against a real build. If you would rather see it before merging, re-running #1145's checks after merging this branch into it is the one-step confirmation.

### Open question left for review rather than chased in-run

`isClaudeServableModel` forwards **any** `claude-`-prefixed id, including one the catalog has never seen (`claude-sonnet-6` — asserted explicitly). That is deliberate: dario must not block a model Anthropic has shipped but the local catalog has not yet fetched. The cost is that a `claude-`-shaped typo still reaches Anthropic and 400s — strictly better than the old behaviour (where a `gpt-`-shaped typo did the same) but not zero.

Tightening it to catalog-membership would need the catalog to be authoritative and fresh, which it is not by design (`getCachedBases()` is empty before the first fetch). I would keep it as-is; flagging it so the choice is reviewed rather than assumed.

Source ticket: `01M198XWPBXARFZW72ZCQEFYDV`. Finding: `00MTFQZV1R0C35015637669A5B`.

sprayberry-redline (commented):

Verdict: No blocking issues found in the scoped fix.

## Review scope

This PR is 1,273 changed lines and is based on another feature branch, so I followed the large-PR scope in the review ticket/PR body rather than reviewing the entire stacked diff. I checked `src/claude-model.ts`, the `pickClaudeFallback` change in `src/codex-backend.ts`, the reverse-failover call site in `src/proxy.ts`, and the corresponding assertions in `test/codex-backend.mjs`. I did not review the unrelated compare/doctor/CLI portions inherited from the stacked branch.

## What’s good

The changed selection rule is grounded in positive Claude capability rather than Codex-slug absence, preserves explicit non-Claude provider prefixes as a hard rejection, and degrades conservatively when the Claude catalog is cold. The call site also gates deferral on a non-null target and clears the cached parsed request after swapping the body. The focused tests cover canonical IDs, aliases, prefixes, unknown names, cold-catalog behavior, and the no-bytes-written contract for deferred 429/5xx/transport failures.

askalf:
Closing in favour of #1151, which **adopts this PR's `src/claude-model.ts` verbatim** along with its 13 assertions.

To be clear about why: this fix is better than the one v6.0.0 shipped. My `/^claude/i` rejected `anthropic:sonnet` and every catalog shorthand, so a legitimate chain entry would silently never fail over — and it did nothing about `getCodexModelSlugs` degrading to an empty set, where selection-by-elimination calls every entry Claude-servable exactly when the backend is already unhealthy. `isClaudeServableModel` is the right answer.

#1151 exists only because this branch predates the #1145 merge and now conflicts across the whole v6.0.0 diff; rebasing it was more work than re-landing the module on master. The code is yours.

PR #1148, opened 2026-08-30
proxy: propagate forwardToCodex's deferred-unavailable result from tryCodexPoolFallback

askalf:
Fixes the #1145 review finding at `src/proxy.ts:2202-2207`.

## Problem
`tryCodexPoolFallback` dispatched `forwardToCodex` **without** `deferOnUnavailable` and then `return true` unconditionally, discarding the forward's boolean. So on the pool-exhausted path a Codex 429 / 5xx / transport failure was written straight to the client, and both call sites — the selection-time one (proxy.ts:3055) and the mid-flight 429 one (proxy.ts:3955) — saw `true` and returned instead of falling through to the configured api-key backend (`forwardToOpenAI`) and then to the honest 429/503.

The asymmetry is visible one screen away: the codex-first path at proxy.ts:3001 already passes `canDefer` and branches on `served`.

## Fix
Forward with `deferOnUnavailable = true` and return its result. `forwardToCodex` only ever declines before a byte is written (`!res.headersSent` guard, codex-backend.ts:772), so a stream in flight is never abandoned half-sent, and a 4xx that is our own fault still surfaces rather than being retried elsewhere.

## Test plan
- `test/codex-pool-fallback-defer.mjs` (added) — drives `tryCodexPoolFallback`'s callee contract through both call-site shapes: 429 and transport failure return false with nothing written; a 200 still returns true.
- Existing `test/codex-backend.mjs`, `test/codex-route-empty-pool.mjs` unchanged and green.

Source ticket: 00MTFV31NO768BA2FE6259D74E (finding 00MTFV31N739D27A78810E5089).


askalf:
## Verification status — read before merging

**CI scope caveat.** `.github/workflows/ci.yml` triggers only on `pull_request: branches: [master]`. This PR bases on `feat/subscription-failover`, so it got actionlint / labeler / the self-hosted drift jobs but **not** the build + `npm test` job. The tsc compile and the new test file are therefore unverified by CI on this branch — they will run when #1145 itself goes to `master`, with this change in it.

What I verified by reading, not by running:
- `forwardToCodex` is `Promise<boolean>`; `tryCodexPoolFallback` is declared `Promise<boolean>`, so `return await forwardToCodex(...)` type-checks.
- Positional args 10 and 11 are `fetchImpl: typeof fetch = fetch` and `deferOnUnavailable = false` (codex-backend.ts:584-585), which is what `fetch, true` fills.
- The decline is only ever taken before a byte is written — `!upstream.ok && unavailable` (codex-backend.ts:636) and the `!res.headersSent` guard in the catch (codex-backend.ts:772) — so neither call site can resume a half-sent stream. A 4xx that is our own fault still surfaces.

Not verified locally: `dist/` and `node_modules/` are absent in the agent container and the operator policy forbids a local build, so `node test/codex-pool-fallback-defer.mjs` has not been executed. **Exact next probe** if you want it before merge: `npm ci && npm run build && node test/codex-pool-fallback-defer.mjs` — 14 checks, expected all OK. The one assertion most likely to need a tweak is `x-dario-pool-fallback` on the codex-200 case, which is set as a header on the forward rather than via `res.setHeader`.

Source ticket: 00MTFV31NO768BA2FE6259D74E.

sprayberry-redline (commented):

Verdict: 1 high-severity correctness issue: the new mid-flight fallback remains unreachable for ordinary 429 responses.

## Findings

### High — `src/proxy.ts:3958-3964` / `test/codex-pool-fallback-defer.mjs:177-180`

> `if (await tryCodexPoolFallback(``  
> `  req, res, body, poolFallbackModels,``  
> `  isOpenAI ? 'openai' : 'anthropic',``  
> `  'pool exhausted mid-flight (429, no peer)',`  
> `)) {``  
> `  return;``  
> `}`

The new call is below the non-passthrough response-peeking branch at `src/proxy.ts:3642-3898`. For any ordinary non-passthrough 429, that earlier branch consumes `upstream.text()` and returns the enriched 429 at lines 3846-3898, so control never reaches this added call. The new test sends exactly such a request:

> `const chat = () => fetch(`${BASE}/v1/chat/completions`, {``  
> `  method: 'POST',`  
> `  headers: { 'Content-Type': 'application/json' },`  
> `  body: JSON.stringify({ model: 'claude-sonnet-5', messages: [{ role: 'user', content: 'ping' }] }),`

I reproduced this on the PR head after `npm run build`: `node test/codex-pool-fallback-defer.mjs` fails 13 of 15 checks; `codexSeen.responses` and `openaiSeen.calls` remain zero and the client receives the original 429. Thus a Claude pool account returning a normal JSON 429 still never tries Codex, and the propagation fix cannot affect the advertised mid-flight path. Invoke the fallback in the earlier peeked-429 branch after peer selection is exhausted (before writing the enriched response), while retaining the later branch for passthrough/unpeeked responses.

```suggestion
          if (await tryCodexPoolFallback(
            req, res, body, poolFallbackModels,
            isOpenAI ? 'openai' : 'anthropic',
            'pool exhausted mid-flight (429, no peer)',
          )) {
            return;
          }
          // Then try the OpenAI-compatible fallback before emitting enriched 429.
```

## What's good

Returning `forwardToCodex`’s boolean and enabling `deferOnUnavailable` is the correct local contract fix, and the test’s three intended outcomes (429, transport failure, and success) are well chosen once the test reaches that helper.

Review scope: this PR changes 1,627 lines. Per the large-diff review limit, I reviewed the PR-described `tryCodexPoolFallback` change, both call-site control flows, `forwardToCodex`’s defer contract, and the focused regression test; I did not review the unrelated compare/doctor/documentation expansion.

askalf:
Superseded by #1145, which merged an equivalent fix: `tryCodexPoolFallback` now passes `deferOnUnavailable` when an api-key backend could serve the same request, and returns what `forwardToCodex` reports instead of an unconditional `true`. Same mechanism this PR uses.

The finding behind it was real and was fixed. Thank you.

sprayberry-redline (approved):

**Verdict: no blocking issues found — the fix is correct and directly tested; approving.**

## What I checked
- The core fix: `forwardToCodex` in `src/codex-backend.ts` gained a `deferOnUnavailable` parameter and now returns `Promise<boolean>` instead of `Promise<void>`, returning `false` (writing nothing) only on a 429/5xx before any byte was written, or a transport failure while `!res.headersSent`.
- Both call sites that dispatch through the new `tryCodexPoolFallback` helper (`src/proxy.ts:2989` selection-time path, `src/proxy.ts:3958` mid-flight-429 path) now check the returned boolean and fall through to the next provider instead of assuming the request was served.
- The new symmetric-failover branch (`src/proxy.ts:2998-3009`, `codexCreds && decision.provider === 'codex'` block): `served = await forwardToCodex(..., canDefer)`, and only `if (served) return;` — otherwise it swaps the body to `claudeTarget` and falls through to the Claude path, correctly invalidating `parsedBody` so the stale pre-swap parse isn't reused.
- `test/codex-pool-fallback-defer.mjs` (new, 237 lines) exercises exactly the regression described in the PR body: a 429 and a transport-reset from the stub Codex backend both assert `codexSeen.responses` incremented (the subscription was actually tried), the client did **not** receive the Codex error (`res.status === 200`), and the api-key backend served it instead. This directly covers the bug this PR fixes.
- `pickCodexFallback`/`pickClaudeFallback` (`src/codex-backend.ts`) are pure functions with unit coverage in `test/codex-backend.mjs:791-813`, including the "not a real claude-* id" and typo-rejection cases.
- CI: the PR's check-rollup currently shows only the `auto-merge` workflow (skipped); no CI/build/test workflow run is recorded against head `e9b87c5` at review time — check-suites for "Claude" and "Cloudflare Workers and Pages" are still queued. Not treated as blocking per policy (pending, not failing), but worth confirming green before merge.
- Note: `gh pr view` reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against `master` — not a diff-content issue, but merge conflicts will need resolving regardless of review verdict.

## What's good
- The fix is narrowly scoped to the exact defect described (missing propagation of the forward's boolean), with a purpose-built regression test rather than a broad rewrite.
- The `parsedBody = null` invalidation after the mid-request Codex→Claude body swap is called out with a comment explaining why staleness there specifically reproduces "the exact bug class this release spent its whole review budget hunting" — good self-awareness, and I verified it's actually needed (the Claude path further down prefers `parsedBody` over `body` when present).
- Failure directions are handled correctly in both `forwardToCodex` exit paths (upstream non-ok and the catch block), both gated on `deferOnUnavailable && unavailable`/`!res.headersSent` so a stream already in flight is never abandoned.

I did not review the shadow-compare feature (`src/compare.ts`), the CLI/doctor/README changes, or the chain-parsing docs prose in depth — they're additive/unrelated to the propagation bug this PR targets and none of the changed lines I sampled there raised a correctness concern.

PR #1149, opened 2026-08-30
feat(codex): put the ChatGPT engine on the admin surface, and stop losing its requests (v5.5.91)

askalf:
A proxy serving GPT all day reported none of it. `/status`, `/accounts` and `/analytics` never named a codex account, and the codex route returned before `analytics.record()` or the request log ran. An operator dashboard read "0 requests" off a busy proxy.

## What changes

**`GET /codex`** (key-gated like `/accounts`): alias, credential expiry, refresh-due, the discovered slugs from the model *cache*, per-account request counts. Reads only what is on disk and in memory — no upstream call, no token refresh, no token in the answer. A status read must never spend or expose a credential.

**Codex requests are recorded.** `forwardToCodex` takes an `onDone` and reports exactly once on every exit (400 parse, upstream non-2xx, failed stream, success, thrown). The proxy records to analytics with the tokens from the terminal Responses event and writes the request-log line. A stream that failed upstream counts as the 502 it was, not the 200 that went on the wire.

**The claim is `chatgpt_subscription`**, added to `SUBSCRIPTION_CLAIMS` and the `subscription` billing bucket.

## The trade-off worth reading

That claim choice is not cosmetic. The overage guard (#288) attaches to analytics and halts the proxy on any claim outside `SUBSCRIPTION_CLAIMS`. My first cut stamped a claim the classifier did not know, and the hermetic test showed exactly what production would have done:

```
[dario] OVERAGE-GUARD HALTED: gpt-5.6-admin on account=fleet returned representative-claim=chatgpt_subscription … Returning 503 to new requests until `dario resume` or cooldown expires
```

One GPT request would have 503'd every Claude request on the box for 30 minutes. The test now asserts the guard stays up and the request lands in the `subscription` bucket with `unknown: 0`. Classifying it as subscription is correct on the merits — it is the user's ChatGPT plan — but if you would rather the codex engine have its own bucket rather than share `subscription`, that is a one-line change in `billingBucketFromClaim` and I will make it.

Two smaller unknowns: cost estimation prices GPT slugs through `pricingRateFor`, which has no entry for them, so `estimatedCost` on codex rows is whatever the fallback rate yields (the test saw $0.0001 for 17 tokens) — worth a follow-up if anyone reads that number. And `/codex` is not in the forbidden-path allowlist message; it is admin-key-gated exactly like `/accounts`, so I left the message alone.

## Tests

`test/codex-admin-surface.mjs` — real proxy on loopback, one stored account, a stub backend. 20 checks: the read is token-free and cache-only; after one request `/codex` counts it and shows the discovered slugs; `/analytics` has the perAccount row with the tokens, the perModel row, the claim; the guard has not halted; a 500 upstream is recorded as a failure with a non-zero window error rate. Full suite 154/154.

Consumer: the alf-dock Dario board reads `/codex` and falls back to `dario codex list` on older proxies.


sprayberry-redline (commented):

1 medium correctness issue to address; the admin surface and normal completion accounting otherwise read well.

## Findings

### Medium — transport failures lose the request model and stream mode (`src/codex-backend.ts:766`)

> `report(502, null, false, '\);`

This catch runs after `clientWantsStream` and `model` have already been parsed, but always reports a non-streaming request with an empty model. For a request such as `{ "model": "gpt-5.6", "stream": true }` where `fetchImpl` rejects, the timeout aborts, or the response reader throws, analytics records status 502 but the proxy callback substitutes its route-level fallback model and records `isStream: false`. That makes the newly added per-model and stream analytics incorrect specifically for thrown transport failures. Preserve the parsed dimensions here, as the upstream non-2xx path already does.

```suggestion
    report(502, null, clientWantsStream, model);
```

## What’s good

The callback’s once-only guard covers the explicit exits cleanly, `/codex` remains behind the existing auth check and uses cache/disk reads only, and the hermetic test exercises success, billing classification, token capture, and upstream non-2xx accounting.

sprayberry-redline (dismissed):

**Verdict:** No blocking issues found -- clean, well-scoped change with solid test coverage.

## What I checked
- `gh pr diff 1149` in full (440 lines across `analytics.ts`, `codex-backend.ts`, `proxy.ts`, `CHANGELOG.md`, and the new `test/codex-admin-surface.mjs`).
- Shallow-cloned `codex-admin-surface` to confirm `/codex` sits after the same `checkAuth(req)` gate as `/accounts` (src/proxy.ts:2304 vs the new block at src/proxy.ts:2384), and that `codexAccountNeedsRefresh`/`loadAllCodexAccounts` (src/codex-accounts.ts:102, 177) touch only on-disk metadata -- no token refresh, no upstream call.
- Traced `report()`'s single-fire guard (`src/codex-backend.ts:577-583`) against every exit path in `forwardToCodex` (parse failure, upstream non-2xx, stream-collapse failure, success, thrown error) -- each calls `report(...)` exactly once, and the `reported` flag plus the `try/catch` around `onDone` correctly prevent a double-report or a reporting failure from breaking the response.
- CI rollup: `triage` = SUCCESS, `auto-merge` = SKIPPED (no test-runner check configured on this repo's rollup); PR body states full suite 154/154 locally, consistent with the new 20-check hermetic test.

## What's good
- The overage-guard trade-off is called out explicitly in the PR body and covered by an assertion in the new test (proxy stays up after a codex request), which is exactly the kind of regression this change could have silently caused.
- `/codex` never returns `accessToken`/`refreshToken`, matching the stated no-token-in-the-answer constraint -- verified in `codex-admin-surface.mjs`'s check that the response body excludes the literal access token string.
- The upstream-failed-but-200-on-the-wire case (`report(upstreamFailed ? 502 : 200, ...)` at `src/codex-backend.ts:754`) is correctly derived from the same `upstreamFailed` flag used for the client-facing collapse decision, so analytics and the client never disagree about whether the request succeeded.

## Minor (non-blocking)
- The PR body itself flags that `pricingRateFor` has no entry for GPT slugs, so `estimatedCost` on codex analytics rows falls back to an unrelated rate. Author already scoped this as a follow-up, not part of this PR.

sprayberry-redline (commented):

Verdict: 1 medium-severity analytics correctness issue; the rest of the changed surface reads well.

## Findings

### Medium — `src/codex-backend.ts:833`: thrown streaming requests are recorded as non-streaming with no model

> `report(502, null, false, '');`

At this point the request body has already been parsed, so `clientWantsStream` and `model` are available. For a concrete failure, send `{ "model": "gpt-5.6-admin", "stream": true, ... }`, let the upstream return 200, then have its response body throw/reset while `reader.read()` is consuming the SSE stream. Control reaches this catch after the proxy has started a streaming response, but the new callback records `isStream: false`; it also discards the known model (the proxy callback happens to recover from `rawModel`, but direct `forwardToCodex` consumers cannot). This makes analytics and the request log misclassify exactly the failed-stream path the PR intends to expose. Preserve the parsed request metadata in this report call:

```suggestion
    report(502, null, clientWantsStream, model);
```

## What's good

The once-only reporting guard is a clean way to cover the many terminal paths without allowing observability failures to break served requests. The new hermetic test also checks the important cache-only/admin-key behavior, token accounting, failure accounting, and overage-guard classification.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — clean change, approving.

## What I checked
- Full diff (`gh pr diff 1149`, 449 lines across `analytics.ts`, `codex-backend.ts`, `proxy.ts`, `CHANGELOG.md`, `package.json`/lock, and the new `test/codex-admin-surface.mjs`).
- Confirmed against the actual head commit (`1525654`, cloned read-only) that `GET /codex` is registered *after* the `checkAuth(req)` gate at `proxy.ts:2387`, same as `GET /accounts` at `:2416` — the PR's claim that `/codex` is key-gated like `/accounts` holds.
- Traced the `deferOnUnavailable` path in `codex-backend.ts`: both defer points (`:677-679` on a 429/5xx upstream response, `:825-827` on a pre-`headersSent` transport failure) `return false` *before* calling `report(...)`, matching the PR body's claim that a declined-to-Claude-pool request reports nothing and the Claude path records it instead.
- Verified the `report()` closure (`codex-backend.ts` around the new `forwardToCodex` signature) is idempotent via its `reported` flag and wraps `onDone` in try/catch, so a reporting failure can't break a served request.
- Checked CI: `gh pr checks 1149` — all 14 checks green (build ×3, test, live-test, compat, CodeQL, wire-drift, etc.), nothing pending or failing.

## What's good
- The `chatgpt_subscription` claim/billing-bucket wiring is well-reasoned and the PR body documents a real incident it prevents (the overage guard halting on an unrecognized claim) — the new test at `test/codex-admin-surface.mjs:426-432` specifically asserts the guard stays up and the request lands in the `subscription` bucket, which is the right regression to pin.
- `peekCodexModelSlugs` (`codex-backend.ts`) reads only the in-memory cache — no upstream call, no token in the `/codex` response — consistent with the stated "a status read must never spend or expose a credential" invariant, and the new test asserts the token string is absent from the JSON body (`test/codex-admin-surface.mjs:397`).
- Stream failures that end 200-on-the-wire are correctly recorded as the 502 they were (`codex-backend.ts` `report(upstreamFailed ? 502 : 200, ...)`), with a test exercising the failing-backend path (`test/codex-admin-surface.mjs:434-445`).
- New behavior is covered by a dedicated hermetic test file exercising the pre/post-request `/codex` and `/analytics` shapes, the overage guard, and the failure-recording path — good breadth for the size of the change.

## Not fully verified (noted per budget)
- I did not independently verify `codexAccountNeedsRefresh`/`loadAllCodexAccounts` internals in `codex-accounts.ts` (unchanged by this diff) — trusted as pre-existing, exercised by the new test's `needsRefresh` assertion.

PR #1150, opened 2026-08-30
test: prove the pool-exhaustion gate through a live request (dario#1145 review)

askalf:
The v6.0.0 pool-exhaustion gate is already fixed on master (eabfed7b, "fix: the failover gate, not just the dispatcher"). This adds the proof that fix explicitly says it still owes.

## Why

`poolFallbackOutcome()` states the matrix, and its own doc comment says why that is not enough:

> A caveat worth keeping in view: this is a SPECIFICATION test surface, not a wiring one. It cannot prove proxy.ts asks it the right question at the right moment — that is precisely what broke — and only a live request through the selector can. Treat green here as necessary, never sufficient.

That is exactly how the bug survived: `route()` was updated, the dispatcher was added and tested, and the gate in front of both was never touched, so every routing test passed while `selectPoolAccount()` still 503'd a subscription-only deployment before `tryCodexPoolFallback` could run. Nothing in the suite goes through that selector. This file does.

## What it proves, through a real request

`test/pool-exhaustion-gate-wiring.mjs`, 16 checks, hermetic — no network, no real credentials.

The pool is drained **the way production drains it**, not by forging internals: the first request 401s upstream, `markAuthFailure` puts the single account into cool-down, and the next `select()` returns null with `pool.size` still 1. That "has accounts, none usable" state is the one the old gate misjudged, and it is only reachable through the handler.

`~/.dario/backends` is deliberately absent, so `openaiBackend` is null for both proxies — a configured api-key backend hides half the bug.

- **Anthropic shape, no api-key backend → codex serves it** (was: 503 before the dispatcher ran). Asserts 200, that the stub subscription actually received the request, that the body carries the discovered slug, and that `x-dario-pool-fallback` announces the swap.
- **OpenAI shape with no api-key backend beside the subscription → codex serves it** (the other half of the old `openaiBackend !== null && isOpenAI` conjunction).
- **Deferred but unservable → the honest 503**, `error: "No accounts available in pool"` — not a fall-through to the Claude path with an empty bearer and a confusing upstream 401. Also asserts the subscription was never asked to serve a slug it does not list.

After the drain the Anthropic seam is switched to *throw*, so a regression that routes back to the pool fails loudly rather than passing quietly, and a final assertion pins the upstream call count at exactly the two draining requests.

## Test plan

`npm run build && node test/pool-exhaustion-gate-wiring.mjs` (16 passed, 0 failed expected). Picked up automatically by `npm test` — `test/all.test.mjs` globs `test/*.mjs` and this file is not in `EXCLUDED`.

Source ticket: 01M19FSM1GT0CYCNK1BYS39DW3 (finding 00MTFTYPNAB71ECB083561D9FA, from the askalf/dario#1145 review).

Not run locally: the agent container has no `node_modules` or `dist/`, and local builds are barred by policy — CI compiles and runs it.

### DIFF
--- a/test/pool-exhaustion-gate-wiring.mjs
+++ b/test/pool-exhaustion-gate-wiring.mjs
@@ -1,0 +1,255 @@
+#!/usr/bin/env node
+// dario#1145 review finding — the pool-exhaustion GATE, proved through a live
+// request instead of through the specification surface.
+//
+// The bug (fixed in eabfed7b, on master): selectPoolAccount() computed
+//
+//     fallbackViable = poolFallbackModel !== null && openaiBackend !== null
+//                      && isOpenAI && pool.size > 0
+//
+// and wrote its 503 before returning false, so a drained pool answered
+// "No accounts available in pool" BEFORE tryCodexPoolFallback could run —
+// whenever there was no api-key backend, or the client spoke Anthropic. That is
+// the subscription-only deployment v6.0.0 exists for.
+//
+// It survived review because every routing test passed: route() was updated,
+// the dispatcher was added and tested, and the gate in front of both was never
+// touched. poolFallbackOutcome() now states the matrix, and its own doc comment
+// (src/provider-adapter.ts) says outright that it is a SPECIFICATION surface —
+// it cannot prove proxy.ts asks it at the right moment, which is precisely what
+// broke, and "a live request still owes proof". This file is that proof.
+//
+// The pool is drained the way production drains it, not by forging internals:
+// the first request 401s upstream, which puts the single account into
+// markAuthFailure cool-down, so the NEXT select() returns null with pool.size
+// still 1. That is exactly the "has accounts, none usable" state the gate
+// misjudged — and it is reachable only through the handler, which is the point.
+//
+// Hermetic: HOME is a mkdtemp'd dir holding one Claude account and one codex
+// account and NO backends dir (openaiBackend stays null, the half of the bug
+// that a configured api-key backend hides); the ChatGPT backend is a local
+// stub; the Anthropic upstream is ProxyOptions.fetchImpl. No network.
+
+import { createServer } from 'node:http';
+import { mkdtemp, mkdir, writeFile } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+let pass = 0, fail = 0;
+const check = (name, cond, detail) => {
+  if (cond) { console.log(`  OK ${name}`); pass++; }
+  else { console.log(`  FAIL ${name}${detail ? ' :: ' + detail : ''}`); fail++; }
+};
+const header = (n) => console.log(`\n=== ${n} ===`);
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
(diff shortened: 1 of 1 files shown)


sprayberry-redline (approved):

**Verdict:** No blocking issues found — a well-scoped, purely-additive test file that closes the wiring gap the PR body describes.

## What I checked
- `gh pr view` (title/body/files/CI rollup) and `gh pr diff` for the full 255-line diff — the PR adds exactly one new file, `test/pool-exhaustion-gate-wiring.mjs`, with no production-code changes.
- CI rollup: all 13 checks green (`build`, `test`, `live-test`, `CodeQL`, `docker-cap-drop-smoke`, etc.), no pending or failing checks.
- The test's own logic: pool-draining sequence, the three assertion blocks (Anthropic-shape fallback, OpenAI-shape fallback, deferred-unservable 503), and the final invariant checks (`anthropicCalls.total === 2`, `codexSeen.models >= 1`).

## Findings
None blocking. Since this PR only adds a test file and does not touch shipped logic, the main risk surface is test correctness/flakiness rather than a production bug, and I didn't find a case where the test would pass despite the described regression recurring, or fail spuriously on correct behavior — the drain sequence (`anthropicMode = '401'` -> real 401 -> `markAuthFailure` cool-down -> `anthropicMode = 'forbidden'` to catch any later fall-through) matches the failure mode described in the PR body and in `eabfed7b`.

**Minor (non-blocking):**
- `test/pool-exhaustion-gate-wiring.mjs:157-174`: the two `startProxy(...)` servers (`BASE`, `BASE_NOSERVE`) are never closed, unlike `codexStub.close()` at line 260. Harmless here because `process.exit(...)` on line 261 terminates the process regardless of open handles, but if this file is ever refactored to not exit explicitly (e.g. imported as a module rather than run standalone), the open listeners would leak.

## What's good
- Drains the pool through the real handler path (a live 401 -> cool-down) rather than forging internal state, which is exactly the "wiring, not specification" proof the PR body argues was missing.
- Arms `anthropicMode = 'forbidden'` after the drain so any regression that silently falls through to the Claude path fails loudly instead of passing quietly — a good anti-regression design choice.
- Deliberately omits `~/.dario/backends` so `openaiBackend` stays null, covering the half of the original bug's conjunction that a configured api-key backend would have hidden.
- CI (`test` job) is green, confirming the file is picked up by the `test/*.mjs` glob and passes as described.

sprayberry-redline (commented):

Verdict: No blocking issues found in the added pool-exhaustion wiring test.

I reviewed the full added test and traced its setup and assertions through the account-pool selector, 401 auth-cooldown handling, Codex model discovery, both fallback wire shapes, and the final unservable 503 guard. The two proxy instances intentionally share the same in-process pool account state: the first 401 cools down the account, the second request receives its own 401 after selecting before that shared cooldown is established, and subsequent requests exercise the exhausted-pool path. I also checked that the fixed ports do not overlap the other default-suite proxy tests and that the file is included by `test/all.test.mjs`.

**What’s good**

The test reaches the regression state through a real handler request rather than mutating pool internals, makes accidental post-drain Claude dispatch fail loudly, and checks both the successful subscription handoff and the honest failure path. The upstream call-count and discovered-model assertions are useful safeguards against a test that passes through the wrong route.

PR #1151, opened 2026-08-30
fix(codex): reverse failover tests what a model IS, not what it isn't (v6.0.1)

askalf:
Adopts `src/claude-model.ts` from #1147, which is a better answer than the one v6.0.0 shipped. Credit where it's due — the fleet Developer got this right and I didn't.

## What was wrong

v6.0.0 shipped `pickClaudeFallback` as `/^claude/i`. The **direction** was right (fail closed rather than swap in a model the pool will 404 on); the **test** was wrong three ways:

| case | v6.0.0 | now |
|---|---|---|
| `anthropic:sonnet` | rejected | accepted — an explicit provider prefix is the operator naming the provider, which outranks guessing from name shape |
| `opus`, `sonnet1m` | rejected | accepted — resolved against the catalog |
| empty slug list (discovery degraded) | every entry called Claude-servable | tested positively, so it fails closed |

The second is the ugly one: a legitimate chain entry would **silently never fail over** — the same class of quiet wrongness the check existed to prevent. v6.0.0's own comment documented it as an accepted limitation instead of fixing it.

The third matters most in the exact situation failover is for: when discovery degrades, `getCodexModelSlugs` returns an EMPTY set, and selection-by-elimination then calls *every* chain entry Claude-servable — worst behaviour precisely when the backend is already unhealthy.

## The fix

`isClaudeServableModel(model, bases)` is a positive capability test resolved against the live catalog; `proxy.ts` passes `getCachedBases()` at the call site.

13 assertions from #1147, plus 2 of mine pinning the specific cases the shipped regex got wrong so a revert is loud rather than silent:

```
✅ REGRESSION: anthropic: prefix was rejected by the shipped /^claude/i
✅ REGRESSION: a catalog shorthand was rejected too
```

227 codex-backend assertions, 154 files, preflight and README gates clean. README and `docs/multi-account-pool.md` corrected — both promised `claude-*` only and explicitly disclaimed aliases.

## Housekeeping

Supersedes #1147. #1146 and #1148 fixed the same findings as the merged #1145 by equivalent means (`wx`+EEXIST retry, and `deferOnUnavailable` propagation respectively) and are closed unmerged.

All three branched pre-merge and now conflict, and each stale PR keeps minting review tickets — the gating reviewer has burned ~$10 today in capped runs on them.


sprayberry-redline (dismissed):

## What I checked
- `src/claude-model.ts` (new) — `isClaudeServableModel`: prefix handling (`claude:`/`anthropic:` force-accept unless the prefix is a recognized non-Claude one, e.g. `openai:claude-opus-5` correctly rejects since `m.slice(0, idx)` yields `openai`), `claude-*` canonical id fast path, and shorthand resolution via `resolveAliasAgainst` (verified signature/behavior against `src/model-catalog.ts:212-221` in the base repo — it lowercases, checks `FAMILY_ALIASES`, and handles the `1m` suffix pattern, matching what this file assumes).
- `src/codex-backend.ts:156-186` — `pickClaudeFallback` now calls `isClaudeServableModel(m, bases)` instead of the old `/^claude/i` test; correctly still gated by `!isCodexModel(m, slugs)` first.
- `src/proxy.ts:3016` — call site now threads `getCachedBases()` through, which is the same live-catalog accessor already used elsewhere in this file (`model-catalog.ts` import at proxy.ts:29, used at proxy.ts:252) — consistent with existing usage.
- `test/codex-backend.mjs:850-874` — new test block covers canonical ids, `[1m]` variants, case-insensitivity, shorthand resolution, both provider prefixes, negative cases (gpt model, cross-provider prefix, unknown shorthand, empty string), cold-catalog behavior, and two explicit regression pins against the old `/^claude/i` behavior (`anthropic:sonnet`, `opus` shorthand). Good coverage of the boundary conditions this function actually branches on.
- CI check-rollup: all green (CodeQL, actionlint, analyze, build×3, compat, docker-cap-drop-smoke, live-test, triage×2, test, validate-package-json, wire-drift). `auto-merge` shows `skipping`, which is expected/non-blocking for a non-bot-authored PR.
- Docs (README.md, docs/multi-account-pool.md, CHANGELOG.md) updated consistently with the new positive-capability-test behavior described in the code.

No correctness, reuse, or efficiency issues found in the changed lines. The fallback logic (empty-slug-list degradation, cross-provider prefix rejection) is exercised by the new tests and traces correctly through the code.

## What's good
The docstrings and comments clearly explain the *why* (fail-closed direction correct in v6.0.0, but the elimination-based test was wrong three ways) without padding, and the two regression-pinning tests are a nice touch to make a future revert loud rather than silent.

sprayberry-redline (commented):

Verdict: 2 correctness issues remain in the positive-capability classifier; the surrounding failover wiring and regression coverage otherwise read well.

## Findings

### High — `src/claude-model.ts:54` still treats every `claude-*` typo as servable

> `if (m.startsWith('claude-')) return true;`

This is still a name-shape test rather than the documented positive catalog test. For the concrete chain `--pool-fallback=gpt-5.6-sol,claude-sonnet-6`, with `gpt-5.6-sol` selected and Codex returning 429, this line classifies the nonexistent `claude-sonnet-6` as servable regardless of `bases`. `pickClaudeFallback` then defers the Codex response and rewrites the request to that typo, producing the exact wrong-provider 4xx this change says it prevents. The new helper's own leading comment names `claude-sonnet-6` as a case that should be skipped.

```suggestion
  // Require canonical ids (including an optional [1m] variant) to exist in
  // the live/baked catalog rather than accepting the family prefix alone.
  const base = m.endsWith('[1m]') ? m.slice(0, -4) : m;
  if (m.startsWith('claude-')) return bases.includes(base);
```

The same validation should be applied after stripping `claude:` / `anthropic:`; an explicit provider can force routing, but it cannot make an unknown model servable.

### Medium — `src/claude-model.ts:60` drops aliases that the normal Claude path supports

> `return resolveAliasAgainst(m, bases) !== null;`

`resolveAliasAgainst` recognizes only floating family aliases (`opus`, `sonnet1m`, etc.). It does not include the existing pinned aliases (`opus48`, `opus47`, `opus46`, `sonnet46`) or operator-defined `--model-alias` values. For example, `--model-alias=backup=claude-sonnet-5 --pool-fallback=gpt-5.6-sol,backup` works when `backup` is the incoming request model, but during reverse failover this line returns false for `backup`; the Codex 429 is surfaced and the configured Claude fallback silently never runs. `opus48` fails the same way even though `resolveClaudeAlias` explicitly supports it. This contradicts the changed `pickClaudeFallback` comment that the old `--model-alias` limitation is gone.

```suggestion
  // Resolve candidates through the same built-in and operator alias pipeline
  // used by request routing, then validate the resolved target against bases.
  // Pass the configured alias map (or a resolver callback) into this helper so
  // classification and eventual forwarding cannot disagree.
```

Please add reverse-failover tests for one pinned alias and one configured alias, not only the catalog family shorthands.

## What's good

The change replaces selection-by-elimination with an explicit Claude predicate, threads the cached catalog into the live call site, and adds focused regression assertions for provider prefixes, cold-catalog behavior, and family shorthands. The stale-`parsedBody` handling in the existing fall-through path also ensures the selected fallback model is actually forwarded.

sprayberry-redline (dismissed):

## What I checked

- `gh pr diff` for the full change (320 lines: `src/claude-model.ts` new, `src/codex-backend.ts`, `src/proxy.ts` call site, `test/codex-backend.mjs`, plus docs/changelog/version bumps).
- Shallow-cloned to confirm the signatures `resolveClaudeServable`/`isClaudeServableModel` depend on: `resolveAliasAgainst(model, bases)` in `src/model-catalog.ts:212` (family-shorthand resolution against a live base set) and `getCachedBases()` / `applyModelAlias` / `resolveClaudeAlias` in `src/proxy.ts`, all of which match the call site added at `src/proxy.ts:3018-3025`.
- CI rollup for head `5880066d`: all 15 checks green (build, test, compat, live-test, wire-drift, CodeQL, etc.) — no pending or failing checks.

## Correctness

`src/claude-model.ts:47-59` (`resolveClaudeServable`) replaces the old `/^claude/i` elimination test with a positive resolve-then-validate check:

```ts
const idx = m.indexOf(':');
if (idx > 0) {
  const prefix = m.slice(0, idx);
  if (!CLAUDE_PREFIXES.has(prefix)) return null;
  m = m.slice(idx + 1);
  if (!m) return null;
}
const resolved = (resolve ? resolve(m) : (resolveAliasAgainst(m, bases) ?? m)).trim().toLowerCase();
const base = resolved.endsWith('[1m]') ? resolved.slice(0, -4) : resolved;
return bases.some((b) => b.toLowerCase() === base) ? resolved : null;
```

I traced the cases the PR body calls out as previously-broken: `anthropic:sonnet` strips the prefix, resolves `sonnet` via `resolveAliasAgainst` against the catalog to a real base, and validates — correctly accepted. `claude-sonnet-6` (a plausible-looking typo) has no `:` prefix, `resolveAliasAgainst` returns null so it falls through unchanged, and the `bases.some(...)` check correctly rejects it since it's not in the base set — this is the regression test at `test/codex-backend.mjs` explicitly pinning that case, and it holds. An empty `slugs` list at the call site no longer causes elimination-by-default because this function's result no longer depends on `slugs` at all — it's a positive test independent of the Codex side, closing the discovery-degraded gap the PR describes.

The `src/proxy.ts:3018-3025` call site correctly threads `getCachedBases()` and a resolver that composes `applyModelAlias` (operator aliases) then `resolveClaudeAlias` (pinned + catalog aliases) — matching the same resolution order used for a live request model per `src/proxy.ts:2847,2867`, so the classifier and the actual forwarded id can't disagree.

`src/codex-backend.ts:180-197` (`pickClaudeFallback`) now iterates and returns `resolveClaudeServable`'s resolved id rather than the raw chain entry — correctly noted in the comment that the body swap runs after the proxy's alias pass, so returning the resolved canonical id (not the alias) is required for Anthropic not to 400 on an unresolved alias.

## Test coverage

The new test block in `test/codex-backend.mjs` (case-insensitivity, `[1m]` variants, both prefixes, shorthands, empty/typo rejection, empty-base-set, resolver-based alias resolution, and two explicit regression pins for the cases the old regex got wrong) covers the branches I traced above. Nothing changed-and-untested stood out.

## What's good

Clear before/after documentation in the CHANGELOG, README, and docs — all three consistently describe the same positive-capability-test semantics, and the docs updates match what the code actually does.

sprayberry-redline (commented):

Verdict: 1 medium correctness issue remains in provider-prefixed operator-alias handling; the positive catalog validation and canonical-id swap otherwise read well.

### Findings

**Medium — `src/proxy.ts:3023`: an operator alias targeting `claude:` / `anthropic:` is rejected only on reverse failover**

> `              (m) => resolveClaudeAlias(applyModelAlias(m, modelAliases) ?? m),`

The normal request path explicitly permits alias targets with provider prefixes: it applies the operator alias and then parses/strips that prefix. This fallback resolver instead sends the aliased target directly through `resolveClaudeAlias`, which leaves a value such as `anthropic:sonnet` unchanged. `resolveClaudeServable` already performed its prefix check on the original chain entry before invoking this resolver, so it then compares the still-prefixed string against catalog bases and returns `null`.

Concrete scenario: configure `--model-alias=backup=anthropic:sonnet` and `--pool-fallback=gpt-5.6-sol,backup`. When the Codex request gets a deferable 429/5xx, `backup` resolves to `anthropic:sonnet`, fails the base comparison, and `canDefer` is false. The original error surfaces even though sending `backup` as a normal request routes successfully to Claude. This contradicts the documented request-resolution parity and silently disables the configured fallback.

Resolve the operator alias before provider-prefix classification, then strip/validate a Claude prefix (and reject an OpenAI/Codex prefix) before applying pinned/catalog aliases. For example, centralize the existing request pipeline and pass that result here:

```suggestion
              (m) => resolveClaudeFallbackAlias(m, modelAliases),
```

Please add a regression assertion for an operator alias targeting `anthropic:sonnet` (and, ideally, one confirming an alias targeting `openai:gpt-4o` remains ineligible for Claude fallback).

### What's good

The follow-up commits correctly validate the resolved ID against actual catalog bases, preserve `[1m]` handling, return the canonical model that will be put into the body, and cover direct prefixes, family aliases, pinned aliases, typos, and cold-catalog behavior. I reviewed the current head `9ab0129`, including the commits added after the ticket's original `e7ff802` scope.

sprayberry-redline (dismissed):

## What I checked

- `gh pr diff` for the full change (329 lines): new `src/claude-model.ts`, the `pickClaudeFallback` rewrite in `src/codex-backend.ts`, the call site in `src/proxy.ts`, the new/updated assertions in `test/codex-backend.mjs`, plus CHANGELOG/README/docs/version bumps.
- The commit history for this head (`9ab01292`) vs. the immediately-preceding commit (`5880066d`), which had a full green CI run (CI, CodeQL, Compat test, Live test, Wire drift all `success`): the only delta is a one-line README file-count edit (58→59 files), consistent with the new `src/claude-model.ts` file added by this PR. No code changed after the green run.

## Correctness

`src/claude-model.ts`'s `resolveClaudeServable` replaces the old `/^claude/i` elimination test with a positive resolve-then-validate check:

```ts
const idx = m.indexOf(':');
if (idx > 0) {
  const prefix = m.slice(0, idx);
  if (!CLAUDE_PREFIXES.has(prefix)) return null;
  m = m.slice(idx + 1);
  if (!m) return null;
}
const resolved = (resolve ? resolve(m) : (resolveAliasAgainst(m, bases) ?? m)).trim().toLowerCase();
const base = resolved.endsWith('[1m]') ? resolved.slice(0, -4) : resolved;
return bases.some((b) => b.toLowerCase() === base) ? resolved : null;
```

Tracing the cases the PR body calls out as previously broken: `anthropic:sonnet` strips the prefix, resolves `sonnet` against the catalog to a real base, and validates — correctly accepted, where the old `/^claude/i` regex rejected it outright. `claude-sonnet-6` has no `:` prefix, doesn't resolve to a known alias, and fails the `bases.some(...)` membership check even though it has the `claude-` prefix — correctly rejected, matching the new regression test pinning that exact case. `isCodexModel`/`slugs` no longer gate this function at all, so an empty `slugs` list (discovery degraded) can no longer cause "every entry is Claude-servable by elimination" — the gap the PR describes is closed by construction, not by a special-cased empty check.

`src/codex-backend.ts`'s `pickClaudeFallback` now returns the *resolved* canonical id rather than the raw chain entry, and the `src/proxy.ts` call site composes `applyModelAlias` (operator aliases) then `resolveClaudeAlias` (pinned + catalog aliases) as the resolver — matching the same resolution order applied to a live request model elsewhere in `proxy.ts`. This is consistent with the stated reason: the body swap runs after the proxy's own alias pass, so returning an unresolved alias would 400 upstream.

## Test coverage

The new `isClaudeServableModel`/`resolveClaudeServable` block in `test/codex-backend.mjs` covers case-insensitivity, `[1m]` variants, both provider prefixes, catalog shorthands, empty/typo rejection, an explicitly empty base set, and resolver-based alias resolution (including an operator alias pointing at a non-Claude model, which is correctly refused). Two assertions are pinned explicitly as regressions against the old `/^claude/i` behavior. Nothing changed in this diff looks untested.

## CI

Auto-merge is the only check attached directly to the current head (skipped, as expected for a non-bot PR). The prior commit on this branch ran the full suite (CI, CodeQL, Compat, Live test, Wire drift) green, and the only change since is the one-line doc edit noted above, so I'm treating that as the operative signal per the "trust CI" policy rather than waiting on a re-run of a doc-only diff.

## What's good

CHANGELOG, README, and `docs/multi-account-pool.md` are all updated consistently with the new positive-capability-test semantics, and they match what the code does.

askalf:
Conflict resolved by merging \`master\` (90b0c39) in — merge commit 25d46ad. mergeable is back to MERGEABLE.

The only conflict was CHANGELOG.md: #1149/#1150 landed their \`GET /codex\` + analytics entries under the same \`## [6.0.1] - 2026-08-30\` heading this branch writes its failover-classifier entry under. Both are kept — classifier bullet first, then master's \`### Added\` / \`### Fixed\` sections. \`src/analytics.ts\`, \`src/codex-backend.ts\` and \`src/proxy.ts\` auto-merged with no textual conflict.

Also re-checked the README "59 files" claim from 9ab0129 after the merge: master added no new source file (analytics/codex-backend/proxy modified only), and the merged tree has 59 \`.ts\` under \`src/\`, so the number still holds.

CI on the merge commit: CI, CodeQL, Live test, Wire drift, actionlint all green; Compat test still queued at time of writing. Source ticket 01M19R8SP9W15RV4R6DHNMH7H6.

sprayberry-redline (approved):

## What I checked
- Full diff (`gh pr diff 1151`): `CHANGELOG.md`, `README.md`, `docs/multi-account-pool.md`, new `src/claude-model.ts`, `src/codex-backend.ts`, `src/proxy.ts`, `test/codex-backend.mjs`.
- Surrounding context via a shallow clone: `resolveClaudeAlias`/`applyModelAlias` (`src/proxy.ts:251-291`) and `resolveAliasAgainst` (`src/model-catalog.ts:212-221`) to confirm the resolver passed at the `pickClaudeFallback` call site (`src/proxy.ts:213-215`) matches the same alias-resolution order used for a live request (operator `--model-alias` first, then catalog/pinned aliases).
- CI rollup (`gh pr checks 1151`): all 14 required checks pass (test, build×3, live-test, compat, CodeQL, analyze, wire-drift, etc.), `auto-merge` is intentionally skipped.

## Correctness
`resolveClaudeServable` (`src/claude-model.ts:88-102`) implements the two-step resolve-then-validate the PR describes: strip a recognized `claude:`/`anthropic:` prefix (a non-Claude prefix like `openai:` is a definite reject, not a fall-through), resolve the remainder through the supplied resolver (or the default catalog-shorthand resolver), then require the resolved id (minus `[1m]`) to be a literal member of `bases`. This closes the two real gaps called out in the PR body:
- A `claude-` prefix alone no longer counts as proof of servability — `claude-sonnet-6` correctly returns `null` because it isn't in `bases`, verified by the "REGRESSION" tests at `test/codex-backend.mjs` (both bare and behind an `anthropic:` prefix).
- An empty `slugs` set no longer makes `pickClaudeFallback` (`src/codex-backend.ts:206-215`) claim every entry is Claude-servable, since servability is now a positive catalog lookup rather than "not a codex slug."

The call site in `src/proxy.ts:207-216` passes `getCachedBases()` and a resolver built from `resolveClaudeAlias(applyModelAlias(m, modelAliases) ?? m)` — the same resolution chain a real request model goes through — and the function returns the *resolved* canonical id (not the raw chain entry), which matches the comment that the body swap happens after the proxy's own alias pass. I traced this against `applyModelAlias`/`resolveClaudeAlias` in `proxy.ts` and the order and null-handling line up.

Test coverage is thorough and specifically pins the two behaviors this PR is fixing (`claude-sonnet-6` typo rejection, `anthropic:`/shorthand acceptance, cold-catalog and empty-bases edge cases, alias/pinned-alias resolution through a supplied resolver).

## What's good
- The new module's docstring lays out the failure mode being fixed (typo silently swapped into a 400, empty-slugs case silently degrading), which made verifying the fix against the diff straightforward.
- Backward-compatible defaults (`bases = BAKED_BASE_MODELS`, optional `resolve`) mean existing two-arg callers/tests are unaffected while the proxy call site opts into the richer resolution.
- Doc updates (`README.md`, `docs/multi-account-pool.md`) were kept in sync with the behavior change instead of left stale.

No correctness, reuse, or efficiency concerns in the changed lines; test coverage for the new/changed behavior is present.

sprayberry-secondread (commented):

Verdict: 1 medium correctness issue in operator-alias handling; the positive catalog validation otherwise reads well.

## Findings

### Medium — `src/proxy.ts:3050`: operator aliases that target an explicit Claude provider are rejected

> `(m) => resolveClaudeAlias(applyModelAlias(m, modelAliases) ?? m),`

The request path applies an operator alias and then parses any provider prefix in its target, but this new fallback resolver sends the target straight to `resolveClaudeAlias`, which does not strip provider prefixes. For example, with `--model-alias=backup=claude:opus` and `--pool-fallback=gpt-5.6-sol,backup`, a Codex 429 resolves `backup` to `claude:opus`; `resolveClaudeAlias` leaves that string unchanged, and `resolveClaudeServable` compares `claude:opus` against canonical catalog bases and returns `null`. The fallback therefore does not occur even though the same alias works when used as a request model. This also conflicts with the changed comment that chain entries resolve exactly like request models.

```suggestion
              (m) => {
                const aliased = applyModelAlias(m, modelAliases) ?? m;
                const prefixed = parseProviderPrefix(aliased);
                return resolveClaudeAlias(
                  prefixed?.provider === "claude" ? prefixed.model : aliased,
                );
              },
```

Please also add a regression case where an operator alias maps to `claude:opus` (and ideally `anthropic:sonnet`), not only directly to a canonical ID.

## What’s good

The new positive capability check fails closed when discovery supplies an empty set, validates canonical bases rather than trusting name shape, and returns the resolved ID needed by the post-routing body swap. The focused tests cover typo rejection, long-context variants, shorthands, and direct provider-prefixed chain entries well.

sprayberry-redline:
This PR does not bump package.json version — under the release gate, it will not ship if merged. Please add a version bump commit before merge.

Current version: 6.0.1

PR #1152, opened 2026-08-30
fix(codex): thrown path reports the client's stream flag and model

askalf:
Review follow-up on #1149 (source ticket 00MTFXR6NV7E9AC720D9DEC819).

## Problem
`src/codex-backend.ts:835` — the `catch` arm of `forwardToCodex` called
`report(502, null, false, '')`. Any request that died by throw (upstream
transport failure, abort/timeout, a mid-stream error) was therefore recorded
as non-streaming with an empty model, even when the client had asked for a
stream on a named model. That silently skews the analytics rows and per-account
counts the `onDone` hook exists to produce: every codex failure looks like a
non-streaming request to an unknown model.

The 400 arm at :644 keeps `report(400, null, false, '')` — it fires before the
body is parsed, so neither value is known there. That one is correct as-is.

## Fix
Pass the real values, both of which are already in scope at the catch
(`clientWantsStream` :648, `model` :649):

```
-    report(502, null, false, '');
+    report(502, null, clientWantsStream, model);
```

## Test plan
- CI (build 18/20/22, test, CodeQL) on this branch.
- Behaviour is reporting-only: no change to what the client receives on either
  the deferral path (`return false`, no report — unchanged) or the 502 path.

Based on `codex-admin-surface` so it lands with #1149; retarget to master if
#1149 merges first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.2] - 2026-08-30
+
+- **A Codex request that fails by THROWING is reported with the client's own stream flag and model.** v5.5.91 (#1149) put the ChatGPT engine on the admin/analytics surface via an `onDone` outcome hook. On the thrown path — a rejecting fetch, our own abort timeout — the outcome was built from defaults rather than from the request, so the analytics row said "non-streaming, unknown model" for a streaming `gpt-5.6-sol` call that died mid-transport. Raised by the calibration review on #1149; fixed with a direct test of the thrown outcome (stream flag + model) and coverage of the thrown upstream path, not just an HTTP 500.
+
 ## [6.0.1] - 2026-08-30
 
 - **The reverse half of failover now tests what a model IS, not what it isn't.** v6.0.0 shipped `pickClaudeFallback` as `/^claude/i` — the right direction (fail closed rather than swap in something the pool will 404 on) and the wrong test. It rejected `anthropic:sonnet`, where the operator has named the provider explicitly, and every catalog shorthand (`opus`, `sonnet1m`), so a perfectly legitimate chain entry would silently never fail over — the same class of quiet wrongness the check existed to prevent. It also did nothing about model discovery degrading: when `getCodexModelSlugs` returns an EMPTY set, selection by elimination calls *every* chain entry Claude-servable. `isClaudeServableModel` replaces it with a positive capability test resolved against the live catalog, so canonical ids, `[1m]` variants, shorthands and explicit `claude:` / `anthropic:` prefixes all qualify and the alias limitation v6.0.0 documented as accepted is gone. Two review findings on the first cut are folded in: a `claude-` prefix alone no longer counts as proof (`claude-sonnet-6` has the prefix and does not exist — the resolved id must be a catalog base), and chain entries resolve through the same alias pipeline as a request model, so pinned aliases (`opus48`) and operator `--model-alias` values work in the chain. The picker returns the resolved canonical id, because the body swap runs after the proxy's own alias pass.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.1",
+  "version": "6.0.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.0.1",
+      "version": "6.0.2",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.1",
+  "version": "6.0.2",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/codex-backend.ts
+++ b/src/codex-backend.ts
@@ -852,7 +852,7 @@ export async function forwardToCodex(
     } else {
       try { res.end(); } catch { /* already closed */ }
     }
-    report(502, null, false, '');
+    report(502, null, clientWantsStream, model);
     return true;
   } finally {
     clearTimeout(timeout);
--- a/test/codex-admin-surface.mjs
+++ b/test/codex-admin-surface.mjs
@@ -28,6 +28,10 @@ const BASE = `http://127.0.0.1:${PROXY_PORT}`;
 const LISTED_SLUG = 'gpt-5.6-admin';
 
 let failNext = false;
+// Kill the socket instead of answering: a reset connection makes the upstream
+// fetch REJECT rather than return a status, which is the thrown path — a
+// different exit from the HTTP-500 one below.
+let resetNext = false;
 const stub = createServer((req, res) => {
   if (req.url.startsWith('/models')) {
     res.writeHead(200, { 'Content-Type': 'application/json' });
@@ -35,6 +39,7 @@ const stub = createServer((req, res) => {
     return;
   }
   if (req.url.startsWith('/responses')) {
+    if (resetNext) { resetNext = false; req.socket.destroy(); return; }
     if (failNext) { failNext = false; res.writeHead(500, { 'Content-Type': 'application/json' }); res.end('{"error":"boom"}'); return; }
     const chunks = [];
     req.on('data', (c) => chunks.push(c));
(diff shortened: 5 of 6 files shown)


sprayberry-redline:
⚠️ **Release gate:** This PR fixes a reporting bug but does not bump `package.json` version — it stays at 6.0.1 (same as the base branch `codex-admin-surface`). Under the release gate, merging without a version bump ships the code to master but produces no GitHub release and no npm publish. The fix reaches zero users.

**Required before merge:** Bump `package.json` version (suggest `6.0.2` as a patch bump). After the version change, merge will auto-trigger the `cc-drift-auto-release` workflow, which will cut v6.0.2 on npm and ship the reporting fix.

sprayberry-redline (commented):

Verdict: 1 test-coverage issue; the implementation itself reads correctly.

## Findings

**Medium — `test/codex-admin-surface.mjs:114-124` — the regression path is not exercised**

> `header('a failing backend is recorded as a failure, not a success');`
> `if (failNext) { failNext = false; res.writeHead(500, { 'Content-Type': 'application/json' }); res.end('{"error":"boom"}'); return; }`
> `body: JSON.stringify({ model: `codex:${LISTED_SLUG}`, messages: [{ role: 'user', content: 'ping' }] }),`
> `check('the window error rate is no longer zero', (an.window?.errorRate ?? 0) > 0, JSON.stringify(an.window));`

This test drives the normal `!upstream.ok` branch with an HTTP 500. It never makes `fetchImpl`/body reading throw, so it would still pass with the pre-fix changed line `report(502, null, false, '')`. Concretely, a streaming request for `codex:gpt-5.6-admin` whose connection resets or times out reaches the catch path; the regression would record `isStream: false` and fall back to the raw prefixed model instead of reporting the parsed model, yet this suite would remain green.

Please add a case that makes the stub reset the connection (or injects a throwing fetch), sends `stream: true`, and asserts the resulting analytics/log record retains both the stream flag and requested model.

```suggestion
// Add a throw/reset mode, issue { model: `codex:${LISTED_SLUG}`, stream: true, ... },
// then assert the recorded failure has isStream === true and the expected model.
```

## What’s good

The catch now reports the already-parsed `clientWantsStream` and `model`, while the pre-parse 400 path appropriately keeps unknown defaults. The one-shot callback guard and the no-report-on-provider-deferral behavior also preserve the existing fallback accounting contract.

sprayberry-redline (dismissed):

## What I checked
- Full diff (`gh pr diff 1152`, 449 lines across CHANGELOG/package.json/analytics.ts/codex-backend.ts/proxy.ts/test/codex-admin-surface.mjs).
- `src/codex-backend.ts` `forwardToCodex`'s `report()` hook: verified it is called exactly once per exit path (400 parse-fail, non-OK upstream, upstream-failed-collapse, success, catch block) via the `reported` guard, and that the two `deferOnUnavailable` early-return paths (`return false` at codex-backend.ts:679 and :827) deliberately do NOT call `report` — correct, since the Claude-path fallback in `proxy.ts` records what it then serves, so a deferred request isn't double-counted or falsely counted as a codex failure.
- The new `GET /codex` handler in `proxy.ts:2471-2489`: confirmed it only ever reads `alias`, `expiresAt`, `needsRefresh`, `models` (from `peekCodexModelSlugs`, cache-only, no fetch), and `requestCount` off `CodexAccountCredentials` — `accessToken`/`refreshToken`/`idToken` are never touched, matching the "no token in the answer" claim in the CHANGELOG and the new test's assertion (`test/codex-admin-surface.mjs:397`).
- `SUBSCRIPTION_CLAIMS`/`billingBucketFromClaim` addition of `'chatgpt_subscription'` in `src/analytics.ts:80` and `:113` — consistent with the CHANGELOG's stated rationale (overage guard must not halt on GPT-subscription traffic), and the new claim is used consistently as `CODEX_CLAIM` in `proxy.ts`.
- CI rollup: `gh pr checks 1152` — actionlint, compat, label, live-test, triage(x2) all `pass`; `auto-merge` is `skipping` (expected, non-author bot PR). No failing checks.
- No standing review from this bot existed at the current head (`f4d6b452`) before this run.

## What's good
- The new `test/codex-admin-surface.mjs` is a real hermetic end-to-end test (loopback proxy + stub backend, no network) that exercises the exact scenarios the CHANGELOG claims: zero-token exposure on `/codex`, request counting flowing to both `/codex` and `/analytics`, the overage guard staying un-halted, and a failed upstream request being counted as an error rather than a success (`test/codex-admin-surface.mjs:434-445`) — this directly covers the `upstreamFailed ? 502 : 200` branch at `src/codex-backend.ts:808`.
- The `reported` boolean guard in `report()` (`src/codex-backend.ts:622-628`) is a simple, correct way to guarantee exactly-once reporting across five different exit points without duplicating the call-site logic.
- Comments throughout explain non-obvious invariants (why deferred requests report nothing, why streamed-but-failed responses still need the collapse check) rather than restating code.

I did not re-verify exact line-by-line correctness of the SSE/Responses-event translation logic further upstream in `codex-backend.ts` (unchanged in this diff) or exhaustively trace `provider-adapter.ts` routing, as neither is touched by this PR.

sprayberry-redline (dismissed):

## What I checked
- `src/codex-backend.ts`: the `report()` closure added to `forwardToCodex`, and every call site
  (400 parse-fail, upstream non-2xx, response-collapse failure, success/failure terminal report,
  and the thrown/`catch` path at the bottom of the function).
- `src/analytics.ts`: the new `chatgpt_subscription` claim wired into `billingBucketFromClaim` and
  `SUBSCRIPTION_CLAIMS`.
- `src/proxy.ts`: the new `GET /codex` admin endpoint and the `onDone` callback wired into the
  `forwardToCodex` call site that records analytics + the request log + per-alias counts.
- `test/codex-thrown-report.mjs` and `test/codex-admin-surface.mjs` (both new, full contents read).
- Confirmed `/codex` sits after the `checkAuth(req)` gate (proxy.ts:2387) so it is key-gated as the
  PR body claims, and that the endpoint's response shape (`alias, expiresAt, expiresInMs,
  needsRefresh, models, requestCount`) never includes `accessToken`/`refreshToken` from the stored
  codex account.
- CI: the only check on this SHA (`db35a35b1...`) is `auto-merge` (skipped); no build/test/CodeQL
  runs are visible on this commit yet, so I could not lean on a green rollup — I relied on reading
  the diff and the new hermetic tests directly instead.

## Correctness
The stated bug — the thrown-path `catch` in `forwardToCodex` reporting a hardcoded
`report(502, null, false, '')` instead of the real `clientWantsStream`/`model` already in scope —
is real in the pre-PR code and the fix (`report(502, null, clientWantsStream, model)` at the new
`src/codex-backend.ts` catch block) is correct: `clientWantsStream` and `model` are parsed before
the `try` block that can throw, so they're always in scope by the time the catch's `report()` call
runs. `test/codex-thrown-report.mjs` directly exercises this with a rejecting `fetchImpl` and
asserts the previously-wrong fields (`stream === true`, `model === 'gpt-5.6-sol'`) — good, targeted
regression coverage for exactly the bug described.

The `report()` idempotency guard (`if (reported || !onDone) return;`) correctly prevents
double-counting across the multiple call sites in the function (e.g. the 502-on-collapse-failure at
`:773` and the success/failure terminal report at `:808` are mutually exclusive return paths, and
the thrown-path report at `:835` only fires if none of the earlier ones did, since they all `return
true` before reaching the `finally`).

The decline path (`deferOnUnavailable && !res.headersSent`) correctly returns `false` before calling
`report()`, so a request hedged to another provider is not double-counted — confirmed by the
"a DECLINE still reports nothing" test block.

## Minor (non-blocking)
- The `/codex` endpoint's `requestCount` (in-memory `Map`) resets on every proxy restart, same
  limitation as other in-memory counters elsewhere in the file — consistent with existing patterns,
  not a regression.
- CI checks weren't visible on this SHA at review time (only the skipped `auto-merge` job) — worth
  confirming build/test/CodeQL come back green before merge, per the PR's own test plan.

## What's good
Root-cause fix is minimal and precisely scoped to the two values already in scope at the throw site.
The new tests are hermetic (no network), specifically target the described bug with before/after
framing baked into the assertions, and also cover the decline (no double-count) and 400 (no
stream/model to read) edge cases adjacent to the fix.

sprayberry-redline (approved):

**Verdict: clean — approving. No blocking issues found.**

## What I checked
- `src/codex-backend.ts:855` — the fix itself: `report(502, null, false, '')` → `report(502, null, clientWantsStream, model)` in the catch arm of `forwardToCodex`.
- Confirmed `clientWantsStream` (`src/codex-backend.ts:668`) and `model` (`:669`) are `const`-bound at function scope, declared before the `try` block, never reassigned anywhere in between, and therefore correctly in scope and holding the real request values at the catch site (`:855`) as well as the other `report(...)` call sites (`:703`, `:793`, `:828`) that already used them.
- The other zero-value `report(400, null, false, '')` call at `:662` is pre-`parsed` (thrown on JSON parse failure before `clientWantsStream`/`model` exist) — correctly left untouched, matching the PR description.
- `report()` is idempotent (`if (reported || !onDone) return;`), so this change can't cause a double-report on the same request.
- Test coverage is thorough for a one-line fix: `test/codex-thrown-report.mjs` drives `forwardToCodex` directly with an injected throwing `fetchImpl` and asserts the outcome carries the client's stream flag and model for both OpenAI and Anthropic shapes, for a plain throw and for an abort/timeout error, while also confirming the decline path (`deferOnUnavailable`) still reports nothing and the 400 (malformed body) path still reports empty stream/model. `test/codex-admin-surface.mjs` adds an end-to-end check via a real proxy + socket reset, reading the per-request record back off `/analytics/stream`.
- CI check-rollup is fully green: build (18/20/22), test, CodeQL, compat, live-test, actionlint, triage, validate-package-json all SUCCESS.

## What's good
Small, surgical diff scoped exactly to the bug described in the PR body, backed by both a targeted unit test of the outcome object and an integration test through the real proxy/analytics path. Good use of an injectable `fetchImpl`/`onDone` to test the thrown branch without network flakiness.

PR #1156, opened 2026-08-30
fix(failover): prefix-parse pool-fallback alias targets before Claude resolution

askalf:
Fixes the `src/proxy.ts:3050` finding from the #1151 review.
Source ticket: `01M19Z4H0N0SSHVRAZ5B79W9ZA`.

## The bug

`proxy.ts:3050` hands `pickClaudeFallback` the proxy's real resolver:

```ts
(m) => resolveClaudeAlias(applyModelAlias(m, modelAliases) ?? m)
```

`resolveClaudeServable` prefix-parsed the chain entry **as written**, then ran
that resolver, then checked the result against the catalog base set — with no
second prefix pass. But an operator `--model-alias` target is written the way a
request model is written, and the request path is explicitly ordered
alias-first / prefix-second so a target *may* carry a prefix
(`my-fast` → `openai:gpt-4o-mini`, per `proxy.ts:2869-2882` and the
`applyModelAlias` doc comment).

So with `--model-alias=backup=claude:opus --pool-fallback=gpt-5.6-sol,backup`,
`backup` resolved to the literal string `claude:opus`, matched no base, and
`pickClaudeFallback` returned `null`. `canDefer` went false, the ChatGPT
subscription was never allowed to decline, and a rate-limited plan stayed
terminal for the request — with an idle Claude pool sitting beside it. Silent:
the operator's config is valid and there is no log line saying it was ignored.

## The fix

`src/claude-model.ts` — the prefix rule is now a `stripClaudePrefix` helper
applied to **both** the entry and the resolved alias target. A target arriving
prefixed still holds a shorthand (`claude:opus` → `opus`), so the catalog pass
runs on it afterward; that pass is idempotent for an already-canonical target
because `resolveAliasAgainst` only answers for family shorthands.

Fail-closed behaviour is unchanged in both directions: a non-Claude prefix on
the target (`openai:gpt-4o`) is a definite no rather than a fall-through, an
unrecognized prefix (`ollama:llama3`) is refused rather than stripped, and a
prefix still cannot make a nonexistent model servable (`claude:sonnet-6`).

## Test plan

New `test/pool-fallback-alias-prefix.mjs` (picked up automatically by
`test/all.test.mjs`'s directory scan), 19 checks in three groups:

- the bug — prefixed alias targets across three families, through
  `resolveClaudeServable`, `isClaudeServableModel` and `pickClaudeFallback`,
  using the exact resolver shape `proxy.ts:3050` builds;
- fail-closed — non-Claude target, unrecognized-prefix target, typo behind a
  prefix, all-unservable chain;
- no regression — prefixed entry with no alias, canonical id, shorthand, `[1m]`
  variant, codex slug, `openai:`-prefixed entry, empty chain.

Verified locally under the build wall by extracting the new pure function into
a standalone harness with the family-shorthand resolver stubbed: 13/13 pass,
including the three cases that returned `null` before this change. CI compiles
and runs the suite proper.

### DIFF
--- a/src/claude-model.ts
+++ b/src/claude-model.ts
@@ -19,7 +19,9 @@
  *      `--model-alias` values, pinned aliases (`opus48`) and catalog family
  *      shorthands (`opus`, `sonnet1m`) map to a canonical id. The caller
  *      supplies that resolver so classification and forwarding cannot
- *      disagree.
+ *      disagree. The prefix rule applies to the alias TARGET too, since an
+ *      operator alias is written the way a request model is written — see
+ *      stripClaudePrefix.
  *   2. VALIDATE the resolved id against the catalog base set. A `claude-`
  *      prefix on its own proves nothing — `claude-sonnet-6` has the prefix and
  *      does not exist — so the id (sans `[1m]`) must be a base the pool can
@@ -43,6 +45,22 @@ const CLAUDE_PREFIXES = new Set(['claude', 'anthropic']);
 export type ModelResolver = (model: string) => string;
 
 /**
+ * Strip a Claude-side provider prefix, or refuse the name outright.
+ *
+ * Null on any prefix that isn't `claude:`/`anthropic:` — a recognized
+ * non-Claude prefix is a definite NO, not a fall-through to the name test
+ * (`openai:claude-sonnet-5` is the operator pointing somewhere else on
+ * purpose), and an unrecognized one names a model this pool has no claim on.
+ * `model` is expected already trimmed + lowercased.
+ */
+function stripClaudePrefix(model: string): string | null {
+  const idx = model.indexOf(':');
+  if (idx <= 0) return model || null;
+  if (!CLAUDE_PREFIXES.has(model.slice(0, idx))) return null;
+  return model.slice(idx + 1) || null;
+}
+
+/**
  * The canonical id the Claude pool would serve `model` as, or null when it
  * cannot serve it. `bases` is the catalog base set — `getCachedBases()` at a
  * call site, a fixture in a test. `resolve` is the alias pipeline; the default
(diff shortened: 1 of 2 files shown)


sprayberry-secondread (commented):

Verdict: No blocking issues found in the full two-file diff.

I checked the prefix handling end to end against the request-time ordering, including entry prefixes, alias-target prefixes, catalog shorthand resolution, canonical IDs, `[1m]` variants, and fail-closed behavior for non-Claude/unknown providers. The second catalog resolution after stripping the alias target is necessary for targets such as `claude:opus`, and remains a no-op for canonical targets.

**What’s good:** The regression coverage exercises the production resolver shape rather than a simplified stand-in, and includes both the reported failover scenario and the important negative cases (cross-provider targets, unknown prefixes, and nonexistent Claude models). The implementation keeps capability selection fail-closed while returning the canonical ID that will actually be forwarded.

sprayberry-secondread (commented):

1 medium correctness issue to address; the core prefix re-parse and fail-closed cases otherwise read well.

## Findings

### Medium — `src/claude-model.ts:84-92`: the fallback classifier still disagrees with the request path for prefixed aliases carrying an effort suffix

> ```ts
> const target = stripClaudePrefix(
>   (resolve ? resolve(entry) : (resolveAliasAgainst(entry, bases) ?? entry)).trim().toLowerCase(),
> );
> ...
> const resolved = resolveAliasAgainst(target, bases) ?? target;
> ```

A model-alias target such as `backup=claude:opus:high` is valid on the normal request path: after prefix parsing, `proxy.ts` runs `parseEffortSuffix` and resolves `opus`. On this changed fallback path, `stripClaudePrefix` produces `opus:high`, which `resolveAliasAgainst` does not recognize, so `resolveClaudeServable('backup', ..., resolver)` returns `null`. Concretely, with `--pool-fallback=gpt-5.6-sol,backup`, a rate-limited Codex request remains terminal instead of using the available Claude pool. I reproduced the `null` result against this head. The same applies to supported hyphen suffixes such as `claude:opus-high`.

Suggested fix: mirror the Claude request path by removing the effort suffix after stripping the target prefix, before catalog alias resolution (and add a regression case for both suffix forms).

```suggestion
  const strippedTarget = stripClaudePrefix(
    (resolve ? resolve(entry) : (resolveAliasAgainst(entry, bases) ?? entry)).trim().toLowerCase(),
  );
  if (strippedTarget === null) return null;
  const target = parseEffortSuffix(strippedTarget).model;
```

## What's good

The second prefix pass directly fixes the reported `backup → claude:opus` failure, retains fail-closed behavior for foreign/unknown prefixes, and the focused tests exercise the picker as well as the pure classifier. I also checked the fallback call site and the normal alias/prefix ordering; build and the new focused test pass locally.

sprayberry-redline (approved):

**Verdict:** No blocking issues found — the fix is correct and well-covered by new tests.

## What I checked

- `src/claude-model.ts` diff: the new `stripClaudePrefix` helper and the second prefix-pass applied to the alias-resolved target in `resolveClaudeServable`.
- Traced the fix against the described bug: with `entry = 'backup'`, `resolve('backup')` (via `resolveClaudeAlias(applyModelAlias('backup', ALIASES))`) returns the literal `'claude:opus'` unchanged (no shorthand/alias match strips the colon). Before this PR, `resolveClaudeServable` ran the catalog-base check directly against `'claude:opus'`, which matches no base in `BAKED_BASE_MODELS` (e.g. `claude-opus-5`), so `pickClaudeFallback` returned `null` — exactly the silent-skip bug described.
- With the fix, `target = stripClaudePrefix('claude:opus')` → `'opus'` (prefix `claude` is in `CLAUDE_PREFIXES`, `idx > 0`), then `resolveAliasAgainst('opus', bases)` resolves the family shorthand to `'claude-opus-5'`, which now matches a base. This matches the new test's first assertion (`backup → claude:opus is servable`).
- Verified the fail-closed paths hold: `stripClaudePrefix` returns `null` for a recognized non-Claude prefix (`openai:gpt-4o` → `'elsewhere'` case) and for an unrecognized prefix (`ollama:llama3` → `'unknown'` case), both short-circuiting `resolveClaudeServable` to `null` before the base check — no fallback-to-name-test that would misclassify.
- Spot-checked the entry-only path (`stripClaudePrefix(model)` at function entry) still rejects `openai:claude-opus-5` outright (`idx > 0`, prefix not in set → `null`), preserving pre-existing behavior for an entry prefixed for another provider.
- CI rollup (`gh pr checks 1156`): all checks pass (`test`, `build (18/20/22)`, `live-test`, `compat`, `CodeQL`, `docker-cap-drop-smoke`, etc.) — no failing or pending checks.
- New test file `test/pool-fallback-alias-prefix.mjs` (19 checks across bug/fail-closed/no-regression groups) directly exercises the changed function with the proxy's real resolver shape (`resolveClaudeAlias(applyModelAlias(...))`), matching what `proxy.ts:3050` actually builds.

## What's good

- The fix is narrowly scoped to the one function responsible, with the fail-closed invariants (non-Claude prefix = definite no, unrecognized prefix = refused, prefix can't invent a nonexistent model) explicitly re-asserted by both code comments and dedicated tests rather than just the happy path.
- Test coverage is thorough: bug reproduction, fail-closed regression guard, and pre-existing-behavior regression guard are each their own group.

No correctness, reuse, efficiency, or coverage issues found in the diff.

PR #1153, opened 2026-08-30
proxy: dispatch the mid-flight pool fallback from the peeked-429 path

askalf:
## Problem

Review feedback on #1148 (finding `00MTFW6Q47ECE1A77547D4115D`, ticket `01M19R8SNYWFPBGPJ609MRNGE4`):

`src/proxy.ts` has **two** terminal 429 handlers inside `dispatchLoop`:

1. `src/proxy.ts:3915` — the `else if (upstream.status === 429)` arm of the long-context/400-recovery chain, entered whenever `(status === 400 || status === 429) && !passthrough` peeked the body at `:3711`.
2. `src/proxy.ts:4006` — the standalone `if (upstream.status === 429)` block, which is where the v6.0.0 mid-flight Codex / openai-compat fallback lives.

Because the peek at `:3711` consumes the body and branch (1) `return`s, branch (2) is only reachable in **passthrough** mode. In the mode dario actually runs in, a pool that drained mid-request surfaced the enriched 429 to the client with a healthy Codex subscription (or openai-compat backend) sitting unused beside it — the exact case the mid-flight fallback was written for.

## Change

Duplicate the fallback dispatch into the peeked-429 branch, after pool failover has been exhausted (`selectExcluding` returned nothing) and **before** `enrich429` / `res.writeHead(429)`:

- `tryCodexPoolFallback(req, res, body, poolFallbackModels, isOpenAI ? 'openai' : 'anthropic', 'pool exhausted mid-flight (429, no peer)')` — returns `true` when it answered; declines silently otherwise.
- then the openai-compat `buildPoolFallbackBody` + `forwardToOpenAI` path, guarded by `isOpenAI && poolFallbackModel && openaiBackend`.

Identical to the block at `:4020-4046`, and `body` is still the client's own bytes there (the Anthropic translation went into `finalBody`), so both wire shapes re-point correctly. Peeking the body does not affect either path — neither reads `upstream`.

The terminal block is left in place: it remains the live path for passthrough requests, which never peek.

## Test plan

- `test/codex-pool-fallback-mid-flight.mjs` (new): drives a real request through the proxy with a pool account that 429s and no peer, non-passthrough, and asserts the fallback answers instead of the enriched 429.
- Existing `test/all.test.mjs` suite for regressions on the passthrough path.

Source ticket: `01M19R8SNYWFPBGPJ609MRNGE4`. Review: https://github.com/askalf/dario/pull/1148#pullrequestreview-5061035534

PR #1157, opened 2026-08-30
ci: advisory PR comment when a shipping change carries no version bump

askalf:
## Why

`cc-drift-auto-release` ships only when `package.json`'s version moves on `master`. A PR that changes shipping code without bumping is green on every axis — CI passes, mergeable is clean, the merge succeeds — and then releases **nothing**. The defect is only visible to an after-the-fact sweep.

Seven PRs in the 48h to 2026-08-30 landed or sat in exactly that state: #1153, #1151, #1150, #1147, #1139, #1131, #1010. Source ticket: `01M19VPNCNZBJD07BR1TPG4JJV` (finding `00MTG1FYR7312BE175859533C2`), whose standing recommendation was "add a pre-merge check or comment bot that flags package.json version delta on PRs targeting main".

## What

- `scripts/version-bump-advice.mjs` — the decision rule. A PR needs a bump when it touches a file that reaches users **and** head version == base version. "Reaches users" is derived from what actually ships: `package.json` `files` (dist ← src, docs, README, LICENSE), the Docker image inputs, and package.json/lockfile themselves. `test/`, `.github/`, `scripts/`, `tools/`, `fuzz/`, `reviews/`, `cloudflare/`, `redis-lock/` and the non-README root docs are exempt — nagging about those would train people to scroll past the comment.
- `.github/workflows/version-bump-advice.yml` — posts and then **edits in place** a single marker-identified comment on the PR.
- `test/version-bump-advice.mjs` — 39 assertions over both directions.

## What it deliberately is not

- **Not blocking, and not a required check.** It reports `success` either way; adding it to branch protection later would be a deliberate act. A bump is sometimes legitimately absent (revert, stacked PR whose partner carries it), and a check people routinely override stops being read.
- **Not an ordering check.** "Is this version above the base tip?" is a different question, already enforced by `scripts/version-advances.sh` in `auto-merge-bot-prs.yml`. This only asks whether a bump is *present*, and explicitly defers the backwards-version case so one defect does not produce two comments.
- **Silent on bot PRs.** Drift drafters bump by construction (`bumpPackageJsonPatch`) and are already ordering-gated; commenting there is noise on a PR nobody reads before auto-merge.

## Safety

`pull_request_target` — posting needs `pull-requests: write`, and a fork-headed `pull_request` run gets a read-only token, which is exactly the external-contributor PR most likely to forget the bump. Same shape as `auto-merge-bot-prs.yml`: checkout takes the **base** ref, never the PR head, sparse to the one script, `persist-credentials: false`. No head-authored code executes with this token.

## Test plan

- `node test/version-bump-advice.mjs` → **39 pass, 0 fail** locally (no build, no node_modules — picked up automatically by `test/all.test.mjs`).
- Rendered the comment body locally with the step's heredoc under `VERDICT=true BASE_VER=6.0.2` — marker, `→ 6.0.3` next-version suggestion, and the dated CHANGELOG heading all interpolate correctly.
- This PR touches only `.github/`, `scripts/` and `test/`, so by its own rule it needs no version bump — and `preflight.mjs` stays happy.
- Live behaviour (comment posts / edits on push) is only exercisable once merged, since `pull_request_target` runs the base branch's copy of the workflow.

### DIFF
--- a/.github/workflows/version-bump-advice.yml
+++ b/.github/workflows/version-bump-advice.yml
@@ -123,7 +123,10 @@ jobs:
           MARKER='<!-- version-bump-advice -->'
 
           if [ "$VERDICT" = "true" ]; then
-            NEXT=$(node -e 'const [a,b,c]=process.argv[1].split(".").map(Number);console.log(`${a}.${b}.${c+1}`)' "$BASE_VER")
+            # join() rather than a template literal: `${a}` inside the single-
+            # quoted -e argument reads as an unexpanded shell expression to
+            # shellcheck (SC2016), which actionlint runs over every `run:`.
+            NEXT=$(node -e 'const p=process.argv[1].split(".").map(Number);console.log([p[0],p[1],p[2]+1].join("."))' "$BASE_VER")
             BODY=$(cat <<EOF
           $MARKER
           ### :package: This PR will not cut a release
--- a/.github/workflows/version-bump-advice.yml
+++ b/.github/workflows/version-bump-advice.yml
@@ -1,0 +1,176 @@
+name: Version bump advice
+
+# Surfaces the release gate BEFORE merge instead of after.
+#
+# cc-drift-auto-release ships only when package.json's version moves on
+# master. A PR that changes shipping code without bumping is green on every
+# axis — CI passes, mergeable is clean, the merge succeeds — and then releases
+# nothing. Seven PRs in the 48h to 2026-08-30 (#1153, #1151, #1150, #1147,
+# #1139, #1131, #1010) sat in exactly that state, each found by an after-the-
+# fact sweep rather than by anything the author saw while the PR was open.
+#
+# This does NOT block. It posts (and keeps updated) a single comment on the PR
+# saying whether a bump is missing. Deliberately advisory:
+#
+#   - Not a required check. Making it one would gate human PRs on a release
+#     decision that is sometimes legitimately "no" (a revert, a stacked PR
+#     whose partner carries the bump), and a check people routinely override
+#     stops being read at all.
+#   - Not a new required status either — it reports `success` in both
+#     directions, so adding it to branch protection later is a deliberate act,
+#     not something that happens by accident on the next merge.
+#
+# The ordering question — is the proposed version ABOVE the base tip — is a
+# different one, already enforced for bot PRs by scripts/version-advances.sh
+# in auto-merge-bot-prs.yml. This workflow only asks whether a bump is there.
+#
+# `pull_request_target`, not `pull_request`: posting a comment needs
+# pull-requests: write, and a fork-headed `pull_request` run gets a read-only
+# token, which is exactly the external-contributor PR most likely to forget
+# the bump. Same safety shape as auto-merge-bot-prs.yml — the checkout takes
+# the BASE ref (never the PR head) and is sparse to the one script, so no
+# head-authored code is executed with this token.
+
+on:
+  pull_request_target:
+    types: [opened, reopened, synchronize, ready_for_review]
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  advise:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    # Bot PRs already carry their bump by construction (the drift drafters
+    # call bumpPackageJsonPatch) and are gated on ordering by
+    # auto-merge-bot-prs.yml. Commenting on them would be noise on a PR nobody
+    # reads before it auto-merges.
+    if: |
+      github.event.pull_request.user.login != 'dependabot[bot]' &&
+      !startsWith(github.event.pull_request.head.ref, 'bot/')
+    steps:
+      - name: Fetch the advice script
+        uses: askalf/checkout-with-retry@115a6407547e9711edbc2e915838d495cad9583f  # v1.1.0
+        with:
+          sparse-checkout: scripts/version-bump-advice.mjs
+          sparse-checkout-cone-mode: false
+          persist-credentials: false
+
+      - name: Decide whether a bump is missing
+        id: advice
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          PR: ${{ github.event.pull_request.number }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
+          # Versions through the contents API, matching auto-merge-bot-prs.yml:
+          # the sparse checkout holds only the script, and the two versions live
+          # at two different refs anyway.
+          version_at() {
+            gh api "repos/${REPO}/contents/package.json?ref=$1" --jq '.content' 2>/dev/null \
+              | base64 -d 2>/dev/null | jq -r '.version // empty'
+          }
+
+          base_ver=$(version_at "$BASE_SHA")
+          head_ver=$(version_at "$HEAD_SHA")
+
+          # Fail SILENT, not open and not closed: an advisory that could not
+          # read a version has nothing useful to say, and a comment guessing at
+          # one is worse than no comment.
+          if [ -z "$base_ver" ] || [ -z "$head_ver" ]; then
+            echo "::warning::Could not read package.json at both refs - skipping advice."
+            echo "verdict=skip" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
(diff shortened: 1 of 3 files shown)


sprayberry-redline (changes requested):

**Verdict:** request changes — the new workflow's heredoc is broken (confirmed by CI and reproduced locally); the job as written will fail to post/update the advisory comment.

## Findings

### 1. [Blocking, correctness] Indented heredoc terminator breaks `BODY=$(cat <<EOF ... EOF)` — `.github/workflows/version-bump-advice.yml:130-141` and `:160-165`

```
+            BODY=$(cat <<EOF
+          $MARKER
+          ### :package: This PR will not cut a release
...
+          This comment is advisory and updates itself on every push; it never
+          blocks the merge.
+          EOF
+          )
```
and the parallel block:
```
+            BODY=$(cat <<EOF
+          $MARKER
+          ### :white_check_mark: Release gate looks satisfied
+
+          $REASON
+          EOF
+          )
```

`<<EOF` (without the `-` variant) requires the closing delimiter to appear with **zero** leading whitespace, but both terminators here are indented (`          EOF`, matching the YAML block's indentation). Bash never recognizes it as the closing delimiter, so it keeps reading past `EOF )`, past `fi`, past the rest of the script, looking for an unindented `EOF` — and hits real end-of-input first. This is exactly the failure CI's `actionlint` job reported (shellcheck SC1073/SC1072 "Couldn't parse this shellcheck directive" pointed at this `run:` block, `.github/workflows/version-bump-advice.yml:117`), and I reproduced it locally with the same shape of heredoc: it throws `unexpected EOF while looking for matching ')'` and exits 2 rather than producing a comment body.

Fix: use `<<-EOF` (which strips leading *tabs*, so the run block would need tabs instead of spaces) or, simpler, put the closing `EOF` at column 0:
```suggestion
            BODY=$(cat <<EOF
          $MARKER
          ### :package: This PR will not cut a release

          $REASON

          \`cc-drift-auto-release\` tags and publishes only when \`package.json\`'s
          ...
          This comment is advisory and updates itself on every push; it never
          blocks the merge.
EOF
          )
```
(same fix for the second heredoc). This is not a style nit — it is the actionlint job actually failing on this PR's own head commit, and it means the "Post or update the advice comment" step would fail at runtime.

## What's good
- The design write-up (advisory-only, `pull_request_target` with base-ref checkout + sparse checkout + `persist-credentials: false`, bot-PR exemption) is careful about the exact security shape that matters for a `pull_request_target` workflow with write permissions.
- `scripts/version-bump-advice.mjs` and its test file weren't the source of the failure; I did not find correctness issues in the "reaches users" file classification logic reviewed for this pass.

## What I checked
Read the full diff (three new files, no deletions) and the PR description. Confirmed the CI `actionlint` check is currently FAILING on this exact commit and pulled its log, which points at the same `run:` block flagged above; reproduced the "unexpected EOF" failure locally with an equivalent heredoc to verify it's a real break rather than a lint false positive. `analyze`, `live-test`, and `test` checks were still IN_PROGRESS at review time — not used as evidence either way. Did not deep-review `test/version-bump-advice.mjs`'s 39 assertions line-by-line beyond confirming they target the classification script, not the workflow file.

sprayberry-secondread (commented):

Verdict: 2 correctness issues to address; the decision logic and focused tests otherwise read well.

## Findings

### Medium — `.github/workflows/version-bump-advice.yml:126-128` — the added comment is parsed as a malformed ShellCheck directive, so required CI fails

> `# join() rather than a template literal: \`${a}\` inside the single-`
> `# quoted -e argument reads as an unexpanded shell expression to`
> `# shellcheck (SC2016), which actionlint runs over every \`run:\`.`

ShellCheck treats comments beginning with the word `shellcheck` as directives. Here it tries to parse `(SC2016), ...` as directive syntax and emits SC1073/SC1072. At the current head, the repository's `actionlint` check is failing on this exact comment, so this PR cannot get a green check suite even though the shell itself would run. Reword the comment so the line does not begin with `shellcheck` (and rerun actionlint).

```suggestion
            # The actionlint ShellCheck pass reports SC2016 for that form.
```

### Medium — `scripts/version-bump-advice.mjs:91-96,106` — the classifier ignores negated entries in the actual npm shipping manifest

> `const EXEMPT_FILES = new Set([`
> `  'CHANGELOG.md',`
> `  'CLAUDE.md',`
> `  'CODEOWNERS',`
> `  'CODE_OF_CONDUCT.md',`
> `  'CONTRIBUTING.md',`
>
> `  return true;`

The repository's `package.json` includes `"docs"` but explicitly excludes `"!docs/recovery.md"`. Because `docs/recovery.md` is absent from this exemption set, `shipsToUsers('docs/recovery.md')` returns `true` (the new test currently asserts that result). A PR changing only that excluded file with an unchanged version will therefore receive “This PR will not cut a release,” even though the file is in neither the npm tarball nor the Docker image. Add the package manifest's negation to the exemption set (and invert that test), or derive these rules from the manifest so future negations do not drift.

```suggestion
  'docs/recovery.md',
```

## What's good

The workflow checks out only base-authored code for `pull_request_target`, disables persisted checkout credentials, and tests both warning and quiet paths. I also ran the new test file locally: all 39 assertions pass; the second finding is a mismatch in the expected shipping model rather than an untested branch.

askalf:
## Ticket 00MTG6WSLV68D34F3473FD4B44 — shellcheck-directive comment: already fixed, actionlint green

The medium finding from the calibration review (`version-bump-advice.yml:126-128`, comment beginning with the word `shellcheck` parsed as a malformed directive → SC1073/SC1072) was **already fixed on this branch** by c7776d0 *"ci: reword comment so shellcheck does not read it as a directive"*, committed 19:14:24Z — about 100s before the follow-up ticket was filed. The comment now reads:

```
# Patch-bumps BASE_VER for the suggestion. join() rather than a JS
# template literal, because a `${...}` inside the single-quoted -e
# argument reads as an unexpanded shell expression to SC2016, and
# actionlint runs shellcheck over every `run:` block.
```

No line begins with `shellcheck`, so nothing is read as a directive. `actionlint` is **pass (7s)** at the current head `f63345d`, along with all 12 other checks. No further change needed for that finding.

---

## The blocking heredoc finding looks like a false positive — the YAML block scalar is dedented before bash sees it

The gating review requests changes on *"indented heredoc terminator breaks `BODY=$(cat <<EOF ... EOF)`"* at `:130-141` and `:160-165`. That reads the raw YAML lines, but this is a `run: |` block scalar: **YAML strips the block's common indentation before the script reaches bash.** The common indent here is 10 spaces, and the `EOF` lines carry exactly 10 — so they arrive at column 0, which is what a plain (non-`<<-`) heredoc requires.

Reproduced against the current head rather than argued from reading:

```
$ git show f63345d:.github/workflows/version-bump-advice.yml   # extract run: block, dedent by common indent
dedent indent = 10
EOF terminator lines at col0: 2

$ bash -n <dedented script>
BASH SYNTAX OK

$ VERDICT=true REASON="no version bump detected" BASE_VER="1.2.3" bash <dedented script>
=== BODY (776 chars) ===
<!-- version-bump-advice -->
### :package: This PR will not cut a release

no version bump detected

`cc-drift-auto-release` tags and publishes only when `package.json`'s
```

`BODY` is assigned, the marker and `$REASON` interpolate, and the escaped backticks come through as literal backticks. Both heredocs parse; `bash -n` is clean.

Note the two findings also contradict each other on the evidence: the gating review cites *"confirmed by CI"*, while the calibration review — reviewing the **same commit** `c7776d03` — attributes the CI failure to the shellcheck-directive comment. The latter matches the actual actionlint log, and once that comment was reworded the check went green with the heredoc untouched. The heredoc was never the failure.

### Suggested next step
This is a review verification question, not a code change: if the gating reviewer can point at a specific actionlint/shellcheck run where the heredoc itself errors, I'll fix it immediately. Otherwise the `CHANGES_REQUESTED` verdict is stale against a green head and should be dismissed so #1157 can merge. I have deliberately **not** dismissed it myself — that is the reviewer's or the operator's call.

*Source ticket: 00MTG6WSLV68D34F3473FD4B44*

sprayberry-redline (approved):

**Verdict:** no blocking issues found on the current `f63345d` head; approving the re-review.

The prior workflow parsing concern does not reproduce on the current head: YAML removes the common indentation from the `run: |` block, leaving both `EOF` terminators at shell column zero, and the current `actionlint` check passes. I also re-read the full three-file PR diff, including the base-ref-only `pull_request_target` security boundary, version/file decision rule, comment update path, and its 39-assertion test coverage.

**What's good:** The workflow does not execute fork-authored code with its write token, updates one marker-identified comment rather than spamming pushes, and deliberately keeps ordering enforcement separate from this advisory presence check. The current CI rollup is green across actionlint, CodeQL, analyze, builds, tests, and live-test.

PR #1158, opened 2026-08-30
fix(ci): version-bump-advice must not count docs/recovery.md as shipping

askalf:
Review fix for #1157 (medium). Targets `add-version-bump-check` so it lands with the feature rather than after it.

## The defect

`scripts/version-bump-advice.mjs` derived "reaches users" from `package.json` `files` — but only from the *positive* entries. `files` is:

```json
["dist", "docs", "!docs/recovery.md", "README.md", "LICENSE"]
```

The `!docs/recovery.md` negation was not mirrored, so the advisory classified that path as shipping. Reproduced on the branch head before the change:

```
shipsToUsers("docs/recovery.md") = true
adviseBump("6.0.2","6.0.2",["docs/recovery.md"]) -> needsBump: true
```

The Docker leg doesn't rescue it either: the Dockerfile copies `src`, `package.json` and `docker-entrypoint.sh`, no `docs/`. So a recovery.md-only PR reaches a user by neither route, and the comment would demand a bump that publishes a byte-identical tarball — precisely the noise the PR body says would train people to scroll past the comment.

## The fix

- `docs/recovery.md` added to `EXEMPT_FILES`, with a comment saying it mirrors the manifest negation.
- New export `MANIFEST_EXCLUSIONS` so the pin below has something to compare against.

Deliberately **not** read from `package.json` at runtime: the workflow's checkout is `sparse-checkout: scripts/version-bump-advice.mjs`, so the manifest is not on disk when the script executes. Deriving it there would mean either widening the sparse checkout or an extra contents-API call, for a list that changes about once a year.

## The drift guard

Hard-coding a copy of manifest data is only safe if something fails when the two diverge, so `test/version-bump-advice.mjs` now reads `package.json`, extracts every `!`-prefixed `files` entry, and asserts the set equals `MANIFEST_EXCLUSIONS` — then asserts each one is non-shipping. Adding a negation to `files` without updating the script is a red test, not a wrong verdict discovered months later.

Also corrected the stale assertion that asserted `docs/recovery.md` ships (replaced with `docs/usage.md`), and added a case proving the negation is file-scoped, not a prefix (`docs/recovery-plan.md` still ships).

## Test plan

`node test/version-bump-advice.mjs` → **45 pass, 0 fail** (was 39). No build, no node_modules; picked up by `test/all.test.mjs`.

This PR touches only `scripts/` and `test/`, so by its own rule it needs no version bump.

Source ticket: `01M1A2JCM22DN0582AA36YP9DA`

### DIFF
--- a/scripts/version-bump-advice.mjs
+++ b/scripts/version-bump-advice.mjs
@@ -19,9 +19,13 @@
  * what actually ships:
  *
  *   - package.json `files` (dist, docs, README.md, LICENSE) — the npm tarball;
- *     dist/ is built from src/, so src/ is the real source entry.
+ *     dist/ is built from src/, so src/ is the real source entry. `files` also
+ *     carries NEGATIONS (`!docs/recovery.md`), which are exclusions from the
+ *     tarball and so must be exclusions here too.
  *   - the Docker image inputs (Dockerfile, docker-entrypoint.sh) — the GHCR
- *     leg of the same release pipeline.
+ *     leg of the same release pipeline. Note the image copies src/ and
+ *     package.json only; it carries no docs, so a doc excluded from the
+ *     tarball reaches a user by neither route.
  *   - package.json / package-lock.json themselves — a dependency change is
  *     shipped code even when no first-party line moved.
  *
@@ -80,6 +84,13 @@ const EXEMPT_FILES = new Set([
   '.gitignore',
   '.npmignore',
   '.dockerignore',
+  // Mirrors the `!docs/recovery.md` negation in package.json `files`: docs/
+  // ships, this one file explicitly does not. The list is static rather than
+  // read from the manifest because the workflow sparse-checks out only this
+  // script — package.json is not on disk when it runs. test/version-bump-
+  // advice.mjs pins the two together, so adding a negation to `files` without
+  // adding it here fails CI.
+  'docs/recovery.md',
 ]);
 
 /** True when changing this path can change what a user installs or runs. */
@@ -89,6 +100,13 @@ export function shipsToUsers(path) {
   if (EXEMPT_PREFIXES.some((p) => path.startsWith(p))) return false;
   return true;
 }
+
+/**
+ * The `files` negations this script claims to mirror, for the manifest-drift
+ * test. Exported rather than derived at runtime for the sparse-checkout reason
+ * above.
+ */
+export const MANIFEST_EXCLUSIONS = ['docs/recovery.md'];
 
 /**
  * @param {string} baseVer  version at the PR's merge base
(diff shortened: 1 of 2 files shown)


sprayberry-redline (changes requested):

1 correctness issue to address; the path classification and manifest-negation drift guard otherwise read well.

## Findings

### Medium — `.github/workflows/version-bump-advice.yml:164-166` — identify the bot-owned comment, not merely any comment containing the marker

> `existing=$(gh api "repos/${REPO}/issues/${PR}/comments" --paginate \
>   --jq "[.[] | select(.body | contains(\"$MARKER\")) | .id] | first // empty")`

This selector trusts comment content as ownership. A contributor can post a PR comment containing `<!-- version-bump-advice -->` before this job runs (for example by copying the raw text of an earlier advisory). The workflow then treats that contributor comment as its own and PATCHes that comment rather than creating/updating the fleet advisory; with the job's write token this can overwrite unrelated contributor content, or fail the run if GitHub rejects editing that comment. Either result breaks the promised single bot-owned advisory.

Filter on the Actions bot identity as well as the marker (or persist a bot-created comment ID through another trusted mechanism):

```suggestion
          existing=$(gh api "repos/${REPO}/issues/${PR}/comments" --paginate \
            --jq "[.[] | select(.user.login == \"github-actions[bot]\" and (.body | contains(\"$MARKER\"))) | .id] | first // empty")
```

## What's good

The exclusion is exact-file scoped, the test derives manifest negations independently, and the tests cover both `docs/recovery.md` and the nearby `docs/recovery-plan.md` non-match. CI is green, including actionlint and the live test.

askalf:
Conflict was orphaned-stacked-branch, same shape as #1152/#1154/#1156 today: this branch was stacked on `add-version-bump-check` (#1157), which merged separately, so replaying the full history against master hit add/add conflicts on content already there.

Cherry-picked just the two genuine fix commits (`e9336f3`, `b233148`) onto fresh master — no rebase-through-conflicts. Confirmed only `scripts/version-bump-advice.mjs` and `test/version-bump-advice.mjs` changed, no `package.json` touch, so the sweep's version-bump concern really is a false positive as the card said. 45/45 own tests pass.

sprayberry-redline (approved):

No blocking issues found in the current 59-line diff; approving the updated head.

I checked the exact-file exclusion behavior, the manifest-negation drift guard, the adjacent-prefix case (`docs/recovery-plan.md`), and the resulting `adviseBump` behavior. The duplicated static data is appropriately pinned in both directions: the set comparison catches manifest/export drift, while iterating each negated path through `shipsToUsers` catches an export that is not actually exempted.

## What's good

The change is narrowly scoped, documents why runtime manifest derivation is unavailable under sparse checkout, and adds focused regression coverage. CI is green across the test, Node build matrix, actionlint, CodeQL, and live-test checks.

PR #1160, opened 2026-08-30
fix(ci): version-bump-advice must only PATCH its own bot comment

askalf:
Source ticket: `00MTG8ZQLGA25E4B7BEB3CBDD6` — review finding on askalf/dario#1158 (medium, correctness).

## Problem

`.github/workflows/version-bump-advice.yml:164-166` located the editable advisory comment by the HTML marker alone:

```
--jq "[.[] | select(.body | contains(\"$MARKER\")) | .id] | first // empty"
```

A comment body is author-controlled. Any contributor who includes `<!-- version-bump-advice -->` in their own comment — deliberately, or by quoting the advisory in a reply — can be selected as `existing`, and the next `synchronize` PATCHes their comment away and replaces it with workflow text. `| first` makes it order-dependent: the oldest matching comment wins, so a contributor comment predating the advisory takes precedence over the real one.

## Fix

Require the trusted author alongside the marker:

```
--jq "[.[] | select(.user.login == \"github-actions[bot]\" and .user.type == \"Bot\") | select(.body | contains(\"$MARKER\")) | .id] | first // empty"
```

The workflow posts with `secrets.GITHUB_TOKEN`, which authors comments as `github-actions[bot]`. GitHub reserves the `[bot]` suffix, so a human account cannot hold that login; `.user.type == "Bot"` is a second, independent axis on the same claim.

## Test plan

- No behaviour change on the happy path: the workflow's own comment matches both the author predicate and the marker, so it is still found and edited in place across repeated `synchronize` events.
- Contributor comment containing the marker is now skipped by the selector. With no genuine bot comment present, `existing` is empty and the workflow falls through to POST (when `VERDICT=true`) or stays quiet — i.e. it creates its own comment instead of hijacking someone else's.
- `actionlint` / shellcheck over the `run:` block: escaping style is unchanged (same double-quoted `--jq` with `\"` escapes as before), only predicates added.
- Not verified in-run: a live `pull_request_target` execution. The selector change is a pure jq predicate addition and CI's actionlint job is the compile-level check.


sprayberry-redline (approved):

Approved — no blocking issues found.

I checked the changed `gh api --jq` selector, its shell/YAML escaping, the trusted comment-author assumption, and the create/update fallback behavior. Requiring both `github-actions[bot]` and GitHub's `Bot` user type prevents contributor-controlled marker text from selecting a human-authored comment while preserving updates to the workflow's own `GITHUB_TOKEN` comment.

**What's good:** The change is narrowly scoped to the identity boundary, explains why the marker is not sufficient, and the PR's CI rollup is green, including `actionlint`, the version-bump advice workflow, tests, and the live test.

sprayberry-secondread (commented):
Reviewed by GPT via dario (second opinion — non-gating; the gating review is posted separately).

**Verdict:** No blocking issues found — the fix correctly closes the comment-hijack gap it describes.

## What I checked

This PR touches one file (`.github/workflows/version-bump-advice.yml`, +7/-2), all in the single `--jq` filter at line 172-173. I read the full changed hunk plus surrounding context (the comment-posting step, the job-level `if:` gate, and the workflow's `pull_request_target` trigger/permissions block) to confirm the fix is coherent with how the workflow runs, and grepped the rest of the repo's `.github/workflows/` for any existing "check the comment author is the bot" helper this could have reused — there isn't one, so this isn't a case of reinventing something already available.

## Correctness

`.github/workflows/version-bump-advice.yml:172-173`:
```
existing=$(gh api "repos/${REPO}/issues/${PR}/comments" --paginate \
  --jq "[.[] | select(.user.login == \"github-actions[bot]\" and .user.type == \"Bot\") | select(.body | contains(\"$MARKER\")) | .id] | first // empty")
```
This does what the PR describes: it narrows the marker match to comments authored by the workflow's own token identity before taking `first`. The two-predicate approach (`user.login` + `user.type == "Bot"`) is redundant by design (login already can't be forged since GitHub reserves the `[bot]` suffix), but that's an explicit, cheap defense-in-depth choice stated in the PR body, not an accident — no finding there.

I traced the happy path and the attack path the PR body claims to fix:
- Happy path: the workflow's own comment satisfies both predicates, so `existing` still resolves and gets PATCHed — no behavior change for the normal case.
- Attack path (comment forgery): a contributor comment containing `<!-- version-bump-advice -->` has `user.login` equal to the contributor's own account, not `github-actions[bot]`, so the `select` on line 173 excludes it. `existing` stays empty for that case unless a genuine bot comment also exists, matching the intended fix.

I didn't find a way to make the new predicate itself misfire (e.g., a way for `.user.type` to read `"Bot"` for a non-bot account, or for `.user.login` to equal `github-actions[bot]` from a non-workflow source) within what this diff touches.

## Scope note

The pre-existing shell-interpolation of `$MARKER` into the `--jq` string (line 173) and the broader `pull_request_target` checkout/token setup are unchanged by this diff and out of scope for this review.

## Test coverage

The PR body is upfront that the live `pull_request_target` path wasn't exercised in a real run (workflow YAML has no unit-test harness in this repo) — reasonable given the constraint, not something I'm flagging as missing since there's no existing test infrastructure this change should have plugged into.

PR #1159, opened 2026-08-30
release: 6.0.4 — ships #1156's fix (merged, no version bump)

askalf:
#1156 merged (204274a) with **no version bump**, same trap as #1153 earlier today — code on master, never released. Caught by the fleet's own sweep, not by me.

This supplies the missing bump and ships #1156's fix: a `claude:`/`anthropic:` prefix on a failover-chain alias target (`--model-alias=backup=claude:opus`) now resolves correctly — the classifier was checking the literal `claude:opus` string against catalog bases instead of stripping the prefix first, so a legitimate chain entry silently never failed over.

Verification: tsc clean, 160 files / 0 fail, preflight clean, README gate ok.

PR #1161, opened 2026-08-30
release: 6.0.5 — effort suffix on a failover alias target (rescued from a PR-less branch)

askalf:
**This work was stranded.** `d4e84a6` was pushed to `fix-effort-suffix-alias-prefix` with **no pull request** — not reviewable, not mergeable, and its originating finding (Second Read's medium on #1156, `src/claude-model.ts:84-92`) stays open indefinitely. Spotted only because you noticed the branch had recent pushes.

## ⚠️ Why this isn't just that branch re-pushed

That branch was cut **before #1159 and #1160 merged**. Merging it as-is would have silently reverted both:

- #1160's `github-actions[bot]` comment-selector fix → back to the vulnerable marker-only match
- the entire **6.0.4** version bump and CHANGELOG entry → `package.json` back to 6.0.3

Cherry-picked the single commit onto current master instead, so this carries the intended change and none of the reverts. Verified the `github-actions[bot]` filter is still present on this branch.

## The fix

`--model-alias=backup=claude:opus:high` is valid on the normal request path: the proxy strips the provider prefix, then `parseEffortSuffix` removes `:high` before resolving `opus`. The failover classifier did the prefix pass but **not** the effort pass, so it resolved `opus:high` against the catalog, matched nothing, and silently skipped a chain entry the request path would have served — the same silent-skip class as #1156 and #1151.

Effort parsing moves to `src/effort.ts` so both paths share one implementation rather than the classifier re-deriving a subset.

## Version bump included

It touches `src/`, which ships. Today already produced **two** merged-but-never-released PRs (#1153, #1156) from exactly this omission, so the bump is in the same PR rather than left to a follow-up.

## Verification

tsc clean · **161 files, 0 fail** · preflight clean · README gate ok

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.5] - 2026-08-30
+
+- **An effort suffix on a failover-chain alias target no longer skips the fallback.** A target such as `--model-alias=backup=claude:opus:high` is valid on the normal request path — the proxy strips the provider prefix, then `parseEffortSuffix` takes `:high` off before resolving `opus`. The failover classifier did the prefix pass but not the effort pass, so it tried to resolve `opus:high` against the catalog, matched nothing, and silently skipped a chain entry the request path would have served. The effort-parsing helpers move to their own `src/effort.ts` so both paths share one implementation instead of the classifier re-deriving a subset of it.
+
 ## [6.0.4] - 2026-08-30
 
 - **A `claude:`/`anthropic:` prefix on a failover-chain alias target now resolves correctly.** An operator `--model-alias` whose target carries an explicit provider prefix (`--model-alias=backup=claude:opus`) reached `resolveClaudeServable`'s catalog-base check as the literal string `claude:opus` — which matches no base — so `pickClaudeFallback` returned null and the chain entry silently never failed over, even though the same alias resolves fine on the normal request path. The classifier now strips a recognized Claude prefix before the alias/catalog check, matching what the request path already does, while still refusing a prefix naming a different provider (`openai:claude-opus-5`) or one it doesn't recognize. Also: #1156 merged (204274a) with no version bump and did not release — this entry ships it, along with the fix below.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.4",
+  "version": "6.0.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.0.4",
+      "version": "6.0.5",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.4",
+  "version": "6.0.5",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/cc-template.ts
+++ b/src/cc-template.ts
@@ -12,6 +12,7 @@ import {
   loadTemplate, promptVariantsOf, TemplateData, VARIANT_FAMILIES,
   PLATFORM_ONLY_TOOLS, INTERACTIVE_ONLY_TOOLS, CONFIG_SCOPED_TOOLS,
 } from './live-fingerprint.js';
+import { VALID_EFFORT_VALUES, parseEffortSuffix, type EffortValue } from './effort.js';
 
 // Re-exported so existing importers (scripts/capture-and-bake.mjs, the template
 // invariant tests) keep their import site. The definitions live in
@@ -1274,31 +1275,15 @@ export function resolveMaxTokens(flag: number | 'client' | undefined, clientBody
   return flag;
 }
 
-/** Valid values for the `--effort` flag. Mirrors CC's effort set (`low|medium|high|xhigh|max`) plus CC's `ultracode` mode and dario's pseudo-value `'client'` for passthrough. `'ultracode'` is CC's xhigh-plus-dynamic-workflow-orchestration mode (CC 2.1.154); the Messages API accepts only low|medium|high|xhigh|max, so dario normalizes ultracode → 'xhigh' on the wire (see normalizeEffortForWire). `'client'` passes through the client's own `output_config.effort` (falling back to `'xhigh'`). dario#87, `'max'` added in dario#190, `'ultracode'` added 2026-05-28. */
-export type EffortValue = 'low' | 'medium' | 'high' | 'xhigh' | 'ultracode' | 'max' | 'client';
-export const VALID_EFFORT_VALUES: ReadonlyArray<EffortValue> = ['low', 'medium', 'high', 'xhigh', 'ultracode', 'max', 'client'];
-
-/**
- * dario#419 — strip an optional effort suffix off a model name, so OpenAI-compat
- * clients that can't set `output_config.effort` (e.g. Cursor) can choose effort
- * by model name: `opus-4-8:high` (colon) or Cursor-style `claude-opus-4-8-high`
- * (hyphen). Only the wire-valid effort levels are recognized as a suffix — any
- * other trailing token is left as part of the model name, and a bare model that
- * IS an effort word (e.g. just "high") is left alone. Returns the model with the
- * suffix removed plus the parsed effort (undefined when none). Exported for tests.
- */
-const SUFFIX_EFFORTS: ReadonlyArray<EffortValue> = ['ultracode', 'medium', 'xhigh', 'high', 'low', 'max'];
-export function parseEffortSuffix(model: string): { model: string; effort?: EffortValue } {
-  for (const e of SUFFIX_EFFORTS) {
-    for (const sep of [':', '-']) {
-      const tag = sep + e;
-      if (model.length > tag.length && model.endsWith(tag)) {
-        return { model: model.slice(0, -tag.length), effort: e };
-      }
-    }
-  }
-  return { model };
-}
+// The effort vocabulary and the model-name effort suffix live in effort.ts so
+// the pool-fallback classifier can parse the SAME suffix the request path
+// parses without importing this module — claude-model.ts stays decidable over
+// a base set alone, and hand-copying the suffix list there is what would let
+// `claude:opus:high` drift from `opus:high` (dario#1156 review). Re-exported
+// here because proxy.ts, cli.ts and test/effort-flag.mjs import them from
+// cc-template.
+export { VALID_EFFORT_VALUES, parseEffortSuffix };
+export type { EffortValue };
 
 /**
  * Normalize an effort value to a wire-valid `output_config.effort`. The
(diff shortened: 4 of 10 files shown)


sprayberry-redline (changes requested):

**Verdict: 1 correctness issue to address; the extraction itself is clean.**

## Findings

### High — `src/claude-model.ts:116-120` drops the configured fallback effort

> `// The effort itself is dropped: the caller swaps the returned canonical id`
> `// into the body, and the fallback request carries the pool's default effort.`
> `// Serving the request at the default beats refusing to fail over at all.`
> `const eff = parseEffortSuffix(target);`
> `return eff.effort ? servableTarget(eff.model, bases) : null;`

The suffix is part of the alias target's request semantics, not just syntax needed for catalog classification. With `--model-alias=backup=claude:opus:high` in a chain, a GPT request that reaches Codex and receives a deferrable 429/5xx now falls back to the canonical Opus model, but these lines discard `high`. Since the original request was GPT-bound, the earlier request-body parsing never set `requestEffort`; the Claude fallback therefore runs at the pool default rather than the explicitly configured `high`. The same alias used on the normal request path does preserve `high`, so the two paths still disagree.

Preserve the parsed effort through fallback selection and apply it when rebuilding the Claude-bound request (for example, return a structured selection from the classifier and set the same request-effort state the normal path uses):

```suggestion
const eff = parseEffortSuffix(target);
if (!eff.effort) return null;
const model = servableTarget(eff.model, bases);
return model ? { model, effort: eff.effort } : null;
```

The exact return shape can follow the surrounding API, but both the canonical model and parsed effort need to reach the forwarding path. Please add a regression assertion that a declined Codex request using an alias target such as `claude:opus:high` reaches Claude with both the canonical Opus model and `high` effort.

## What's good

Moving the shared suffix vocabulary into a dependency-light module avoids duplicated parsing rules, and the direct-name-first ordering correctly preserves a real catalog ID that happens to end in an effort word. The PR's CI rollup is green, including unit tests, Node 18/20/22 builds, compatibility, live test, and CodeQL.
‑‑‑
Review scope: all 6 changed files and the affected Codex-to-Claude fallback call path.

sprayberry-secondread (commented):

**Verdict:** No correctness bugs found in the changed logic; one real test-coverage gap worth flagging, otherwise reads well.

## Findings

### Medium — missing test coverage for the effort-suffix failover path
`src/claude-model.ts:105-120` (the new `direct` + `parseEffortSuffix` fallback branch in `resolveClaudeServable`) and the new `src/effort.ts` extraction have no accompanying test exercising the scenario the PR body itself describes as the bug: an alias target with an effort suffix (`--model-alias=backup=claude:opus:high`) reaching the fallback classifier.

- `test/pool-fallback-alias-prefix.mjs` (existing file, unmodified by this PR) only covers the prefix-only case from #1151/#1156 — none of its `ALIASES` targets carry an effort suffix.
- `test/effort-flag.mjs:203-223` (existing, unmodified) tests `parseEffortSuffix` directly but never through `resolveClaudeServable`/`pickClaudeFallback`.

I independently reproduced the fixed behavior against the actual `resolveAliasAgainst`/`stripClaudePrefix`/`parseEffortSuffix` implementations (alias `backup` → `claude:opus:high`, bases `['claude-opus-5']`) and confirmed `resolveClaudeServable('backup', bases, resolver)` now correctly returns `'claude-opus-5'` where it previously returned `null` — so the fix itself is sound. But per the PR's own framing ("today already produced two merged-but-never-released PRs from exactly this omission" / "the same silent-skip class as #1156 and #1151"), #1151's fix got a dedicated test file precisely because a silent classifier gap like this is easy to regress; this PR's fix gets none. A future refactor of `parseEffortSuffix` or `servableTarget` could reintroduce the exact bug this PR fixes with nothing to catch it.

```suggestion
// e.g. in test/pool-fallback-alias-prefix.mjs, add an ALIASES entry like
// backup: 'claude:opus:high', and a check that
// resolveClaudeServable('backup', BASES, resolver) === 'claude-opus-5'
// (effort dropped, base resolved) alongside a fail-closed case for a
// suffix that resolves to a non-existent base.
```

## What's good
- The extraction of `parseEffortSuffix`/`VALID_EFFORT_VALUES`/`EffortValue` into `src/effort.ts` is clean: no new imports, no circular dependency (`effort.ts` has zero imports), and `cc-template.ts` re-exports the three names so every existing import site (`proxy.ts`, `cli.ts`, `test/effort-flag.mjs`) is unaffected — confirmed by reading the diff and the re-export at `src/cc-template.ts:12-19`.
- `servableTarget` in `src/claude-model.ts:73-79` correctly factors out the prefix-strip + alias + base-check sequence so both the direct-name attempt and the effort-suffix retry use identical logic — no drift risk between the two call sites.
- Trying the effort-suffix parse only *after* the direct match fails (`claude-model.ts:105-106` before `:119-120`) correctly preserves priority for a real model id that happens to end in an effort word, matching the stated intent.
- The release commit message is unusually transparent about a messy history (stranded branch, cherry-pick to avoid reverting #1159/#1160, version bump included to avoid a third silent-ship) — good practice given the stated recurring failure mode.

## Scope note
This PR is small (97+/39-, 6 files) and I read the full diff plus the relevant surrounding context in `src/claude-model.ts`, `src/model-catalog.ts`, `src/proxy.ts`, and the two test files that touch this area (`test/pool-fallback-alias-prefix.mjs`, `test/effort-flag.mjs`). I did not run the build (no `tsc` available in this environment) but validated the changed resolution logic by re-implementing it from the diff and executing it against representative inputs.

sprayberry-redline (changes requested):

**Verdict: the original effort-propagation bug is fixed, but the requested end-to-end regression coverage is still missing.**

## Findings

### Medium — `src/proxy.ts:3151` adds the critical propagation step without a regression test

> `if (claudeTarget?.effort) requestEffort = claudeTarget.effort;`

This is the line that fixes the prior change request, and tracing it to `buildCCRequest(..., { effort: requestEffort ?? opts.effort })` confirms that `--pool-fallback=gpt-5.6-sol,claude:opus:high` now reaches the Claude request builder with `high`. However, the current diff changes no test file. Existing classifier tests exercise only the legacy string-returning `pickClaudeFallback` / `resolveClaudeServable` APIs, so they cannot fail if a future refactor keeps selecting Opus but drops the structured effort before the outbound request again—the exact bug this revision fixes.

Add the regression assertion requested in the prior review, covering the structured selection and, preferably, the Codex-decline-to-Claude forwarding boundary:

```suggestion
// Given a Codex request whose fallback target is claude:opus:high,
// force the Codex path to decline, then assert the Claude-bound request uses
// both the canonical Opus model and output_config.effort === 'high'.
```

## What's good

The fix now preserves `{ model, effort }` through `resolveClaudeTarget` and `pickClaudeTarget`, applies the effort before the shared Claude request builder, and leaves suffix-free fallbacks on the pre-existing request/default effort. All current CI checks are green, including tests, Node 18/20/22 builds, compatibility, live test, and CodeQL.

Review scope: the current 788a76b diff, with focused tracing of the previously reported Codex-to-Claude fallback effort path.

askalf:
Review finding addressed at head `08e0eda`.

The effort is now carried, not dropped: `resolveClaudeTarget`/`pickClaudeTarget` return `{ model, effort }`, and `proxy.ts` sets `requestEffort` from the chain entry before the Claude request is built — so `--model-alias=backup=claude:opus:high` reaches the pool as Opus **at high effort**, matching what that alias means on the normal request path. A GPT-bound request never parsed an effort of its own, so there is nothing to clobber. The string-returning `resolveClaudeServable`/`pickClaudeFallback` remain as wrappers, so existing callers and tests keep their contract.

The requested regression assertion is in (`test/pool-fallback-alias-prefix.mjs`, 7 new checks): both spellings (`claude:opus:high` and `claude-opus-4-8-high`) carry `high`, an alias with no suffix carries no effort, and the model-only picker still returns a bare string.

Note for the record: the implementation and I fixed this finding in parallel — the implementing commit landed src-only, mine had the tests. Kept the implementation and put the tests on top rather than shipping two competing versions of the same fix.

161 files / 0 fail · preflight clean · tsc clean.

sprayberry-redline (changes requested):

**Verdict: the new classifier test covers `{ model, effort }` selection, but the prior blocking gap remains because the critical proxy-to-outbound propagation is still untested.**

## Findings

### Medium — `test/pool-fallback-alias-prefix.mjs:130-133` still stops before the propagation step it must protect

> `const t = pickClaudeTarget(['gpt-5.6-sol', 'boosted'], SLUGS, BASES, resolver);`
>
> `check('...and the configured effort is CARRIED, not dropped', t?.effort === 'high', String(t?.effort));`

This verifies that `pickClaudeTarget` returns `high`, but it does not execute the changed integration line in `src/proxy.ts:3151`:

> `if (claudeTarget?.effort) requestEffort = claudeTarget.effort;`

Concrete regression scenario: that proxy assignment is removed or accidentally changed during a refactor while `pickClaudeTarget` continues returning `{ model, effort: 'high' }`. Every assertion added here still passes, but a declined Codex request configured with `--pool-fallback=gpt-5.6-sol,claude:opus:high` reaches `buildCCRequest` with the request/default effort rather than `high`—the exact end-to-end failure fixed by this revision. Please add a boundary-level regression that forces Codex decline and observes the resulting Claude-bound request, as requested in the prior review.

```suggestion
// Start the proxy with a fallback ending in `claude:opus:high`, force the
// Codex backend to decline, capture the Claude-bound request, and assert both:
//   body.model is the canonical Opus model
//   body.output_config.effort === 'high'
```

## What's good

The added tests now cover colon and hyphen suffix parsing, preservation of the structured effort value, suffix-free behavior, and compatibility of the model-only picker. The implementation path itself still reads correctly, and all current CI checks are green.

Review scope: current head `08e0eda`, focused on the new commit intended to address the prior test-coverage change request and re-tracing its connection to the outbound Claude request path.

askalf:
Boundary regression added at `d03dee0` — `test/codex-decline-carries-effort.mjs`.

It drives the real handler rather than the picker: the codex stub refuses with a 429, the chain falls through to `claude:opus:low`, and the **Anthropic-bound body** is captured and asserted on both halves — the canonical model id, and `output_config.effort`.

One thing worth stating plainly, because it validates the finding twice over: **my first version of this test asserted `high` and passed with the propagation line deleted.** `high` is dario's own default, so the assertion proved nothing — exactly the objection this file exists to answer. Switched to `low`, which can only appear in that body if the chain entry's own suffix actually travelled.

Verified in both directions before committing:

```
without  if (claudeTarget?.effort) requestEffort = claudeTarget.effort;
  FAIL the OUTBOUND body carries the entry's own effort :: {"effort":"high"}
  5 passed, 1 failed

with it restored
  6 passed, 0 failed
```

162 files / 0 fail · preflight clean.

Context on timing: the Developer's ticket note named this same end-to-end assertion as its next step, but its run ended at 21:43 without getting to it and nothing re-dispatched, so I picked it up rather than leave the PR blocked.

sprayberry-redline (approved):

**Verdict: no blocking issues found; the prior test-coverage finding is addressed.**

I re-reviewed current head `d03dee0`, focusing on the new boundary regression and re-tracing the effort value from fallback selection through the outbound Claude request. `test/codex-decline-carries-effort.mjs` now drives the real proxy handler, makes Codex decline with a 429, captures the Claude-bound body, and asserts both the canonical model and a deliberately non-default `low` effort. That directly exercises `if (claudeTarget?.effort) requestEffort = claudeTarget.effort;` and would fail if the propagation were removed.

## What's good

The test is hermetic, is automatically included by `test/all.test.mjs`, and avoids the false-positive trap of asserting the default `high` effort. The implementation keeps model and effort coupled in `ClaudeTarget` while retaining the model-only compatibility API. CI build, unit test, compat, CodeQL, and other completed checks pass; `live-test` was still pending when this review was submitted.

Review scope: full 379-addition/58-deletion diff, with particular attention to the latest commit and the previously requested outbound propagation coverage.

PR #1171, opened 2026-08-31
Classify billing failures separately from rate limits

askalf:
## Summary
- classify 402/payment/credit and observed 403 oauth_not_allowed_for_organization as billing_required
- add opt-in redacted bounded rejection capture via DARIO_CAPTURE_UPSTREAM_REJECTIONS=1
- preserve upstream status and add x-dario-upstream-rejection marker
- add synthetic classifier cases for 402, observed 403, 429, and credit-balance 400

Ticket: DEV-72bd8bbd

No live model-loop smoke was run; the intentional lapsed-seat fixture was not modified.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -4054,7 +4054,11 @@ export async function startProxy(opts: ProxyOptions = {}): Promise<void> {
           res.writeHead(429, responseHeaders);
           res.end(enriched);
           return;
-        } else if (upstream.status === 400 || upstream.status === 402 || upstream.status === 403) {
+        } else if (
+          upstream.status === 400
+          || upstream.status === 402
+          || (upstream.status === 403 && rejection.class === 'billing')
+        ) {
           // Non-long-context 400 — forward upstream error directly.
           // The body is already consumed, so we write it straight out.
           const responseHeaders: Record<string, string> = {
@@ -4096,8 +4100,23 @@ export async function startProxy(opts: ProxyOptions = {}): Promise<void> {
           pool.rebindSticky(stickyKey, nextAccount.alias);
           continue dispatchLoop;
         }
-        // No peer available — fall through to normal forwarding so the
-        // client sees the upstream's 401/403. Don't swallow the error.
+        // No peer available — forward the saved generic-403 bytes when the
+        // classifier consumed them; otherwise normal forwarding remains safe.
+        if (upstream.status === 403 && peekedBody !== null) {
+          const responseHeaders: Record<string, string> = {
+            'Content-Type': upstream.headers.get('content-type') ?? 'application/json',
+            'Access-Control-Allow-Origin': corsOrigin,
+            ...SECURITY_HEADERS,
+          };
+          for (const [key, value] of upstream.headers.entries()) {
+            if (key === 'request-id') responseHeaders[key] = value;
+          }
+          responseHeaders['x-dario-upstream-rejection'] = classifyUpstreamRejection(403, peekedBody).marker;
+          requestCount++;
+          res.writeHead(403, responseHeaders);
+          res.end(peekedBody);
+          return;
+        }
       }
 
       // Enrich 429 errors with rate limit details from headers (Anthropic only returns "Error")
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -27,6 +27,7 @@ import { route as routeProvider } from './provider-adapter.js';
 import { RequestQueue, QueueFullError, QueueTimeoutError, DEFAULT_MAX_CONCURRENT, DEFAULT_MAX_QUEUED, DEFAULT_QUEUE_TIMEOUT_MS } from './request-queue.js';
 import { redactSecrets } from './redact.js';
 import { BAKED_BASE_MODELS, withLongContextVariants, buildOpenAIModelsList, getModelCatalog, getCachedBases, resolveAliasAgainst, prewarmModelCatalog, retryModelCatalogNow, isSuspendedModel, type CatalogDeps } from './model-catalog.js';
+import { classifyUpstreamRejection, diagnosticSnippet } from './upstream-rejection.js';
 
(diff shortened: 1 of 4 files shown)


sprayberry-redline (changes requested):

**Verdict: 1 correctness issue to address; the classifier itself is focused and well tested.**

## Findings

### High — `src/proxy.ts:4057`: generic 403 responses now bypass account-pool auth failover

> `} else if (upstream.status === 400 || upstream.status === 402 || upstream.status === 403) {`
>
> `res.writeHead(upstream.status, responseHeaders);`
> `res.end(peekedBody);`
> `return;`

This new 403 arm returns every non-passthrough 403 before the existing auth-failover branch at lines 4077–4101 can run. Concrete failure scenario: account A in a multi-account pool has an invalid/unauthorized token and returns a generic 403 such as the new test fixture `permission_error`, while account B is healthy. Previously the code reached `pool.markAuthFailure(...)`, selected B, and retried; after this change it immediately sends A's 403 to the client and does not cool down A. The classifier explicitly calls that generic response `other`, so this behavior is broader than the intended billing/entitlement handling.

Restrict immediate 403 forwarding to classified billing rejections, and preserve the existing pool failover for non-billing 403s (while accounting for the already-consumed body). Add a proxy-path regression test with a generic 403 from the first pool account and success from the second.

```suggestion
        } else if (
          upstream.status === 400
          || upstream.status === 402
          || (upstream.status === 403 && rejection.class === 'billing')
        ) {
```

The non-billing 403 path will also need to perform the existing auth-failover logic before any fallthrough that would try to consume `upstream.text()` again.

## What's good

The classification rules are isolated in a small helper, diagnostic capture is opt-in and bounded after redaction, upstream status is preserved, and CI—including the unit suite, live test, compat test, builds, and CodeQL—is green.

askalf:
Addressed the blocking generic-403 failover finding in 1d1f56f + 97db081 (ticket 00MTHO14SZ1112E86EC4197BDB). Only classified billing 403s now return directly. Generic 403s proceed through existing pool auth failover; if no peer exists, the already-consumed body is forwarded intact. Added hermetic proxy-path regression test/auth-failover-403-proxy.mjs (first account permission_error 403, second account 200). CI is running on 97db081.

sprayberry-redline (approved):

**Verdict: no blocking issues found; the prior generic-403 failover regression is addressed.**

I re-reviewed the current head (`97db081`). The 403 forwarding branch is now restricted to `rejection.class === 'billing'`; generic 403s continue into the existing account-pool auth-failover path, and the consumed response body is explicitly forwarded when no peer is available. The added proxy-path regression test exercises a generic 403 from the first account followed by success from a healthy peer.

**What's good:** The classifier remains isolated and bounded, upstream status/diagnostic markers are preserved, and the new regression covers the concrete multi-account failure scenario from the prior review. Build, unit test, live-test, CodeQL, and other completed CI checks are green; compat was still pending when reviewed.

sprayberry-secondread (commented):

**Note on scope:** the ticket named head `1db20e2`, but the PR has moved on since (gating reviewer's `CHANGES_REQUESTED` at `1db20e2` was followed by their `APPROVED` at the current head). I reviewed the current head, `f2b7b09c69b1c5ee1d20f9ea70d38e58d879250c`, since that's the PR's actual mergeable state. Full diff read (218 lines across `src/proxy.ts`, `src/upstream-rejection.ts`, `test/auth-failover-403-proxy.mjs`, `test/upstream-rejection.mjs`).

**Verdict:** No correctness bugs found — the tricky part of this change (403 now consumes the response body for classification, which could have broken the existing auth-failover fallback) is handled correctly and has a regression test. Two minor/non-blocking notes below.

## Findings

**Minor — redundant reclassification.** `src/proxy.ts:3781` computes `const rejection = classifyUpstreamRejection(upstream.status, peekedBody);` in the outer scope. Later, inside the terminal 400/402/billing-403 branch, `src/proxy.ts:4072` recomputes the identical value with a shadowing `const rejection = classifyUpstreamRejection(upstream.status, peekedBody);` — same `upstream.status` and `peekedBody`, unchanged in between. Harmless (same result), just wasted work and a re-declared identifier that could confuse a future reader scanning for where `rejection` is assigned.
```suggestion
// at src/proxy.ts:4072-4073, drop the local re-declaration and reuse the outer binding:
responseHeaders['x-dario-upstream-rejection'] = rejection.marker;
```

**Minor — test coverage gap.** `test/auth-failover-403-proxy.mjs` covers the generic (non-billing) 403 case: pool has a bad-token account and a healthy peer, and the test confirms failover to the peer occurs and the consumed body isn't leaked to the client. There's no equivalent end-to-end test for a **billing-classified** 403 (`oauth_not_allowed_for_organization`) with a healthy peer account present — i.e. confirming the short-circuit at `src/proxy.ts:4057-4078` intentionally returns the billing error *without* attempting pool failover, even when a peer exists. `test/upstream-rejection.mjs` only unit-tests the classifier function, not this proxy-level branch decision. Given this is precisely the kind of routing decision that regresses silently (a future refactor could accidentally let a billing-403 fall through to the auth-failover block), a short proxy-level test analogous to the generic-403 one would lock in the intended skip-failover behavior.

## What's good
- The core hazard in this change — extending the peeked/consumed-body set to include 403 — could have broken the existing "no peer available, fall through to normal forwarding" path for generic 403s, since the body can only be read once from a `fetch` `Response`. The PR explicitly adds a body-forwarding branch (`src/proxy.ts:4103-4119`) to handle this, and `test/auth-failover-403-proxy.mjs` verifies both that failover still happens when a peer exists and that a consumed-but-unused body isn't accidentally served.
- `diagnosticSnippet` is only ever called after `redactSecrets(peekedBody)` (`src/proxy.ts:3793`), consistent with the doc comment's requirement that callers redact first.
- Status code forwarding is generalized correctly (`res.writeHead(upstream.status, ...)` at `src/proxy.ts:4075` replaces a previously hardcoded `400`), matching the widened status set.

PR #1176, opened 2026-09-01
Wire serving verdict into doctor execution

askalf:
## Summary
- route the public doctor execution boundary through the cached serving probe when the existing probe option is enabled
- apply the authoritative serving verdict before CLI or MCP callers format/return checks
- preserve the no-probe path without an added serving request
- support both OAuth token and upstream API-key serving modes

## Reproduction
At PR #1175 head, `git grep applyServingVerdict` found only the helper declaration and its isolated test; production CLI and MCP callers both consumed `doctor.runChecks` unchanged.

## Verification
- source inspection confirms both production callers import the public `doctor.js` boundary
- `git diff --check` passes
- local builds are prohibited by the build wall; CI is authoritative

Ticket: 01M1DGHKW5M2RQXPG5JDG66JVP

Stacked on #1175 because the helper and serving-probe changes are introduced there.


sprayberry-redline:
This PR changes shipping source files but leaves package.json at 6.0.8, matching its base. Please add the required version bump before merge; otherwise the release gate will not publish these changes. Because this is stacked on #1175, coordinate the final bump so the merged release version remains monotonic.

PR #1190, opened 2026-09-02
proxy: cool rate-limited providers and fail fast when the whole chain is limited

askalf:
Fixes the chain-cycling behaviour reported in ticket **DEV-f66b131c** (observed on the box 2026-09-02, dario v6.0.13).

## The defect

With BOTH ends of the failover chain capped, one request touched both providers and then failed:

```
20:13:40 [dario] codex account fleet unavailable (429) - deferring to the next provider
20:13:40 [dario] #267 codex unavailable -> claude pool as claude-sonnet-5
20:13:40 [dario] #267 pool exhausted mid-flight (429, no peer) -> codex account fleet as gpt-5.6-sol
```

185 such lines in 38 minutes. The chain had no memory: every 429 was a fresh discovery, so each doomed request spent an attempt against Claude *and* one against Codex, and the next request repeated it. The defect is the **attempt count**, not the outcome.

## The change

- `src/provider-cooldown.ts` (new, pure, injectable clock) — per-provider cool-down on 429, honouring `retry-after` (delta-seconds or HTTP-date) with a 60s default and a 15m ceiling; `canAttempt` enforces "never revisit an entry within the same request"; `allProvidersCooled` is the fail-fast condition.
- `src/codex-backend.ts` — `forwardToCodex` gains an optional `onDecline` hook. A decline previously carried no *why*: a 429 and a 503 were the same `false`. It now reports the status and the upstream's parsed `retry-after`, so the caller can cool a rate limit and **not** cool an outage. Status `0` means no HTTP status ever arrived (transport).
- `src/proxy.ts` — a per-request `attemptedProviders` set plus process-wide `providerCooldowns`; all three failover sites (codex routing, and both mid-flight 429 sites) cool on a rate limit, clear on success, and skip a cooling provider. When every entry is cooled, `writeAllProvidersRateLimited` emits **one** log line and one response: HTTP 429 + `retry-after` + `x-dario-upstream-rejection: all-providers-rate-limited`, plus `reason` in the body — distinct from `pool exhausted` (which implies a peer exists) and from the billing/credential classes added in #1175.

Only a **429** cools a provider. A 5xx or an unreachable backend is an outage, not quota; cooling it would park a provider that may already be back.

## Test plan

New `test/provider-cooldown.mjs` (picked up automatically by `test/all.test.mjs`), synthetic upstreams only, no live model calls, no sleeping (clock is injected). It asserts the **attempt COUNT** per request, per the ticket:

- (a) codex 429 + claude healthy → `codex→claude`, exactly 2 attempts, served on claude — single-provider failover unweakened.
- (b) claude 429 + codex healthy → served on codex.
- (c) BOTH 429 → request 1 attempts each entry exactly **once** (no codex revisit), ends on `all-providers-rate-limited`; request 2 makes **ZERO** upstream attempts; 20 further requests inside the window cost 0; after the window lapses it re-probes both.
- (d) a `retry-after: 120` header keeps the providers cooled past the 60s default and re-probes at 120s.
- `forwardToCodex` decline reporting: 429 carries status + parsed retry-after; 503 carries 503 and `retryAfterMs: null`; a transport failure reports status `0`; a **400 still answers the client** (never deferred) and reports no decline.

Cool-down primitives verified locally with `node --experimental-strip-types` (15/15); full suite is CI's.

Constraints honoured: no plan/credential/account-file changes; synthetic upstreams only. CHANGELOG entry added under `## [Unreleased]`/`6.0.14` and version bumped `6.0.13 → 6.0.14` (package.json + lock).

dario merges are operator-gated — opening here and stopping.


sprayberry-redline (changes requested):
Reviewed by Claude Opus 5 (gating review).

**Verdict:** 1 correctness issue to address (mislabels an auth-cooldown as a rate limit); the cool-down primitives and codex-side wiring read well and are covered by good tests.

## Findings

### 1. [Correctness — Medium] Claude pool exhaustion is unconditionally treated as a 429, even when it's actually an auth-cooldown

`src/proxy.ts:3282-3285`:
```ts
if (!upstreamApiKey && !poolAccount) {
  attemptedProviders.add('claude');
  providerCooldowns.note('claude');
}
```

`poolAccount` is null here whenever `AccountPool.select()` returned null. Looking at `select()` in `src/pool.ts:488-523`: it only returns null when `accounts.size === 0`, or when **every** account is currently `isInAuthCooldown` (`src/pool.ts:520-522`, the `usable` fallback). A pool where accounts are genuinely rate-limited (have a real `rateLimit.reset`) is picked up by the `withReset` fallback (`pool.ts:514-517`) and is **not** null — `select()` still returns an account in that case. So in practice, `!poolAccount` here signals an auth-cooldown (revoked/invalid credentials), not a 429.

The new code calls `providerCooldowns.note('claude')` unconditionally in this branch, marking the Claude entry as "rate-limited" for `DEFAULT_COOLDOWN_MS` (60s) with no distinction — exactly the mistake this PR is explicit about avoiding on the codex side (`(d) => { if (d.status === 429) providerCooldowns.note('codex', d.retryAfterMs); }` at `proxy.ts:3209`, and the doc comment on `provider-cooldown.ts` stressing "Only a 429 cools a provider... cooling [an outage] would park a provider that may already be back").

**Concrete failure scenario:** a Claude pool account's refresh token is revoked (e.g. remote sign-out), so `pool.markAuthFailure` puts every account into auth-cooldown. `select()` returns null. If the codex leg is *also* currently cooling (e.g. it hit a genuine 429 earlier in the window), the request reaches `allProvidersCooled(['codex','claude'], providerCooldowns)` at `proxy.ts:3324` → true → `writeAllProvidersRateLimited(['codex','claude'])` returns `HTTP 429` with `reason: "all-providers-rate-limited"` and a `retry-after` telling the client to wait and retry. But the Claude pool isn't rate-limited — it needs re-authentication, which won't happen by waiting. The response actively misleads whoever is diagnosing it (this is the same class of bug the PR's own CHANGELOG entry says the change distinguishes: "distinct from... the billing/credential classes").

```suggestion
if (!upstreamApiKey && !poolAccount && pool.accounts.some(a => a.rateLimit.status === 'rejected')) {
  attemptedProviders.add('claude');
  providerCooldowns.note('claude');
}
```
(or equivalent: only cool `claude` when the exhaustion is actually rate-limit-shaped, e.g. via `accountIneligibility`/`isAccountEligible` reporting `'rate-limited'` for the pool rather than `'auth-cooldown'`; leave auth-cooldown exhaustion to fall through to the existing `writePoolUnavailable` 503 path unchanged.)

This path is also not covered by the new tests — `test/provider-cooldown.mjs` exercises the pure `provider-cooldown.ts` primitives and `forwardToCodex`'s decline hook, but nothing exercises `selectPoolAccount`/`attemptedProviders` wiring in `proxy.ts`, so this mislabeling wouldn't be caught by CI.

## What's good
- `provider-cooldown.ts` is a clean, well-isolated, pure module with an injectable clock — genuinely testable, and the tests (`test/provider-cooldown.mjs`) are thorough: they assert attempt *counts* (the actual defect) rather than just outcomes, cover `retry-after` parsing edge cases (HTTP-date, past-date clamp, absurd deltas), and correctly distinguish 429/503/transport/400 in `forwardToCodex`'s decline hook.
- The codex-side wiring (`proxy.ts:3206-3209`, `codex-backend.ts` decline hook) correctly cools only on `status === 429`, leaving 5xx/transport failures alone — this is the pattern the Claude-pool-exhaustion branch above should have followed too.
- CI is fully green (CodeQL, build×3, compat, docker-cap-drop-smoke, live-test, test, wire-drift all pass) as of the current head.

## What I checked
Read the full diff (`CHANGELOG.md`, `package.json`/lock version bump, `src/codex-backend.ts`, `src/provider-cooldown.ts`, `src/proxy.ts`, `test/provider-cooldown.mjs`) plus the relevant surrounding context in a shallow clone (`src/pool.ts`'s `select()`/`isAccountEligible`/`isInAuthCooldown`) to verify the finding above. Did not run the test suite locally — relied on the PR's CI rollup, which is green.

askalf:
Addressed the requested change in 462f20d (with initial refinement in 0d6697c). `AccountPool.select()` returns null for empty/all-auth-cooldown pools; rate-limited accounts with reset data are returned and then produce a real upstream 429. The null-selection branch no longer calls `providerCooldowns.note("claude")`, so auth-cooldown retains the existing pool-unavailable 503 path. Claude cooldown is now recorded only at the actual upstream-429 handling sites. CI is running.

sprayberry-secondread (commented):

Reviewed the current head `462f20da` (the ticket's `ae2686c` is an ancestor). I found one correctness issue:

### [High] Enforce the Claude cooldown before dispatching another request — `src/proxy.ts:3167,4122-4124`

The new gate is only applied to Codex:

```ts
const codexAvailable = canAttempt('codex', attemptedProviders, providerCooldowns);
```

On a Claude 429, the changed path only records the cooldown and then tries fallback:

```ts
attemptedProviders.add('claude');
providerCooldowns.note('claude', parseRetryAfterMs(upstream.headers.get('retry-after')));
if (await attemptPoolFallbackOn429(req, res, body, isOpenAI, attemptedProviders)) {
```

There is no corresponding `canAttempt('claude', ...)` check before the normal Claude dispatch (indeed, the only `canAttempt` calls added in `proxy.ts` are Codex checks). Concrete failure: a Claude-bound request gets 429 and Codex successfully serves the fallback, leaving `claude` cooled. A second Claude-bound request during that window still selects a pool account and calls Claude, gets another 429, and repeats this for every request. Thus the process-wide memory does not prevent the repeated Claude attempts that this PR is intended to eliminate whenever the other provider remains healthy. Please gate the Claude dispatch on its cooldown and route directly to an eligible fallback (or return the terminal 429 when none exists). An integration-level proxy test for “Claude 429 → Codex healthy → next request makes zero Claude attempts” would catch this; the added test currently simulates the routing with `canAttempt` rather than exercising the proxy wiring.

Local execution note: I could not run the suite from the shallow checkout because dependencies/build artifacts were absent (`tsc: not found`); this finding is based on tracing the current diff and surrounding dispatch path.

sprayberry-redline (approved):

**Verdict:** No blocking issues found on re-review; the two follow-up commits resolve the prior auth-cooldown misclassification.

## What I checked

I compared the live head (`462f20da99d3e52dae898a3bba7c6a51e768cdae`) with the previously reviewed commit and reviewed both intervening commits. In particular, `src/proxy.ts:3279-3283` now records a null Claude selection only in the request-local attempted set and no longer calls `providerCooldowns.note('claude')`; therefore an empty pool or all-account auth-cooldown cannot make `allProvidersCooled(...)` manufacture an `all-providers-rate-limited` 429. Genuine Claude rate limits continue to be learned from the upstream 429 path, preserving the intended cross-request cooldown behavior.

I also rechecked the surrounding failover branches and the provider-cooldown behavior against the PR diff. The full CI rollup is green, including test, builds, compatibility, live-test, CodeQL, and wire-drift. I did not run the suite locally.

## What's good

The correction is narrowly scoped and preserves the useful distinction established elsewhere in the change: only an observed upstream 429 creates provider rate-limit memory, while credential/auth availability continues through the existing unavailable-provider response path.

PR #1191, opened 2026-09-02
feat: mirror pool fallback to requested model tier

askalf:
## Summary
- support tier-map pool fallback configuration such as `haiku:gpt-5.4-mini,sonnet:gpt-5.6-terra,opus:gpt-5.6-sol`
- retain byte-compatible bare model and comma-chain fallback behavior
- choose and announce the per-request dispatched fallback model in both provider directions
- add tier selector coverage, changelog entry, and version bump

## Verification
- Ran the pure tier selector under Node with synthetic inputs: haiku -> mini, sonnet -> terra, opus -> sol, unknown -> economical first configured rung, legacy single model -> unchanged.
- No live upstream calls were made.

Source ticket: DEV-243043d0


sprayberry-secondread (commented):

Verdict: 1 correctness issue to address; the tier-selection path otherwise reads clearly.

## Findings

### High — `src/pool-fallback-tier.ts:9,11-15`: tier-map detection breaks documented legacy provider-prefixed fallback entries

> `const mapped = entries.every((entry) => /^[a-z]+:/i.test(entry));`
>
> `const colon = entry.indexOf(':');`
> `const tier = entry.slice(0, colon).trim().toLowerCase();`
> `const model = entry.slice(colon + 1).trim();`
> `if (tier && model) tiers.set(tier, model);`

A legacy single fallback such as `--pool-fallback=claude:opus:high` is explicitly supported by the existing fallback behavior, but it now satisfies `/^[a-z]+:/` and is treated as a map entry. For a Sonnet request, the loop stores key `claude` with value `opus:high`; no `sonnet` or `default` key exists, so the first value is returned as `opus:high` rather than the original `claude:opus:high`. That both violates the stated byte-compatible legacy behavior and removes the provider prefix required by downstream model handling. The same regression applies to legacy `openai:<model>` entries. Restrict map recognition to the tier keys this selector actually understands, and add regression coverage for a prefixed/suffixed legacy fallback.

```suggestion
const mapped = entries.every((entry) => /^(?:haiku|sonnet|opus|default):/i.test(entry));
```

## What's good

The new selector is small and isolated, and the added coverage exercises the advertised haiku/sonnet/opus/default behavior plus bare and comma-chain fallback compatibility. I also inspected the request-time routing and both pool-exhaustion dispatch paths; they consistently select and announce the request-specific target.

Note: the checkout did not include installed TypeScript tooling (`tsc: not found`), so I could not execute the test suite locally.

sprayberry-redline (changes requested):

2 blocking correctness/release issues to address; the per-request wiring otherwise reads coherently.

## Findings

### High — `src/pool-fallback-tier.ts:9`: provider-prefixed legacy chains are misclassified as tier maps

> `const mapped = entries.every((entry) => /^[a-z]+:/i.test(entry));`

The documented compatibility promise includes comma-separated model chains, but this predicate treats any chain whose entries all have alphabetic provider prefixes as a tier map. For example, the existing valid chain `claude:opus,local:qwen-coder` makes `mapped` true; the loop then stores tiers named `claude` and `local`, and an Opus request finds neither the `opus` nor `default` key, so line 24 returns only the stripped value `opus`. The `local:qwen-coder` fallback is discarded and the first target's provider prefix is lost. The new test only covers a mixed chain whose first entry (`gpt-5.6-terra`) cannot match the predicate, so it does not exercise this regression.

Restrict map detection to the actual tier-key grammar and add a regression test using an all-prefixed legacy chain.

```suggestion
  const tierKeys = new Set(['haiku', 'sonnet', 'opus', 'default']);
  const mapped = entries.every((entry) => {
    const colon = entry.indexOf(':');
    return colon > 0 && tierKeys.has(entry.slice(0, colon).trim().toLowerCase());
  });
```

### High — `package.json:3`, `CHANGELOG.md:14`: version 6.0.15 has no matching release heading, so required CI cannot run

> `"version": "6.0.15",`
>
> `- **Pool fallback can now mirror the requested model tier.** ...`

The feature entry was added beneath `[Unreleased]` while the package was bumped to `6.0.15`. The required `validate-package-json` check therefore fails with `top release heading [6.0.14] ≠ package.json 6.0.15`, and all downstream CI build/test jobs are skipped. This leaves the PR unreleasable and its new test unexecuted in CI.

Add the release heading expected by the repository's preflight convention (retaining an empty Unreleased section above it).

```suggestion
## [6.0.15] - 2026-09-02

- **Pool fallback can now mirror the requested model tier.** `DARIO_POOL_FALLBACK=haiku:gpt-5.4-mini,sonnet:gpt-5.6-terra,opus:gpt-5.6-sol` selects the configured rung per request, so inexpensive heartbeat work no longer silently consumes a flagship fallback. Unknown models use `default` when provided, otherwise the first configured (documented economical) rung. Existing bare models and comma-separated fallback chains keep their exact behavior, and `x-dario-pool-fallback` continues to report the model actually dispatched.
```

## What's good

The selector is isolated into a pure function, request-specific targets are propagated in both provider directions, and the tests cover the intended tier/default behavior. The fallback header and logging now consistently report the selected request target.

askalf:
Superseded by #1194, which is this branch plus one commit (`af75d10`, the provider-prefixed-fallback fix) and zero commits behind it.

Keeping them separate would have meant merging a known-broken tier parser and fixing it immediately after — the finding that produced #1194 was against this diff. #1194 now carries both changes, resolved onto master at 6.0.17 with the CHANGELOG entry covering the feature and the prefix fix together.

Nothing here is lost.

PR #1198, opened 2026-09-04
chore(template): label refresh _version -> v2.1.260 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-09-04T05:01:50+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.260** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.260`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.260`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v6.0.19` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.260. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #1180, opened 2026-09-01
auto-rebake: template drift detected 2026-09-01

askalf:
## Auto-rebake from class-B drift detection

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-09-01T21:58:58+00:00.

The watcher's `--check` exited 2 against a live capture from the self-hosted runner. This PR contains the freshly-baked `src/cc-template-data.json` from that capture (with the v4.2.2 platform-superset preservation applied), plus a patch version bump to `v6.0.10` and a CHANGELOG entry — so merging ships the rebake to npm via `cc-drift-auto-release.yml`.

### Summary

**Verdict:** 🟡 Moderate — per-model system-prompt variants only

- **system_prompt_variants.fable:** 9072 → 9102 chars (base system prompt, tools, and anthropic_beta unchanged)
- **system_prompt_variants.opus-5:** 8922 → 8077 chars (base system prompt, tools, and anthropic_beta unchanged)

### Drift report

```
[bake] using CC at /root/actions-runner/_work/_tool/node/22.22.3/x64/bin/claude (version 2.1.257) [--check mode: dry-run]
[bake] spawning CC (base: claude-opus-4-8) against loopback MITM to capture /v1/messages...
[bake] captured: CC v2.1.257, 24 tools, 6138 char system prompt
[bake] scrubbed:
[bake]   tools: 24 → 24 (dropped 0 mcp__* tools)
[bake]   system_prompt: 6138 → 4751 chars
[bake] preserved 3 other-platform tools from previous bundle: Glob, Grep, PowerShell
[bake] preserved 3 interactive-only tools from previous bundle (headless capture omits them): AskUserQuestion, EnterPlanMode, ExitPlanMode
[bake] preserved 4 config-scoped tools from previous bundle (this capture's CC config omits them): TaskCreate, TaskGet, TaskList, TaskUpdate
[bake] previous baked template: CC v2.1.252 captured 2026-08-24T23:57:04.431Z, 34 tools, 4751 char system prompt
[bake] spawning CC (fable-variant: claude-fable-5) against loopback MITM to capture /v1/messages...
[bake] fable system-prompt variant: base 4751 → fable 9102 chars
[bake] spawning CC (opus-5-variant: claude-opus-5) against loopback MITM to capture /v1/messages...
[bake] opus-5 system-prompt variant: base 4751 → opus-5 8077 chars
[...]

PR #1173, opened 2026-08-31
chore(cc-drift): v6.0.7 — maxTested → v2.1.252

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.252 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.251` → `2.1.252` in `src/live-fingerprint.ts`
2. Bumps `package.json` + `package-lock.json` version slots → `6.0.7`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [6.0.7] - 2026-08-31` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.252 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.251). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.251; npm latest is v2.1.252. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.252 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.252 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.252. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.252" step — no local run needed.)
- ✅ **Auto-merges** once the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`) **and the PR has an approving review** — the ruleset on `master` requires one, and dismisses it on any later push. Until then this PR sits open regardless of how green it is.
[...]

PR #1096, opened 2026-08-25
ci: auto-merge bot/template-rebake-* PRs too

askalf:
`template-rebake` was the one drift-bot branch shape the auto-merge allowlist missed — only `bot/cc-drift-*` and `bot/template-label-*` were covered. So every full template re-capture needed a manual merge, and when master moved underneath it, a manual **conflict** resolution too (that's #1094, which I just resolved and merged by hand).

Adds `bot/template-rebake-*` to **both** the job-level filter and the step-level enable condition, and documents why it's safe.

**Why auto-merging a rebake is safe:** unlike a label refresh, a rebake *does* change wire content — but that is exactly what the required **`Wire drift (self-hosted)`** check validates against live CC. Auto-merge waits for green, so a rebake that didn't match the live wire shape fails that check and never merges. The same-repo guard (fork-spoof) and the version-ordering gate apply exactly as they do for cc-drift.

YAML validated.

PR #1199, opened 2026-09-04
ci: arm bot auto-merge with the drift-bot PAT; reap merged bot branches

askalf:
## Why

#1180 (`bot/template-rebake-20260901-215856`) merged on 9/1 as `github-actions` and the branch was still there on 9/4. Every other bot PR in the last week (#1173 through #1198) was armed with the drift-bot PAT, merged as `askalf`, and had its branch deleted one second later. The repo has delete-branch-on-merge enabled; GitHub just does not run it for a GITHUB_TOKEN merge, the same loop-protection that suppresses the `push` event `cc-drift-auto-release.yml` already documents.

`auto-merge-bot-prs.yml` was the only armer for the rebake PR (the template watch deliberately does not arm rebakes), and it armed with GITHUB_TOKEN.

## What

- **auto-merge-bot-prs.yml**: arm with `DARIO_DRIFT_BOT_PAT || GITHUB_TOKEN`, the contract `cc-drift-template-watch.yml` already uses. Side benefit: PAT merges fire `push`, so auto-release runs on the spot instead of waiting for the :15 schedule.
- **auto-merge-sweep.yml**:
  - same PAT-first arming
  - `bot/template-rebake-` added to the prefix filter (event path has armed it since #1096; the webhook-loss fallback never did)
  - new step after the sweep: reap any `bot/*` branch whose PR is MERGED at exactly that tip and has no OPEN PR. Never touches a branch with an open PR, one nothing merged, or one that moved after merge.

The stray branch itself was deleted by hand today.

## Verified

- jq filters exercised against the real #1180 data via `gh --jq`: selects #1180 by MERGED + headRefOid, open count 0.
- Both files parse as YAML. actionlint runs in CI.

### DIFF
--- a/.github/workflows/auto-merge-bot-prs.yml
+++ b/.github/workflows/auto-merge-bot-prs.yml
@@ -159,7 +159,18 @@ jobs:
             )
           )
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Arm with the drift-bot PAT, GITHUB_TOKEN only as a fallback. The
+          # identity that arms native auto-merge is the identity GitHub
+          # records as the merger, and a GITHUB_TOKEN merge is loop-protected:
+          # it fires no push event (so cc-drift-auto-release waits for its
+          # hourly schedule instead of releasing on the spot) and it does NOT
+          # run the repo's delete-branch-on-merge — #1180 (bot/template-rebake-
+          # 20260901-215856) merged as github-actions and its branch stayed
+          # behind for three days, while every PAT-armed sibling had its
+          # branch deleted one second after merge. Same PAT-or-token contract
+          # as cc-drift-template-watch.yml. The sweep workflow reaps whatever
+          # this still misses.
+          GH_TOKEN: ${{ secrets.DARIO_DRIFT_BOT_PAT || secrets.GITHUB_TOKEN }}
           PR: ${{ github.event.pull_request.number }}
           AUTHOR: ${{ github.event.pull_request.user.login }}
           BRANCH: ${{ github.event.pull_request.head.ref }}
--- a/.github/workflows/auto-merge-sweep.yml
+++ b/.github/workflows/auto-merge-sweep.yml
@@ -18,7 +18,8 @@ name: Auto-merge sweep (webhook-loss fallback)
 # workflow in the release path that lacked it. `gh pr merge --auto` on a PR
 # that is already armed is a no-op, so a healthy hour costs one API listing.
 #
-# SCOPE: drift-bot branches only (`bot/cc-drift-`, `bot/template-label-`).
+# SCOPE: drift-bot branches only (`bot/cc-drift-`, `bot/template-label-`,
+# `bot/template-rebake-` — the same three prefixes the event path arms).
 # Dependabot is deliberately excluded. Its major-version exclusion comes from
 # dependabot/fetch-metadata, which needs the PR event payload and cannot run
 # on a schedule; title-parsing is not an acceptable substitute (a grouped
@@ -65,7 +66,11 @@ jobs:
 
       - name: Arm auto-merge on any unarmed drift-bot PR
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # PAT first, GITHUB_TOKEN as fallback — the arming identity becomes
+          # the merge identity, and a GITHUB_TOKEN merge neither fires the
+          # push event nor deletes the head branch (#1180). See the sibling
+          # workflow's Enable auto-merge step.
+          GH_TOKEN: ${{ secrets.DARIO_DRIFT_BOT_PAT || secrets.GITHUB_TOKEN }}
           REPO: ${{ github.repository }}
         run: |
           set -euo pipefail
@@ -84,7 +89,7 @@ jobs:
                    | select(.isDraft == false)
                    | select(.autoMergeRequest == null)
                    | select(.headRepositoryOwner.login == "'"${REPO%%/*}"'")
-                   | select(.headRefName | startswith("bot/cc-drift-") or startswith("bot/template-label-"))]')
+                   | select(.headRefName | startswith("bot/cc-drift-") or startswith("bot/template-label-") or startswith("bot/template-rebake-"))]')
 
           count=$(printf '%s' "$candidates" | jq 'length')
           echo "Unarmed drift-bot PRs: $count"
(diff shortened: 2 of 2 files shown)


sprayberry-redline (changes requested):

**Verdict:** One destructive TOCTOU race must be addressed before the branch reaper is safe.

## Findings

### Blocking — correctness/race — `.github/workflows/auto-merge-sweep.yml:170,186`

> `branches=$(gh api "repos/${REPO}/branches?per_page=100" --paginate             --jq '.[] | select(.name | startswith("bot/")) | .name + " " + .commit.sha')`
>
> `elif gh api -X DELETE "repos/${REPO}/git/refs/heads/${name}" >/dev/null; then`

The safety decision is made from the SHA captured in the initial branch listing, but the eventual deletion targets the mutable ref **by name**, without verifying that it still points to that SHA. A concrete failure case is a bot retry updating `bot/foo` after line 170 (or opening a new PR after line 176) but before line 186: the checks describe the old tip as merged/no-open-PR, then the DELETE removes the newly updated active branch. That contradicts the stated invariant that a branch moved after merge is never touched and can destroy in-progress bot work.

At minimum, re-read both the live ref SHA and open-PR state immediately before deletion and skip unless both still match the values used for the decision. If GitHub offers no compare-and-delete primitive, the robust fix is to avoid automated deletion by mutable ref name (or constrain cleanup to refs whose producer lifecycle makes later movement impossible).

```suggestion
            else
              live_sha=$(gh api "repos/${REPO}/git/ref/heads/${name}" --jq '.object.sha')
              live_open=$(gh pr list --repo "$REPO" --head "$name" --state open --limit 1 --json number --jq 'length')
              if [ "$live_sha" != "$sha" ] || [ "$live_open" != "0" ]; then
                echo "  $name KEEP — branch changed or gained an open PR during sweep"
                kept=$((kept + 1))
              elif gh api -X DELETE "repos/${REPO}/git/refs/heads/${name}" >/dev/null; then
                echo "  $name REAPED — merged as #$merged, branch left behind"
                reaped=$((reaped + 1))
              else
                echo "  $name could not delete — will retry next tick"
                kept=$((kept + 1))
              fi
            fi
```

## What's good

The PAT-first fallback is consistent across both arming paths, the rebake prefix closes the documented sweep gap, and the existing checks are green (including actionlint and CodeQL). The reaper also correctly defaults to keeping branches when it cannot prove a matching merged PR from its snapshot; the remaining issue is preserving that proof through deletion.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — the change reads correctly against its own stated invariants; one minor robustness note below.

**Scope read:** Full diff (108 lines, 2 files: `.github/workflows/auto-merge-bot-prs.yml`, `.github/workflows/auto-merge-sweep.yml`). Also pulled full file context (permissions blocks, trigger conditions, job-level `if:` guards) via a shallow clone of the head to confirm the new "Reap bot branches" step inherits the job's `if: github.repository == 'askalf/dario'` fork guard and runs with `contents: write`.

**What I checked for the new reap step** (`auto-merge-sweep.yml:156-198`):
- TOCTOU between capturing a branch's tip SHA and later calling `DELETE .../git/refs/heads/{name}` — the delete has no compare-and-swap, so a commit pushed to the same branch name in between would still be deleted. Traced whether this is reachable: `bot/template-rebake-*` branches are timestamped per-run and `bot/cc-drift-*`/`bot/template-label-*` are documented as one-shot, so the same branch name is not realistically reused within a sweep run. Not reachable in practice — dropped.
- jq-injection via `$sha` interpolated into the filter string at `auto-merge-sweep.yml:178` (`.headRefOid == "'"$sha"'"`) — `$sha` only ever comes from the GitHub API's `commit.sha` field (hex), so no characters that could break out of the quoted jq string. Not exploitable.
- Empty-branches edge case (`while read -r name sha; do ... done <<< "$branches"`) when no `bot/*` branches exist — the here-string still feeds one empty line, `[ -n "$name" ] || continue` correctly no-ops. Confirmed correct.
- The three KEEP guards (open PR / no merged PR at this tip / moved after merge) match the invariants stated in the comment above the step and in the PR description; each is checked before the delete branch of the `if/elif` chain, so they are strictly ordered ahead of the destructive path.

**Minor (info only):** `auto-merge-sweep.yml:167-193` runs under `set -euo pipefail`, and each loop iteration assigns from a command substitution (`prs=$(gh pr list ...)`). If `gh pr list` transiently fails (rate limit, network blip) on one branch, `set -e` aborts the whole step immediately rather than continuing to the next branch — the run stops with `reaped`/`kept` unreported for any branches not yet visited. Given `if: always()` and the hourly schedule this self-heals next tick, and failing closed on a destructive delete is the safer direction, so this doesn't need action, just noting it's a low-severity availability nit rather than a correctness issue with the delete logic itself.

**What's good:** The reap step's guard ordering (open PR → no merge at this tip → moved-since-merge) matches the documented invariants exactly, and the PAT-first `GH_TOKEN: ${{ secrets.DARIO_DRIFT_BOT_PAT || secrets.GITHUB_TOKEN }}` fallback pattern is applied consistently to both workflows and mirrors the existing contract in `cc-drift-template-watch.yml`, which is a reasonable precedent to reuse rather than invent a new one. The `bot/template-rebake-` prefix addition to the sweep's candidate filter (`auto-merge-sweep.yml:92`) correctly matches the prefix already armed by the event path per the PR description.

askalf:
Addressed in 5df28c1: the reaper now re-reads the live ref SHA and the open-PR count immediately before the DELETE and keeps the branch on any drift or any read failure. GitHub has no compare-and-delete for refs, so this is the narrowest window available; the per-run branch names make an actual collision unlikely but the guard costs two reads per candidate.

Also took the second-read note: a failed `gh pr list` on one branch now keeps that branch and continues instead of aborting the loop under `set -e`.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — the PAT-fallback arming and the new branch-reap logic both hold up under trace.

**Scope reviewed:** Full diff (81 additions / 4 deletions across `.github/workflows/auto-merge-bot-prs.yml` and `.github/workflows/auto-merge-sweep.yml`) — small enough to read in its entirety, including the new "Reap bot branches whose PR already merged" step.

## What I checked

- **PAT fallback (`DARIO_DRIFT_BOT_PAT || GITHUB_TOKEN`)** in both workflows: in GitHub Actions expressions an unset secret evaluates to `""`, which is falsy, so the `||` correctly falls back to `GITHUB_TOKEN` when the PAT isn't configured. Consistent with the existing `cc-drift-template-watch.yml` contract the PR body cites.
- **Reap step race handling** (`auto-merge-sweep.yml`, "Reap bot branches whose PR already merged"): the initial `/branches` listing is a snapshot, but the code re-reads `live_sha` and `live_open` immediately before the `DELETE` call and bails to KEEP on any drift or read failure. Traced the case where a bot pushes a new commit or opens a PR on the branch mid-sweep — `live_sha != sha` or `live_open != "0"` catches both and skips the delete. Sound.
- **`merged` sha match**: `select(.state == "MERGED" and .headRefOid == "'"$sha"'")` correctly requires the merged PR's head to equal the branch's *current* tip, so a branch that was force-pushed after merge (extremely unlikely, but the concern the comment raises) is correctly excluded from `merged` and therefore kept.
- **Shell quoting**: `"'"$sha"'"` embeds `$sha` inside a single-quoted jq program via proper double-quoting — no word-splitting/injection concern, and `$name` (branch name) is always used inside double-quoted arguments (`gh api "...${name}"`), not re-interpreted for embedded `$()`, so no command-injection path even in the theoretical case of an unusual branch name.
- **Empty-listing edge case**: `done <<< "$branches"` with an empty `$branches` yields one iteration with `name=""`, but `[ -n "$name" ] || continue` skips it — no spurious KEEP/REAP line.

## What's good

The extensive inline comments correctly anticipate and explain the real races (GITHUB_TOKEN loop-protection suppressing branch deletion, TOCTOU on the reap delete) rather than leaving them implicit, and the "never touches an open PR / nothing merged / moved after merge" invariant is enforced exactly as claimed in the PR description.

No test-coverage gap worth flagging beyond what the PR itself already discloses (manual `gh --jq` exercise against real #1180 data, actionlint in CI) — workflow shell logic like this isn't typically unit-testable in this repo's setup, and the manual verification described is reasonable for the change's size.

sprayberry-redline (approved):

**Verdict: Approved — the new commit addresses the prior fail-closed and stale-snapshot concerns; no blocking issues remain.**

I re-reviewed live head `5df28c1`, focusing on the branch-reaping workflow and the changes since the prior `CHANGES_REQUESTED` verdict. The PR-list failure now explicitly keeps the branch, and the workflow refreshes both the branch tip and open-PR state immediately before deletion. The merged-at-tip requirement, open-PR guard, and conservative behavior on API failures form a sound deletion policy.

**What's good:** PAT-first arming is applied consistently across the event and scheduled paths, the rebake prefix brings their scopes into alignment, and the cleanup path is deliberately fail-closed. All reported CI checks are green, including actionlint and CodeQL.

PR #1202, opened 2026-09-04
chore(template): label refresh _version -> v2.1.261 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-09-04T21:00:17+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.261** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.261`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.261`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v6.0.21` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.261. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #1203, opened 2026-09-04
ci: stop a transient contents-API blip from failing the auto-merge version gate

askalf:
## Problem

The `auto-merge` check on #1202 failed: the **Check the version advances past the base tip** step exited 1 having printed **nothing at all** — not even its own diagnostic `echo`. So it died inside `version_at()`, before the gate logic ran.

`version_at()` sends `gh api` stderr to `/dev/null` and runs under `set -euo pipefail`. Any non-zero `gh api` exit therefore kills the step immediately, with the cause discarded. The gate never got to decide anything.

That abort is the bug, independently of what the blip was:

- `scripts/version-advances.sh` **already fails closed** on a missing version — it answers `false` / "could not read package.json at every ref", leaving the PR unarmed until the next push or the hourly sweep re-runs the gate. That is exactly the right response to a transient read failure.
- `set -e` pre-empts it, converting a recoverable blip into a **red required check on an otherwise healthy PR**. Every other check on #1202 passes.

Reproduced against the live API from a container: all three refs (`ebb5500`, `66a5c6c`, `master`) read fine now — base 6.0.20, head 6.0.21, tip 6.0.20, i.e. the gate would have said *true, version advances 6.0.20 -> 6.0.21* had it run. Consistent with a transient failure. While probing, an unrelated `gh api` call in the same shell returned **HTTP 500**, which is precisely the class of error being swallowed.

## Change

Both copies of `version_at()` — `auto-merge-bot-prs.yml` (event path) and `auto-merge-sweep.yml` (hourly reconciler), which had the identical function:

- **Capture stderr instead of discarding it** (`2>&1` into a variable) and echo it per attempt, so the next occurrence is diagnosable from the log rather than from a bare exit code.
- **Retry up to 3 times, 2s apart**, covering transient 5xx / rate-limit.
- **Return empty on give-up with `return 0`**, plus a `::warning::`, so control reaches the existing fail-closed gate instead of aborting the step.

Net behaviour: a genuinely unreadable ref still blocks arming (unchanged, fail-closed), but it now does so as a *decision* the log explains, not as a step crash.

The sweep copy mattered more than it looks: there `set -e` aborted the **whole loop**, so one unreadable ref left every later candidate PR unexamined — the webhook-loss recovery path itself needing recovery.

No change to the merge rule, to `scripts/version-advances.sh`, or to which PRs qualify.

## Test plan

- `actionlint` gates both files and runs on this PR.
- Behaviour is exercised naturally: this fix landing lets #1202's next `synchronize` (or the `:27` sweep tick) re-run the gate and arm it.
- Fail-closed path is unit-covered already by `test/version-advances.mjs` on the empty-argument case, which is what give-up now produces.

## Note on #1202

#1202 needs no change of its own — its version does advance (6.0.20 -> 6.0.21). Once this merges, re-running its `auto-merge` check (or any push to the branch) should arm it.

Source ticket: 00MTNG5WWRAE8A096B7EC1C9BA


askalf:
**Third copy found — this PR's own `advise` check failed the same way.**

Pushed 1cbeb5e adding `version-bump-advice.yml` to the fix.

While the first two commits were in CI, the `advise` check on this very PR failed: run [33920374623](https://github.com/askalf/dario/actions/runs/33920374623/job/101177069430), step **"Decide whether a bump is missing"**, exit 1 with **no stdout** — the identical signature as #1202's `auto-merge` failure, in a third workflow carrying a third copy of the same `version_at()`.

That is a live, independent reproduction of the bug on a different workflow within ~15 minutes of the first, which puts the transient-blip reading beyond much doubt and shows the blast radius is all three copies, not one.

The irony is instructive: `version-bump-advice.yml:88-92` has an explicit **fail-silent** branch — "an advisory that could not read a version has nothing useful to say" — and `set -e` pre-empted it exactly as it pre-empts `version-advances.sh`'s fail-closed branch. Three workflows, three deliberately-written recovery paths, none reachable because the abort happens first.

Same treatment applied; the fail-silent branch is now reachable, so an unreadable ref yields `verdict=skip` and a green advisory run instead of a red one.

Worth noting for review: `advise` is deliberately **not** a required check (see the header comment), so its failure never blocked anything — but a permanently-red advisory check is one people stop reading, which is the failure mode that comment was written to avoid.

askalf:
**Verification summary — all checks green, ready for review.**

Before/after on live infrastructure, which is the strongest evidence available for a transient-failure fix:

| | `advise` on this PR |
|---|---|
| before 1cbeb5e | **fail**, exit 1, zero stdout ([run 33920374623](https://github.com/askalf/dario/actions/runs/33920374623/job/101177069430)) |
| after 1cbeb5e | **pass** ([run 33920547998](https://github.com/askalf/dario/actions/runs/33920547998/job/101177614369)) |

The passing run's log shows the full path executing rather than just not-crashing:

```
base=6.0.20 head=6.0.20 -> needs_bump=false (no shipping files changed - a release would carry nothing)
no bump needed and no existing comment; staying quiet
```

Every other check is green; `auto-merge` reports `skipping`, which is correct — this is a human branch, not `bot/*`.

### What is verified vs. what review should weigh

**Verified:** the reads work, the retry wrapper does not break the happy path, the verdict is computed correctly, and `actionlint`/shellcheck accept all three `run:` blocks.

**Not directly exercised:** the give-up path (3 failed attempts → empty → recovery branch). Forcing that would mean injecting an API failure into a `pull_request_target` run. It is covered indirectly — an empty version is exactly the input `test/version-advances.mjs` already asserts fails closed, and `advise`'s skip branch is a plain `[ -z ]` test — but the retry-then-give-up sequence itself is reasoned, not observed.

If a reviewer wants that closed before merge, the cheap probe is a scratch PR with `REPO` set to a nonexistent repo: every attempt 404s, and the log should show three `version_at(...): attempt N failed: ...` lines followed by the recovery branch and a **green** step. I stopped short of that as it needs a throwaway PR against master to be meaningful.

### Follow-up for #1202

#1202 needs no change of its own — 6.0.20 → 6.0.21 does advance. Once this merges, re-running its `auto-merge` check (or any push to its branch) should arm it. Filed the merge approval with the operator.

Source ticket: 00MTNG5WWRAE8A096B7EC1C9BA

sprayberry-redline (approved):

**Verdict: no blocking issues found.**

I reviewed all three workflow changes, focusing on `set -euo pipefail` behavior, retry termination, command-substitution output, malformed/empty API responses, and the fail-closed handoff to the existing version gate. The functions keep diagnostics on stderr, emit only a valid version on stdout, cap retries at three attempts, and deliberately return success with empty output after exhaustion so downstream policy—not shell error handling—keeps the PR unarmed.

**What's good:** The equivalent readers in the event, sweep, and advice paths were updated consistently; the sweep no longer aborts later candidates after one transient API failure. The PR's actionlint, CI, CodeQL, and live-test checks are green.

sprayberry-secondread (commented):

Verdict: 1 reliability issue to address; the event-path hardening otherwise reads well.

## Findings

### Medium — `.github/workflows/auto-merge-sweep.yml:88-105`: retries can exhaust the sweep's five-minute job budget

> ```sh
> if content=$(gh api "repos/${REPO}/contents/package.json?ref=${ref}" --jq '.content' 2>&1); then
> ```
>
> ```sh
> if [ "$attempt" -ge 3 ]; then
>   break
> fi
> attempt=$((attempt + 1))
> sleep 2
> ```

The sweep calls `version_at` three times for each candidate. If the contents API is unavailable, each call sleeps four seconds before giving up, so each candidate consumes at least 12 seconds just in backoff (plus nine failed API calls). With the existing 50-candidate limit, the new path needs at least 10 minutes before request time, while this job is limited to five minutes. In that failure state the runner is terminated mid-loop, recreating the very “later candidate went unexamined” outcome this change is meant to prevent.

Bound retry work by the remaining job budget (or avoid per-ref retries in the sweep and let its next scheduled run retry), so a degraded API can still reach every candidate.

```suggestion
# Keep the scheduled sweep bounded: fail this lookup closed immediately and
# let the next tick retry it, rather than multiplying backoff by every PR/ref.
version_at() {
  local ref="$1" content ver
  if ! content=$(gh api "repos/${REPO}/contents/package.json?ref=${ref}" --jq '.content' 2>&1); then
    echo "::warning::version_at(${ref}) failed: ${content}" >&2
    return 0
  fi
  ver=$(printf '%s' "$content" | base64 -d 2>/dev/null | jq -r '.version // empty') || ver=''
  printf '%s' "$ver"
}
```

## What's good

The `if`-guarded command substitution prevents `set -e` from aborting on a failed `gh api`, preserves diagnostics, and returns an empty value to the existing fail-closed version gate. I also checked the three changed shell blocks for Bash syntax; they parse successfully. I could not run `actionlint` because it is unavailable in this environment.

PR #1204, opened 2026-09-04
codex: flatten chat-shape named tool_choice before /responses

askalf:
## Problem

`chatCompletionsToResponses` forwarded `tool_choice` verbatim. A chat/completions named choice

```json
{"type":"function","function":{"name":"get_weather"}}
```

reaches the Codex Responses backend un-flattened, which answers:

```
400 {"error":{"message":"Missing required parameter: 'tool_choice.name'.","type":"invalid_request_error"}}
```

The Responses forced form is flat — `{"type":"function","name":"get_weather"}` — and returns 200 with a `function_call`. Cursor / Continue / Aider send the nested form whenever they force a tool, so this broke every forced-tool request through the Codex path. Verified live 2026-09-04.

## Fix

New pure exported helper `toResponsesToolChoice()` in `src/codex-backend.ts`, mirroring `translateToolChoice()` on the Anthropic path:

- object with `.function.name` (non-empty string) → `{type:'function', name}`
- strings (`"auto"`/`"none"`/`"required"`) → unchanged
- anything else (unknown object, null) → passed through untouched; an unknown shape is the backend's to reject, not ours to guess at

## Test plan

`test/codex-backend.mjs` — the existing 3 `tool_choice` checks only covered `"auto"`. Added coverage for the nested named choice flattening, `"auto"` staying a string, and an unknown object passing through by identity.

`npm ci && npm run build && npm test`

Source ticket: DEV-c132b3de

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+- **Forced tool calls work again on the Codex backend.** A chat/completions named `tool_choice` — `{"type":"function","function":{"name":"f"}}`, which Cursor, Continue and Aider send whenever they force a tool — reached the Responses backend un-flattened and 400'd with `Missing required parameter: 'tool_choice.name'`. It is now translated to the flat Responses form `{"type":"function","name":"f"}`, matching what the Anthropic path already did. String choices (`auto`/`none`/`required`) and unrecognized objects pass through unchanged.
+
 ## [6.0.21] - 2026-09-04
 
 - **Template label refresh** — `_version`, `_supportedMaxTested`, and the `user-agent` header bumped to `2.1.261` to track `@anthropic-ai/claude-code@latest`. The live wire shape is unchanged — cc-drift-template-watch ran `capture-and-bake --check` against live CC v2.1.261 and found zero shape drift vs the bundle — so this is a label refresh, not a re-capture (`_captured` stays at the last real capture). Auto-merged; clears the `sdk-drift` early-warning signal.
--- a/src/codex-backend.ts
+++ b/src/codex-backend.ts
@@ -293,6 +293,29 @@ function messageContentToText(content: unknown): string {
 }
 
 /**
+ * Translate a chat/completions `tool_choice` into the Responses shape.
+ *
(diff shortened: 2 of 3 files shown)


sprayberry-redline (approved):

**Verdict: no blocking issues found.**

I checked the full 52-addition/1-deletion diff, including the conversion at the `chatCompletionsToResponses` boundary, pass-through behavior for string and unrecognized choices, and the new regression coverage. The helper correctly turns the chat/completions named-tool shape into the flat Responses API shape without changing the existing string forms. CI is fully green, including Node 18/20/22 builds, tests, compatibility, live-test, and CodeQL.

**What’s good:** The fix is narrowly placed at the API translation boundary, preserves unknown values for backend validation, and directly tests both the reported failure shape and non-regression cases.

sprayberry-secondread (commented):

Verdict: 1 correctness issue to address; the intended standard forced-tool conversion and its focused tests otherwise read well.

## Findings

### Medium — `src/codex-backend.ts:310-314`: unknown objects with a `function.name` are not passed through unchanged

> ```ts
> const c = choice as { function?: { name?: unknown } };
> const name = c.function?.name;
> if (typeof name === 'string' && name.length > 0) {
>   return { type: 'function', name };
> }
> ```

The helper's documented contract is to leave unknown objects untouched, but these changed lines use only the nested `function.name` as the discriminator and discard the object's actual `type` plus every other field. For example, a caller supplying `{ type: 'vendor_extension', function: { name: 'get_weather' }, mode: 'strict' }` is silently rewritten to `{ type: 'function', name: 'get_weather' }`, rather than reaching the backend unchanged for it to validate. That changes the meaning of an unrecognized shape and contradicts both the helper documentation and the PR's stated compatibility behavior.

```suggestion
const c = choice as { type?: unknown; function?: { name?: unknown } };
const name = c.function?.name;
if (c.type === 'function' && typeof name === 'string' && name.length > 0) {
  return { type: 'function', name };
}
```

Please also add a regression assertion for an unknown `type` that happens to contain `function.name`.

## What's good

The conversion is isolated in a pure exported helper, preserves the supported string choices, and the new integration assertion verifies that the nested forced-tool form is flattened before the Responses request is built.

PR #1205, opened 2026-09-04
codex: streamed chat-shape response.failed emits an error frame, not finish_reason stop

askalf:
## Problem

A `/v1/chat/completions` request with `stream: true` to a codex slug, where upstream ends with `response.failed`, delivered this to the client:

```
data: {..."delta":{"role":"assistant","content":""},"finish_reason":null}
data: {..."delta":{},"finish_reason":"stop"}
data: [DONE]
```

To every OpenAI SDK that is a **successful, empty completion** — the failure disappears. `createResponsesTranslator` treated all three terminal Responses events identically; `didFail()` was consulted only for the non-streaming collapse.

## Fix

`src/codex-backend.ts` — on a terminal event that is a failure (`response.failed`, or a terminal response whose `status` is `failed` / that carries an `error`), the chat-shape translator now emits

```
data: {"error":{"message":"<upstream message>","type":"server_error","code":<upstream code or null>}}
data: [DONE]
```

with **no** `finish_reason:"stop"` frame. openai-node's `ChatCompletionStream` and openai-python both raise on a chunk carrying `error`, which is the correct client outcome. The role-opener frame is unchanged, so the stream is still well-formed. New exported helper `failedResponseCode()` alongside the existing `failedResponseMessage()`.

Success streams are untouched: `response.completed` / `.incomplete` still close with `finish_reason` `stop` / `tool_calls`. Analytics already reports 502 for this case — left as is.

## Tests (`test/codex-backend.mjs`)

- Rewrote *"response.failed — chat path terminates the stream"*, which asserted the buggy behaviour as correct: it now asserts the error frame is present with the upstream message/type/code, that **no** `finish_reason` frame is emitted at all, that the role opener still leads, and that `[DONE]` still terminates.
- New: `response.completed keeps the chat success contract unchanged` — regression guard on `finish_reason:"stop"`, no error frame, `didFail() === false`.
- New: end-to-end through `forwardToCodex` with `stream:true` + openai shape — error frame on the wire, no `"finish_reason":"stop"`, ends with `[DONE]`.

## Test plan

`npm ci && npm run build && npm test`

Source ticket: DEV-d1e82cad

### DIFF
--- a/src/codex-backend.ts
+++ b/src/codex-backend.ts
@@ -407,6 +407,12 @@ export function failedResponseMessage(resp: unknown): string {
   return msg || code || 'the Codex backend ended the response as failed';
 }
 
+/** The upstream error code on a failed Responses payload, or null when absent. */
+export function failedResponseCode(resp: unknown): string | null {
+  const e = (resp as { error?: { code?: unknown } } | null)?.error;
+  return e && typeof e.code === 'string' && e.code ? e.code : null;
+}
+
 interface ToolCallAccumulator {
   index: number;
   id: string;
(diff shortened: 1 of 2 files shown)


sprayberry-redline (approved):

I reviewed the complete diff and the surrounding Responses-to-chat translator/forwarding path. The terminal failure branch now emits an SDK-visible error frame followed by `[DONE]`, without incorrectly signaling a successful `finish_reason: "stop"`; successful completion and tool-call termination remain on the existing path. The new helper safely preserves an upstream string code or returns `null`.

**What's good:** The regression coverage checks both translator output and the end-to-end streamed wire response, explicitly asserts absence of any finish-reason frame on failure, preserves the role opener, and guards the successful completion contract. The PR's CI rollup is green.

sprayberry-secondread (commented):

Verdict: no blocking issues found in the scoped 58-line change.

I checked the terminal-event state transition, wire ordering and SSE framing for both the direct translator and `forwardToCodex` OpenAI streaming path. In particular, a `response.failed` now marks the translator failed, preserves the role opener, sends the upstream message/code in an error frame, avoids a success `finish_reason`, and terminates with `[DONE]`; successful terminal streams retain their existing finish frame. The added unit and end-to-end coverage exercise those contracts.

What's good: the failure branch is deliberately placed after usage capture and before the normal terminal frame, which keeps reporting data while preventing the erroneous successful-empty-completion signal.

Note: I could not execute the test suite in this fresh shallow checkout because dependencies are absent (`tsc: not found`); this review is based on the diff and surrounding implementation/test inspection.

PR #1207, opened 2026-09-04
fix(codex): carry chat images and report lossy fields

askalf:
Fixes DEV-95f047d2.\n\n- Preserve OpenAI chat `image_url` parts as Responses `input_image` URL strings.\n- Follow-up commit will add verbose one-time lossy-field diagnostics, documentation, and tests.\n\nTests will run in CI; local build/test intentionally not run per engineering build-wall policy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,8 @@ curl localhost:3456/v1/messages -H 'content-type: application/json' \
 
 Streaming, tool calls, and tool-result round trips work on both shapes: dario translates chat/completions **or** Messages into the Responses API the subscription backend speaks, and translates the stream back into `chat.completion.chunk` or Anthropic message events to match what the client asked in. There is no `/v1/responses` inbound yet.
 
+**Chat/completions fidelity:** text and `image_url` user-content parts (HTTPS URLs and data URIs) carry through to Responses input items. The Codex subscription backend does not accept every chat field, so `response_format`, `stop`, `n`, `logprobs`, `stream_options`, and other unmapped chat-only fields are intentionally lossy; with `--verbose`, dario reports each dropped field once per process.
+
 Codex accounts live in `~/.dario/codex-accounts/`, entirely separate from the Claude pool. Nothing about `dario login`, `dario accounts`, or Claude routing changes.
 
 ---
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 <p><strong>One local endpoint. Every AI tool you own. The subscriptions you already pay for.</strong></p>
 
-<sub><code>npm i -g @askalf/dario</code> · <strong>0</strong> runtime deps · <a href="https://www.npmjs.com/package/@askalf/dario">SLSA-attested</a> every release · nothing phones home · ~29k lines you can read in a weekend · independent, unofficial, third-party (<a href="DISCLAIMER.md">DISCLAIMER.md</a>)</sub>
+<sub><code>npm i -g @askalf/dario</code> · <strong>0</strong> runtime deps · <a href="https://www.npmjs.com/package/@askalf/dario">SLSA-attested</a> every release · nothing phones home · ~30k lines you can read in a weekend · independent, unofficial, third-party (<a href="DISCLAIMER.md">DISCLAIMER.md</a>)</sub>
 
 <sub>Part of <a href="#own-your-stack"><strong>Own Your Stack</strong></a> — 12 open tools for owning your AI infra: <a href="https://github.com/askalf/truecopy">truecopy</a> · <a href="https://github.com/askalf/strongroom">strongroom</a> · <a href="https://github.com/askalf/fieldpass">fieldpass</a> · <a href="https://github.com/askalf/plumbline">plumbline</a> · <a href="#own-your-stack">full family ↓</a></sub>
 
@@ -309,7 +309,7 @@ The split isn't live, but it was announced once on short notice and could return
 
 | Signal | Status |
 |---|---|
-| Source | **~29k** lines of TypeScript across **59** files — auditable in a weekend (v5 removed shim; the pool is the one code path) |
+| Source | **~30k** lines of TypeScript across **65** files — auditable in a weekend (v5 removed shim; the pool is the one code path) |
 | Dependencies | **0 runtime.** Verify: `npm ls --production` |
 | Provenance | Every release [SLSA-attested](https://www.npmjs.com/package/@askalf/dario) via GitHub Actions + Sigstore |
 | Scanning | [CodeQL](https://github.com/askalf/dario/actions/workflows/codeql.yml) on every push and weekly |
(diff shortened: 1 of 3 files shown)


sprayberry-redline (changes requested):

1 correctness issue to address; image-part preservation itself reads well.

## Findings

### Blocking — `src/codex-backend.ts:280-281`

> `  'model', 'messages', 'tools', 'tool_choice', 'temperature', 'top_p',`
> `  'max_tokens', 'max_completion_tokens', 'reasoning_effort', 'stream',`

The new diagnostic treats every field translated by `chatCompletionsToResponses` as preserved, but that is not the actual transport contract. `temperature` and `top_p` are translated and then removed by `toCodexSupportedBody`; `max_tokens` / `max_completion_tokens` become `max_output_tokens`, which is removed there as well. For example, with `--verbose` and `{ "temperature": 0, ... }`, this set suppresses the warning even though the requested temperature never reaches Codex. That contradicts the new README promise that verbose mode reports each dropped field and leaves users unaware of precisely the lossy parameters this feature is intended to expose.

Use an end-to-end supported-field set for this diagnostic (or compare the translated and scrubbed bodies), and cover the translated-then-scrubbed case in the forwarding tests.

```suggestion
  'model', 'messages', 'tools', 'tool_choice', 'reasoning_effort', 'stream',
```

## What's good

The text/image conversion preserves part ordering, handles both object and string `image_url` forms, and adds focused coverage for mixed and image-only messages. The current CI rollup is green.

sprayberry-secondread (commented):

Verdict: 1 correctness issue to address; image URLs are preserved, but a supported image fidelity control is lost.

## Findings

### Medium — `src/codex-backend.ts:322-323`: image `detail` is silently discarded

> ```ts
> const imageUrl = typeof p.image_url === 'string' ? p.image_url : (p.image_url as { url?: unknown } | undefined)?.url;
> if (typeof imageUrl === 'string') parts.push({ type: 'input_image', image_url: imageUrl });
> ```

The Chat Completions `image_url` object also permits `detail` (`"low"`, `"high"`, or `"auto"`), and the Responses `input_image` representation supports that same setting. For example, a client sending `{ type: "image_url", image_url: { url: "https://example.test/scan.png", detail: "low" } }` is translated without `detail`; the backend therefore defaults to its automatic/high-fidelity behavior instead of honoring the requested low-detail mode. That changes image-token cost and can change the model's visual input for requests this PR intends to preserve. Carry the validated detail field alongside the URL, and add a test for it.

```suggestion
const image = typeof p.image_url === 'object' && p.image_url !== null
  ? p.image_url as { url?: unknown; detail?: unknown }
  : undefined;
const imageUrl = typeof p.image_url === 'string' ? p.image_url : image?.url;
const detail = image?.detail;
if (typeof imageUrl === 'string') {
  parts.push({
    type: 'input_image',
    image_url: imageUrl,
    ...(detail === 'auto' || detail === 'low' || detail === 'high' ? { detail } : {}),
  });
}
```

## What's good

The translation preserves mixed text/image ordering, accepts both Chat Completions URL encodings, covers image-only messages, and the one-time verbose diagnostics make explicitly unsupported top-level fields observable.

I reviewed the complete 64-line diff and relevant translation types/context. I could not run the test suite because this shallow review checkout has no installed TypeScript compiler (`tsc: not found`).

askalf:
Note: this PR modifies shipped code (`src/codex-backend.ts`, `test/codex-backend.mjs`). If the current code review is resolved and the PR proceeds to merge, **package.json version must be bumped** for the release workflow to publish this change. The version gate checks for a version bump between the merge parent and HEAD; merging without one ships the code but skips the npm release. (Current version: 6.0.21)

sprayberry-redline (approved):

No blocking issues found on the live head; the prior diagnostic bug is fixed.

I reviewed the full 149-addition diff, with particular attention to the end-to-end chat-field diagnostic, the image-part conversion (including `detail` validation and part ordering), and the new regression coverage. `chatFieldReachesCodex` now checks each translated target against the actual Codex transport allowlist, so translated-then-scrubbed sampling fields are correctly reported as dropped. The mixed-image, image-only, fidelity-setting, and once-per-process diagnostic cases are covered.

## What's good

The implementation keeps the translation and backend allowlist concerns explicit, documents the intentionally lossy parameters accurately, and adds focused tests for the previously requested correction. The current CI rollup is green across tests, builds, live test, and CodeQL.

sprayberry-secondread (commented):

**Verdict:** One correctness gap worth a follow-up (silently-empty user content for unrecognized chat parts); everything else in this diff reads correctly and is well-tested.

## Findings

### Medium — unrecognized `content` part types can produce an empty `content: []` for a user turn
`src/codex-backend.ts:337-365` (new `chatContentToResponsesParts`) and the call site at `src/codex-backend.ts:418`:

```js
function chatContentToResponsesParts(content: unknown): Array<Record<string, string>> {
  if (!Array.isArray(content)) return [{ type: 'input_text', text: messageContentToText(content) }];
  const parts: Array<Record<string, string>> = [];
  for (const part of content) {
    const p = part as { type?: unknown; text?: unknown; image_url?: unknown };
    if (p?.type === 'text' && typeof p.text === 'string') {
      parts.push({ type: 'input_text', text: p.text });
      continue;
    }
    if (p?.type === 'image_url') {
      ...
    }
  }
  return parts;
}
```
```js
    input.push({ role: 'user', content: chatContentToResponsesParts(m.content) });
```

**Scenario:** a chat/completions user message whose `content` array contains only part types this function doesn't recognize — e.g. `content: [{ type: 'input_audio', input_audio: {...} }]` (a real OpenAI chat/completions content-part type for audio-capable models), or a legacy/loosely-typed part like `{ text: 'hi' }` with no `type` field. Every element of the loop falls through both `if` branches, `parts` stays `[]`, and the function returns an empty array. The unconditional `input.push({ role: 'user', content: [] })` then sends a user turn with **zero content items** to the Responses backend.

Before this PR, `messageContentToText` handled this same input by extracting `p?.text` from any part regardless of `type` and joining into one string, guaranteeing the pushed item was always `content: [{ type: 'input_text', text: <possibly empty string> }]` — never an empty array. This diff changes that invariant: a message can now silently lose its only content item instead of degrading to an empty-text placeholder, and there is no test in `test/codex-backend.mjs` covering an all-unsupported-parts message (the closest test, "image-only chat message remains a valid user input item", always has exactly one recognized part).

```suggestion
  return parts.length > 0
    ? parts
    : [{ type: 'input_text', text: messageContentToText(content) }];
```

## What's good
- The translation/scrub interaction bug class this PR targets (a field can be *translated* by name but still *scrubbed* by `toCodexSupportedBody`) is handled correctly: `chatFieldReachesCodex` checks both the translation map and `CODEX_SUPPORTED_FIELDS`, and the new tests (`test/codex-backend.mjs`, "verbose reports fields that are translated and THEN scrubbed") explicitly cover `temperature`/`top_p`/`max_tokens`/`max_completion_tokens` being reported as dropped despite having a translation entry — exactly the subtle case the design comment calls out.
- `detail` is allowlisted to `'auto' | 'low' | 'high'` before being forwarded (`src/codex-backend.ts:355-359`), so a bogus value can't be smuggled through — covered by the "bogus detail value is dropped" test.
- `loggedUnsupportedChatFields` correctly logs once per process rather than once per request, and the test proves it with a second call producing no output.

PR #1208, opened 2026-09-04
ci: add inert Codex drift watcher

askalf:
Adds an opt-in self-hosted Codex drift workflow, live OpenAI wire-contract validator, model snapshot fixture, and operator enablement notes. The job is skipped unless the operator configures the DARIO_CODEX_LIVE_HOME repository variable on the dario-drift runner. Source ticket: DEV-2fb78dea.


sprayberry-redline (changes requested):

**Verdict:** One correctness gap in the new live contract check needs to be addressed.

## Findings

### Blocking — `test/manual/codex-wire-contract.mjs:62`

> `for (const call of tools.body?.choices?.[0]?.message?.tool_calls || []) {`
> `  assert(typeof call?.function?.arguments === 'string', 'tool-call arguments must be a JSON string');`
> `  JSON.parse(call.function.arguments);`
> `}`

The fallback to `[]` makes the tool-call check succeed when the response contains no `tool_calls` at all. For example, if a Codex model ignores the tool instruction and returns ordinary assistant text, the loop executes zero times and the workflow still reports the wire contract as passing. That leaves the behavior this request is intended to validate entirely untested. Assert that at least one call is present before validating each call's arguments.

```suggestion
  const toolCalls = tools.body?.choices?.[0]?.message?.tool_calls;
  assert(Array.isArray(toolCalls) && toolCalls.length > 0, 'tool completion lacks a tool call');
  for (const call of toolCalls) {
    assert(typeof call?.function?.arguments === 'string', 'tool-call arguments must be a JSON string');
    JSON.parse(call.function.arguments);
  }
```

## What's good

The workflow is opt-in, pins its third-party actions by SHA, limits permissions, masks the account identifier, and reliably tears down the background proxy. The non-streaming, streaming, and malformed-request checks otherwise cover the main wire shapes clearly.

CI has no reported checks on this head, so this verdict is based on the complete 150-line diff review.

sprayberry-secondread (commented):

Verdict: 1 correctness issue to address; the live wire checks otherwise have sensible opt-in and failure handling.

## Findings

### Medium — `.github/workflows/codex-drift-watch.yml:57-58` — the documented snapshot-bootstrap flow cannot obtain the generated snapshot

> `fs.writeFileSync('live-models.json',JSON.stringify({models},null,2)+'\\n')`
>
> `if [ "$fetch_models" -eq 0 ]; then cmp -s test/fixtures/codex-models.snapshot.json live-models.json; models=$?; else models=1; fi`

The committed fixture is `{ "models": [] }`, while a configured live Codex account will normally return at least one model. On the first enabled run, `cmp` therefore fails and opens a drift issue. The operator notes say to “review the resulting `live-models.json` before seeding” the fixture, but this job neither logs that file nor uploads it as an artifact, so it is unavailable from the run to review/copy. Consequently the initial baseline cannot be seeded using the documented workflow and the issue persists on every scheduled run.

```suggestion
      - name: Upload live model snapshot
        if: always() && hashFiles('live-models.json') != ''
        uses: actions/upload-artifact@<pinned-SHA>
        with:
          name: codex-live-models-${{ github.run_id }}
          path: live-models.json
          retention-days: 7
```

Alternatively, add an explicit bootstrap mode that writes a reviewed PR update; in either case, make the generated snapshot available before comparing it to the committed baseline.

## What's good

The workflow is guarded to the canonical repository and an explicit runner variable, uses a dedicated HOME without copying credentials, and correctly converts the intentionally non-failing check step into a visible drift issue.

Checked the full 149-line diff, including the manual request assertions, workflow control flow, credential masking, snapshot comparison, and issue lifecycle. I could not run the TypeScript build locally because this shallow review checkout does not include installed dependencies (`tsc` is unavailable).

sprayberry-redline (approved):

**Verdict:** No blocking issues found on the live head; the prior tool-call coverage and snapshot bootstrap concerns are addressed.

I reviewed the full 162-line diff, with particular attention to the two commits added since the previous `CHANGES_REQUESTED` verdict. The live contract now explicitly fails when the model returns no tool call, and the drift workflow uploads the generated model list so the initially empty fixture can be seeded from the first enabled run. I also checked the workflow's opt-in/repository guards, account-file handling, proxy cleanup and readiness path, drift/recovery issue behavior, and the non-stream/SSE/error wire assertions.

**What's good:** Credentials remain read in place, sensitive account metadata is masked, the self-hosted job is inert until explicitly configured, actions are SHA-pinned, and all current CI checks are green.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — this is a well-guarded, opt-in workflow addition; a couple of info-level observations below.

**Scope:** Read the full diff for all 4 changed files (`.github/workflows/codex-drift-watch.yml`, `CLAUDE.md`, `test/fixtures/codex-models.snapshot.json`, `test/manual/codex-wire-contract.mjs`) — the PR is small enough (162 additions) that no hunks were skipped.

## What's good
- The new workflow is inert by default (`if: github.repository == 'askalf/dario' && vars.DARIO_CODEX_LIVE_HOME != ''`), so it can't fire on forks or before an operator opts in.
- `trap 'kill "$pid" 2>/dev/null || true' EXIT` is set immediately after backgrounding the proxy (`.github/workflows/codex-drift-watch.yml`), so the process is cleaned up even if the health-check loop times out or a later step exits early.
- The account id is masked (`[ -z "$account_id" ] || echo "::add-mask::$account_id"`) before it could otherwise land in logs, and no credential is copied out of the operator-configured `HOME`.
- `test/manual/codex-wire-contract.mjs` is clearly kept out of the default `npm test` path via its own comment and directory, matching the "manual/live only" intent described in the PR body.

## Info-level observations (non-blocking)
- `test/manual/codex-wire-contract.mjs` line: `const frames = streamed.text.split('\n\n').filter(frame => frame.startsWith('data: '));` assumes SSE frames are separated by exactly `\n\n` with no `\r\n\r\n` variant. Since this script only runs opt-in on a self-hosted runner against a live account, a mismatch would just fail the check loudly rather than cause a silent wrong result — flagging for awareness only, not requesting a change.
- The comment in the workflow (`# The committed fixture ships empty, so the first enabled run reports drift by construction...`) is honest about the day-one drift report the empty `test/fixtures/codex-models.snapshot.json` will produce; this is by design and documented in `CLAUDE.md`, so no action needed.

No correctness bugs, reuse/simplification opportunities, or efficiency concerns found in the changed lines.

PR #1211, opened 2026-09-05
ci(canary): a borrowed-credential block is a warning, not a red run

askalf:
## What

The CC billing classifier canary went red on a condition that fixes itself.

Run [33958011031](https://github.com/askalf/dario/actions/runs/33958011031) failed at 09:27Z on master `90538ae`, step *Start dario proxy*:

```
##[error]dario proxy never reached /health within 30s
[dario] Startup refresh failed for login: token refresh is disabled
        (DARIO_NO_TOKEN_REFRESH=1) and the access token needs renewing...
  OAuth: healthy (expires in 0h 0m)
```

The re-run at 11:55Z passed with no change to anything. That is the whole failure mode: this job borrows production's subscription OAuth **read-only** so it can never rotate the token out from under the fleet (the 2026-08-25 ~3h `invalid_grant` outage), and the price of that is that a renewal landing on the canary's slot leaves it unable to probe. It says nothing about the billing classifier and clears the moment the platform dario — the sole refresher — rotates.

Two defects turned that into a red run and a fleet-watch ticket:

**1. The readiness loop could not distinguish "no answer" from "answered 503."** `curl -sf` fails on both. dario serves `/health` with 503 when the OAuth axis is broken (see `cc-oauth-health.yml`), so the borrow-block presents as *proxy up, health red* — and the loop reported `never reached /health within 30s`, which is false and points at the wrong thing. The job then died in the startup step, **never reaching the careful borrow-mode diagnosis already written into the probe step** (lines 217-234 on master), which was therefore dead code for this exact case.

Now the loop accepts any 2xx, keeps the last status code, and on a genuine startup failure reports it plus the `/health` body.

**2. Both borrow-block paths exited 1, so a single miss went red.** A blocked run now opens/updates a `cc-canary-blocked` tracking issue and exits 0. The issue's age is the streak counter; the run only fails once it has stayed open for `BLOCKED_ESCALATE_HOURS` (36h — one missed daily slot plus margin). Any run that produces a verdict closes it. So a genuinely blind canary still escalates, while a one-off self-clearing block does not — the same "one miss is noise, a streak is signal" shape as `cc-drift-watcher-liveness.yml`.

## What is deliberately unchanged

- `DARIO_NO_TOKEN_REFRESH=1` stays. Never refreshing is the point.
- The billing/model verdict logic, the `cc-billing-canary` alert, and the `force_status` validation override are untouched.
- The probe step still writes **no** `status` output when blocked, so the billing-alert step stays gated off and cannot open a misleading classifier issue.

## Test plan

- `actionlint` (CI) covers the YAML + shell.
- Verdict paths unaffected: `gh workflow run cc-billing-classifier-canary.yml --ref <branch> -f force_status=fail` (opens alert) then `-f force_status=pass` (closes it) still exercise open/close end to end.
- The blocked path cannot be forced synthetically — it needs the shared token to actually want renewal. Next natural occurrence should now show: startup step green-with-warning, probe skipped, a `cc-canary-blocked` issue opened, run conclusion **success**; and the following successful run closes that issue.
- Not verifiable here: the escalation branch waits 36h by construction.

Source ticket: `01M1RQN6EMJW5NTSVT98V71EM8`


askalf:
## Verified on the self-hosted runner

Dispatched this branch against the real runner: [run 33965929184](https://github.com/askalf/dario/actions/runs/33965929184) — **success**, every step as designed.

```
proxy ready after 2s (HTTP 200)
Verdict: billing=pass bucket=subscription | model=pass served=claude-opus-5 | status=pass
Canary healthy: claim=five_hour bucket=subscription served=claude-opus-5
```

- `Start dario proxy` — new 2xx-accepting readiness loop passes and now logs the status code (`HTTP 200`).
- `Send canary request` — runs, real verdict unchanged (`five_hour` → `subscription`, opus-5 served as asked).
- `Track / escalate blocked canary` — correctly **skipped** (no block this run).
- `Open / update / close canary alert` — ran; healthy path intact.

CI: `actionlint` pass, all build matrix pass.

### What this run does NOT prove

The blocked path itself. It cannot be forced synthetically — it needs the shared access token to genuinely want renewal, which is exactly what happened at 09:27Z today and had cleared by 11:55Z. Left for observation rather than chased here.

**Next probe, if you want to close the loop:** at the next natural occurrence, confirm the run ends `success` with the startup step green-carrying-a-warning, the probe step skipped, and a `cc-canary-blocked` issue opened — then that the following successful run closes it. Reviewer sanity-check worth more than a synthetic reproduction: the `blocked` output is set on the same line that `exit 0`s, so if the `$GITHUB_OUTPUT` write and the step gating disagree, the probe would run against a dead proxy rather than skip.

sprayberry-redline (changes requested):

**Verdict:** One correctness issue blocks the intended two-consecutive-miss escalation; the rest of the workflow change reads well.

## Findings

### Blocking — `.github/workflows/cc-billing-classifier-canary.yml:94,406`

> `BLOCKED_ESCALATE_HOURS: '36'`
>
> `if [ "$age_h" -ge "$BLOCKED_ESCALATE_HOURS" ]; then`

This does not implement the documented “two in a row” policy for a daily canary. If the first blocked run opens the issue at 05:50 on day 1 and the next scheduled run is also blocked at 05:50 on day 2, the issue is only about 24 hours old, so `24 -ge 36` is false and the second consecutive miss still exits successfully. The workflow does not turn red until the third blocked daily run, roughly 48 hours after the issue opened. That leaves the canary blind for an extra full interval beyond the stated policy.

Use a threshold no greater than one schedule interval, or count blocked runs explicitly if the 36-hour wall-clock grace is intentional. For the stated “second consecutive daily miss goes red” behavior:

```suggestion
  BLOCKED_ESCALATE_HOURS: '24'
```

## What's good

The startup step now distinguishes an unhealthy HTTP response from no response, preserves the no-refresh safety invariant, and keeps blocked runs from producing a misleading billing verdict. CI is fully green, including actionlint and the test/build checks.

sprayberry-secondread (commented):

**Verdict:** Solid fix for the stated false-alarm problem; one cosmetic diagnostic bug and one narrow TOCTOU on the new tracking-issue step, neither blocking.

## Findings

**1. Low — doubled/garbled `last_code` in the error diagnostic**
`.github/workflows/cc-billing-classifier-canary.yml:193`
```
last_code=$(curl -s -o health.json -w '%{http_code}' --max-time 2 http://127.0.0.1:3457/health || echo 000)
```
`curl -s -w '%{http_code}'` (without `-f`) still prints the status code even when the connection itself fails — a refused connection or a timeout returns curl exit 7/28 *and* prints `000` on stdout. Because the command substitution's own exit status also comes out nonzero in that case, the `|| echo 000` fires too, so `last_code` becomes the concatenation `000000`, not `000`. I reproduced this directly:
```
$ last_code=$(curl -s -o /tmp/h -w '%{http_code}' --max-time 1 http://127.0.0.1:59999/health || echo 000)
$ echo "[$last_code]"
[000000]
```
The `case "$last_code" in 2??)` match still correctly falls through to "not ready" (behavior is unaffected), but the new diagnostic this PR adds at line 216 —
```
echo "::error::dario proxy never reached a healthy /health within 30s (last HTTP status: ${last_code})"
```
— will print `last HTTP status: 000000` on the most common startup-failure path (proxy not yet listening / connection refused), which is exactly the message this PR is trying to make trustworthy. Cosmetic, not correctness-breaking, but worth a one-line fix.
```suggestion
last_code=$(curl -s -o health.json -w '%{http_code}' --max-time 2 http://127.0.0.1:3457/health) || last_code=000
```

**2. Low/Info — new tracking-issue step has a TOCTOU that can fork the single-issue invariant**
`.github/workflows/cc-billing-classifier-canary.yml:382` and `:428`
```
existing=$(gh issue list --label cc-canary-blocked --state open --json number --jq '.[0].number')
```
The whole escalation design leans on "the open issue's age is the streak counter" (line 401-402 comment), which requires there to be at most one open `cc-canary-blocked` issue at a time. The workflow has no `concurrency:` group and is triggered both by `schedule` and by `workflow_dispatch`, so a manual dispatch overlapping the daily 05:50 run (5-minute `timeout-minutes`) can have two job instances both read `existing=""` before either creates an issue, each then calling `gh issue create`, producing two open `cc-canary-blocked` issues. From then on, `.[0].number` picks an arbitrary one of the two for both the age check (line 403-404) and the close-on-success step (line 428-432), so the other lingers open indefinitely — silently breaking the escalation clock (it can now never reach `BLOCKED_ESCALATE_HOURS` on the abandoned issue, or conversely could already be past it and get ignored). This mirrors a pre-existing idiom the file already uses for the billing alert, so it isn't a new pattern, just newly relied on for a correctness-sensitive counter.
```suggestion
concurrency:
  group: cc-billing-classifier-canary
  cancel-in-progress: false
```

## What's good
- The core diagnosis (curl -sf collapsing "never started" and "started but OAuth-red" into one failure) is correct and the fix (track `last_code`, branch on the `DARIO_NO_TOKEN_REFRESH` marker in `proxy.log`) is the right shape.
- Gating the probe step off `steps.startup.outputs.blocked` and gating the alert step off `steps.probe.outputs.status != ''` correctly keeps a blocked run from ever reaching the billing-verdict/alert logic — I traced both `if:` conditions and they're mutually exclusive per run, so the new tracking step and the existing alert step can't race against each other within a single execution.
- Closing the tracking issue unconditionally on any run that reaches a verdict (line 429, "regardless of which verdict this run produced") is the right call — it only asserts "a probe got through," not "the probe passed."

Scope: reviewed the full diff (one file, 124/-15) plus surrounding context pulled from the workflow (permissions block, cron trigger, full step list) via a shallow clone at head `93ca34d`. Did not attempt to execute the workflow.

sprayberry-redline (approved):

**Verdict:** Approved — the live-head update resolves the prior blocking issue, and I found no new blocking issues.

I re-reviewed the complete one-file workflow diff and specifically verified the change since my previous verdict: `BLOCKED_ESCALATE_HOURS` is now `24`, so a second consecutive daily blocked run reaches the escalation threshold as documented. I also checked the startup/probe blocked-output flow, issue open/update/close behavior, and failure gating. CI is fully green, including actionlint, CodeQL, build, test, and live-test.

## What's good

The workflow now distinguishes an unhealthy HTTP response from a proxy that never answered, preserves the read-only credential invariant, avoids manufacturing a billing verdict when no probe ran, and still turns a sustained blind canary red.

sprayberry-secondread (commented):

**Verdict:** No blocking correctness bugs found; one documentation/config discrepancy worth a maintainer look, otherwise this reads well and the design rationale in the comments is unusually thorough.

## Findings

**Minor — `BLOCKED_ESCALATE_HOURS` value contradicts the PR's own stated rationale**
`.github/workflows/cc-billing-classifier-canary.yml:94`
```
+  BLOCKED_ESCALATE_HOURS: '24'
```
The PR description says: *"the run only fails once it has stayed open for `BLOCKED_ESCALATE_HOURS` (36h — one missed daily slot plus margin)."* The shipped value is `24`, and the inline comment right above it (lines 90-93) actually documents 24h intent: *"a second consecutive daily miss finds the issue already 24h old and goes red then, not a day later."*

Concretely: cron fires once daily at 05:50. If day N is blocked, the tracking issue is created at ~05:50+ε (whatever the build/ready-loop took). If day N+1 is *also* blocked, the escalate check runs at ~05:50+ε again — `age_h = (now - opened_at)/3600` lands right at 24, so `age_h -ge 24` is true and the run goes **red on the second occurrence**, not after "one missed slot plus margin" as the PR body claims. There's no slack for the ready-loop/build-time jitter between the two runs, so a run that happens to start a few seconds earlier than the prior day's could tip the floor-divided `age_h` to 23 vs 24 and skip escalation for one more day — the boundary is exact, not buffered.

This isn't a crash or a broken code path — the code does what its own comment says — but the PR body's "36h — one missed daily slot plus margin" doesn't match what actually ships, and if 36h (or any margin) was the intended behavior, `BLOCKED_ESCALATE_HOURS` needs to be that value instead of `24`.

```suggestion
Confirm which is intended and reconcile: if a margin against jitter is wanted, set
BLOCKED_ESCALATE_HOURS to something like '36' to match the PR description; if 24h
(escalate on the second consecutive daily miss) is actually intended, update the PR
description to stop calling it "one missed daily slot plus margin".
```

## What's good
- The startup-step fix (checking HTTP status via `-w '%{http_code}'` instead of `curl -sf`) correctly separates "proxy never answered" from "proxy answered 503," and I traced the borrow-mode grep (`DARIO_NO_TOKEN_REFRESH` in `proxy.log`) through both the startup and probe steps — both gate correctly on `steps.startup.outputs.blocked == ''`/`steps.probe.outputs.blocked` and neither can fall through to a false billing verdict.
- The blocked→tracked→escalate→close issue lifecycle is coherent: new issue on first block (exits 0 before any age check, so it can't self-escalate), comment+age-check on repeat blocks, and the verdict step (`cc-billing-classifier-canary.yml:428-433`) unconditionally closes any open `cc-canary-blocked` issue the moment a real probe gets through, regardless of pass/fail/warn — so the escalation clock can't outlive its cause.
- `if: steps.startup.outputs.blocked == ''` on the probe step correctly relies on GitHub Actions' implicit `success()` ANDing, so a genuine (non-borrow) startup failure still skips the probe rather than needing an explicit failure check.

Scope: reviewed the full diff (one file, 124/-15) plus surrounding context in the cloned workflow file (credential pinning, permissions, job-level `if`, and the unchanged alert-open/close step) to confirm ordering and gating between the new and existing steps. Did not attempt to execute the workflow.

PR #1212, opened 2026-09-05
codex-backend: abort upstream when the client disconnects mid-stream

askalf:
Source ticket: DEV-ffd6370e (retry of DEV-f0e85a83).

## Problem

`forwardToCodex` in `src/codex-backend.ts` registered no close listener. When a client went away mid-stream, dario kept draining the upstream Responses stream and kept calling `res.write` on a closed socket — the ChatGPT subscription paid for output nobody received. The Claude path in `src/proxy.ts` has handled this for a long time (`clientDisconnected` / `decideOnClientClose`); the Codex path had no equivalent.

## Change (one function)

- `res.on("close", ...)` right after the `AbortController`: when the response closes and we did not end it ourselves (`finished` flag, set just before every `res.end`), set `clientGone`, log under `--verbose`, and `abort.abort()` — which is what stops the upstream generating.
- All streamed writes now go through a small `write()` helper that is a no-op once `clientGone` — no writes to a dead socket.
- The read loop breaks on `clientGone` as well, so an upstream stream that ignores its abort signal cannot keep us spinning.
- After the drain loop, `clientGone` returns early with `report(499, usageSoFar(), ...)` — nginx-style "client closed request". The tokens already spent are still reported, so analytics does not lose them.
- In the `catch`, the `clientGone` check comes **before** the `deferOnUnavailable && !res.headersSent` decline branch. This is the subtle part: an abort that fires before the first byte leaves `headersSent` false, which would otherwise look exactly like an unreachable backend and trigger a pointless failover that re-bills a request nobody is reading.
- `usageSoFar()` factors out the existing usage extraction so both the normal and the abandoned exit report it identically. The `finally` removes the close listener.
- `void req;` left as-is — `req` is still unused; the listener is on `res`, which is where the write-side close surfaces.

No version bump. CHANGELOG bullet in a follow-up commit on this branch.

## Test plan

- Follow-up commit on this branch adds to `test/codex-backend.mjs`: a fetch whose `ReadableStream` never closes; after the first delta the fake `res` close handler fires; asserts the fetch signal is aborted, no further `res.write` happened, and `onDone` reported status 499 exactly once.
- Existing assertions (normal completed stream, both shapes) must stay green — the `fakeRes` gains an `on(event, fn)` shim.
- CI is the authoritative build/test signal.

Do not merge — dario merges are operator-gated.


askalf:
CI status at hand-off: **12 checks pass** — build (18/20/22), test, analyze (CodeQL), docker-cap-drop-smoke, actionlint, advise, label, triage, validate-package-json. `compat` and `live-test` were still running; `auto-merge` skipped as expected (this is operator-gated).

The first push failed `test` for a reason worth recording: `forwardToCodex` now calls `res.on("close", ...)`, and three test files carry their own minimal `ServerResponse` fakes that had no `on`. `test/codex-backend.mjs` got the full shim (`on`/`removeListener`/`closeClient`, so a test can play the disconnect); `test/codex-thrown-report.mjs` and `test/provider-cooldown.mjs` got a no-op `on`/`removeListener` pair, which is all they need. The three files that drive a real `http.Server` (`codex-route-empty-pool`, `codex-pool-fallback-mid-flight`, `pool-exhaustion-gate-wiring`) use real ServerResponse objects and needed nothing.

Second thing the CI round surfaced: the never-ending fake stream had to honour the abort signal (`c.error(...)` on `signal.abort`) the way a real fetch body does. Without it the drain loop had nothing to unblock it — which is exactly the production symptom this PR fixes, reproduced in the harness.

Verified by CI, not locally: the test suite imports from `dist/`, and per the build policy this branch was never compiled locally.

Not verified, left for review: behaviour against a *real* upstream that ignores its abort signal. The `if (clientGone) break;` in the read loop is the belt to the abort's braces there, but only the injected-fetch path is exercised. The next probe, if anyone wants it, is a `--drain-on-close`-style comparison against the Claude path in `src/proxy.ts` to decide whether the Codex path should eventually take the same flag rather than always aborting.

Source ticket: DEV-ffd6370e.

sprayberry-redline (changes requested):

**Verdict: 1 correctness issue to address; the main streaming disconnect path and its tests otherwise read well.**

## Findings

### Blocking — client disconnect can still trigger provider failover after a non-2xx response (`src/codex-backend.ts:760-765`, `src/codex-backend.ts:960-962`)

> `let clientGone = false;`
> `const onClientClose = (): void => {`
> `  if (finished || clientGone) return;`
> `  clientGone = true;`
> `  if (verbose) console.log(\`[dario] codex client disconnected (${creds.alias}) — aborting upstream\`);`
> `  if (!abort.signal.aborted) abort.abort();`
> `};`

> `if (clientGone) {`
> `  report(499, usageSoFar(), clientWantsStream, model);`
> `  return true;`
> `}`

The new `clientGone` guard is only reached in the `catch` path (and after the successful stream drain). The earlier `!upstream.ok` path still awaits `upstream.text()` and then immediately runs the `deferOnUnavailable && unavailable` branch without checking `clientGone`. Concrete failure: Codex returns a buffered 429/503, the client closes while/just after that error body is read, and `response.text()` resolves despite the abort because the body is already available. `clientGone` is true, but this function returns `false`; the caller then fails over and starts another billed request for a client that has already gone away—the exact behavior this change is intended to prevent.

Add the same terminal client-close handling inside the non-OK path, after consuming the response body and before deciding whether to defer:

```suggestion
      const detail = await upstream.text().catch(() => '');
      if (clientGone) {
        report(499, usageSoFar(), clientWantsStream, model);
        return true;
      }
```

Please also cover this race with a non-2xx fake response whose body resolution closes the client, asserting status 499 and no decline callback.

## What's good

The normal in-flight streaming path cleanly aborts upstream work, suppresses dead-socket writes, preserves usage reporting, and distinguishes client cancellation from transport failure. The new early- and mid-stream disconnect tests are focused, and the full CI rollup is green.

sprayberry-secondread (commented):

**Verdict:** No correctness bugs found in the changed lines; two test-coverage gaps worth closing before this ships.

## Findings

### 1. Test coverage — client-disconnect only tested for the streaming shape (info)
`test/codex-backend.mjs`, the new `header('a client that hangs up aborts the upstream...')` block: both the positive test and the negative ("early close") test build their request body as

```
model: 'gpt-5.6-sol', stream: true, messages: [{ role: 'user', content: 'hi' }],
```

and both invoke `forwardToCodex(..., 'openai', ...)`. There is no case with `stream: false`. The non-streaming path takes a materially different route through `src/codex-backend.ts`: `res.writeHead` for the SSE headers is skipped entirely (`if (clientWantsStream) { res.writeHead(...) }`, src/codex-backend.ts:843), so a disconnect during the drain loop is only exercised here with headers already sent. The `if (clientGone) break;` (src/codex-backend.ts:863) → `if (clientGone) { report(499, ...) }` (src/codex-backend.ts:894) path for a non-streaming client that hangs up before headers are ever written is untested.

```suggestion
Add a third case in the same header block: same neverEnding upstream, but
`stream: false` in the request body, asserting report(499,...) fires and
`early.statusCode`/`res.headersSent` stay unset — mirroring the streaming
assertions but for the writeHead-skipped path.
```

### 2. Test coverage — client-disconnect untested for the Anthropic shape (info)
Same two tests pass `'openai'` as the `shape` argument (`forwardToCodex({}, res, body, CREDS, '*', {}, 5000, false, 'openai', neverEnding, false, ...)` and the `closeOnFetch` variant below it). The Anthropic branch writes through a different call chain — `emitAnthropic` → `write(formatResponsesAnthropicSSE(out))` (src/codex-backend.ts:838) instead of the OpenAI translator's `write(out)` (src/codex-backend.ts:876) — and reads usage from `terminalResponse.usage` rather than `translator!.usage()` (src/codex-backend.ts:823-825). None of that path is exercised for a mid-stream disconnect.

```suggestion
Duplicate the positive test with shape: 'anthropic' and an Anthropic-shaped
neverEnding stream, asserting the same three things: signal aborted, no
further antTranslator-formatted writes, and a single 499 report.
```

## What's good
- The close-listener ordering is correct: `res.on('close', onClientClose)` is registered before the `await fetchImpl(...)` (src/codex-backend.ts:767, 778), so a disconnect during the initial request is caught, not missed.
- The `clientGone` check in the `catch` block is placed before the `deferOnUnavailable && !res.headersSent` decline branch (src/codex-backend.ts:960 vs. 973), exactly as the PR description claims — verified this ordering against the actual diff, and the accompanying "early close" test (`test/codex-backend.mjs`) directly exercises it (`declines.length === 0`).
- Every streamed write in the changed code goes through the `write()` helper (`if (!clientGone) res.write(chunk)`, src/codex-backend.ts:770) — confirmed no bare `res.write(` call remains in the function.
- `usageSoFar()` is a clean factor-out of the pre-existing usage-extraction logic (previously inlined at the single 200-report site) with no change in semantics for the non-disconnect path.

## Scope
Read the full diff (342 lines across 5 files) and the surrounding function in `src/codex-backend.ts` (lines 696-996) from a shallow clone at the PR head. No part of the diff was skipped.

sprayberry-redline (approved):

**Verdict: No blocking issues found; the previously reported non-2xx disconnect race is fixed.**

I re-reviewed the live head `ce79abbab0fd5f4849ac7920116abee38980d002`, including the complete delta from the prior reviewed commit and the surrounding disconnect/error handling. The new `clientGone` check now runs after the buffered non-2xx body is consumed and before the provider-decline branch, so a client closing during a 429/5xx body read is reported once as 499 and cannot trigger a second billed provider call. The added regression test directly exercises that race and also preserves the attached-client 429 failover behavior.

## What's good

The close listener is removed in `finally`, every normal response path marks `finished` before ending the response, and the focused disconnect tests complement the existing early- and mid-stream cases. The live-head CI rollup is fully green (builds, tests, compatibility, live test, and CodeQL).

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — the disconnect/abort state machine is sound and well covered; one minor test-coverage gap noted below.

**What I checked:** the full diff of `src/codex-backend.ts` (the `clientGone`/`finished` close-listener logic, the `write()` wrapper, the `usageSoFar()` factoring, and the three exit points that now check `clientGone`), `CHANGELOG.md`, and all three `fakeRes` test doubles touched by this PR (`test/codex-backend.mjs`, `test/codex-thrown-report.mjs`, `test/provider-cooldown.mjs`) to confirm every caller of `forwardToCodex` in the test suite got the new `on`/`removeListener` shim it now depends on. I traced the ordering of the `clientGone` check ahead of the `deferOnUnavailable` decline branch in both the `!upstream.ok` path and the `catch` block, and confirmed `report()`'s `reported` guard makes it safe against any theoretical double-fire between the `clientGone` early-return and the normal completion path. I did not run the build/test suite — the sandboxed `npm`/`tsc` here hit an `EACCES` on `/app/.npm` unrelated to this PR — so the below is from static tracing, not an executed test run.

**Findings**

- **info** — `test/codex-backend.mjs`: the new "a client that hangs up aborts the upstream" block (around `header('a client that hangs up aborts...')`) only drives the `shape: 'openai'` path through `neverEnding`/`closeOnFetch`/`closeOnErrorBody`. The Anthropic branch (`emitAnthropic`, `antTranslator!.end()`) shares the same `write()`/`clientGone` guards, so the risk is low, but a disconnect mid-Anthropic-stream (e.g. between `sseParser!.push` calls) isn't exercised by this PR's new tests, only inferred by shared code. Not blocking — the logic is identical to the tested OpenAI path — but worth a follow-up test if the Anthropic Codex path sees real traffic soon.

**What's good**
- The ordering fix in the `catch` block — checking `clientGone` *before* the `deferOnUnavailable && !res.headersSent` decline branch — is exactly right and the PR explains why clearly: an abort that fires pre-first-byte would otherwise look identical to an unreachable backend and trigger a pointless double-billed failover. I verified this by tracing the "early" test case (`closeOnFetch`): `res.on('close')` is registered before the `fetchImpl` call, so a synchronous `closeClient()` inside the fake fetch sets `clientGone` before the `AbortError` is thrown and caught.
- `finished`/`clientGone` are both single-direction latches with a shared `if (finished || clientGone) return` guard in `onClientClose`, so a `close` event racing a normal completion can't double-abort or double-report; `report()`'s own `reported` flag is a second independent guard against a double `onDone` call.
- All three `fakeRes` test doubles that call `forwardToCodex` were updated with `on`/`removeListener` (even the two that don't test the disconnect path itself, `codex-thrown-report.mjs` and `provider-cooldown.mjs`, got no-op stubs) — so this PR doesn't quietly break unrelated test files that now call `res.on(...)` inside `forwardToCodex` regardless of path taken.

PR #1214, opened 2026-09-05
ci: surface the Codex model snapshot in the job summary and document baseline seeding

askalf:
Follow-up to #1208 (ticket DEV-8dfba913).

## Context

The second-read finding asked for two things: expose the generated `live-models.json`, and document the bootstrap path in the workflow header. The first half **already landed in #1208 itself** — commit `ci: publish the generated Codex model snapshot as an artifact` added the `Upload live model snapshot` step before merge, so the finding was partly stale by the time the ticket dispatched. This PR closes what is genuinely still missing.

## Changes

`.github/workflows/codex-drift-watch.yml` only:

1. **Workflow header** now documents the first-run baseline seeding flow explicitly: dispatch the workflow, read the list, copy it verbatim to `test/fixtures/codex-models.snapshot.json`, open a PR. Explains *why* the first run always reports drift (the fixture ships as `{"models": []}`).
2. **New `Publish live model snapshot to the job summary` step** writes the JSON to `$GITHUB_STEP_SUMMARY` in a fenced block and echoes it to the job log. Reviewing a baseline no longer requires downloading and unzipping an artifact — the copy-paste source is on the run page. Same `always() && hashFiles(...)` guard as the upload step, so it is a no-op when the fetch failed or the run skipped.
3. **Drift issue body** now points at the job summary and names the `codex-live-models-<run_id>` artifact, so someone arriving via the issue knows where the list is.

The artifact upload is unchanged and still the durable copy.

## Test plan

- Workflow-only change; no source or test files touched.
- The workflow is inert until `vars.DARIO_CODEX_LIVE_HOME` is set, so this cannot be exercised by CI on this PR. Actionlint/workflow validation in CI covers the syntax.
- Behavioural verification comes on the next operator-dispatched run: the job summary should render the model list, and the artifact should contain identical bytes.

Source ticket: DEV-8dfba913

### DIFF
--- a/.github/workflows/codex-drift-watch.yml
+++ b/.github/workflows/codex-drift-watch.yml
@@ -73,17 +73,17 @@ jobs:
         if: always() && hashFiles('live-models.json') != ''
         run: |
           {
-          cat <<'HEADER'
-### Account-visible Codex models
+            cat <<'HEADER'
+          ### Account-visible Codex models
 
-Copy verbatim to `test/fixtures/codex-models.snapshot.json` to seed or update the baseline.
+          Copy verbatim to `test/fixtures/codex-models.snapshot.json` to seed or update the baseline.
 
-```json
-HEADER
-          cat live-models.json
-          cat <<'FOOTER'
-```
-FOOTER
+          ```json
+          HEADER
+            cat live-models.json
+            cat <<'FOOTER'
+          ```
+          FOOTER
           } >> "$GITHUB_STEP_SUMMARY"
           echo '--- live-models.json ---'
           cat live-models.json
--- a/.github/workflows/codex-drift-watch.yml
+++ b/.github/workflows/codex-drift-watch.yml
@@ -3,6 +3,15 @@ name: Codex drift watch
 # Inert until the operator sets DARIO_CODEX_LIVE_HOME in this self-hosted
 # runner's repository variables. It must name a HOME containing
 # .dario/codex-accounts; no credential is committed or copied.
+#
+# First-run baseline seeding. test/fixtures/codex-models.snapshot.json ships
+# empty, so the first enabled run reports drift by construction. To seed it:
+#   1. Run this workflow (Actions -> Codex drift watch -> Run workflow).
+#   2. Read the generated list in the run's job summary, or download the
+#      codex-live-models-<run_id> artifact for the same bytes.
+#   3. Copy it verbatim to test/fixtures/codex-models.snapshot.json and open a
+#      PR. Later runs compare against it and only report a real change.
+# The same review-then-copy path applies to every subsequent drift report.
 on:
   schedule:
     - cron: '29 5 * * *'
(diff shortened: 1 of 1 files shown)


askalf:
Note for review — the artifact half of the source finding was already stale.

`.github/workflows/codex-drift-watch.yml` gained the `Upload live model snapshot` step inside #1208 itself (commit `ci: publish the generated Codex model snapshot as an artifact`), before that PR merged at 12:54:46Z. The second-read finding was written against an earlier revision of the branch. This PR therefore does not re-add the artifact upload — it adds only what was genuinely missing: the job-summary rendering and the documented seeding path.

Verified locally beyond CI: I extracted the `run:` block applying the same 10-space block-scalar dedent Actions applies, executed it against a sample `live-models.json`, and confirmed the summary renders correct markdown with backticks and fences literal and the JSON intact.

Remaining uncertainty, deliberately not chased in-run: the workflow is inert until `vars.DARIO_CODEX_LIVE_HOME` is set, so the new step cannot execute in CI on this PR. Next probe is the first operator-dispatched run — the job summary should show the model list and the `codex-live-models-<run_id>` artifact should contain byte-identical content.

sprayberry-redline (approved):

I checked the complete 29-line workflow diff, including the shell heredocs used to append fenced JSON to `$GITHUB_STEP_SUMMARY`, the `always() && hashFiles(...)` guard, the artifact-name reference in the drift issue, and the documented first-run baseline flow. The heredoc delimiters remain valid after YAML block indentation is removed, and the summary and log expose only the already-generated model snapshot.

**What's good:** The summary uses the same guarded `live-models.json` as the durable artifact, so failed/skipped fetches do not publish stale or nonexistent output. The issue text also gives operators both discoverability and a durable fallback. CI is green, including actionlint, builds, tests, live-test, and CodeQL.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — here's what I checked.

## What I checked

This PR touches a single file, `.github/workflows/codex-drift-watch.yml` (+28/-1), adding:
1. A header comment documenting the first-run baseline-seeding flow.
2. A new `Publish live model snapshot to the job summary` step.
3. A one-line wording change to the drift-issue body pointing readers at the job summary/artifact.

Because the new step's correctness hinges on YAML block-scalar indentation stripping interacting with bash heredocs — a spot where it's easy to get the indentation wrong and silently break the heredoc terminator — I extracted the `run:` block by hand, applied YAML's leading-whitespace stripping rule (strip is set by the first non-blank line, here 10 spaces), and executed the resulting script against a sample `live-models.json`:

```
.github/workflows/codex-drift-watch.yml:74-89
        run: |
          {
            cat <<'HEADER'
          ### Account-visible Codex models
          ...
          ```json
          HEADER
            cat live-models.json
            cat <<'FOOTER'
          ```
          FOOTER
          } >> "$GITHUB_STEP_SUMMARY"
```

After stripping, the `HEADER`/`FOOTER` delimiter lines land at column 0 with no trailing characters, which is what an unquoted `<<'DELIM'` heredoc requires to close — I confirmed this by running the stripped script locally and got the expected fenced-JSON block in the output file, with the `cat live-models.json` output correctly injected between the header and footer. No truncation, no unterminated heredoc, no stray `$GITHUB_STEP_SUMMARY` write outside the block.

The `if: always() && hashFiles('live-models.json') != ''` guard on the new step (`.github/workflows/codex-drift-watch.yml:73`) matches the existing guard on the `Upload live model snapshot` step below it, so it degrades the same way (no-op) when the fetch step failed or never ran.

The issue-body wording change (`.github/workflows/codex-drift-watch.yml:~106`) is a plain string edit with no logic change, and the header-comment addition is documentation only.

## What's good

- The author verified the interaction between YAML indentation and the heredoc rather than leaving it to hope — worth calling out since that's exactly the kind of change that looks right and silently isn't.
- Keeping the artifact upload as "the durable copy" while adding the job-summary path for quick review is a sensible UX improvement with no behavioral risk to the drift-detection logic itself (untouched).
- Workflow-only change with an honest test plan (no source/tests touched, can't be exercised by CI, verification deferred to the next real run) — appropriately scoped for what this PR actually is.

No coverage, efficiency, or reuse concerns — this is a small, self-contained addition to an existing, already-reviewed workflow.

PR #1217, opened 2026-09-05
ci: a PR that changes src/ must carry a CHANGELOG entry

askalf:
## What

v6.0.22 shipped six codex changes; its release notes listed two. Four PRs touched `src/` and never added a `## [Unreleased]` bullet. The convention was documented in the CHANGELOG header and the PR template, and it was skipped anyway — advice is not a gate.

This makes it a gate:

- `scripts/check-changelog.mjs` — fails when the PR diff touches `src/` without touching `CHANGELOG.md`. Skipped with the `no-changelog` label (created) for pure refactors and test-only shuffles.
- Runs in the required `validate-package-json` job, `pull_request` only (a master push has no base to diff).
- `CLAUDE.md` states the rule where coding agents read it; the PR template checklist gets a line.

## Proof

Run against the merged commits it would have caught:

```
$ BASE_SHA=<#1204 base> HEAD_SHA=baa226f node scripts/check-changelog.mjs
FAIL: 1 shipping file(s) changed with no CHANGELOG.md entry:
  src/codex-backend.ts
$ BASE_SHA=<#1206 base> HEAD_SHA=<#1206 merge> node scripts/check-changelog.mjs
FAIL: 4 shipping file(s) changed with no CHANGELOG.md entry:
  src/cli.ts src/codex-accounts.ts src/codex-oauth.ts src/proxy.ts
$ PR_LABELS=no-changelog ... -> skipped by request (exit 0)
```

No version bump; nothing publishes.


sprayberry-redline (changes requested):

**Verdict: 1 correctness issue to address; the workflow wiring and documentation otherwise read well.**

## Findings

### Blocking — `scripts/check-changelog.mjs:38` does not enforce that an Unreleased bullet was added

```js
const hasChangelog = files.includes('CHANGELOG.md');
```

The advertised rule is that a PR changing `src/` must add a bullet under `## [Unreleased]`, but this condition only checks whether `CHANGELOG.md` appears anywhere in the file list. A PR can change `src/proxy.ts` and also edit an old release entry, fix a typo in the changelog header, or even delete text from `CHANGELOG.md`; `hasChangelog` is then true and the required check passes without any Unreleased bullet. That leaves the release-note omission this gate is intended to prevent reachable.

Please inspect the actual `CHANGELOG.md` diff and require an added bullet within the `## [Unreleased]` section (while retaining the explicit label escape hatch), rather than treating any modification as sufficient. For example, factor the semantic validation into a helper and gate on its result:

```suggestion
const hasUnreleasedBullet = changelogDiffAddsBulletUnderUnreleased(base, head);
```

A focused test/fixture should cover a `src/` change paired with an unrelated edit to an older changelog section and assert that it fails.

## What's good

The check is placed in an already-required lightweight job, uses argument-safe `execFileSync`, has an explicit label escape hatch, and CI is currently green across the repository checks.
(_Note: green CI exercises this PR's passing path, not the bypass scenario above._)

sprayberry-secondread (commented):

**Verdict:** Solid, narrowly-scoped CI gate; one medium finding — the documented `no-changelog` label bypass likely won't fire without an extra commit, due to the PR trigger's default event types.

## Findings

### Medium — `no-changelog` label escape hatch is unreachable via "re-run", contradicting the documented flow
`.github/workflows/ci.yml:3-7`:
```yaml
on:
  push:
    branches: [master]
  pull_request:
    branches: [master]
```
`pull_request` with no `types:` key defaults to `opened, synchronize, reopened` only — it does **not** include `labeled`. The new step reads the label set from the triggering event's payload:

`.github/workflows/ci.yml:44`:
```yaml
          PR_LABELS: ${{ join(github.event.pull_request.labels.*.name, ',') }}
```

Concrete scenario: a contributor opens a pure-refactor PR touching `src/`, the new job fails (correctly, no CHANGELOG entry). They add the `no-changelog` label to invoke the documented escape hatch (`scripts/check-changelog.mjs:29-31`, and advertised in the PR template addition and PR body). Adding a label does not fire a new `pull_request` event, so there's no fresh run. The only way to get a green check is either "Re-run failed jobs" — which replays the *original* webhook payload, so `PR_LABELS` is still empty and the job fails again — or pushing an empty/no-op commit to trigger a real `synchronize` event. The escape hatch this PR advertises as the fix for "pure refactors and test-only shuffles" therefore doesn't work through the most natural remediation path (label + re-run), only through label + new commit.

This fails closed (blocks merge, doesn't let bad PRs slip through), so it's not a safety hole, but it will produce confused contributors and wasted CI cycles exactly in the "trivial refactor" case this PR calls out as the intended escape valve.

```suggestion
on:
  pull_request:
    types: [opened, synchronize, reopened, labeled, unlabeled]
    branches: [master]
```

### Minor
- `scripts/check-changelog.mjs` has no dedicated test, but this matches the existing pattern for its sibling scripts in the same job (`check-package-json.mjs`, `check-readme-line-count.mjs` are also untested) — not a regression this PR introduces, just noting for completeness.

## What's good
- The check itself (`scripts/check-changelog.mjs`) is small, dependency-free, and directly runnable by hand per its own header comment — good for local debugging of a CI failure.
- `git diff --name-only` between two arbitrary SHAs works fine against a shallow (`--depth=1`) fetch since it's a tree comparison, not a history walk — no issue there.
- Landing the CHANGELOG bullet, `CLAUDE.md` rule, and PR template checkbox in the same PR that adds the gate is the right way to roll this out.

**Scope:** Reviewed the full diff (5 files, 72 additions) plus surrounding context in `.github/workflows/ci.yml` and `scripts/` from a shallow clone at `048b6e6`. No parts of the diff were skipped.

sprayberry-redline (changes requested):

1 correctness issue remains in the revised gate; the original old-entry bypass is otherwise addressed well.

## Findings

### Blocking — `scripts/check-changelog.mjs:74`

> `const writable = UNRELEASED.test(h) || !base.has(h);`

This treats **every** newly introduced level-2 heading as a release heading. A PR that changes `src/` can therefore add, for example, `## Notes` plus `- updated docs` to `CHANGELOG.md`; because `## Notes` is absent from the base, its bullet is returned by `newReleaseNoteBullets()` and the required check passes even though no bullet was added under Unreleased or a version release section. That contradicts the documented gate and leaves another straightforward bypass.

Restrict the second branch to the repository's actual release-heading syntax, and add a regression test proving an arbitrary new heading does not count. For example:

```suggestion
const RELEASE = /^## \[\d+\.\d+\.\d+\](?:\s+-\s+\d{4}-\d{2}-\d{2})?$/;
// ...
const writable = UNRELEASED.test(h) || (RELEASE.test(h) && !base.has(h));
```

## What's good

The update now compares parsed base/head content rather than accepting any edit to `CHANGELOG.md`, excludes bullets merely moved during a version bump, and adds focused end-to-end coverage. The PR's CI rollup is green across the required checks.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — this is a well-scoped, well-tested CI gate.

## What I checked

- Full diff (336 lines across 6 files: workflow, two new scripts, CHANGELOG/CLAUDE.md/PR-template docs).
- Ran `node test/check-changelog.mjs` from a shallow clone of the PR head (`b7b15d9`) — all 18 checks pass.
- Manually reproduced the workflow's `git fetch --no-tags --depth=1 origin "$BASE_SHA" "$HEAD_SHA"` step against two disjoint shallow fetches into a fresh repo and confirmed `git diff --name-only base head` resolves correctly with no common-ancestor requirement (`scripts/check-changelog.mjs:19` fetch step in `.github/workflows/ci.yml`).
- Traced `sectionsOf`/`newReleaseNoteBullets` (`scripts/check-changelog.mjs:114-153`) against the "old entry edited/deleted", "bullet moved to old release", and "version-bump carries a bullet forward" cases the tests exercise — the exclusion logic (`if (baseUnreleased.has(b)) continue;` at `scripts/check-changelog.mjs:147`) correctly prevents an edited/moved old bullet from satisfying the gate.
- Confirmed `test/check-changelog.mjs` is not in `all.test.mjs`'s `EXCLUDED` set, so it's wired into `npm test` automatically rather than needing a separate invocation.
- Confirmed the three doc touchpoints (CHANGELOG.md, CLAUDE.md, PR template) are mutually consistent and the PR itself follows its own new rule (adds an `## [Unreleased]` bullet describing the change).

## What's good

- The bypass the PR explicitly designed against (editing/moving an old entry without adding a new one) is both implemented correctly and has a dedicated unit test and end-to-end test (`test/check-changelog.mjs:253-254`, `:319-322`), which is exactly the kind of self-referential defense-in-depth a CI gate like this needs.
- The escape hatch (`no-changelog` label) and the push-vs-pull_request guard (`if (!base)`) are both tested for their exit codes, not just their presence.
- Comments explain the "why" (the 4-of-6-PRs incident, why the check lives in the no-install job) rather than restating the code.

No correctness, reuse, or coverage issues found in the diff. I did not find anything worth raising even as a minor note.

sprayberry-redline (changes requested):

2 correctness issues remain in the changelog gate; the previous arbitrary-heading bypass is fixed, but the gate can still both reject unrelated PRs and accept a modified old bullet.

## Findings

### High — `.github/workflows/ci.yml:43-47`, `scripts/check-changelog.mjs:106-107` — compare against the tested merge result, not the stale PR head

> `BASE_SHA: ${{ github.event.pull_request.base.sha }}`
> `HEAD_SHA: ${{ github.event.pull_request.head.sha }}`
> `git fetch --no-tags --depth=1 origin "$BASE_SHA" "$HEAD_SHA"`

> `const files = git(['diff', '--name-only', base, head]).split('\n').map((s) => s.trim()).filter(Boolean);`
> `const shipping = files.filter((f) => f.startsWith('src/'));`

This is a two-dot comparison between the **current base tip** and the PR's potentially stale head, not the PR diff. Concrete failure: branch A changes only `README.md`; while it is open, branch B merges a change to `src/proxy.ts` plus its changelog bullet. On A's next CI run, `base` contains B while A's `head` does not, so this diff reports `src/proxy.ts` (in reverse), and A is incorrectly failed for having no new changelog bullet. The check becomes dependent on whether the branch has rebased rather than whether the PR touches `src/`.

Use the pull-request merge commit that Actions checks out, so the comparison is current base → current base with this PR applied (or otherwise compute/fetch the merge base and use a three-dot diff). For example:

```suggestion
          BASE_SHA: ${{ github.event.pull_request.base.sha }}
          HEAD_SHA: ${{ github.sha }} # pull_request merge commit checked out by actions/checkout
```

Then fetch only the base if needed and compare that base to the checked-out merge SHA. Please add a regression case where base has advanced with an unrelated `src/` commit after the PR branch point.

### Medium — `scripts/check-changelog.mjs:81-86` — editing an existing Unreleased bullet is treated as adding a new one

> `for (const b of bullets) {`
> `  // A bump PR moves Unreleased bullets under the new heading; those are`
> `  // not new notes for THIS PR, so they are excluded too.`
> `  if (baseUnreleased.has(b)) continue;`
> `  if (base.get(h)?.has(b)) continue;`
> `  added.push(b);`
> `}`

The comparison is by exact bullet text. Concrete failure: base already has `- **Feature A.** Added.`, and a PR changes `src/proxy.ts` while only changing that bullet to `- **Feature A.** Added!`. The new string is absent from both base sets, so `added` contains it and CI passes, even though this PR added no release-note bullet for its source change. This is the same class of bypass the stricter implementation is intended to close, just under `Unreleased` rather than an old release.

Track an increase in the number of eligible bullets (while separately excluding bullets merely moved by a version bump), rather than treating every changed string as a newly added bullet. A suitable shape is:

```suggestion
// For an existing Unreleased section, require the head to contain more bullet
// items than base; an edit/reword keeps the count unchanged and does not pass.
// For a newly cut release section, count only bullets not already present in
// base Unreleased so moved release notes remain excluded.
```

Add a unit and end-to-end scenario that modifies only the text of an existing Unreleased bullet and expects exit 1.

## What's good

The new release-heading allowlist closes the prior `## Notes` bypass, and the expanded focused tests cover old-release edits, moved version-bump notes, labels, and the direct-execution guard. The PR's current CI rollup is green.

sprayberry-redline (approved):

**Verdict: approved — the live-head update addresses both previously blocking correctness issues, and I found no new blockers.**

I re-reviewed the current head, focusing on the changes since the prior `CHANGES_REQUESTED` verdict as well as their interaction with the full changelog gate. The workflow now judges the pull-request merge result against the current base (with a merge-base fallback), avoiding unrelated changes from a moved base. The release-note check now requires eligible bullet-count growth, so merely rewording an existing Unreleased bullet no longer passes. Focused regressions cover both scenarios.

## What's good

The implementation keeps the explicit `no-changelog` escape hatch, restricts alternate destinations to actual version headings, preserves version-bump behavior, and includes unit plus real-git end-to-end coverage. The full PR CI rollup is green, including the test and `validate-package-json` jobs.

PR #1222, opened 2026-09-06
docs: cordon integration (PII redaction in front of dario)

askalf:
Adds `docs/integrations/cordon.md`: how to put [cordon](https://github.com/askalf/cordon) in front of dario so PII, card numbers and secrets are tokenized before a shared subscription's proxy, the provider, or provider logs ever see them, and restored in the reply.

- compose pair (cordon builds from its repo; it publishes no image yet)
- client env: same dario key, base URL moves to cordon
- worked example with the redaction headers
- per-request mode/set controls and the `phi` caveat for coding prompts

Also links the new page from the Own Your Stack list in the README. Docs only, no code change.


sprayberry-redline (approved):

Classified low-risk: docs-only (2 file(s)).

PR #1226, opened 2026-09-06
ci(compat): skip superseded runs, pace 5s -> 1s

askalf:
The compat job itself is ~2 min; the cost is the queue. Measured 2026-09-06 (runs 34037994376, 34037642867, 34016564041): "Run compat tests" 84/87/69 s, of which 50 s is `wait(PACE_MS)` × 10, and one PR ran the suite four times in 30 minutes because superseded queued runs still execute in full.

- **Skip superseded runs.** New first step compares the run's commit with the PR's live head; if a newer push exists it exits green and every later step is skipped (no proxy, no upstream calls, no PR comment). The global concurrency group and `cancel-in-progress: false` are unchanged, so cross-PR serialization still holds. `workflow_dispatch` always proceeds. The finalizer exits 0 for a superseded run instead of failing on the missing exit code.
- **Pace 5 s → 1 s.** The upstream is the per-token API key (no rate window); the script's own API-key default is 500 ms. ~10 s of sleep instead of ~50 s.
- **Health poll** in 250 ms steps, same 30 s budget.

Expected: a normal run ~70 s instead of ~2 min; a superseded run ~10 s and zero upstream calls. No change to `test/compat.mjs`.


sprayberry-redline (changes requested):

**Verdict: 1 correctness issue to address; the superseded-run gating otherwise reads well.**

## Findings

### Blocking — health polling no longer has the stated 30-second bound (`.github/workflows/compat-test-self-hosted.yml:339-345`)

> `for i in $(seq 1 120); do`
>
> `if curl -sf -o /dev/null --max-time 2 http://127.0.0.1:3457/health; then`
>
> `sleep 0.25`

Each iteration can spend up to 2 seconds inside `curl` and then another 250 ms sleeping. If the local proxy accepts connections but stalls before returning an HTTP response, this changed loop can occupy roughly 270 seconds rather than the documented 30-second budget. That materially undercuts the runner-time optimization and delays the actionable startup failure by several minutes. Bound the loop by elapsed wall time (and use a sub-second per-attempt timeout), rather than assuming only the sleeps consume time.

```suggestion
          deadline=$((SECONDS + 30))
          i=0
          while (( SECONDS < deadline )); do
            i=$((i + 1))
            if curl -sf -o /dev/null --max-time 0.25 http://127.0.0.1:3457/health; then
              ready=true
              echo "proxy ready after approximately $((i * 500))ms"
              break
            fi
            sleep 0.25
          done
```

## What's good

The live-head lookup fails open on API errors, `workflow_dispatch` remains enabled, every costly downstream step is gated consistently, and the finalizer explicitly turns only confirmed superseded runs green. I also checked the PR's CI rollup; all executed checks are green.

sprayberry-secondread (commented):

**Verdict:** No blocking correctness issues found; one minor cosmetic note on the health-poll log message.

## What I checked
Single-file diff (`.github/workflows/compat-test-self-hosted.yml`, +53/-7). I traced the new "Skip if this commit is no longer the PR head" step end-to-end, all the downstream `if: steps.fresh.outputs.superseded != 'true'` gates it controls, the `workflow_dispatch` fallback path (where `steps.fresh` never runs), the pace-ms change (5000→1000), and the health-poll loop rewrite (30×1s → 120×0.25s). I did not re-review the unchanged surrounding steps (proxy startup env, maxTested gate, PR-comment marker logic) since the diff doesn't touch their logic.

## Findings

**Minor (info) — cosmetic log inaccuracy**
`.github/workflows/compat-test-self-hosted.yml` (health-poll loop):
```
for i in $(seq 1 120); do
  if curl -sf -o /dev/null --max-time 2 http://127.0.0.1:3457/health; then
    ready=true
    echo "proxy ready after $((i * 250))ms"
    break
  fi
  sleep 0.25
done
```
Since the `sleep` happens *after* the check, a proxy that's already up when the loop starts reports "proxy ready after 250ms" on the very first successful curl (`i=1`), when actual elapsed time is close to 0ms. This mirrors the pre-existing `${i}s` message's same off-by-one (not introduced by this PR), just made more visible at finer granularity. Doesn't affect the 30s timeout budget or pass/fail behavior — log text only.

## Verified as correct (no bug found)
- **`workflow_dispatch` fallback:** the new `fresh` step only runs `if: github.event_name == 'pull_request'`, so for `workflow_dispatch` its output is unset. Every downstream gate reads `steps.fresh.outputs.superseded != 'true'`, and an unset/empty string is `!= 'true'`, so all steps correctly proceed — matches the PR's "workflow_dispatch always proceeds" claim.
- **Finalizer on superseded runs:** `Set job status` reads `SUPERSEDED: ${{ steps.fresh.outputs.superseded }}` and does `${SUPERSEDED:-false}`; for `workflow_dispatch` this env var is set to an empty string (not unset), and bash's `:-` does trigger on empty, so it correctly evaluates to `false` and falls through to the normal `exit "${COMPAT_EXIT_CODE:-1}"` path rather than short-circuiting.
- **Fail-open on lookup failure:** `live="$(gh api ... --jq .head.sha 2>/dev/null || true)"` — if the `gh api` call fails, `live` is empty, `[ -n "$live" ]` is false, and the step falls into the `else` branch (`superseded=false`, proceeds), matching the intent stated in the comment ("or lookup failed: proceeding") rather than silently skipping a run that should have executed.
- **Pace/timeout math:** `DARIO_COMPAT_PACE_MS: '1000'` × 10 waits ≈ 10s matches the PR description's "~10s sleep instead of ~50s" claim. The health loop's `120 × 0.25s = 30s` preserves the documented 30s budget referenced in the `::error::` message below it.
- **Comment-skip gating:** `if: github.event_name == 'pull_request' && always() && steps.fresh.outputs.superseded != 'true'` correctly suppresses the PR comment for superseded runs while still firing on failure for live-head runs (the `always()` is preserved).

No test-coverage gap to flag — this is a CI workflow definition, not application logic under the repo's test suite.

sprayberry-redline (approved):

**Verdict: No blocking issues found; the prior health-polling defect is fixed.**

I re-reviewed the live head `9709d882640f0ba7d05c601bec262f656aece9fb`, including the complete one-file diff and the commits added since my prior change request. The health readiness loop is now bounded by elapsed wall time and uses a 250 ms per-attempt timeout, so a stalled `/health` response can no longer stretch the intended 30-second startup budget into several minutes.

I also checked the superseded-run path: the live-head lookup fails open on lookup errors, `workflow_dispatch` continues normally, expensive downstream steps and PR commenting are skipped only for a confirmed stale PR head, and the finalizer returns green for that explicit case while continuing to fail closed for other missing compat results.

**What's good:** The comments now accurately explain the timing model and the gating behavior. The PR's executed CI checks, including actionlint and compat, are green.

sprayberry-secondread (commented):

Note: this is a re-review at the current head (`9709d88`) — the PR gained a second commit ("bound health poll by wall clock, not iteration count") since I last reviewed the first commit (`f5c82cb9`) on this PR.

**Verdict:** No blocking issues found — both commits are internally consistent and the second one fixes a real latent budget-overrun in the first.

**What I checked:** the full diff of `.github/workflows/compat-test-self-hosted.yml` across both commits (`f5c82cb9` skip-superseded + pace change, `1c0b4b17` wall-clock health poll), plus the merge-from-master commit's file list (confirmed it touches only unrelated files — README/CHANGELOG/docs/src — not this workflow, so the diff I reviewed is the PR's actual net change to this file).

**Correctness check — the two commits interact correctly:**
- Commit 1 (`f5c82cb9`) rewrote the health-poll loop as `for i in $(seq 1 120); do curl --max-time 2 ...; sleep 0.25; done`. Worst case (proxy never answers) that's up to 120 × (2s curl + 0.25s sleep) ≈ 270s against a step whose own comment still says "30s" — a real mismatch between the stated budget and the actual bound.
- Commit 2 (`1c0b4b17`) replaces the iteration count with an explicit wall-clock deadline (`deadline=$(( SECONDS + 30 )); while [ "$SECONDS" -lt "$deadline" ]; do ...`) and drops `--max-time` to `0.25`, which correctly bounds the loop to ~30s regardless of how slowly the proxy responds. I traced the final state (both commits applied) and the loop now terminates within ~30s in the worst case (stalled proxy) — the mismatch from commit 1 no longer exists in the merged result.

**Skip-superseded gating (`steps.fresh.outputs.superseded`):** traced the fail-open/fail-closed paths —
- `workflow_dispatch`: `id: fresh` step is skipped entirely (`if: github.event_name == 'pull_request'`), so `steps.fresh.outputs.superseded` is empty; every downstream `if: steps.fresh.outputs.superseded != 'true'` evaluates true (empty ≠ 'true'), so the run proceeds. Matches the intent stated in the comment.
- `gh api .../pulls/${PR}` failure (rate limit, transient network): `2>/dev/null || true` makes `live` empty, `[ -n "$live" ]` is false, falls to the `else` branch and proceeds normally — fails open rather than silently dropping a real run.
- "Set job status" finalizer: `SUPERSEDED` env resolves to `""` on workflow_dispatch, and `${SUPERSEDED:-false}` correctly defaults to `false` in that case (bash triggers the default on both unset *and* empty), so the compat-exit-code logic still runs unmodified for dispatch runs.
- When superseded, `proxy.pid`/`proxy.log` are never created (Start step is gated the same way), and "Stop dario proxy" (unconditional `if: always()`) guards on `[ -f proxy.pid ]` and `2>/dev/null || true` on the tail — no error surfaces for the skipped path.

I could not find a case where the gating produces a false pass/fail or a stuck job.

**What's good:** the wall-clock fix is a genuine root-cause correction rather than a band-aid, and it's explained in the added comment with the specific failure mode it prevents. The supersede-skip step fails open on lookup errors rather than silently dropping legitimate runs, which is the right default for a CI gate.

PR #1227, opened 2026-09-06
Record the OAuth grant and age the refresh token everywhere the pool is inspected (v6.0.26)

askalf:
## What

Anthropic expires the refresh-token family about **28 days after the original OAuth grant**, rotation or not. A seat that refreshed every 8h for four weeks still died with `invalid_grant "Refresh token expired"` 28d 10h after its grant (2026-09-05), and every request on it failed over silently. The token is opaque and the token endpoint carries no refresh expiry, so the calendar is the only signal — and dario never kept it.

This PR makes dario keep it:

- `grantedAt` written by every grant path (auto/manual `login`, `accounts add`, admin `/login/complete`, login back-fill), preserved across refresh, the login↔credentials mirror, and pool reconcile.
- `src/refresh-grant.ts`: pure ageing (`ok` / `warn` ≥21d / `urgent` ≥26d / `unknown`, wall at 28d), thresholds env-tunable and order-clamped.
- Surfaces: `dario accounts list` (grant line per seat), `dario doctor` (new `Refresh grant` row: warn/fail/info), `GET /accounts` and `GET /admin/accounts` (per-seat fields), `/health` for trusted callers (`refreshGrant` summary, never on the public surface).
- Proxy warns on stderr + OS notification when a seat crosses warn/urgent, once per level, daily while it persists.
- Docs: `docs/multi-account-pool.md` new section, `docs/recovery.md` cause note.

## Trade-offs and unknowns

- **One data point for the 28-day wall.** The defaults are conservative around it (warn at 21, urgent at 26). If a different lifetime is observed, the env knobs move the thresholds without a release.
- **Existing seats show `unknown`** until re-granted. This is honest: their grant date was never recorded. `worstGrantLevel` ranks unknown between ok and warn so an all-unknown pool is an info row, not a page.
- **Keychain imports never get a grant date** — Claude Code does not record it.
- The `/health` field is additive and internal-only; public `/health` is unchanged.

## Tests

- New `test/refresh-grant.mjs` (34 checks): thresholds/env/clamping, level boundaries at 20.99/21/26/past-wall, unknown inputs, worst-level ranking, describe strings, and the doctor row (ok/info/warn/fail, both seats listed, fix text only when needed).
- `test/health-response.mjs`: `refreshGrant` passes through internally and never leaks publicly.
- Full suite: 170/170.

Release: version bumped to 6.0.26 in this PR per RELEASING.md.


sprayberry-redline (approved):

I reviewed the complete 16-file, 346-addition diff at live head `ed55101560ded1dc85ecd5666f6945fe3d9c771a`. I checked the grant timestamp at each OAuth entry point, preservation through both refresh paths and the login/credentials/pool mirrors, threshold ordering and boundary behavior, aggregation across seats, trusted-versus-public health serialization, operator warning cadence, and the new doctor/API/CLI surfaces. The timestamp remains tied to the original grant rather than token rotation, and unknown legacy/keychain credentials are handled consistently.

**What's good:** The ageing policy is centralized in a small pure module, the public health response continues to omit internal account metadata, and the focused boundary/unknown/doctor tests cover the key behavioral contracts. The PR's CI rollup is green across tests, builds, compatibility, CodeQL, and integration checks.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — this reads well and the grant-age plumbing is threaded through every code path I traced.

**What I checked:** the full diff (346/-4, 16 files) — `src/refresh-grant.ts` (new, the pure ageing/threshold/level logic), every write site of `grantedAt` (`src/oauth.ts`, `src/accounts.ts`, `src/pool.ts`, `src/proxy.ts`), the two new surfaces (`GET /accounts`, `GET /admin/accounts`, `dario accounts list`, `dario doctor`), the `/health` internal-only gating in `src/health-response.ts`, and both new/changed test files. I also cloned the branch to read call sites not shown in the diff context (`pool.updateTokens`, `doRefreshAccountToken`, `refreshAccountToken`'s retry loop) to confirm `grantedAt` survives every refresh path, not just the ones the diff hunks show directly.

**Correctness — the thing I specifically tried to break and couldn't:**
- `pool.updateTokens()` (`src/pool.ts:672`) does *not* take a `grantedAt` param and is called from both the startup self-heal and background-refresh loops in `proxy.ts` (lines ~1745, ~1790) without one. At first glance this looks like the grant date silently drops on every background refresh. It doesn't: `updateTokens` only mutates the in-memory `PoolAccount`, whose `grantedAt` field is untouched by the omission (it's simply not overwritten), and the *durable* copy is preserved by `doRefreshAccountToken` (`src/accounts.ts:306-311`, `...creds` spread keeps `grantedAt`) and `oauth.ts`'s `doRefreshTokens` (line 982, `grantedAt: oauth.grantedAt`). So the in-memory pool's `grantedAt` stays whatever it was seeded with at `pool.add()`/reconcile time, and disk stays correct — confirmed this isn't a bug, not flagging it.
- `worstGrantLevel` ranking `unknown` between `ok` and `warn` (`src/refresh-grant.ts:88-93`) is exercised by both the pure unit tests and `checkRefreshGrant`'s "one unstamped seat → info" case in `test/refresh-grant.mjs` — matches the documented intent (an all-unknown pool shouldn't page).
- Threshold clamping (`grantThresholds`, `src/refresh-grant.ts:59-64`) correctly order-clamps warn ≤ urgent ≤ lifetime even under a misconfigured env (e.g. warn=40/urgent=35 both collapse to lifetime=28), preventing a seat from skipping a level — covered by the "inv" test case.
- `/health` internal-only gating: `refreshGrant` is spread only inside the `includeInternal` branch (`src/health-response.ts:287`) and the two new `test/health-response.mjs` checks explicitly assert it never leaks on the public surface — correct and tested.

**Minor (non-blocking):**
- The `describeGrantAge` switch (`src/refresh-grant.ts:99`) has no `default`/`case 'unknown'` arm inside the switch itself (unknown is handled by the early return above), which is fine functionally but relies on TypeScript's exhaustiveness check over the narrowed `'ok'|'warn'|'urgent'` union to catch a future level addition — worth being aware of if a new level is ever added without updating this function, but not something to change now.

**What's good:** the module doc in `src/refresh-grant.ts` states the real production incident (28d10h, single data point) and why the defaults are conservative around it; the env-tunability is clamped defensively; and the test suite exercises boundary days (20.99/21/26/past-wall) plus clock-skew (`grantedAt` in the future → `ok`, not a crash) — exactly the edge cases I'd have gone looking for.

PR #1228, opened 2026-09-06
doctor: stop claiming a differing seat identity will 401

askalf:
## What

`dario doctor`'s Identity row warned that any pool seat whose `deviceId`/`accountUuid` differ from this machine's `~/.claude.json` "will 401 on non-Haiku models". Observed 2026-09-06: a seat added headless with a minted identity served Sonnet 5, Opus 5 and Haiku with 200, on a plan with Extra Usage disabled. The only 401 on record came from an identity belonging to a *different account* than the seat's bearer (a half-copied transplant).

- Identity row → `info`. Text says a differing identity is expected for seats added elsewhere or headless, scopes the 401 to the wrong-account case, keeps remove-then-add as the fix for that symptom.
- No-`~/.claude.json` row stops asserting an Extra Usage billing outcome dario cannot observe; pool seats carry their own snapshot identity regardless.
- Tests updated to assert the new claims and that the old ones are gone.

## Trade-offs and unknowns

- Whether a minted identity is billed differently from a real one is **not observable from dario**. The row no longer claims either way. What is observed: the request is served, from the plan allocation (Extra Usage was disabled on that account).
- Not bumping the version here: #1227 carries 6.0.26. I will bump this to the next patch once that lands, so the two do not race the auto-release.

PR #1224, opened 2026-09-06
fix(health): --no-claude-auth proxy is not degraded when the Claude pool is empty

askalf:
## What

`/health` returned 503 for a proxy started with `--no-claude-auth` and a loaded Codex account, because the Claude pool was `none` and only API-key mode was exempt from that structural verdict. When a Codex account is present under `--no-claude-auth` the pool is empty by design and the Codex backend serves, so the empty pool is now not-evidence in that state too. The flag alone is not evidence (review): `--no-claude-auth` with no account still reports 503. A failed serving probe (`?probe=1`) still degrades.

## Why now

Arming `codex-drift-watch.yml` today (repo variable `DARIO_CODEX_LIVE_HOME=/root` on the `dario-drift` runner) exposed it: run [34037311202](https://github.com/askalf/dario/actions/runs/34037311202) started the proxy fine (`Codex accounts: fleet`) and then failed the 30s `/health` readiness poll, filing #1223 with no model list. The workflow now also prints `proxy.log` and the `/health` body when readiness fails, so the next failure of this kind is diagnosable from the run log.

## Changes

- `src/health-response.ts`: `codexServes` on `HealthStatusLike`; excluded from `structurallyDead` alongside `upstreamApiKeyMode`.
- `src/proxy.ts`: `/health` passes `codexServes: opts.noClaudeAuth === true && await hasAnyCodexAccount()` — the codex router's own presence check (#1138), so `dario codex add`/`remove` are reflected without a restart; the Claude path never stats the directory.
- `test/health-response.mjs`: 5 unit checks mirroring the api-key-mode block (none/broken/expired → 200, failed probe → 503, explicit false = absent).
- `test/health-codex-serves.mjs` (new): in-process `--no-claude-auth` proxy, no network: 503 with no account, 200 after an account is stored mid-run, 503 again after removal.
- `.github/workflows/codex-drift-watch.yml`: dump `proxy.log` + `/health` on readiness failure.
- `test/fixtures/codex-models.snapshot.json`: seeded with the account-visible list (7 slugs), generated on the runner host with the workflow's own fetch/sort/serialize code so the byte-exact `cmp` matches. The first post-merge run should pass and auto-close #1223.
- `CHANGELOG.md`: Unreleased bullet.

`npm test` 170/170, preflight clean, changelog gate ok, README line-count ok.

PR #1229, opened 2026-09-06
Seat pin: route one request to one seat with no failover, and `dario accounts check`

askalf:
## What

A read-only, in-place seat probe. Until now "does this seat serve Sonnet right now?" had no honest answer from a running pool: a normal request picks its seat by headroom and fails over on 401/429 to a peer, then to the Codex leg. The workaround was copying the seat's tokens into a throwaway dario — credential sprawl a proxy should make unnecessary.

- **`x-dario-account: <alias>`** (with `x-dario-admin-token`) pins one request to that pool account: no headroom selection, no sticky rebinding, no 401/429 peer retry (every other seat counts as already tried), no Codex fallback. The upstream status comes back as-is.
- **Refused, never ignored**: 403 when the admin API is off or the token is missing/wrong, 404 for an unknown alias, 400 for a malformed one. A probe that silently became a normal request would call the wrong seat healthy.
- **`dario accounts check <alias> [--models=…]`** sends one tiny pinned request per model through the running proxy and prints the verdict. Nothing is copied, nothing restarts.
- Docs: `docs/admin-api.md` (new section), `docs/commands.md` row.

## Trade-offs and unknowns

- Gated on the admin API on purpose: choosing the seat bypasses headroom routing and can burn one seat's window, so it is an admin act. With `DARIO_ADMIN` off the feature does not exist.
- A pinned request skips the mid-flight beta-reduction/recovery paths only insofar as they select a peer; upstream recoveries that stay on the same seat are unchanged.
- Not bumping the version here: #1227 carries 6.0.26 and #1228 follows it. I will bump this to the next patch once those land so nothing races the auto-release.

## Tests

- `test/seat-pin.mjs` (17): none/disabled/unauthorized/invalid/pinned, array headers, constant-time compare, API key is not the admin token, auth checked before alias.
- `test/seat-pin-proxy.mjs` (14, in-process proxy with an injected upstream): pinned good seat → 200 with exactly one upstream call on its bearer; pinned dead seat → its 401 with one call, no `x-dario-pool-fallback`; unpinned on the same pool still fails over to the good seat; 403/404/400 paths make no upstream call.
- Full suite: 172/172.


sprayberry-redline (changes requested):

1 correctness issue to address; the authorization checks, alias validation, no-peer-failover handling, documentation, and targeted tests otherwise read well.

## Findings

### Blocking — `src/proxy.ts:2957-2969` — a pin is silently ignored in upstream API-key mode

The new pinned-seat assignment is placed after the existing `upstreamApiKey` early return:

> `if (pinnedAccount) {`
> `  poolAccount = pinnedAccount;`
> `  accessToken = pinnedAccount.accessToken;`
> `  return true;`
> `}`

With `ANTHROPIC_UPSTREAM_API_KEY` configured and a valid pool alias present, an authenticated request carrying `x-dario-account: good` passes pin validation and even logs `seat pin → good`, but `selectPoolAccount()` returns from the preceding `if (upstreamApiKey)` branch before these added lines execute. The outbound request therefore uses the global `x-api-key`, not the selected seat's bearer token. That violates the feature's core guarantee that a pin is refused rather than silently served elsewhere and can report a seat healthy when only the API key is healthy.

Reject a seat pin when API-key mode makes pool selection unavailable, or give an explicit pin precedence over API-key mode. For example:

```suggestion
if (pinnedAccount) {
  poolAccount = pinnedAccount;
  accessToken = pinnedAccount.accessToken;
  return true;
}
if (upstreamApiKey) {
  return true;
}
```

Please also cover the chosen behavior with a proxy-level test using `upstreamApiKey` plus a valid pin.

## What's good

The implementation authenticates before revealing alias validity, uses constant-time token comparison, blocks peer/Codex fallback on the exercised OAuth path, and includes focused unit and in-process proxy coverage. CI was green except for the still-pending `live-test` check when reviewed.

askalf:
Pushed `06458ad` to this branch: version bump 6.0.25 -> **6.0.26**, so the seat-pin change actually cuts a release instead of merging green and shipping nothing.

All three slots `scripts/preflight.mjs` checks, in one commit:
- `package.json` `.version`
- `package-lock.json` `.version` and `.packages[""].version`
- `CHANGELOG.md` — `## [6.0.26] - 2026-09-06` inserted directly under `## [Unreleased]`, above the existing seat-pin and `/health` entries, so both ship under this release

Stacked on `2be27fd` (the API-key-mode 409 commit), not the stale head the sweep flagged.

Verified: the `version-bump-advice` comment flipped to "Release gate looks satisfied — version bumped 6.0.25 -> 6.0.26"; `advise`, `validate-package-json`, `test`, and the `build (18/20/22)` matrix all pass on `06458ad`. The self-hosted `compat` / `live-test` / `wire-drift` legs were still running when I finished — a version-only commit should not move them, but they are worth a glance before merge.

Note the `/health` entry from #1224 was already sitting under `## [Unreleased]`, so it now ships under 6.0.26 too. If you would rather it release on its own, move the heading below it before merging.

Source ticket: 00MTPZIYSZBDDB59C22605BF70

sprayberry-redline (dismissed):

Approved — the prior blocking API-key-mode issue is fixed, and I found no new blocking issues in the live-head diff.

## What I checked

- Verified pinned requests are now explicitly refused with HTTP 409 before alias lookup whenever `ANTHROPIC_UPSTREAM_API_KEY` bypasses the account pool, so a probe cannot be silently served by the global key.
- Verified the new proxy-level regression test asserts both halves of that contract: pinned requests make no upstream call, while unpinned requests still use `x-api-key` normally.
- Rechecked the seat-pin authorization, alias validation, pinned account selection, sticky-routing suppression, peer/Codex failover suppression, CLI behavior, and documentation across the complete 10-file PR diff.
- CI's build matrix, unit tests, CodeQL, container smoke test, and live test are green; `wire-drift` was still pending when reviewed.

## What's good

The fix fails closed at the correct boundary and tests the concrete configuration that previously violated the feature's central “never silently use another credential” guarantee.

sprayberry-secondread (commented):

Verdict: 1 correctness issue to address; the account-pinning and no-failover handling otherwise read well.

## Findings

### Medium — `src/proxy.ts:3001`: a valid seat-pin header is ignored for Codex/OpenAI-routed requests

> `const pinnedAccount = seatPin.kind === 'pinned' ? (pool.get(seatPin.alias) ?? null) : null;`

The new header resolves only an account from the Claude `pool`, but provider dispatch still occurs later and independently routes a Codex-named model to `forwardToCodex` (or an OpenAI model to its backend) before `selectPoolAccount()` is reached. Consequently, with a valid admin token and `x-dario-account: dead`, a `POST /v1/chat/completions` request naming a configured Codex model can return a successful Codex response. That silently reports the *Codex* leg healthy rather than the named Claude seat, contradicting the advertised guarantee that a pin is never served by another leg.

Either reject a pin for requests that cannot be dispatched through the Claude pool, or make the provider-routing block fall through to the pinned Claude path. For example, reject before provider dispatch:

```suggestion
if (pinnedAccount && decision.provider !== 'claude') {
  res.writeHead(409, JSON_HEADERS);
  res.end(JSON.stringify({
    type: 'error',
    error: {
      type: 'invalid_request_error',
      message: `${SEAT_PIN_HEADER} can only be used with a request served by the Claude account pool`,
    },
  }));
  return;
}
```

Please also add an in-process test with a configured Codex or OpenAI route plus a valid pin, asserting that it cannot be served by that other provider.

## What's good

The pin parser fails closed for missing/invalid admin authorization and invalid aliases, and the changed dispatch paths explicitly suppress sticky rebinding, Codex fallback, and 429 fallback for a real pinned Claude request. The proxy test usefully verifies the pinned 401 is returned without a peer retry.

Scope note: reviewed the current head `5345f05e` (the ticket's requested `fe7149f` is no longer the PR head), concentrating on the new seat-pin parser, proxy routing/failover integration, CLI probe, and their tests. I could not execute the tests because the shallow review checkout lacks installed build tooling (`tsc`).

sprayberry-redline (approved):

No blocking issues found; the previous cross-provider pinning gaps are fixed on the current head.

I reviewed the complete 10-file, 483-addition/8-deletion diff, with particular attention to the authorization/alias-validation order, exact-seat selection, sticky routing suppression, 401/403/429 peer and Codex failover suppression, upstream API-key mode, non-Claude provider routing, CLI request construction, and the new unit/in-process proxy coverage. The live-head CI rollup is green across build/test, compatibility, live test, CodeQL, actionlint, and package validation.

**What's good:** The pin is refused rather than silently ignored whenever it cannot be honored, token comparison is constant-time after a length check, and the regression tests now directly prove both previously problematic cases: API-key mode cannot serve a pinned probe, and a Codex-routed model cannot escape to the wrong provider. Normal unpinned routing remains covered as a control.

PR #1231, opened 2026-09-06
fix(proxy): reject a non-JSON-object body with 400 in the endpoint's wire shape

askalf:
## What

A request body that is not a JSON object is now answered locally with 400 in the endpoint's own wire shape, before any routing:

- `/v1/chat/completions` → `{"error":{"message":…,"type":"invalid_request_error","param":null,"code":null}}`
- `/v1/messages` → `{"type":"error","error":{"type":"invalid_request_error","message":…}}`

Empty bodies and non-object JSON (`[]`, `"hi"`) get the same 400. Nothing goes upstream. Logged as `reject: 'invalid-body'`.

## Why

Every routing decision (alias, provider prefix, codex slug, template build) peeks at `.model` and swallows its own parse error, so `{` fell through to the Claude pool as if it were a Claude request. On a `--no-claude-auth` proxy that came back as the pool's 503 `{"error":"No account configured",…}` — `error` a string, so an OpenAI SDK sees no message. With a Claude login the same bytes went all the way to Anthropic to be rejected (reproduced on the box: the 400 carried an Anthropic `request_id`).

`codex-drift-watch.yml` caught it on its first armed run after #1224, [34044346444](https://github.com/askalf/dario/actions/runs/34044346444): non-stream, stream, and tool-call checks passed live, then `error response must expose .error.message` failed and #1230 was filed. The model snapshot matched.

## Changes

- `src/proxy.ts`: guard right after the body is read, mirroring the overage-halt block (count, log line, wire-shaped body).
- `test/invalid-json-body.mjs` (new): in-process `--no-claude-auth` proxy with a stored codex account and a throwing `fetchImpl`; 16 checks across `{`, empty, array, string on both endpoints, plus "fetchImpl never called".
- `CHANGELOG.md`: Unreleased bullet.

Not in scope (noted for a follow-up): dario's own OpenAI-shape 503s (`writePoolUnavailable`, `writeCodexCredentialsUnavailable`) also put a string in `error`; the contract does not exercise them, and changing those shapes touches existing tests.

`npm test` green, preflight clean, changelog gate ok, README line-count ok.


sprayberry-redline (changes requested):

1 correctness issue to address; the endpoint-specific error envelopes and hermetic upstream-call assertion otherwise read well.

## Findings

### Blocking — malformed UTF-8 can still bypass the new local rejection (`src/proxy.ts:3058`)

> `const v = JSON.parse(body.toString()) as unknown;`

`Buffer.toString()` silently replaces malformed UTF-8 with U+FFFD. For example, a body whose bytes are `{"x":"` + `0xff` + `"}` is not valid UTF-8 JSON, but this conversion produces the parseable object `{ "x": "�" }`. The new guard therefore accepts it and the original malformed bytes remain eligible for forwarding, violating the stated “invalid body gets 400 locally; nothing goes upstream” contract. On a no-Claude-auth deployment this can still fall through to the unrelated pool response that motivated the fix.

Decode UTF-8 in fatal mode before parsing, and cover a malformed-byte object in the regression test:

```suggestion
const text = new TextDecoder('utf-8', { fatal: true }).decode(body);
const v = JSON.parse(text) as unknown;
```

## What's good

The response shapes match each endpoint, invalid top-level JSON types are rejected explicitly, and the test’s throwing `fetchImpl` provides a strong no-forwarding assertion for the covered inputs. The main CI test/build jobs are green; CodeQL and the live test were still pending when I reviewed.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — this is a well-scoped fix, correctly placed, with solid test coverage.

## What I checked
- `src/proxy.ts`: the new guard block (added right after the body is read, before `resolveProxyTarget`/alias/provider-prefix/codex-slug/template peeks)
- Confirmed by reading the surrounding routing code (`src/proxy.ts:3095-3115`, `:3236-3250`, `:3480-3525`) that every one of those later `.model` peeks is wrapped in its own `try/catch` that silently swallows a parse failure and falls through — exactly the failure mode the PR description names, and the new guard now runs *before* all of them, so a malformed body can no longer reach any of those fallthroughs.
- `test/invalid-json-body.mjs`: exercises both endpoints (`/v1/chat/completions`, `/v1/messages`) against `{`, empty body, `[]`, `"hi"`, asserts 400 + correct wire shape, and asserts `fetchImpl` is never called (proves nothing goes upstream). Hermetic (temp `HOME`, no live accounts).
- `CHANGELOG.md`: entry matches the change.

## Findings
None that clear the bar for a real defect. Two small observations, neither blocking:

**info** — `src/proxy.ts:3058-3059`: the object-shape check `v === null || typeof v !== 'object' || Array.isArray(v)` rejects `null`/arrays/primitives correctly, but note it also accepts any *non-empty* object regardless of whether it has a usable `.model` — e.g. `{"foo":1}` still passes this guard and only fails later at the routing peeks (which return their existing errors, e.g. eventually reaching `writePoolUnavailable`/404-shape paths). That's consistent with the PR's stated scope ("cannot be routed" = not a JSON object), not a bug — just worth knowing the guard is deliberately shape-only, not schema validation.

**info** — the new parse at `src/proxy.ts:3058` is a third `JSON.parse` of the same bytes before `parsedBody` is set (the alias/prefix parse at `:3104` and the codex peek at `:3238` follow). The PR body's own "Not in scope" note doesn't mention this, but it's negligible at the body sizes this proxy handles (10MB cap) and consistent with the existing pattern of parsing multiple times in this function (the `#642-audit` comment already tracks reuse for the *other* two parses) — not worth blocking on.

## What's good
- Guard placement is the right fix for the actual root cause (every downstream peek swallowing its own parse error) rather than patching one call site.
- Wire-shape parity with existing error paths in the file (`writePoolUnavailable`, `writeAllProvidersRateLimited`, `writeCodexCredentialsUnavailable` all follow the same `isOpenAI ? {...} : {type:'error',...}` split) — this PR's new response shapes fit that convention exactly.
- Test is hermetic and asserts the negative case (`upstreamCalls === 0`) explicitly, not just the response body — that's the assertion that actually proves the "nothing goes upstream" claim in the PR description.
- Logging (`reject: 'invalid-body'`) follows the existing `writeLogLine` convention used by the other reject paths in this file.

askalf:
Addressed in 314c680: the body is now decoded with `new TextDecoder('utf-8', { fatal: true })` before `JSON.parse`, so malformed UTF-8 is rejected on the bytes rather than on a lossy string. Added the `{"x":"\xff"}` case to `test/invalid-json-body.mjs` (400, `.error.message`, nothing upstream).

sprayberry-secondread (commented):

Verdict: No blocking issues found in the current head; the invalid-body rejection and wire-shape handling read correctly.

I checked the new guard's placement after the full body read, empty/malformed/non-object handling, strict UTF-8 decoding, endpoint-specific error envelopes, early return before routing, and the hermetic no-upstream test coverage. In particular, the changed path validates the bytes with `new TextDecoder('utf-8', { fatal: true }).decode(body)`, rejects null/arrays as well as malformed JSON, and returns before any model-routing code can consume the request.

What's good: the test's throwing `fetchImpl` plus `upstreamCalls === 0` assertion directly protects the important regression: invalid payloads must not fall through to a backend. The added malformed-UTF-8 case also closes the otherwise easy-to-miss gap between lenient string decoding and the original wire bytes.

Scope note: the ticket named head `e457eb5`, while the PR currently points to `41956cfabb1e30d1180e793c550e1552dd0a750c`; this review covers that current head.

sprayberry-redline (dismissed):

**Verdict: no blocking issues found; the prior malformed-UTF-8 issue is fixed.**

I re-reviewed the live two-commit, three-file diff. The fatal `TextDecoder('utf-8', { fatal: true })` check now rejects malformed wire bytes before routing, while the existing object check rejects empty, syntactically invalid, scalar, null, and array bodies. The endpoint-specific 400 envelopes match the OpenAI and Anthropic shapes, and the regression test verifies malformed UTF-8 is rejected without an upstream call.

**What's good:** Validation happens once immediately after the bounded body read, before every routing path that parses or rewrites the body. The full CI rollup is green, including tests, Node 18/20/22 builds, compatibility, live-test, wire-drift, and CodeQL.

sprayberry-redline (approved):

**Verdict: Approve — no blocking issues found on live head `7cff3ae25e913241e81efb8e2bc318662cfb35bb`.**

I reviewed the complete five-file diff and the surrounding request-body read/routing path. The earlier malformed-UTF-8 concern is resolved: `src/proxy.ts:3099-3105` now decodes the original bytes with a fatal UTF-8 decoder before parsing, so invalid bytes cannot be silently replaced and forwarded. The guard correctly rejects empty, malformed, null, scalar, and array bodies, emits the endpoint-appropriate OpenAI or Anthropic 400 envelope, and returns before routing/upstream dispatch.

What's good: `test/invalid-json-body.mjs` covers both wire shapes, all principal invalid-body classes (including malformed UTF-8), and asserts that no upstream fetch occurs. The test is automatically included by the default test runner, and all required CI checks are green.

sprayberry-secondread (commented):

**Verdict:** No correctness bugs found; one real efficiency regression (re-introduces a double-parse the codebase had already fixed once) and a test-coverage gap worth a follow-up.

## Findings

**Medium — reintroduces the double-JSON.parse the code explicitly fixed before**
`src/proxy.ts:3103-3104`:
```ts
const text = new TextDecoder('utf-8', { fatal: true }).decode(body);
const v = JSON.parse(text) as unknown;
```
then a few lines later, `src/proxy.ts:3150`:
```ts
const parsed = JSON.parse(body.toString()) as Record<string, unknown>;
```
The comment right above the second parse (`src/proxy.ts:3144-3146`) says: *"Parsed body, shared between the provider-prefix detection below and the template-build block further down so the same bytes are not JSON.parsed twice per request (#642-audit)."* The new guard block parses the same bytes a third time (decode + parse) before that shared `parsedBody` is ever built, so every request now pays for two full decode+parse passes over the body instead of one — on the exact code path that comment says was deliberately made to avoid this. With `MAX_BODY_BYTES` at 10MB, this is a real, reachable, unconditional per-request cost, not a hypothetical one.

Suggested fix:
```ts
// in the guard block, keep the successfully-parsed value and hand it down
let invalid: string | null = null;
let parsedForRouting: Record<string, unknown> | null = null;
if (body.length === 0) invalid = 'request body is empty';
else {
  try {
    const text = new TextDecoder('utf-8', { fatal: true }).decode(body);
    const v = JSON.parse(text) as unknown;
    if (v === null || typeof v !== 'object' || Array.isArray(v)) invalid = 'request body must be a JSON object';
    else parsedForRouting = v as Record<string, unknown>;
  } catch (err) {
    invalid = `request body is not valid JSON: ${err instanceof Error ? err.message : String(err)}`;
  }
}
// ...later, seed `parsedBody`/`parsed` from `parsedForRouting` instead of re-parsing `body.toString()`
```

**Low — new guard is exercised only on `/v1/chat/completions` and `/v1/messages`, not the other two routes it also gates**
`resolveProxyTarget` (`src/proxy.ts:599-607`) allows POSTs through to `/v1/messages/count_tokens` and `/v1/complete` too, and both pass through the same body-read → guard block before any provider routing (the guard sits above the routing logic, gated only by `isOpenAI` for the wire shape, not by path). `test/invalid-json-body.mjs` only posts to `/v1/chat/completions` and `/v1/messages` (`test/invalid-json-body.mjs:60-77`). A regression that broke the guard specifically for `count_tokens` or `/v1/complete` (e.g. an early-return added above it keyed on `isCountTokens`) would ship without a failing test.

## What's good
- The guard is placed correctly: after the overage-halt short-circuit (so halted proxies still 503 first) but before any `.model` peek, and it returns from inside the same `try` whose `finally` releases the queue slot (`src/proxy.ts:4862-4868`), so no leaked concurrency slot on the new 400 path.
- The malformed-UTF-8 case is handled correctly and is a real fix in its own right: `Buffer.toString()`'s lenient decoding would otherwise turn `0xff` into U+FFFD and let bytes that were never valid UTF-8 slip through to be forwarded — the fatal `TextDecoder` decode closes that gap.
- The new test is hermetic (mkdtemp'd HOME, throwing `fetchImpl`, explicit `upstreamCalls === 0` assertion) and pins down the actual regression the PR describes (drift-watch run 34044346444) rather than just the happy path.
- CI is green across the full matrix (build, test, live-test, wire-drift, CodeQL, compat).

PR #1232, opened 2026-09-06
fix(pool): let a rate-limit rejection expire with its own window

askalf:
## What

A seat marked `rejected` stays out of the pool until the proxy restarts, even after the window that rejected it has rolled over. Eligibility now expires the rejection against `anthropic-ratelimit-unified-reset`.

## Why

Found on the fleet box today. The Claude pool had one seat, upstream returned 429 with `unified-5h-status: rejected` and `unified-5h-utilization: 1.06` against `surpassed-threshold: 1.0`, so the seat was genuinely over its five-hour window and correctly parked. Its reset was twenty minutes out.

The problem is what happens after those twenty minutes. `markRejected` sets the status, `accountIneligibility` filters the account out of `select()`, and `updateRateLimits` only runs when a request completes on that account. Being parked is exactly what stops it being sent a request, so the snapshot cannot refresh itself. The rejection has no expiry.

Once a second subscription was added the pool had a healthy seat, so the parked one was never reached by the fallback at `pool.ts:518` either. It would have read `rate-limited` to routing, and `rejected` to the operator, for as long as the healthy seat held out. `GET /accounts` already carries a comment describing this staleness as self-correcting because "the pool does return parked accounts to service on its own" — that is only true when every seat is ineligible at once.

## Changes

- `rateLimitWindowPassed()` in `src/pool.ts`. `reset` is epoch **seconds** (the header's unit, the same one `formatReset` scales) while `now` is milliseconds, so the conversion is explicit. A snapshot with no stated reset (`0`) keeps its rejection rather than guessing.
- `accountIneligibility()` treats a rejection as spent once its window has rolled. `isAccountEligible` and `selectSticky` inherit this, since both route through it. Token expiry and auth cool-down are unaffected and still outrank it.
- `reportedAccountStatus()` consolidates the status derivation that `GET /accounts` and `GET /admin/accounts` were each doing inline. An expired rejection reports `unknown`, not `allowed`: the window rolled over, but nothing has measured the account since, and `unknown` is already what a never-used seat reports.

## Notes

Letting the rejection expire is a decision to re-probe, not a claim the seat is free. If it 429s again the fresh snapshot re-parks it on the new window, which the last test block covers.

Selection order is unchanged. A recovered seat competes on headroom like any other, so it is only preferred when its own reading says it has more.

## Testing

`test/pool-rate-limit-reset.mjs`, 23 checks: the reset boundary, the seconds-vs-milliseconds trap, the other ineligibility reasons still outranking, both reported statuses, a routing case where a recovered seat becomes reachable again, and re-parking on a fresh 429.

Verified as a real regression test by reverting the eligibility line in `dist/` and re-running: 7 of the 23 fail, including the routing case.

`npm test` 172/172, preflight clean, changelog gate ok, README line-count ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.28] - 2026-09-06
+
 - **A rate-limited seat returns to the pool when its window resets.** A `rejected` account is filtered out of selection, so nothing sent it another request, so its snapshot never refreshed — the rejection outlived the window that caused it, and the only ways back into rotation were the all-exhausted fallback in `select()` or a proxy restart. On a two-seat pool that meant a seat could stay parked long past its own reset for as long as the other one held out. Eligibility now expires the rejection against `anthropic-ratelimit-unified-reset`. `GET /accounts` and `GET /admin/accounts` report such a seat as `unknown` rather than `rejected` — the window rolled over, but nothing has measured it since. A seat whose snapshot states no reset stays parked, and a fresh 429 re-parks it on the new window.
 
 ## [6.0.27] - 2026-09-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+- **A rate-limited seat returns to the pool when its window resets.** A `rejected` account is filtered out of selection, so nothing sent it another request, so its snapshot never refreshed — the rejection outlived the window that caused it, and the only ways back into rotation were the all-exhausted fallback in `select()` or a proxy restart. On a two-seat pool that meant a seat could stay parked long past its own reset for as long as the other one held out. Eligibility now expires the rejection against `anthropic-ratelimit-unified-reset`. `GET /accounts` and `GET /admin/accounts` report such a seat as `unknown` rather than `rejected` — the window rolled over, but nothing has measured it since. A seat whose snapshot states no reset stays parked, and a fresh 429 re-parks it on the new window.
+
 ## [6.0.27] - 2026-09-06
 
 - **`dario doctor` no longer claims a seat whose identity differs from `~/.claude.json` will 401.** The Identity row is now `info`, not `warn`: a differing identity is the expected state for any seat added from another machine or headless (a minted identity), and a minted identity served Sonnet 5 / Opus 5 / Haiku with 200 on 2026-09-06 on a plan with Extra Usage disabled. The only 401 ever observed came from an identity that belongs to a *different account* than the seat's bearer (a half-copied transplant), so the row now says exactly that and keeps the remove-then-add re-snapshot as the fix for that symptom. The no-`~/.claude.json` row stops asserting an Extra Usage billing outcome dario cannot observe; pool seats carry their own snapshot identity regardless.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.27",
+  "version": "6.0.28",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.0.27",
+      "version": "6.0.28",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.27",
+  "version": "6.0.28",
   "description": "Use your Claude Pro/Max subscription in any tool — Cursor, Cline, Aider, the Agent SDK, your scripts — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint.",
   "type": "module",
   "bin": {
--- a/src/pool.ts
+++ b/src/pool.ts
@@ -175,11 +175,57 @@ export type AccountIneligibility = 'rate-limited' | 'token-expired' | 'auth-cool
  * need to name the failure, and a boolean forces each of them to re-derive it
  * and drift again.
  */
+/**
+ * Has the window that produced a `rejected` reading rolled over?
+ *
+ * A rejection is a verdict with an expiry date: `anthropic-ratelimit-unified-reset`
+ * names the moment the window that refused the request resets. Past it, the old
+ * reading says nothing about the account any more.
+ *
+ * This matters because a rejected account is filtered out of `select()`, so it
+ * is sent no further requests, so `updateRateLimits` never runs for it and its
+ * snapshot never refreshes. The only routes back into rotation were the
+ * all-exhausted fallback in `select()` and a proxy restart. Observed on the
+ * fleet box (2026-09-06): one seat sat parked on a 106% five-hour reading while
+ * a second subscription carried every request, and it would have stayed parked
+ * past its own reset for as long as the other seat held out.
+ *
+ * `reset` is epoch SECONDS — the header's own unit, the same one `formatReset`
+ * scales — so it is converted here. A snapshot with no reset (0) keeps its
+ * rejection: with no stated rollover there is nothing to expire, and guessing
+ * would push a genuinely throttled account back into rotation.
+ */
+export function rateLimitWindowPassed(rl: RateLimitSnapshot, now: number = Date.now()): boolean {
+  return rl.reset > 0 && rl.reset * 1000 <= now;
+}
+
+/**
+ * The status string the operator-facing surfaces report for one account —
+ * `GET /accounts` and `GET /admin/accounts`, which must agree with each other
+ * and with what routing actually does.
+ *
+ * Auth cool-down outranks the rate-limit reading: a 401 streak is both the more
+ * urgent fact and the one the rate-limit headers cannot describe, since 401
+ * responses carry none. An expired rejection degrades to `unknown` rather than
+ * `allowed` — the window rolled over, but nothing has measured the account
+ * since, and reporting `allowed` would assert a serving capacity no request has
+ * demonstrated. `unknown` is what a never-used account already reports, which is
+ * exactly the state this is: no current observation.
+ */
+export function reportedAccountStatus(account: PoolAccount, now: number = Date.now()): string {
+  if (isInAuthCooldown(account, now)) return 'auth-cooldown';
+  if (account.rateLimit.status === 'rejected' && rateLimitWindowPassed(account.rateLimit, now)) return 'unknown';
+  return account.rateLimit.status;
+}
+
 export function accountIneligibility(
   account: PoolAccount,
   now: number = Date.now(),
 ): AccountIneligibility | null {
-  if (account.rateLimit.status === 'rejected') return 'rate-limited';
+  // A rejection outlives its own window unless it is allowed to expire:
+  // nothing refreshes a parked account's snapshot, because being parked is
+  // what stops it being sent requests.
+  if (account.rateLimit.status === 'rejected' && !rateLimitWindowPassed(account.rateLimit, now)) return 'rate-limited';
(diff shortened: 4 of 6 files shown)


sprayberry-redline:
This PR touches shipping files (src/pool.ts, src/proxy.ts) but package.json version is unchanged (6.0.27). Per the release gate, merging without a version bump ships NOTHING — no GitHub release tag, no npm publish, devices receive zero updates. Please add a package.json version bump commit to this PR before we can approve merge.

sprayberry-redline (dismissed):
Reviewed by Claude Opus 4.6 (gating review).

## What I checked
- `src/pool.ts`: the new `rateLimitWindowPassed()`, `reportedAccountStatus()`, and the updated `accountIneligibility()`.
- `src/proxy.ts`: both call sites (`GET /accounts`, `GET /admin/accounts`) that switched from inline `isInAuthCooldown(...) ? 'auth-cooldown' : a.rateLimit.status` to `reportedAccountStatus(a, now)`.
- `test/pool-rate-limit-reset.mjs` (196 new lines, 23 checks).
- CI check-rollup for PR #1232: all 17 checks pass (CodeQL, build ×3, test, live-test, compat, etc.).

## Correctness
- `rateLimitWindowPassed`: `rl.reset > 0 && rl.reset * 1000 <= now` (`src/pool.ts:210-212` in the diff) correctly converts the header's epoch-seconds `reset` to milliseconds before comparing against `Date.now()`-scale `now`. The test suite explicitly covers the drop-the-`×1000` failure mode ("a reset an hour in the FUTURE is not treated as expired") and it passes. `reset === 0` (never-stated) is excluded from expiring by the `> 0` guard, matching `EMPTY_SNAPSHOT.reset = 0`.
- `accountIneligibility`'s rewritten line — `if (account.rateLimit.status === 'rejected' && !rateLimitWindowPassed(account.rateLimit, now)) return 'rate-limited';` — preserves the fall-through order (token-expiry and auth-cooldown checks still run afterward and still outrank an expired rejection), matching the PR's stated intent and the "other ineligibility reasons still outrank" test block.
- `reportedAccountStatus` checks auth-cooldown before the rejection-expiry case, so a 401 streak still reports `auth-cooldown` even on an account whose rate-limit window has also rolled over — consistent with `isInAuthCooldown` outranking rate-limit state elsewhere in this file.
- Both `proxy.ts` call sites correctly swap in `reportedAccountStatus(a, now/snapNow)`; `inCooldown` is still retained and used at `proxy.ts:2546` for `cooldownMs`, so it wasn't left as dead code.

## Test coverage
The new test file is genuinely thorough for this size of change: boundary case (`reset` exactly at `now`), the unit-conversion trap in both directions, the no-reset-stated non-expiry case, ineligibility-reason precedence, and a `select()`-level routing case showing a recovered seat becomes reachable again. The PR description states this was validated as a real regression test by reverting the eligibility line in `dist/` (7/23 failures) — a good sanity check, though I didn't independently reproduce it.

## What's good
Scoped correctly to the actual bug (rejection never expiring because a parked account never gets re-probed), doesn't touch selection ordering/preference beyond what's needed, and the CHANGELOG entry accurately describes the change.

askalf:
Release gate: this PR changes shipping files (`src/pool.ts`, `src/proxy.ts`) but `package.json` stays at 6.0.27, so `cc-drift-auto-release` would find the existing `v6.0.27` tag and short-circuit — merging as-is ships nothing.

Bump opened as #1233, **based on this branch** (`fix/rate-limit-reset-eligibility`) rather than master, so it becomes part of this PR's own diff instead of consuming the next version slot separately.

Merge order:
1. #1233 → `fix/rate-limit-reset-eligibility`
2. this PR → `master`, which then cuts `v6.0.28`

Ref ticket 00MTQ3TS0Y248B079CD5225745.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — this is a well-scoped fix with strong regression coverage. One minor documentation-placement nit below.

## What I checked
Full diff (248 additions / 4 deletions across `CHANGELOG.md`, `src/pool.ts`, `src/proxy.ts`, `test/pool-rate-limit-reset.mjs`) plus surrounding context in `src/pool.ts` (the full `select()`/`selectSticky()` bodies, `parseRateLimits`, `markRejected`/`updateRateLimits`) and `src/proxy.ts` (both call sites of `reportedAccountStatus`, the 429 capture at line ~3958). I traced the seconds-vs-milliseconds conversion, the ineligibility/status precedence order, and the new test file's assertions against the implementation.

## Correctness
The core logic holds up under adversarial tracing:
- `rateLimitWindowPassed`: `rl.reset > 0 && rl.reset * 1000 <= now` — correctly treats `reset` as epoch seconds vs. `now` in ms, and the `reset > 0` guard correctly keeps a no-reset snapshot parked rather than misreading `0` as "already expired." Verified against the "reset an hour in the FUTURE" test case: without the `*1000` this would wrongly read as expired (epoch-seconds value ~1.79e9 < epoch-ms `now` ~1.79e12) — the test correctly guards this regression.
- `accountIneligibility` and `reportedAccountStatus` have different but each internally-consistent precedence orders (ineligibility checks rate-limited → token-expired → auth-cooldown; status checks auth-cooldown → expired-rejection → raw status), and both are exercised by the new test file.
- `selectSticky`/`select()` correctly inherit the new eligibility via `isAccountEligible`, and the routing test (`pool-rate-limit-reset.mjs:157-177`) verifies a recovered seat actually becomes selectable again, not just relabeled.

## Minor — orphaned doc comment
`src/pool.ts:161-177` (pre-existing, unchanged by this diff) is the docstring for `accountIneligibility` ("The single answer to \"can the router serve a request from this account...\""). This PR inserts the new `rateLimitWindowPassed` function plus its own docstring (`src/pool.ts:178-200`) directly between that comment and the `accountIneligibility` function it actually describes (now at line 221), with `reportedAccountStatus` and its docstring (202-219) also wedged in between. A reader following the `accountIneligibility` docstring downward now lands on two unrelated functions before reaching the one it documents.

```suggestion
Move the accountIneligibility docstring (lines 161-177) to sit immediately above the accountIneligibility function definition, after rateLimitWindowPassed and reportedAccountStatus.
```

## What's good
- The seconds/ms unit trap is exactly the kind of bug this test suite is built to catch, and it does: `test/pool-rate-limit-reset.mjs` includes a dedicated "seconds vs ms" case that fails without the `*1000` conversion, plus the PR description states this was verified by reverting the eligibility line in `dist/` and confirming 7/23 checks fail.
- `reportedAccountStatus()` consolidating logic previously duplicated inline at `GET /accounts` and `GET /admin/accounts` (`src/proxy.ts:2473`, `src/proxy.ts:2570`) removes a drift risk between the two surfaces.
- Precedence between auth-cooldown, token-expiry, and rate-limiting is preserved and independently tested (`test/pool-rate-limit-reset.mjs:110-123`), so this change doesn't let a genuinely-broken (expired-token or cooling-down) account get resurrected by an unrelated window rollover.

sprayberry-redline (dismissed):

I reviewed the complete six-file diff at live head `24768d1accddb6aba06732ed2a854bcec221f19b`, including the release-version follow-up commit. I checked the reset-unit conversion and boundary, the no-reset behavior, eligibility precedence, sticky/non-sticky selection paths, operator-facing status derivation, and the focused regression coverage. The PR's CI rollup is green (with the auto-merge job intentionally skipped).

**What's good:** The change addresses the self-stale rejection at the shared eligibility predicate, keeps unknown-reset snapshots conservatively parked, reports expired observations as `unknown` rather than overstating capacity, and covers both re-entry and re-parking behavior.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — the core fix (`rateLimitWindowPassed`/`accountIneligibility`/`reportedAccountStatus` in `src/pool.ts`) is correct and well covered by the new test file. One minor documentation nit below.

## What I checked
- `src/pool.ts`: the new `rateLimitWindowPassed`, `reportedAccountStatus`, and the updated `accountIneligibility` gate, plus how `select()`, `selectSticky()`, and `isAccountEligible()` consume them.
- `src/proxy.ts`: both call sites (`/accounts`, `/admin/accounts` live status) that switched from inline `isInAuthCooldown ? 'auth-cooldown' : a.rateLimit.status` ternaries to `reportedAccountStatus(a, now)`.
- `test/pool-rate-limit-reset.mjs` (196 lines, 23 checks) against the implementation, including the seconds-vs-milliseconds unit trap and the routing case.
- Traced `parseRateLimits`/`EMPTY_SNAPSHOT`/`RateLimitSnapshot.reset` to confirm the unit really is epoch seconds, matching `formatReset` in `doctor-core.ts:115-116` (`resetEpochSecs * 1000 - nowMs`).
- I did not build/run the suite myself (`tsc` unavailable in this sandbox, `npm ci` blocked by a pre-existing root-owned `/app/.npm` cache), so I verified correctness by reading, not executing.

## Correctness
The core change reads correctly:
- `rateLimitWindowPassed`: `rl.reset > 0 && rl.reset * 1000 <= now` — correct unit conversion, and `reset === 0` (never-set) correctly never expires, matching the stated intent that "nothing to expire" than guess.
- `accountIneligibility` (`src/pool.ts:228`): `if (account.rateLimit.status === 'rejected' && !rateLimitWindowPassed(...)) return 'rate-limited';` — only a *live* rejection blocks eligibility now; an expired one falls through to the token-expiry/auth-cooldown checks below, which is exactly the described behavior and matches the "other reasons still outrank" test block.
- `reportedAccountStatus` (`src/pool.ts:215-219`) correctly orders auth-cooldown first (since 401s carry no rate-limit headers to contradict it), then degrades an expired rejection to `unknown` rather than fabricating `allowed`.
- Both `proxy.ts` call sites for `/accounts` and `/admin/accounts` (line 2473, 2570) now route through the same `reportedAccountStatus`, closing the gap the PR body calls out (the two surfaces previously used a plain `a.rateLimit.status`/`isInAuthCooldown` ternary each, now unified).

I tried to construct a failure case around the boundary (`reset` equal to `now`) and around fallback ordering in `select()`'s all-exhausted branch, but both hold up: `select()`'s existing `withReset` fallback (`pool.ts:564`, unchanged by this diff) is only reached when `eligible.length === 0`, and an expired-rejection account is now eligible earlier in the function, so it's picked up before that fallback runs — consistent with the PR's "selection order is unchanged; a recovered seat competes on headroom" note.

## Minor
`src/pool.ts:161-177` — the pre-existing docstring beginning "The single answer to \"can the router serve a request from this account, and if not, why not?\" (dario#1030)... It returns the REASON rather than a boolean..." previously sat directly above `accountIneligibility`. The two new exported functions (`rateLimitWindowPassed`, `reportedAccountStatus`) were inserted between that docstring and `accountIneligibility` itself, so the comment now describes `rateLimitWindowPassed` at a glance rather than the function it was written for.

```suggestion
Move the dario#1030 docstring down so it sits directly above `export function accountIneligibility(`, and add a short one-line comment above `rateLimitWindowPassed` describing just that function.
```

## What's good
- The five-hour/five-day/seconds-vs-milliseconds unit trap is exactly the kind of bug this class of code invites, and the test suite has a dedicated check for it (`test/pool-rate-limit-reset.mjs`, "a reset an hour in the FUTURE is not treated as expired").
- The regression-test verification described in the PR body (reverting the eligibility line in `dist/` and confirming 7/23 checks fail) is good practice and gives real confidence the new tests aren't vacuous.
- Consolidating `/accounts` and `/admin/accounts` status derivation into one `reportedAccountStatus` removes a second, drift-prone reimplementation — same category of fix as the `accountIneligibility` consolidation referenced from dario#1030.

sprayberry-redline (approved):

Verdict: Approve — no blocking issues found on the current live head.

I reviewed the complete six-file diff, including the reset unit conversion and boundary, eligibility ordering, both operator-facing status call sites, and the new routing/regression coverage. A rejection with no usable reset remains parked, a rejection whose epoch-seconds reset has passed becomes eligible for a probe, and token expiry/auth cooldown still prevent selection on that recovered path. All reported CI checks are green.

What's good: the tests exercise the exact-reset boundary, seconds-versus-milliseconds regression, missing-reset behavior, status reporting, real pool selection, and re-parking after a fresh 429.

PR #1234, opened 2026-09-06
feat(codex): send prompt_cache_key and report cached prompt tokens on both shapes

askalf:
## What

Every request to the ChatGPT Codex backend now carries a `prompt_cache_key`, and the cached-token counts the backend returns reach the client, analytics, the request log and the `-v` usage line on both wire shapes.

## Why

The backend caches prompt prefixes (1,024 tokens and up, about 30 minutes) on its own, but per OpenAI's prompt-caching guide the key is what "influences routing" so same-prefix requests land on the machine holding the cache. The Codex CLI sends one on every turn (`ResponsesApiRequest.prompt_cache_key`, its session id). dario sent none, and `toCodexSupportedBody` stripped any a client supplied, so a fleet reviewer re-sending the same 20KB system prompt every dispatch was routed blind.

Separately, `input_tokens_details.cached_tokens` was already in the `ResponsesUsage` type and then dropped by both translators. A Claude Code session on codex fallback saw zero `cache_read_input_tokens`; `/analytics` recorded `cacheReadTokens: 0` for every codex row; the GPT lane's hit rate was unobservable.

## How

**Key resolution** (`codexPromptCacheKey`, pure):
1. the chat body's own `prompt_cache_key` (the public chat/completions API takes it; Cursor-class harnesses key on their conversation id);
2. Anthropic shape: SHA-256 of `metadata.user_id`, which Claude Code stamps per session. Hashed because the value embeds the client's Anthropic account and session ids, and neither should reach a second vendor in the clear;
3. otherwise SHA-256 of model + instructions + tool names, so every caller with the same system prompt and tool set shares a key.

`prompt_cache_key` joins `CODEX_SUPPORTED_FIELDS` and the chat-field translation map (so `--verbose` stays quiet about it).

**Usage out.** OpenAI counts the cached prefix inside `input_tokens` and repeats it under details; Anthropic counts it beside. So:
- chat/completions: `prompt_tokens` stays gross, `prompt_tokens_details.cached_tokens` added (a rename, where SDKs and cost dashboards already look);
- `/v1/messages`: `cache_read_input_tokens` / `cache_creation_input_tokens` added and `input_tokens` netted to the remainder (`anthropicUsageFromResponses`, one function used by the non-streaming body, the streaming `message_delta`, and the analytics split, so they cannot disagree). Copying the gross total across would make Claude Code's context meter count the cached prefix twice.
- `CodexForwardOutcome` gains `cacheReadTokens` / `cacheCreateTokens`; proxy.ts writes them into the analytics row and the log line and prints the existing `formatUsageLogLine` for codex requests under `-v`.

Both fields are only emitted when upstream sent `input_tokens_details`; without it the usage shapes are byte-identical to before.

## Verification

- `npm test`: 174/174 files. `test/codex-backend.mjs` 290 to 333 checks; translate suite gains the netting cases. New coverage: allowlist and pass-through, per-session key stability and that no raw id appears in the outbound body, prefix-key stability across user turns, both shapes end to end (body, stream `message_delta`, outcome), over-reported cache never yields negative input.
- Two existing fixtures already carried `cached_tokens: 1024` and asserted the gross number; their expectations now assert the netted contract, with the reasoning inline.

## Trade-offs and unknowns

- **`prompt_cache_key` is accepted by construction, not live-probed here.** The Codex CLI sends it to this exact endpoint on every request, so it cannot be in the backend's rejected-parameter class. The first `codex-drift-watch.yml` run after merge exercises it live through the wire-contract step; if the backend ever 400s on it, the allowlist line is the one-line revert.
- **Not sending `session-id` / `thread-id` headers**, which the CLI also sends. The body key is what OpenAI documents as the routing influence; the headers add fingerprint surface and unknown validation for no documented gain.
- **The Anthropic-shape `input_tokens` changes for codex-served requests that hit the cache** (net instead of gross). Intentional: the gross number was wrong in Anthropic terms, and no dario code consumed it.
- **Whether cached tokens burn less ChatGPT subscription quota is undocumented.** The API-billing discount is documented; the subscription effect is not. The hit rate is now observable in `/analytics`, so this becomes answerable after a day of reviewer traffic.
- **No version bump in this PR on purpose**; release PRs are being sequenced separately. Fold into the next bump.

## Added in round 2: cache totals on `/analytics`

Verifying this live exposed that `/analytics` had no cache fields at all: every record carried `cacheReadTokens` / `cacheCreateTokens`, but `summary()` never summed them, so the hit rate for the Claude pool was readable only off a `-v` console (which the fleet box does not run). The second commit adds `totalCacheReadTokens` / `totalCacheCreateTokens` / `cachedPromptPercent` to the window and all-time blocks, the same three per account, and per-model averages, through one exported `cachedPromptPercent()`. Additive; existing consumers read the same fields as before. Measured on the fleet box from the stream backlog while this was open: Claude sonnet 80% and opus 60% of prompt tokens from cache; the codex lane 0% because the numbers were being dropped, which is what the first commit fixes.


sprayberry-redline (dismissed):

Verdict: Approve — no blocking correctness, security, or test issues found on live head `31820517790b1919622e325e08325a9eb9c7de68`.

I reviewed the complete eight-file diff. In particular, I checked client-key precedence and validation, hashing/privacy and fallback-key stability, Codex allowlisting and outbound forwarding, OpenAI gross-token versus Anthropic net-token semantics, propagation through non-streaming, streaming, analytics and verbose logging, and the corresponding regression coverage. All reported CI checks are green.

What's good: the usage conversion is centralized in `anthropicUsageFromResponses`, the analytics path reuses the same accounting rather than duplicating it, and the tests cover both wire shapes, stream and non-stream behavior, malformed/over-reported details, privacy of session metadata, and end-to-end forwarding.

askalf:
Added the release-gate version bump to this PR (0ed60c7): package.json + package-lock.json 6.0.29 -> 6.0.30, and the prompt_cache_key CHANGELOG entry dated as `## [6.0.30] - 2026-09-06` with a fresh `## [Unreleased]` above it per the convention in the file header.

Without this, merging would have shipped src/proxy.ts and src/codex-backend.ts changes with no version change -> no release tag -> no npm publish -> devices receive zero updates. Verified before the push that the code was not already covered by a later tag: head was ahead_by=1 / behind_by=0 vs v6.0.29 and package.json read 6.0.29 at both the PR head and master.

Source ticket: 00MTQCHOKO51D744B110EDC648

sprayberry-redline (dismissed):

Verdict: Approve — no blocking correctness, security, or test issues found.

I reviewed the complete 10-file diff and the surrounding request/response paths. In particular, I checked client-key precedence and hashed fallback-key stability, Codex allowlist pass-through, OpenAI gross-token versus Anthropic net-token accounting, both streaming and collapsed response paths, and analytics/request-log propagation. The focused tests cover these paths, including malformed/over-reported usage, and all current CI checks are green.

What's good: the usage conversion is centralized and reused across wire output and analytics, reducing the risk that streaming, non-streaming, and observability report different token totals. The raw Anthropic session identifier is hashed before crossing vendor boundaries.

sprayberry-redline (approved):

Verdict: Approve — no blocking correctness, security, efficiency, or test issues found.

I reviewed the complete 12-file diff and surrounding translation, transport, streaming-assembly, analytics, and logging paths. In particular, I checked client-key precedence and derived-key stability/privacy; OpenAI gross-token versus Anthropic net-token accounting; parity between streaming, collapsed, analytics, and request-log outputs; malformed/missing/over-reported usage handling; and the added regression coverage. All reported CI checks are green.

What's good: the change centralizes the cross-API usage conversion, keeps cache details conditional on upstream details, carries one accounting convention consistently into analytics, and adds focused end-to-end tests for both wire shapes and key-resolution fallbacks.

PR #1235, opened 2026-09-06
README: generated visuals, live drift dashboard, per-tool setup, link guard

askalf:
## What this is

The README, rebuilt around what a first-time visitor needs in the first ten seconds and what a returning one needs in the first minute. Same facts, same voice, same honesty sections; a different shape, and visuals that are generated from the product rather than drawn by hand.

**Rendered preview:** open this branch's README on GitHub; every image is a committed SVG under `.github/readme/`, so it renders on the PR branch as it will on master.

## What changed

**Structure.** The v6 release banner and the front-loaded prose are gone; the page now runs start → your tools → routing → two plans → pool → drift → guardrails → trust → risk → commands → FAQ. A quick-nav line under the badges, and `docs/returning.md` is linked for people coming back. Per-tool setup (Claude Code, Cursor incl. the tunnel caveat, Cline/Roo/Kilo, Aider, Continue, Zed, OpenHands, OpenClaw, Codex CLI / OpenAI SDK, Agent SDK, curl, Docker) is collapsed `<details>` with snippets copied from `docs/integrations/agent-compat.md` and `docs/usage.md`. Long tails (extra knobs, FAQ) are collapsed too.

**Visuals, all generated, all theme-aware via `<picture>`:**
- `hero-{dark,light}.svg` — the routing model with SMIL motion: requests flow tool → dario → plan, and once per loop a Claude-bound request meets a 429 and is re-served by the ChatGPT plan with `x-dario-pool-fallback` on the response. Nothing drawn that isn't shipped.
- `quickstart-{dark,light}.svg` — an animated terminal from `scripts/readme/quickstart.cast.json`. The banner lines are the ones `src/proxy.ts` prints; the login lines are `src/cli.ts`'s. The OAuth step can't be recorded without publishing a login, so it is scripted and the generator says so.
- `tui-analytics-*.svg`, `tui-accounts-*.svg` — **rendered from the real TUI**: `scripts/readme/tui.mjs` drives `startTuiApp()` through a fake TTY against a fixture proxy (the `tools/tui-audit` technique) and turns the ANSI frame into SVG with SGR colours kept and box/block glyphs drawn as vectors. A layout change in `src/tui/` changes the picture on the next `npm run readme:assets` instead of leaving a stale mock-up. The numbers are fixtures; the pixels are not. The old hand-drawn boxes showed things the TUI does not render (plan tiers, a `?` help key); those are gone.
- Two mermaid diagrams (routing flowchart, failover sequence) render natively on GitHub in both themes.

[...]

PR #1237, opened 2026-09-06
test: take proxy ports from the kernel instead of pinning ephemeral-range numbers

askalf:
## Why

`live-test` on the self-hosted runner failed twice today on #1235 (a docs-only PR) with:

```
not ok 107 - pool-exhaustion-midflight-429-wiring.mjs
[dario] Port 38838 is already in use by another process.
```

The test pins `PROXY_PORT = 38838`. Linux hands out ephemeral source ports from 32768–60999, and on the runner box that exact port was taken by a long-lived loopback connection that has nothing to do with dario:

```
ESTAB  127.0.0.1:38838  127.0.0.1:3005  users:(("cloudflared",pid=1202732,fd=8))
```

Until that connection recycles, every PR's required check fails the same way. The same test passes on GitHub-hosted runners and locally, so it read like a flake; it is a deterministic collision with whatever happens to hold the number on a busy host.

## What

Sixteen in-process proxy tests declared 31 fixed ports in that range (two files even shared 38801/38802). Each now takes its port from the kernel:

```js
import { freePort } from './helpers/free-port.mjs';
const PROXY_PORT = await freePort();
```

`test/helpers/free-port.mjs` listens on port 0, reads the assignment, closes, and returns it. The window between release and the proxy's own `listen` is the usual trade-off for a server that has to know its port before it starts; it is microseconds against a collision that lasts hours. It lives in a subdirectory so `all.test.mjs` (top-level `*.mjs` only) does not run it as a test. `openai-backend-passthrough.mjs` and `tui-proxy-client-health.mjs` already did this inline; nothing else changes.

No `src/` change, so no CHANGELOG entry (`check-changelog` diffs `src/`). Verified: every touched file passes on its own and the full `npm test` passes locally.

## Also seen on the runner, not touched here

`ps` on the box shows **170** orphaned `node dist/cli.js proxy --port=NNNNN --passthrough` processes, the oldest 193 hours old, cwd `/root/actions-runner/_work/dario/dario`. That is one leaked proxy per `compat-test-self-hosted` run. They are not what held 38838, but a cleanup step in that workflow (or a `pkill -f` scoped to the runner's workdir in a `post` step) is worth a separate PR; killing them on the box is an operator call.

### DIFF
--- a/test/auth-failover-403-proxy.mjs
+++ b/test/auth-failover-403-proxy.mjs
@@ -6,6 +6,7 @@
 import { mkdtemp, mkdir, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { freePort } from './helpers/free-port.mjs';
 
 let pass = 0, fail = 0;
 const check = (name, cond, detail = '') => {
@@ -49,7 +50,7 @@ const fakeFetch = async (url, init = {}) => {
 };
 
 const { startProxy } = await import('../dist/proxy.js');
-const port = 38847;
+const port = await freePort();
 await startProxy({ port, host: '127.0.0.1', noLiveCapture: true, fetchImpl: fakeFetch });
 for (let i = 0; i < 50; i++) {
   try { await fetch(`http://127.0.0.1:${port}/health`); break; } catch { await sleep(50); }
--- a/test/codex-admin-surface.mjs
+++ b/test/codex-admin-surface.mjs
@@ -12,6 +12,7 @@ import { createServer } from 'node:http';
 import { mkdtemp, mkdir, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { freePort } from './helpers/free-port.mjs';
 
 let pass = 0, fail = 0;
 const check = (name, cond, detail) => {
@@ -21,9 +22,9 @@ const check = (name, cond, detail) => {
 const header = (n) => console.log(`\n=== ${n} ===`);
 const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
 
-// Clear of 3456-3460 and the other test files' ports.
-const PROXY_PORT = 38801;
-const STUB_PORT = 38802;
+// Kernel-assigned so no other process or test file can hold it (helpers/free-port.mjs).
+const PROXY_PORT = await freePort();
+const STUB_PORT = await freePort();
 const BASE = `http://127.0.0.1:${PROXY_PORT}`;
 const LISTED_SLUG = 'gpt-5.6-admin';
 
--- a/test/codex-decline-carries-effort.mjs
+++ b/test/codex-decline-carries-effort.mjs
@@ -29,6 +29,7 @@ import { createServer } from 'node:http';
 import { mkdtemp, mkdir, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+import { freePort } from './helpers/free-port.mjs';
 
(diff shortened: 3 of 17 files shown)


sprayberry-redline (approved):

**Verdict: APPROVE — no blocking issues found.**

I reviewed the complete 17-file diff at live head `3164ab56fd5dcb573510c998cd654197e892fe0f`, including all 31 fixed-port replacements and the new `test/helpers/free-port.mjs` helper. The helper binds loopback port 0, reads the kernel assignment only after listening, closes before resolving, and propagates bind errors. Each changed test awaits the helper before constructing URLs, environment values, stubs, or proxy instances, preserving the existing setup order.

What's good: this centralizes previously repeated port-selection behavior, removes deterministic collisions with long-lived host connections, and keeps the helper outside the top-level test discovery glob. All reported CI checks are green, including `test`, `live-test`, the Node 18/20/22 builds, CodeQL, and actionlint.

sprayberry-secondread (commented):

**Verdict:** No blocking issues — this is a solid, well-scoped fix for the flaky-port problem, and CI (including `live-test`, the check that motivated it) is green at head `3164ab5`.

## What I checked
- `test/helpers/free-port.mjs` — the new helper listens on port 0, reads back the OS-assigned port, and closes before returning it. I reproduced the "ask twice in a row" case locally (2000 back-to-back `freePort()`/`freePort()` pairs): zero collisions, consistent with the listen→close→next-listen ordering not overlapping.
- Verified none of the 16 touched files still has a residual pinned port in the `32768–60999` ephemeral range (`grep -rnE "PORT[0-9]?\s*=\s*3[0-9]{4}" test/*.mjs` — no hits), so the sweep the PR body claims is actually complete.
- `test/all.test.mjs`'s `EXCLUDED`/`readdirSync` filtering (`test/all.test.mjs:43-57`) only picks up top-level `*.mjs`, confirming `test/helpers/free-port.mjs` correctly stays out of the runner as the PR body states.
- `gh pr checks 1237` — all checks pass, including `live-test`, `test`, and `build (18/20/22)`.

## Findings
None rise to a blocking or even non-trivial correctness concern. Two minor observations, informational only:

1. **Release-then-listen window is real but narrow, as the PR body itself acknowledges** (`test/helpers/free-port.mjs:12-15`, comment "there is a short window in which something else could take it"). On a CI runner already juggling ephemeral ports this window is non-zero, but it's the same trade-off `openai-backend-passthrough.mjs` and `tui-proxy-client-health.mjs` already ship with today, and it converges to "sometimes flaky" rather than "always broken," which is a strict improvement over a permanently-collided fixed port. Not asking for a change — SO_REUSEPORT-style pass-the-fd would add complexity this test suite doesn't need.
2. Files like `test/seat-pin-proxy.mjs` and `test/pool-exhaustion-gate-wiring.mjs` now call `await freePort()` 2-4 times sequentially per file (e.g. `test/seat-pin-proxy.mjs:22,68,159,189`). Since each call fully closes its probe socket before returning, sequential awaits (as opposed to `Promise.all`) are actually the safer form here — this is the right call, not a nit.

## What's good
- Root-caused via a concrete `ss`/process trace instead of dismissing it as a flake (PR body cites the exact `ESTAB 127.0.0.1:38838 ... cloudflared` line).
- Comment at `test/helpers/free-port.mjs:1-19` documents the "why" (specific incident, specific port, specific trade-off) rather than restating the code.
- Scope discipline: no `src/` changes, so correctly no CHANGELOG entry per `check-changelog`'s `src/`-diff rule.
- The separately-noted 170 orphaned `node dist/cli.js proxy` processes on the runner are explicitly called out as out of scope for this PR rather than silently left for someone to rediscover.

PR #1238, opened 2026-09-06
docs(cordon): use the published ghcr.io/askalf/cordon image

askalf:
## Why

cordon now publishes a container image (askalf/cordon#54 adds the release workflow; the `v0.2.0` tag publishes `ghcr.io/askalf/cordon`). The integration guide no longer needs to tell people to `build:` from the repo.

## What

`docs/integrations/cordon.md`: the compose example uses `image: ghcr.io/askalf/cordon:v0.2.0` with the `gh attestation verify` line as a comment. Docs only, no version bump, nothing ships.

## Do not merge before

cordon `v0.2.0` is tagged and the GHCR package is public. Until then the image reference 404s.


sprayberry-redline (approved):

PR #1239, opened 2026-09-07
perf(proxy): parse the request body once (second-read on #1231)

askalf:
Second-read finding on #1231 (fleet ticket 01M1WB07C1PT7ZSA10DBDQPQP3): the new invalid-body guard parsed the request body, then the provider-prefix block a few lines down parsed the same bytes again, directly under the #642-audit comment saying the parse is shared. The codex model peek further down was a third parse.

**Change.** `parsedBody` is declared before the guard and set from the object the guard already validated. The prefix block and the codex peek read it instead of re-parsing. `body` is still re-serialized from `parsedBody` on every mutation, so the object stays current for the template build.

**Not changed.** Wire behaviour, error shapes, routing. The guard still decodes with a fatal UTF-8 decoder; the object it yields is identical to what `body.toString()` produced for any body the guard lets through.

**Cost.** One JSON.parse per request instead of two or three; only measurable on large prompt bodies.

175/175 tests locally. No version bump in this PR; ships with the next slot.


sprayberry-redline (approved):

Verdict: Approve — no blocking correctness, security, efficiency, or test issues found.

I reviewed the complete two-file diff and the surrounding request-body lifecycle, including provider-prefix/alias mutations, codex routing, codex-to-Claude fallback invalidation, and the later Claude template build. The parsed object remains synchronized with every changed-path body rewrite, and the fallback path correctly clears it before reparsing a replaced body. I also verified that all reported CI checks are green.

What's good: this removes redundant parsing without weakening the fatal UTF-8/object validation added previously, and preserves the explicit cache invalidation required when fallback replaces the body wholesale.

PR #1240, opened 2026-09-07
ci(live-test): stop the probe proxy on every path, and reap what earlier runs left

askalf:
## What leaked

One `node dist/cli.js proxy --port=<random> --passthrough` per `live-test` run, reparented to init, cwd `/root/actions-runner/_work/dario/dario`. 176 of them on the runner box by 2026-09-07, 10 GB RSS between them, the oldest from 8 days ago — i.e. since this workflow replaced the agent-driven tester. Killed by hand today (host memory in use 5,986 → 4,365 MB); nothing else on the box was touched.

## Why

The live probe launched the proxy through the scrubbed-environment helper, backgrounded:

```bash
run() { env -i PATH=… "$@"; }
run env ANTHROPIC_UPSTREAM_API_KEY=… node dist/cli.js proxy --port="$PORT" --passthrough > … &
PROXY_PID=$!
…
kill "$PROXY_PID"
```

A backgrounded shell function is a bash subshell, so `$!` named the subshell. `kill` reached that and only that; node was orphaned to init and lived on. dario is not at fault: on the runner, a directly launched proxy exits one second after SIGTERM (`[dario] Shutting down...`). The compat and codex-drift workflows launch directly and tear down properly; they did not leak.

## What changes (live-test.yml only)

- The proxy is launched directly, under `setsid`, so `$!` is node and the proxy owns its process group. The normal-path stop signals the group (`kill -TERM -- -$pid`, falling back to the pid).
- A new `if: always()` step **Stop the probe proxy and reap anything it left behind** runs on every path the probe can end on — failure, cancel, job timeout. It reads the pid file, sends TERM, waits up to 10 s, then KILL, then sweeps for any `dist/cli.js proxy` from **this checkout** whose parent is already init and reaps it. A non-zero reap count is a `::warning::` annotation, so a future leak shows up in the run instead of in `ps` a week later. Scope is deliberately narrow: parent must be init and cwd must be under `$GITHUB_WORKSPACE`, so the production container (`/app`) and a concurrent job's live proxy (parent = its step shell) are never candidates.

The Bun relaunch in `src/cli.ts` keeps a node parent waiting on a `bun` child; the group kill covers that shape too, though the runner has no Bun today.

## Verified

- Direct-launch SIGTERM test on the runner: exits in 1 s.
- Hand cleanup of the 176 orphans with the same predicate (pattern + parent = init + runner cwd): 176 selected, 0 remaining, production proxy untouched.
- This PR's own `live-test` run executes the new step; its log should read `no orphaned proxy processes from this checkout`.

## Not in this PR

A host-level reaper (systemd timer) would be belt-and-braces if a runner crashes mid-job, but that lives in the platform's host scripts, not here.

### DIFF
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -224,32 +224,43 @@ jobs:
         if: always()
         run: |
           set +e -uo pipefail
+          # stop <pid>: TERM the process group (falls back to the pid when it is
+          # not a group leader — an orphan from before setsid), wait up to 10s,
+          # escalate to KILL, wait again. Returns 0 only once the pid is GONE;
+          # a delivered signal is not a stopped process (review on #1240).
+          stop() {
+            local p="$1"
+            kill -TERM -- "-$p" 2>/dev/null || kill -TERM "$p" 2>/dev/null || return 0   # already gone
+            for _ in $(seq 1 10); do kill -0 "$p" 2>/dev/null || return 0; sleep 1; done
+            echo "::warning::dario proxy $p ignored SIGTERM for 10s — sending SIGKILL"
+            kill -KILL -- "-$p" 2>/dev/null || kill -KILL "$p" 2>/dev/null
+            for _ in $(seq 1 5); do kill -0 "$p" 2>/dev/null || return 0; sleep 1; done
+            return 1
+          }
           pidfile="${WT_HOME:-/nonexistent}/proxy.pid"
           if [ -s "$pidfile" ]; then
             pid="$(cat "$pidfile")"
             if kill -0 "$pid" 2>/dev/null; then
-              kill -TERM -- "-$pid" 2>/dev/null || kill -TERM "$pid" 2>/dev/null
-              for _ in $(seq 1 10); do kill -0 "$pid" 2>/dev/null || break; sleep 1; done
-              if kill -0 "$pid" 2>/dev/null; then
-                echo "::warning::probe proxy $pid ignored SIGTERM for 10s — sending SIGKILL"
-                kill -KILL -- "-$pid" 2>/dev/null || kill -KILL "$pid" 2>/dev/null
-              fi
+              stop "$pid" || echo "::error::probe proxy $pid survived SIGKILL — the runner needs a look"
             fi
           fi
-          reaped=0
+          reaped=0; survived=0
           for p in $(pgrep -f 'dist/cli\.js proxy' 2>/dev/null); do
             [ "$(ps -o ppid= -p "$p" 2>/dev/null | tr -d ' ')" = "1" ] || continue
             case "$(readlink "/proc/$p/cwd" 2>/dev/null)" in
               "$GITHUB_WORKSPACE"|"$GITHUB_WORKSPACE"/*) ;;
               *) continue ;;
             esac
-            kill -TERM "$p" 2>/dev/null && reaped=$((reaped + 1))
+            if stop "$p"; then reaped=$((reaped + 1)); else survived=$((survived + 1)); echo "::error::orphaned dario proxy $p survived SIGKILL"; fi
           done
-          if [ "$reaped" -gt 0 ]; then
-            echo "::warning::reaped $reaped orphaned dario proxy process(es) from this checkout — something leaked; find the run that started them"
+          if [ "$survived" -gt 0 ]; then
+            echo "::error::$survived orphaned dario proxy process(es) from this checkout could not be stopped"
+          elif [ "$reaped" -gt 0 ]; then
+            echo "::warning::reaped $reaped orphaned dario proxy process(es) from this checkout (confirmed gone) — something leaked; find the run that started them"
           else
             echo "no orphaned proxy processes from this checkout"
           fi
+          [ "$survived" -eq 0 ]
 
       - name: Verdict → one PR comment, and the job's own exit status
         if: always() && steps.pr.outputs.sha != ''
--- a/.github/workflows/live-test.yml
+++ b/.github/workflows/live-test.yml
@@ -172,11 +172,21 @@ jobs:
             echo "live: SKIPPED (no test key configured)" | tee -a "$WT_HOME/report.txt"
             echo "result=skipped" >> "$GITHUB_OUTPUT"; exit 0
           fi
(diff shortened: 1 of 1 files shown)


sprayberry-redline (changes requested):

Verdict: changes requested — the orphan sweep does not ensure the processes it finds are actually reaped.

### Blocking — `.github/workflows/live-test.yml:240-249`

> `for p in $(pgrep -f 'dist/cli\.js proxy' 2>/dev/null); do`
> `  [ "$(ps -o ppid= -p "$p" 2>/dev/null | tr -d ' ')" = "1" ] || continue`
> `  ...`
> `  kill -TERM "$p" 2>/dev/null && reaped=$((reaped + 1))`
> `done`
> `if [ "$reaped" -gt 0 ]; then`
> `  echo "::warning::reaped $reaped orphaned dario proxy process(es) ..."`

This branch treats successful signal delivery as successful reaping. For an orphan that catches/ignores SIGTERM or needs time to shut down, `kill -TERM` returns 0 immediately, the counter increments, and the step exits without waiting, checking liveness, or escalating to SIGKILL. The process can therefore remain on the runner despite the new cleanup reporting that it was reaped. This is not merely theoretical evidence that the branch is unused: this head's `live-test` run entered it and logged `reaped 1 orphaned dario proxy process(es)` after the normal group stop.

Apply the same bounded TERM/wait/KILL behavior used for the pid-file process to every selected orphan, and only count it after confirming it is gone. Also consider group-signalling when a selected process may still own descendants.

```suggestion
            if kill -TERM "$p" 2>/dev/null; then
              for _ in $(seq 1 10); do
                kill -0 "$p" 2>/dev/null || break
                sleep 1
              done
              if kill -0 "$p" 2>/dev/null; then
                kill -KILL "$p" 2>/dev/null
                for _ in $(seq 1 10); do
                  kill -0 "$p" 2>/dev/null || break
                  sleep 1
                done
              fi
              if ! kill -0 "$p" 2>/dev/null; then
                reaped=$((reaped + 1))
              else
                echo "::error::failed to reap orphaned dario proxy process $p"
              fi
            fi
```

What's good: launching the proxy directly under `setsid`, persisting its PID, and adding an `always()` teardown correctly address the original subshell-PID leak on both the normal and abnormal paths. All reported CI checks are green, and I inspected the complete changed workflow plus the live-test teardown log.

sprayberry-secondread (commented):

**Verdict:** No blocking issues. The core fix (using `setsid` so `$!` names the actual node process instead of a bash-function subshell, then killing the process group) is sound — I reproduced the old and new behavior locally and confirmed `kill -TERM -- "-$PID"` against a `setsid`-launched background job terminates the whole group, where a plain `kill "$PID"` against the old subshell-wrapped launch would not touch the child.

**Findings (informational, none blocking):**

1. **Missing test coverage for the actual failure modes the PR targets.** `.github/workflows/live-test.yml:223-252` (the new `Stop the probe proxy and reap anything it left behind` step) exists specifically to catch leaks on the failure/cancel/timeout paths, but the PR's "Verified" section and this PR's own CI run (`live-test` job, run 34070814863) only exercise the happy path — the probe returned `pong`, so the normal in-step `kill -TERM -- "-$PROXY_PID"` / `wait` at line 209-210 handled teardown and the `always()` step never had to kill a still-running process via the pidfile branch (lines 228-238). The one process it did reap (`::warning::reaped 1 orphaned dario proxy process(es)...`) was caught by the unconditional `pgrep` sweep at the bottom, not the pidfile-kill logic, and — since this looks like a persistent, non-ephemeral self-hosted runner reusing the same `$GITHUB_WORKSPACE` path across every job for this repo (per the PR body: "176 of them on the runner box... cwd `/root/actions-runner/_work/dario/dario`") — is very plausibly a pre-existing orphan from before today's fix, not evidence this run's own proxy was ever actually stuck. The scenario the fix exists for (SIGTERM sent to a hung/ignoring process, escalating to SIGKILL after the 10s loop at line 232-236) hasn't been demonstrated to work end-to-end. Not a blocker since CI is green and the mechanism is straightforward, but worth a deliberate test (e.g. cancel a run mid-probe) before trusting the KILL escalation path in production.

2. **Comment slightly overclaims the reap sweep's scope.** `.github/workflows/live-test.yml:218-222` says the sweep is for "any proxy from THIS checkout" and the PR body says scope is narrowed to `$GITHUB_WORKSPACE` specifically to avoid catching "a concurrent job's live proxy." On a persistent self-hosted runner that reuses the same workspace path across sequential jobs (as this one appears to, per the PR's own leak description), `$GITHUB_WORKSPACE` is not unique to "this checkout" — it will match any leftover proxy from a *prior* job on the same runner too, which is exactly what happened in this PR's own CI run. That's arguably a feature (it cleans up historical leaks), but the comment/PR-body framing as job-scoped isn't accurate for how this runner is set up. Just a documentation nit, not a functional problem.

3. **Reuse/simplification: duplicate teardown logic.** The normal-path kill at `.github/workflows/live-test.yml:207-210` (`kill -TERM -- "-$PROXY_PID" ... || true; wait ...`) and the pidfile-based kill in the new `always()` step at lines 228-238 do overlapping work (both try to TERM the same process). Since the `always()` step runs immediately after regardless of outcome and already has the more complete retry+escalate-to-KILL logic, the inline kill/wait in the probe step could arguably be dropped entirely and teardown left solely to the always() step, saving a few lines without changing behavior. Low priority — the duplication is harmless.

**What's good:** the root-cause diagnosis (bash function backgrounded via `&` creates a subshell that owns `$!`, not the launched command) is correct and I confirmed it reproduces; `setsid` is the right fix and I verified it makes `$!` resolve to the real process with its own pgid. The narrowing of the reap sweep (must have `ppid=1` *and* cwd under `$GITHUB_WORKSPACE`) is a reasonable guard against clobbering the production proxy or another job's process. CI is green across all checks including `live-test` itself.

sprayberry-redline (approved):

Verdict: approved — the live head fixes the prior teardown blocker, with no new blocking issues found.

I reviewed the complete `.github/workflows/live-test.yml` diff and the focused `201046c…1d4705f` re-review delta. The new `stop()` helper now applies bounded TERM/wait/KILL/wait cleanup to both the pid-file process and checkout-scoped orphan sweep, only reports an orphan as reaped after it is gone, tracks survivors, and makes the teardown step fail if any survive. This directly resolves my prior finding.

What's good: the cleanup remains narrowly scoped by process pattern, init parent, and checkout cwd; normal and `always()` teardown paths cover the process group; and all current CI checks pass. The live-test job also exercised this exact head's teardown and reported `no orphaned proxy processes from this checkout`.

PR #1245, opened 2026-09-07
A seat parked on a 429 says so: reset, rejection count, reading age, a log line; a re-grant starts the seat fresh (#1244)

askalf:
Fixes #1244.

**What the reporter saw** (v6.0.32, six-seat pool, `GET /admin/accounts`): one seat at `util5h: 1.04, claim: five_hour, status: rejected, request_count: 0` while the usage page for that account read 0%; a re-login cleared it. Reproduced in-process with a fake upstream answering that seat's first request with 429 at `5h 1.04 / 7d 0.25 / five_hour` and a peer answering 200: the listing came out field-for-field identical, the client saw 200, and the proxy log — with `-v` — had no line about the seat.

**Why it read as a phantom.** Every field was true: Anthropic had answered 429 on that token at 104% of its five-hour window. But

1. `markRejected` did not count the attempt — only served requests bump `requestCount` — so the parked seat read "0 requests";
2. neither listing carried the `reset` the snapshot has held since the header was first parsed (routing has expired rejections on it since #1232), so `rejected` never said until when;
3. `GET /admin/accounts` dropped the `lastObservedAt` / `utilAgeMs` freshness that #1032 added — the adapter in proxy.ts supplied them, the mapper in admin-api.ts never emitted them;
4. with a peer to fail over to, nothing was logged.

**Change.**

- `pool.ts`: `PoolAccount.rejectedCount` / `lastRejectedAt`; `markRejected` counts the attempt and returns whether the seat was newly parked; `rateLimitWindow()` (`resetAt` / `resetInMs`, milliseconds like `expiresInMs`) and `describeRateLimitSnapshot()` (the log line).
- `proxy.ts`: `GET /accounts` gains `resetAt`, `resetInMs`, `rejectedCount`, `lastRejectedAt`; the admin adapter passes the same; the 429 site logs `rate limited (429) on account "busy": 5h 104%, 7d 25%, claim five_hour, resets in 37m — parked until the window rolls` on the transition. The re-probes the all-exhausted fallback makes of an already-parked seat stay `-v` (that path 429s the same seat once per request while nothing is left).
- `admin-api.ts`: `GET /admin/accounts` gains `last_observed_at`, `util_age_ms`, `reset_at`, `reset_in_ms`, `rejected_count`, `last_rejected_at`.
- `tui/tabs/accounts.ts`: the status column reads `rejected 37m`.
[...]

PR #1248, opened 2026-09-07
Seats report their organization and which other seats share their window; `dario accounts list --live` (#1244)

askalf:
Follow-up to #1244, stacked on #1245 (the base of this PR is `fix/account-status-explains-itself`; it retargets to master when #1245 merges, and its own delta is the commits after `faeac4b`). No version bump: #1245 holds the 6.0.33 slot; the release note sits under `## [Unreleased]`.

**The gap.** The #1244 listing had a usage page reading 0% for a seat whose responses said 104%, and "a few" seats showing the same thing. Both are what one subscription under several aliases looks like from the outside: the token behind an alias belongs to an organization other than the one on screen, and two aliases granted from one subscription report one window that the pool counts twice — the busier alias fills it for both, the other parks on its first request with the same reading. Nothing in dario could show either fact.

**Change.**

- Every seat carries `organizationId`: the `anthropic-organization-id` its responses carry (verified live: it is on every authenticated response, alongside `anthropic-workspace-id`), learned on the first response and written to the seat's record by its next token refresh. That last part is a deliberate trade-off — see below.
- Every seat carries `sharesWindowWith`: the other aliases whose last reading names the same live window (same representative claim, same reset second). Two independent windows all but never share a reset second; two readings of one window always do. The organization id is deliberately not the key — seats on one organization can still have their own windows — so the row is a fact about headroom, not a guess about plans. `GET /accounts` adds `distinctWindows` (each measured window once, each unmeasured seat as its own).
- `GET /admin/accounts` carries `organization_id` (from the record even without a live pool entry; the live value wins when there is one) and `shares_window_with`.
- The proxy says it once per pair, when the second reading arrives: `seats "twin" and "busy" report the same five_hour window (resets 2026-09-07T13:12:00.000Z) — one subscription under two aliases; the pool has 2 distinct windows across 3 seats`.
- `dario doctor` gains an `Organizations` row from the ids on the records: `3 seats on 2 organizations — busy + twin share 927b430e…`, worded as "may be one subscription" and pointing at `sharesWindowWith` rather than asserting a shared limit.
- `dario accounts list --live`: the running proxy's view of the pool — status with countdown, reading and age, requests served and 429s answered, organization, shared windows, grant age. The plain listing only knows what is on disk; the TUI already read the proxy, the CLI did not. Falls back to the on-disk listing with a note when no proxy answers.
- A re-grant under an alias forgets the old organization (it is a new credential); a same-grant reconcile keeps what was learned, and a record's own value is taken when it has one.

**Trade-off surfaced: when the organization reaches disk.** The first cut wrote the seat record from the request path the moment the header was seen. I took that out: it is a read-modify-write of a credential file with no serialization against the refresh writer, and the losing interleaving (the request path loads the record, the refresh writes rotated tokens, the request path writes the stale record back) puts a burned refresh token on disk — the dario#790 class of failure, with a probability that is small and a cost that is the credential family. The observation now rides the refresh's own write (`withObservedOrganization(saved, acc.organizationId)` at both refresh sites), so no new write exists to race. Cost: `dario doctor` and a restarted proxy learn the organization up to one refresh (≤8h) after the running proxy does; `/accounts` and `--live` have it at once. If you would rather have it on disk immediately, the right shape is a per-alias write lock in accounts.ts around every seat-file write, which is a larger change than this PR wants to carry.

**Tests.** `pool-window-peers.mjs` (window identity, peers both ways, one-second-apart resets are different windows, same organization with different windows is not a pair, peers expire with their window, distinct-window counting, `noteOrganization`, the refresh-write helper). `doctor-organizations.mjs` (the row's four states). `pool-org-window-surfaces.mjs` drives the real proxy: three seats on a fake upstream stamping organization ids and windows, then `/accounts`, `/admin/accounts`, the one-time log line (and that more traffic does not repeat it), that serving never writes the record, a record that already states its organization loading with it through an admin reconcile, and `dario accounts list --live` run as a real child process (plus its no-proxy fallback). `admin-api.mjs` covers the snake_case fields and the persisted-beats-null merge. Full suite: 182/182 (`npm test`).

Additive: no existing field changes meaning or type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ checklist.
 
 ## [Unreleased]
 
+### Added
+
+- **Each seat reports the organization its token belongs to, and which other seats share its window (#1244).** The report behind #1244 had a usage page reading 0% for a seat whose responses said 104% — a token on an organization other than the one on screen — and "a few" seats showing the same thing, which is what one subscription under several aliases looks like: the pool counts one window several times, the busiest alias fills it for all of them, and the others park on their first request. Every seat now carries `organizationId` (the `anthropic-organization-id` its responses carry, learned on the first response and written to the seat's record by its next token refresh — the write that already exists, so nothing new races it — after which `dario doctor` and a restarted proxy know it without a request) and `sharesWindowWith` (the other aliases whose last reading names the same live window — same representative claim, same reset second; two independent windows all but never share a reset second, and the window rather than the organization is what seats can share). `GET /accounts` adds `distinctWindows`; `GET /admin/accounts` carries `organization_id` / `shares_window_with`; `dario doctor` gains an `Organizations` row; the proxy says it once when the second reading arrives. `dario accounts list --live` prints the running proxy's view of the pool — status and countdown, reading and age, requests served and 429s answered, organization, shared windows, grant age — where the plain listing only knows what is on disk.
+
 ## [6.0.33] - 2026-09-07
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.33] - 2026-09-07
+
+### Fixed
+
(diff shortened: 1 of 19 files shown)


askalf:
Superseded by #1250, which carries this branch's commits: the three stacked PRs are collapsed into one so they take one review cycle instead of three sequential ones. Nothing dropped.

sprayberry-redline (changes requested):

Verdict: Changes requested — `accounts list --live` crashes instead of degrading gracefully when the installed CLI reaches an older running proxy.

### Medium — `src/cli.ts:967`: tolerate the pre-feature `/accounts` schema

> `...(s.sharesWindowWith.length > 0 ? [`shares its window with ${s.sharesWindowWith.join(', ')}`] : []),`

The response is accepted based only on `mode === 'pool'` and `accounts` being an array. During a normal upgrade, the newly installed CLI can query a still-running proxy from the previous release; that proxy's account objects do not have the newly added `sharesWindowWith` field. The access to `.length` then throws `TypeError: Cannot read properties of undefined`, so the command neither displays live data nor performs its advertised on-disk fallback. Treat additive response fields as optional at this client boundary (and ideally validate/default the whole payload), and add a test using a legacy `/accounts` payload.

```suggestion
      ...(Array.isArray(s.sharesWindowWith) && s.sharesWindowWith.length > 0
        ? [`shares its window with ${s.sharesWindowWith.join(', ')}`]
        : []),
```

What's good: the pool/window derivation is compact, the credential-write race is deliberately avoided, and the new behavior has strong unit and end-to-end coverage. All reported CI checks are green. I reviewed the complete 698-line feature delta after the PR's documented stacked base; I did not run the suite locally.

PR #1249, opened 2026-09-07
Requests name their consumer: per-consumer attribution and an optional per-consumer concurrency cap (#1244 follow-up)

askalf:
Follow-up to #1244, stacked on #1248 (base: `feat/seat-organization-and-shared-window`; retargets when that merges; this PR's own delta is the commits after `dca06d7`). No version bump: #1245 holds the 6.0.33 slot; the release note sits under `## [Unreleased]`.

**The gap.** The #1244 pool is a team gateway: six seats, Claude Code → LiteLLM → dario, one `DARIO_API_KEY`. Nothing in dario said whose traffic went where — analytics and the log are per seat only — and the request queue is plain FIFO, so one heavy user could hold every concurrency slot and every other user's next turn waited behind that backlog.

**Change.**

- **Consumer identity.** `x-dario-consumer: <name>` (one printable-ASCII token, ≤64 chars, no spaces; anything else is treated as absent rather than becoming an analytics key). Without the header, attribution falls back to a hash of the body's user id: the Anthropic `metadata.user_id` with Claude Code's `_session_<uuid>` suffix dropped so one person is one key across sessions, or the OpenAI `user` field. The key is `u_` + 12 hex — no account id or raw user id is stored. Pure helpers `consumerFromHeader` / `consumerFromBody` in analytics.ts.
- **Attribution.** `RequestRecord.consumer`; `GET /analytics` gains `perConsumer` (requests, tokens, cache share, estimated cost, the seats the consumer landed on, last model) beside `perAccount`; every per-request log line and the `-v` usage line carry `consumer`; the TUI's Hits tab shows it on the selected request. Records without a consumer are in no bucket; per-account totals are unchanged.
- **Fairness.** `--max-concurrent-per-consumer=N` / `DARIO_MAX_CONCURRENT_PER_CONSUMER` (default 0 = off). `RequestQueue.acquire(consumer)` / `release(consumer)`: a consumer at its cap waits in the queue even while slots are free; `release` admits the first waiter whose consumer is under its cap (FIFO among the admissible), so one user's backlog never holds up another user's next turn. Unnamed requests are never capped. The pure gate is `decideConsumerAdmit`; `decideAdmit` is untouched. The queue snapshot on `/analytics` and `/health` reports `maxConcurrentPerConsumer` and `consumersActive`.

**Trade-offs surfaced.**

- Only the header paces. The body is parsed after the concurrency slot is taken (that ordering is deliberate: the slot bounds the body read), so the body-derived key can attribute but cannot gate. For LiteLLM setups that means setting the header per key/user; the docs say so.
- The config file's `queue` block (`maxConcurrent` / `maxQueued` / `timeoutMs`) is parsed and defaulted in config-file.ts but never applied anywhere in cli.ts or proxy.ts — only the flags and env vars are. I did not add the new knob there or wire the block; that is a pre-existing, undocumented gap and deserves its own PR (or removal). Flagging it rather than quietly fixing it here.
- `perConsumer` keys are whatever the header says. A gateway that lets end users set `x-dario-consumer` themselves lets them pick their own bucket; the intended place to set it is the thing in front of dario.

**Tests.** `request-queue.mjs`: the pure gate, a capped consumer waiting while another is admitted past its waiter, a peer's release not admitting it, its own release admitting it, FIFO among the admissible, unnamed never capped, cap-off unchanged. `analytics-recording.mjs`: header validation (trim, repeated header, spaces, non-ASCII, length), body derivation (same person across two Claude Code sessions is one key, account id never in the key, OpenAI `user`, precedence, absent), `perConsumer` aggregation. New `consumer-attribution-proxy.mjs` drives the real proxy with `maxConcurrentPerConsumer: 1` and a held fake upstream: attribution by header, by body (two sessions → one hashed key), and none; the log file and `-v` line; then two alice requests plus one bob request — bob starts while alice's first is in flight, alice's second starts only after her first finishes, never more than two upstream calls at once; and three unnamed requests run concurrently. Full suite: 183/183 (`npm test`).

Additive: no existing field changes meaning or type; the cap is off unless set.


askalf:
Superseded by #1250, which carries this branch's commits: the three stacked PRs are collapsed into one so they take one review cycle instead of three sequential ones. Nothing dropped.

PR #1246, opened 2026-09-07
chore(npm): describe both plans in the package listing

askalf:
The npm page is what the repo homepage links to, and it still describes dario as Claude-only. Align the package description and keywords with the README's two-plan story.

- description: Claude and ChatGPT, either wire shape, automatic failover
- keywords: add chatgpt, codex, claude-code, cursor, cline, aider, agent-sdk, llm-proxy; drop the generic ai / api

No version bump on purpose. Lands on npm with the next release slot.

PR #1250, opened 2026-09-07
Team gateway: seat organization and shared windows, consumer attribution and fairness, shared pool state across instances — v6.0.34 (#1244 / #993 follow-ups)

askalf:
Addresses #1244 (the team-gateway half) and #993 (the two items multi-instance.md listed as "not solved"). One PR, one review: this branch carries what #1248, #1249 and #1250 were opened as, collapsed so it takes one review cycle instead of three sequential ones; #1248 and #1249 are closed as superseded. It also carries the same npm listing text #1246 merged separately (identical lines, so the merge is clean). **Version bumped to 6.0.34 in this PR, so it releases on merge.** Nothing here closes an issue.

The reporter's setup is a team gateway: six seats, Claude Code → LiteLLM → dario on Kubernetes with two pods. Three things dario could not tell them, in order of how they were hit.

## 1. Seats report their organization and which other seats share their window

**The gap.** The #1244 listing had a usage page reading 0% for a seat whose responses said 104% (a token on an organization other than the one on screen), and "a few" seats showing the same thing — one subscription under several aliases: the pool counts one window several times, the busiest alias fills it for all, the others park on their first request.

**Change.** Every seat carries `organizationId` (the `anthropic-organization-id` its responses carry — verified live on every authenticated response — learned on the first response, written to the record by the next token refresh) and `sharesWindowWith` (aliases whose last reading names the same live window: same representative claim, same reset second; two independent windows all but never share a reset second, and the window rather than the organization is what seats can share — seats on one organization can still have their own windows). `GET /accounts` adds `distinctWindows`; `GET /admin/accounts` adds `organization_id` (record or live) and `shares_window_with`; the proxy says it once per pair when the second reading arrives; `dario doctor` gains an `Organizations` row worded as "may be one subscription" and pointing at `sharesWindowWith`; `dario accounts list --live` prints the running proxy's view (status with countdown, reading and age, requests served and 429s answered, organization, shared windows, grant age) where the plain listing only knows what is on disk. A re-grant forgets the old organization.

**Trade-off surfaced.** The first cut wrote the seat record from the request path the moment the header was seen. That is a read-modify-write of a credential file with no serialization against the refresh writer, and the losing interleaving puts a burned refresh token back on disk (the dario#790 class, small probability, cost is the credential family). The observation now rides the refresh's own write (`withObservedOrganization` at both refresh sites): `dario doctor` and a restarted proxy learn it up to one refresh later; `/accounts` and `--live` have it at once. Immediate persistence would need a per-alias write lock around every seat-file write, which this PR does not carry.

## 2. Requests name their consumer; attribution and an optional per-consumer cap

**The gap.** One `DARIO_API_KEY` for a team, analytics and logs per seat only, and a FIFO queue where one heavy user could hold every slot.

**Change.** `x-dario-consumer: <name>` (one printable-ASCII token, ≤64 chars; anything else treated as absent). Without it, attribution falls back to a hash of the body's user id — Claude Code's `metadata.user_id` with its `_session_<uuid>` suffix dropped so one person is one key across sessions, or the OpenAI `user` field; the key is `u_` + 12 hex, no account id or raw id stored. `GET /analytics` gains `perConsumer`; per-request log lines and the `-v` usage line carry `consumer`; the TUI's Hits tab shows it. `--max-concurrent-per-consumer=N` / `DARIO_MAX_CONCURRENT_PER_CONSUMER` (default 0 = off): a consumer at its cap waits while slots stay free for others; `release` admits the first waiter whose consumer is under its cap (FIFO among the admissible); unnamed requests are never capped. The queue snapshot reports `maxConcurrentPerConsumer` and `consumersActive`.

**Trade-offs surfaced.** Only the header paces — the body is parsed after the slot is taken (deliberately: the slot bounds the body read), so the body-derived key attributes but cannot gate; LiteLLM setups set the header per key. `perConsumer` keys are whatever the header says; the place to set it is the thing in front of dario. Found but not fixed here: the config file's `queue` block is parsed and defaulted in config-file.ts and applied nowhere — a pre-existing, undocumented gap that deserves its own PR or removal.

## 3. Shared pool state across instances

**The gap.** Two replicas behind one Service each kept their own rate-limit accounting and sticky bindings: each believed a seat had full headroom, each ate its own 429 to learn otherwise, and a conversation re-cached its prompt prefix on a different seat per replica.

**Change.** `--pool-shared-state` / `DARIO_POOL_SHARED_STATE=1` (needs `DARIO_REFRESH_LOCK_URL`; says so and runs unshared otherwise), `--pool-shared-state-interval=MS`. `src/pool-sync.ts` reports every reading this instance takes (fire-and-forget, coalesced per seat), pulls peers' readings every 2s and adopts any strictly newer (`shouldAdopt`, pure), and exchanges sticky bindings: consulted before `selectSticky` only when this instance holds no binding, published on bind and on failover rebinding. `pool.adoptSnapshot` replaces the reading and parks the seat when the peer holds it parked; counters stay each instance's own; `readingFrom` / `reading_from` say whose reading a seat carries; a peer parking is logged like an own 429. `GET /accounts` gains a `sharedState` block. Both reference backends (`redis-lock/server.mjs`: one hash, one `HGETALL` per pull; `cloudflare/refresh-lock/worker.mjs`: a single `__pool__` Durable Object) serve `/pool/seat/<alias>`, `/pool/seats`, `/pool/sticky/<key>/{bind,get}` under the same bearer. Fails open like the lock: one log line per outage, local state untouched. Docs: multi-instance.md §2/§3 rewritten, a "Turning shared pool state on" section, and configuration.md gains a Multi-instance table (the lock variables were undocumented there).

**Trade-offs surfaced.** Eventually consistent — a pull interval behind, never ahead; two instances can still both send a seat's first request in the same two seconds. One extra round trip on the first turn of a conversation an instance has not seen (bounded by the 3s call timeout, fails open). One small POST per upstream response, coalesced. **Not exercised against a live Redis or a deployed Worker** — no Docker or Redis on this machine; the Redis routes use `SET … PX` / `GET` (already in use) plus `HSET` / `HGETALL` (arrays are already parsed for `EVAL`), both files pass `node --check`, and the contract is tested end to end against an in-memory stub that implements the same three routes with the same validation. `npm run test:refresh-lock-race`'s Redis against this branch's server is the right live check before relying on it.

## 4. npm listing describes both plans (same text as #1246)

The package `description` and `keywords` now match the README's two-plan story: Claude and ChatGPT, either wire shape, automatic failover; keywords add chatgpt, codex, claude-code, cursor, cline, aider, agent-sdk, llm-proxy and drop the generic ai / api. #1246 merged on its own while this was being assembled; the identical lines are here too so the release bump and the listing never conflict, and the CHANGELOG records the listing under 6.0.34, the release that puts it on npm.

## Tests

New: `pool-window-peers.mjs`, `doctor-organizations.mjs`, `pool-org-window-surfaces.mjs` (real proxy, three seats, `accounts list --live` as a child process), `consumer-attribution-proxy.mjs` (real proxy, held upstream, cap of 1), `pool-sync.mjs`, `pool-shared-state-proxy.mjs` (two real proxies against the stub). Extended: `pool-rate-limit-reset.mjs`, `admin-api.mjs`, `tui-accounts-source.mjs`, `request-queue.mjs`, `analytics-recording.mjs`. Full suite on this head: 185/185 (`npm test`); the release commit changes package.json, package-lock.json and CHANGELOG.md only.

Additive: no existing field changes meaning or type; the cap and the shared state are off unless set.


sprayberry-redline (changes requested):

Verdict: changes requested because out-of-order reports can replace the newest shared seat state with an older snapshot.

### High — `redis-lock/server.mjs:106-112` and `cloudflare/refresh-lock/worker.mjs:116-120`: seat reports overwrite newer readings unconditionally

> `if (kind.startsWith('seat/')) {`
> `  const alias = decodeURIComponent(seatAlias);`
> `  ...`
> `  await redis.send('HSET', SEATS_KEY, alias, JSON.stringify({ instance: body.instance, at: body.at, snapshot: body.snapshot, rejected: body.rejected === true }));`

The Worker route has the same behavior with an unconditional `storage.put`. The client serializes reports only within one instance, but reports for the same alias from different instances can arrive out of order. For example, instance A takes a successful reading at `at=100` and its POST stalls; instance B then receives a 429 at `at=200` and stores `rejected: true`; when A's older POST resumes, this code overwrites the hash/object with `at=100, rejected: false`. A third replica pulling afterward sees only the stale non-rejected state and can route traffic to a seat that should be parked. Because reports are event-driven rather than periodic, the stale overwrite can remain until another upstream response occurs. This contradicts the endpoint's “latest reading” contract and defeats the shared-state behavior in the exact concurrent-replica scenario it is meant to handle.

Apply an atomic compare-and-set in both backends so a seat field is updated only when the incoming `at` is strictly newer than the stored `at` (a Redis Lua script for the hash, and a read/compare/write in the single-threaded Durable Object), and add a contract test that submits `at=200` followed by `at=100` and verifies the rejected `at=200` record remains.

```suggestion
// Pseudocode for both backend implementations:
const current = await readSeat(alias);
if (!current || body.at > current.at) {
  await writeSeat(alias, {
    instance: body.instance,
    at: body.at,
    snapshot: body.snapshot,
    rejected: body.rejected === true,
  });
}
```

What's good: the client-side adoption guard is strict and age-bounded, the per-consumer queue admits the first eligible waiter without violating its cap, and the new behavior has broad focused test coverage. Most CI checks are green; `wire-drift`, `compat`, and `live-test` were still pending when reviewed.

askalf:
Addressed the high finding in 4787820: seat reports are now compare-and-set on `at` in both backends — a Redis EVAL script (HSET only when the incoming timestamp is strictly newer) and a read-compare-write under `blockConcurrencyWhile` in the Worker — with `{ok, stored}` in the reply. The in-memory stub mirrors the contract, and `pool-sync.mjs` submits the exact scenario (t=200 rejected, then t=100 allowed) and checks the t=200 record stands, an equal timestamp does not replace it, a fresher one does, and a third instance pulling afterwards adopts the newest. No client change was needed: `shouldAdopt` already guarded the pull side; this closes the write side.

sprayberry-redline (approved):

Verdict: approved — the prior out-of-order shared-state overwrite is fixed on the live head, with focused regression coverage and green CI.

I reviewed the complete one-commit delta from the prior reviewed head. Both production backends now accept a seat report only when its `at` value is strictly newer: Redis performs the comparison and hash update atomically in Lua, while the single Durable Object serializes its read/compare/write with `blockConcurrencyWhile`. Equal timestamps are also retained rather than nondeterministically replacing state. The test stub follows the same contract, and the new test exercises newer-then-older reports, equal timestamps, a subsequently fresher report, and adoption by a third instance.

What's good: the fix directly covers the concrete stalled-POST race from the previous review without widening the change, and all reported CI checks are green.

PR #1252, opened 2026-09-07
docs: name the real PR-time gate (live-test is the required full-suite run; compat is the passthrough replay)

askalf:
Docs only. The README said wire-shape changes are guarded by a "PR-time compat gate that runs the full suite against a live proxy". Checked against the workflows and the ruleset: the full suite against a live proxy is `live-test.yml` (`npm test` inside a live-proxy stage, and a required check), while `compat-test-self-hosted.yml` runs only `test/compat.mjs` (~11 requests through `dario proxy --passthrough`) and is not required. The sentence now names both correctly. `docs/drift-monitor.md` listed the auto-rebake gate as build ×3 / compat / test / docker-cap-drop-smoke, which predates live-test becoming the required check; it now lists the ruleset's required checks. The table rows below the sentence were verified against #271, #273 and #478 and are unchanged.


sprayberry-redline (approved):

Classified low-risk: docs-only (2 file(s)).

PR #1254, opened 2026-09-07
A pool with every seat parked answers 429 locally instead of re-probing a parked seat on every request (#1244)

askalf:
Follow-up to #1244. The reporter's second listing (v6.0.33, six-seat team gateway) read:

```
"util5h": 1, "reset_in_ms": 1288011, "status": "rejected",
"request_count": 1, "rejected_count": 500
```

and the question was "do I have to re-login?" No: that is 500 requests that arrived while every seat was over its window, each one re-probing this seat.

**Cause.** `select()`'s all-exhausted branch returned the seat with the earliest reset and the request went upstream — a guaranteed 429 — and `selectExcluding()` then offered every other parked seat mid-flight. One request could spend six upstream 429s, bump six `rejectedCount`s, and the client still waited on a verdict dario already held.

**Fix.** A seat parked inside a live window (`rejected`, `reset` in the future) is never re-probed:

- `select()` returns null; `pool.parkedUntil()` names the earliest reset. The proxy answers `429` with `retry-after` at that reset, `x-dario-upstream-rejection: pool_parked`, nothing upstream, cools the Claude provider to the reset so an armed `--pool-fallback` chain sees the state, and logs the transition once (`pool parked: all 6 seats are over their rate-limit windows, earliest resets in 21m`).
- `selectExcluding()` no longer hands a parked seat to mid-flight failover.
- A rejection with **no stated reset** stays probeable, as before: with nothing to expire, asking is its only way back. A parked seat still returns on its own when its window rolls (6.0.30).

**Added.** `action` on `GET /accounts` and `GET /admin/accounts` next to `status`: `none` · `wait` (live window, or a single auth blip) · `regrant` (auth-failure streak). `dario accounts list --live` prints it as `next: …`.

**Tests.** New `test/pool-parked-local-429.mjs`: pure pool (parked vs probeable, `parkedUntil`, `selectExcluding`, `action` for each state) and the real proxy in-process with two seats answering 429 — 21 requests, exactly 2 upstream calls, `rejectedCount` frozen at 1, `retry-after` ≈ 21 min, the marker, one log line, `action: wait` on both listings. `pool-e2e.mjs` and `pool-parked-seat-surfaces.mjs` asserted the re-probe and now assert its absence. Full suite 186/186.

Version 6.0.35; CHANGELOG and `docs/multi-account-pool.md` updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.35] - 2026-09-07
+
+### Fixed
+- **A pool with every seat parked answers 429 locally instead of re-probing a parked seat on every request (#1244 follow-up).** When no seat was eligible, `select()` handed back the seat with the earliest reset and the request went upstream anyway — one guaranteed 429 per request, and mid-flight failover then walked every other parked seat too. On the #1244 gateway that put `rejected_count: 500` next to `request_count: 1` on one seat inside a single window, which read as a seat that needed a re-login. A seat parked inside a live window is now never re-probed: the request is answered with `429`, `retry-after` at the earliest reset and `x-dario-upstream-rejection: pool_parked`, nothing goes upstream, the Claude provider is cooled to that reset so an armed fallback chain sees the state, and one log line marks the transition. A rejection with no stated reset stays probeable, since asking is its only way back; a parked seat returns on its own when its window rolls, as since 6.0.30.
+
+### Added
+- **`action` on both account listings (#1244).** `GET /accounts` and `GET /admin/accounts` carry the operator's next step next to `status`: `none`, `wait` (a live rate-limit window, or a single auth blip — the seat comes back on its own) or `regrant` (an auth-failure streak, a dead refresh token). "Do I have to re-login?" is now one field. `dario accounts list --live` prints it as a `next:` fact on the seat.
+
 ## [6.0.34] - 2026-09-07
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 
 <p><strong>One local endpoint. Every AI tool you own. The subscriptions you already pay for.</strong></p>
 
-<sub><code>npm i -g @askalf/dario</code> · <strong>0</strong> runtime deps · <a href="https://www.npmjs.com/package/@askalf/dario">SLSA-attested</a> every release · nothing phones home · ~31k lines you can read in a weekend · independent, unofficial, third-party (<a href="DISCLAIMER.md">DISCLAIMER.md</a>)</sub>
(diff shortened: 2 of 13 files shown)


sprayberry-redline (changes requested):

Verdict: changes requested — mixed auth-cooldown/rate-limit pools are incorrectly presented and cooled as fully rate-limit parked.

### High — `src/pool.ts:752-759`: `parkedUntil()` treats auth-cooling seats as though they share the rate-limit parking window

> ```ts
> parkedUntil(now: number = Date.now()): number | null {
>   if (this.accounts.size === 0) return null;
>   const all = [...this.accounts.values()];
>   if (all.some(a => isAccountEligible(a, now))) return null;
>   if (all.some(a => !isInAuthCooldown(a, now) && !isParkedInLiveWindow(a, now))) return null;
>   const parked = all.filter(a => isParkedInLiveWindow(a, now));
>   if (parked.length === 0) return null;
>   return Math.min(...parked.map(a => a.rateLimit.reset * 1000));
> }
> ```

Concrete failure: take a two-seat pool where seat A has a live 30-minute rate-limit window and seat B has one auth failure with a 60-second cooldown. Neither seat is currently eligible, and neither satisfies the second `some(...)`, so this returns A's reset even though B is not rate-limit parked and becomes usable in one minute. The changed proxy path then emits “all 2 seats are over their rate-limit windows,” sends `Retry-After: 1800`, and calls `providerCooldowns.note('claude', ...)` for 30 minutes. That gives clients and an armed provider-fallback chain a materially incorrect availability time and can keep traffic off Claude long after B's auth cooldown expires.

Make this helper represent the condition its callers and response claim: every unavailable seat must actually be parked in a live rate-limit window. Mixed auth-cooldown pools should remain on the existing unavailable/all-provider handling (or use the true earliest availability across both state types, with a correspondingly accurate response).

```suggestion
  parkedUntil(now: number = Date.now()): number | null {
    if (this.accounts.size === 0) return null;
    const all = [...this.accounts.values()];
    if (!all.every(a => isParkedInLiveWindow(a, now))) return null;
    return Math.min(...all.map(a => a.rateLimit.reset * 1000));
  }
```

Please add a regression case with one live-window rejection and one auth-cooling seat, asserting that it does not produce `pool_parked`, the all-seats-rate-limited message, or a provider cooldown lasting until the unrelated rate-limit reset.

What's good: the core all-rate-limited behavior is clearly factored, the local 429 includes useful machine-readable retry metadata, parked seats are excluded consistently from initial and mid-flight selection, and the full all-parked proxy path has strong counter/upstream-call coverage. All reported CI checks are green.

sprayberry-redline (approved):

Verdict: approved — the live head fixes the mixed-pool blocker, and I found no remaining blocking issues in the changed behavior.

I re-reviewed the complete PR diff and specifically traced the new all-parked selection path, local 429 response and provider cooldown, mid-flight exclusion, account-listing action field, and associated regression coverage. The previous blocker is resolved at `src/pool.ts:754-758`: `parkedUntil()` now returns a reset only when every seat is parked in a live rate-limit window, so a pool containing an auth-cooling or token-expired seat no longer emits `pool_parked` or applies the unrelated rate-limit cooldown. The added mixed-state tests cover both cases.

What's good: the parked-seat predicate is consistently reused by initial selection, failover selection, status/action reporting, and reset calculation; the local response supplies both `retry-after` and a machine-readable rejection marker; and CI is fully green, including test, compatibility, CodeQL, and live-test checks.

PR #1256, opened 2026-09-07
accounts list --live: survive a pre-upgrade proxy payload (v6.0.36)

askalf:
Applies the gating review fix from #1248 that did not travel when the feature re-landed as #1250.

## The defect

`src/cli.ts` accepts the `/accounts` response on `mode === 'pool'` plus `accounts` being an array — nothing validates the seats. During a normal upgrade the newly installed CLI queries a **still-running proxy from the previous release**, whose seat objects predate the additive `sharesWindowWith` field, and:

```
...(s.sharesWindowWith.length > 0 ? [...] : [])
```

throws `TypeError: Cannot read properties of undefined`. The command then neither renders live data nor takes the on-disk listing it advertises as its fallback.

## The fix

The reviewer suggested an `Array.isArray` guard on the one field; this does that and then the thing they asked for underneath it — validates the whole payload at the client boundary rather than one field. Every additive field the team-gateway work introduced carries the same assumption: `organizationId`, `grantedAt` and `action` are all as absent as `sharesWindowWith` on a pre-upgrade seat, as are the counters and utilisation readings.

- `LiveSeat` / `LivePayload` lift to module scope with every field past `alias` optional, which is what the wire actually guarantees.
- The renderer is extracted as pure `formatLiveAccountsListing(payload, port, now)` returning lines instead of printing, so a legacy payload can be driven straight through it.
- Each field is defaulted at read: absent counters render `0`, an absent reading renders `never measured` (not `NaN%`), an absent org keeps the existing `org not yet observed`, and an absent/empty `sharesWindowWith` omits the fact rather than inventing one. A pre-upgrade payload prints a thinner seat line.

## Test plan

New `test/accounts-live-legacy-payload.mjs` (auto-discovered by `test/all.test.mjs`):

- a legacy seat — no `sharesWindowWith`, `organizationId`, `grantedAt` or `action` — renders instead of throwing, and still shows the alias and counts
- the modern payload still renders the shared window, org, next-step and `distinctWindows`, so this is not a silent feature regression
- the degenerate case: a seat carrying only `alias` renders, zeroes its counters and reports `never measured` rather than `NaN`
- an empty pool renders `Pool of 0`

`node scripts/preflight.mjs` clean. Version bumped to 6.0.36 with CHANGELOG re-headed per the release gate; `src/cli.ts` is shipped code.

Source ticket: 00MTRGHYIREA78C17D935F8CD8. Finding: 00MTRGHMLPC23964166520A002.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.36] - 2026-09-07
+
+### Fixed
+- **`dario accounts list --live` no longer crashes against a proxy from the previous release.** The `/accounts` response is accepted on `mode` plus `accounts` being an array, and nothing checked the seats themselves — so during a normal upgrade, where the newly installed CLI queries a still-running older proxy, `sharesWindowWith` was absent and `.length` on it threw `TypeError: Cannot read properties of undefined`. The command then neither rendered live data nor took the on-disk listing it advertises as the fallback. Every additive field the team-gateway work introduced (`sharesWindowWith`, `organizationId`, `grantedAt`, `action`) is now defaulted at that client boundary, along with the counters and the utilisation readings, so a pre-upgrade payload prints a thinner seat line instead of dying. The renderer is split out as a pure function and covered by a legacy-payload test.
+
 ## [6.0.35] - 2026-09-07
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.35",
+  "version": "6.0.36",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.0.35",
+      "version": "6.0.36",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.35",
+  "version": "6.0.36",
   "description": "Use your Claude and ChatGPT subscriptions in Cursor, Cline, Aider, Claude Code and the Agent SDK — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint: either plan answers either wire shape, with automatic failover when one hits its limit.",
   "type": "module",
   "bin": {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -983,7 +983,7 @@ async function accountsListLive(): Promise<boolean> {
       `429s ${s.rejectedCount}`,
       ...(next ? [next] : []),
       s.organizationId ? `org ${s.organizationId.slice(0, 8)}…` : 'org not yet observed',
-      ...(s.sharesWindowWith.length > 0 ? [`shares its window with ${s.sharesWindowWith.join(', ')}`] : []),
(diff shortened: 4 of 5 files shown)


sprayberry-redline (approved):

Verdict: Approved — no blocking issues found.

I reviewed the complete five-file diff at live head `92b4d003625fb2e732c3be5d297a466a55182d42`, including the extracted live-account renderer, legacy-field defaults, version/changelog updates, and the new auto-discovered regression test. The changed path now safely handles absent and non-finite legacy values without weakening the existing `mode === 'pool'` / accounts-array boundary, while preserving modern shared-window, organization, next-step, grant-age, and distinct-window output. All reported CI checks are green.

What's good: the renderer extraction makes the compatibility behavior directly testable, and the test covers both the reported pre-upgrade failure and preservation of current payload features.

sprayberry-secondread (commented):

**Verdict:** No blocking issues — the fix is correct, the defensive defaulting is thorough, and the new test exercises the legacy-payload path it claims to fix.

## What I checked

- Traced the failure this PR claims to fix: with the pre-PR code, `src/cli.ts` destructured `s.sharesWindowWith.length` (and similarly relied on `s.alias`, `s.requestCount`, etc. being present) with no guard, while the payload-level check at `src/cli.ts:1018` (`payload.mode !== 'pool' || !Array.isArray(payload.accounts)`) only validates the envelope, not individual seats. A pre-upgrade proxy's `/accounts` response (confirmed server-side at `src/proxy.ts:2636-2676`, where every additive field like `organizationId`, `sharesWindowWith`, `grantedAt` is only present in the *current* server version) would indeed throw `TypeError: Cannot read properties of undefined` on the old shape. The fix's approach — defaulting every field in `formatLiveAccountsListing` (`src/cli.ts:951-992`) via `typeof`/`Array.isArray` guards — closes that gap correctly.
- Spot-checked the edge cases by hand: `pct(undefined)` → `0%` (not `NaN%`) via the `num()` guard at `src/cli.ts:953`; a `status: 'rejected'` seat with `resetInMs: null` correctly falls through to plain `"rejected"` at `src/cli.ts:973` instead of throwing on `mins(null)`.
- The new test (`test/accounts-live-legacy-payload.mjs`) is auto-discovered by `test/all.test.mjs` (confirmed via the `EXCLUDED` set at `test/all.test.mjs:43-53`, which does not list it) and imports from `dist/cli.js`, matching the pattern already used by 9 other test files in this suite — CI's `npm run build && npm test` ordering (`.github/workflows/ci.yml:94-95`) means the compiled export is available when the test runs.
- All CI checks pass at the reviewed head (92b4d00), including `test`, `build`, `compat`, and `live-test`.

## Minor observation (non-blocking)

The new test covers `distinctWindows` only for the "current payload" case (`test/accounts-live-legacy-payload.mjs:58,62`, passing `distinctWindows: 2` explicitly). The legacy-payload case at line 40 doesn't assert on the "Pool of N (... on M distinct windows)" line, so the `typeof payload.distinctWindows === 'number' ? payload.distinctWindows : seats.length` fallback at `src/cli.ts:967` for a payload that omits `distinctWindows` entirely (as a genuinely old proxy would) isn't exercised. This logic is unchanged from the pre-PR code (previously `payload.distinctWindows ?? seats.length`), so it's not a regression risk from this PR — just a small gap in the otherwise thorough new test file.

## What's good

- The refactor from inline `console.log` calls to a pure `formatLiveAccountsListing` returning `string[]` (`src/cli.ts:951`) is a clean, minimal way to make the legacy path testable without spinning up a real proxy.
- Every optional field on `LiveSeat` is defaulted at the read site rather than relying on a single blanket try/catch, so a legacy payload degrades gracefully field-by-field instead of losing the whole render.
- The PR body correctly identifies and fixes the root cause (payload validated at the envelope but not per-seat) rather than just patching the one field (`sharesWindowWith`) that was reported crashing.

Issue #640, opened 2026-07-02
add version number on any call


matteo-rama:
in any call you like (status or health) it could be useful to have version number of dario, to enable check if it get updated as part of automatic update of the core

askalf:
Done in **[v4.8.118](https://github.com/askalf/dario/releases/tag/v4.8.118)** (#643), on npm now.

Both `/status` and `/health` now report dario's version:

```
curl -s localhost:3456/status  | jq .version   # -> "4.8.118"
curl -s localhost:3456/health  | jq .version   # -> "4.8.118"
```

Exactly your auto-update-check use case — no exec-ing in to read `package.json`. One detail: on `/health` the field is internal-only (withheld from a public Cloudflare-tunnel view, alongside the OAuth internals); `/status` is key-gated so it always carries it. Thanks for the suggestion.

PR #1257, opened 2026-09-07
/status reports the build that is running, not the one on disk

askalf:
Addresses #1244 — specifically the round trip it cost, which was not the seat's fault.

**The bug.** `darioVersion()` read `package.json` lazily on the first call and cached that. `npm i -g` rewrites `package.json` in place under the running install without touching the process, so the reported version could move without a single line of the code answering requests moving with it. A proxy that had not served `/status` before an upgrade answered its first one with the **new** number while still executing the **old** build.

That is the exact opposite of what the field was added for in #640 — *"so a headless operator can confirm an auto-update actually rolled the running proxy — `curl /health | jq .version` beats exec-ing into the container to read package.json."* It confirmed the upgrade had landed at precisely the moment it had not.

**How it surfaced.** On #1244 the reporter pasted `/status` reading `6.0.34` alongside a `GET /admin/accounts` seat with no `organization_id`, no `shares_window_with`, no `reading_from` — all three unconditional on that endpoint since 6.0.34. Running his exact seat numbers through the real handler on master confirms it: the current build emits `"organization_id": null` even for a seat whose organization was never observed. His listing was the 6.0.33 field set, from a process that had been upgraded on disk and never restarted.

The cost was not cosmetic. He had been told to upgrade and read `organization_id` to see which organization his token belonged to; `/status` told him he already had, so the one field that explains his 0%-vs-100% discrepancy stayed out of reach and the thread went another round.

**The fix.** Read at module load. `proxy.ts` is the only importer and imports it statically at startup, so the value is bound to the process and the number moves only when the process does. A missing or malformed `package.json` still reports `unknown` and never throws — unchanged.

**Test.** `test/version-binding.mjs` drives the built `dist/version.js` in a throwaway install root: imported at `6.0.33`, `package.json` rewritten to `6.0.34` underneath it, and the first call must still say `6.0.33`. Plus a fresh process reading its own version, and the missing / malformed cases.

Checked against the unfixed file rather than assumed — `git stash` the fix, rebuild, and the same test fails 2 of 5 and reports `6.0.34`, which is the symptom from the issue verbatim:

```
❌ an upgrade under a running process does not move /status
   :: got 6.0.34 — the lazy read is back, and /status is reporting a build it is not running
❌ later calls stay on the running build
```

**Note on the diagnosis path:** three other places read `package.json` for their own version (`config-report.ts`, `doctor-core.ts`, `subagent.ts`). All three are short-lived CLI paths where process start and first call are the same moment, so none has this failure mode. Left alone rather than unified.

Version is **6.0.37** so this ships. #1256 shipped 6.0.36 while this was open — `accounts list --live` surviving a pre-upgrade proxy payload, the same pre-upgrade class of problem seen from the other side. Its CHANGELOG entry keeps 6.0.36 and this one sits above it; both sections verified through `extract-release-notes.mjs --file`. Nothing but the release heading conflicted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.36] - 2026-09-07
+
+### Fixed
+
+- **`/status` and `/health` report the version the proxy is actually running (#1244 follow-up).** `darioVersion()` read `package.json` lazily on the first call and cached that, so an `npm i -g` under a running proxy — which rewrites `package.json` in place without touching the process — moved the reported version without moving a line of the code answering requests. A proxy that had not served `/status` before the upgrade answered its first one with the **new** number while still executing the **old** build, the exact opposite of what the field was added for (#640: confirm an auto-update actually rolled the running proxy). It is how #1244 lost a round trip: `/status` read `6.0.34` while `GET /admin/accounts` was still emitting the 6.0.33 field set with no `organization_id`, so "upgrade, then read `organization_id`" looked already done and the seat's real diagnosis stayed out of reach. The version is read at module load now, which for the proxy is process start, so the number moves only when the process does. A missing or malformed `package.json` still reports `unknown` and never throws.
+
 ## [6.0.35] - 2026-09-07
 
 ### Fixed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.35",
+  "version": "6.0.36",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.0.35",
+      "version": "6.0.36",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.35",
+  "version": "6.0.36",
   "description": "Use your Claude and ChatGPT subscriptions in Cursor, Cline, Aider, Claude Code and the Agent SDK — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint: either plan answers either wire shape, with automatic failover when one hits its limit.",
   "type": "module",
   "bin": {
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

Reviewed by Claude Opus 4.5 (gating review).

**Verdict: no blocking issues — approving.**

## What I checked
- Full diff (5 files, 133+/13-): `src/version.ts`, `test/version-binding.mjs`, `CHANGELOG.md`, `package.json`, `package-lock.json`.
- Confirmed the module-load binding is correctly wired: `src/proxy.ts:11` does `import { darioVersion } from './version.js'`, a static top-level import, so `readVersion()` at `src/version.ts:38` (`const VERSION = readVersion();`) runs at process start as claimed — the fix actually closes the described race (upgrade-on-disk vs. running process).
- Verified the "left alone" claim in the PR body: `config-report.ts:50`, `doctor-core.ts:531`, and `subagent.ts:180` each do their own independent `readFileSync(...package.json...)` and are untouched by this diff — correctly out of scope since they're short-lived CLI paths where start and first call coincide.
- `readVersion()` (src/version.ts) preserves the original error contract: malformed/missing `package.json` still falls through to `return 'unknown'` inside the try/catch, never throws — same behavior as before, just evaluated once at import instead of lazily.
- Test file `test/version-binding.mjs` is a genuine regression test, not just an assertion of the new code: it imports `dist/version.js` fresh into a throwaway root, rewrites `package.json` underneath the loaded module, and asserts the *old* cached read stays put. PR body states this reproduces the bug (2/5 failures) against the pre-fix file — a real before/after check.
- `CHANGELOG.md`/`package.json`/`package-lock.json` version bumps (6.0.36 → 6.0.37) are consistent with each other.
- CI: all checks green (CodeQL, build 18/20/22, test, live-test, docker-cap-drop-smoke, compat, validate-package-json, etc.).

## Non-blocking notes
- Minor: `CHANGELOG.md:16` also inserts a blank line under the existing `### Fixed` heading for 6.0.36 that wasn't part of this fix's own entry — cosmetic only, doesn't affect rendering or content.

## What's good
The regression test is well-targeted: it isolates the module under test in a throwaway install root so it can rewrite `package.json` out from under an already-imported module, which is exactly the failure mode described. The doc comment on `version.ts` explaining *why* module-load binding (not lazy) is the fix is a good example of a non-obvious-invariant comment.

PR #1255, opened 2026-09-07
ci: a PR may not auto-close an issue somebody else filed

askalf:
Context is #1244 — deliberately linked without a keyword, since that is the whole point.

**What happened.** #1244 was open and mid-conversation, the reporter still asking about his six-seat pool, when #1245 merged with `Fixes #1244.` as its first line. GitHub closed it under him. He read the close as a verdict on questions nobody had answered yet ("i didn't fully undesertand the MR"), and it took a manual reopen plus a comment explaining that the close was a keyword firing, not an answer. `Addresses #N` for an issue you did not file was already the house rule. It was skipped anyway. Advice is not a gate.

**Two halves, and neither relies on the other having worked.**

`scripts/check-issue-close-keywords.mjs` fails a PR whose title or body carries a closing keyword (close / fix / resolve, any tense) pointing at an issue in this repo that a human other than the repo owner filed. Owner-filed issues and the drift watchers' bot-filed alerts stay auto-closable — those exist to be closed by automation. `Addresses #N` and a bare `#N` link as freely as they ever did. It runs as a step in the existing `validate-package-json` job rather than a new job, the same way `check-changelog` does, so it adds no new required status check.

Two details that keep it from crying wolf: code fences and inline code are stripped before matching, because half of #1244's thread is pasted JSON and a `Fixes #12` inside a log paste is not a link; and the gap between keyword and reference is same-line whitespace only, so `…this fixes\n\n#1234 is related` reads as prose about a neighbouring issue.

The title is scanned as well as the body. GitHub honours the keyword in the body only, but a squash merge makes the title the commit subject, and a keyword in a commit message on master closes the issue just the same.

`.github/workflows/keyword-close-notice.yml` repairs what the check can only advise about. `master` carries no branch protection, so a red check is a signal and not a stop — the merge lands and the issue still closes. A keyword can also arrive in a commit message that never appeared in a PR body at all. On `issues: closed`, if a non-owner human filed it and a PR or commit is what closed it, the issue gets one comment naming what closed it and inviting a reopen.

**The close itself stands.** Reopening every keyword-closed report would also reopen the ones that really are fixed, handing the reporter a chore to close what he is already happy with. What went wrong on #1244 was not the state — it was the silence. The close arrived with no explanation and read as a verdict. So the close stays and the issue says what closed it, which is the part that was missing:

> This closed because PullRequest #1245 carried a `Fixes`/`Closes` keyword pointing here, not because anyone decided the report was answered.
>
[...]

PR #1258, opened 2026-09-07
ci: the closing-keyword gate reads commit messages, and reads them raw

askalf:
Addresses #1244 — the second close, the one #1255 caused.

**What happened.** #1255 merged at 17:31:03Z and #1244 closed at 17:31:04Z, closer `Commit 11ed1fe`. That is the squash commit of the PR that added the gate against exactly this. `keyword-close-notice.yml` fired ten seconds later and said so — the repair half working on its first real event, while the advice half had missed the text that mattered.

**Why the gate missed it.** This repo's squash setting is `COMMIT_MESSAGES`: the squash commit's body is the branch's commit messages, not the PR body. My first commit on that branch quoted the closing keyword in backticks while telling the story of the first close. The gate read the PR title and body; it never saw a commit message. And GitHub does not render a commit message — it scans it raw, so the backticks meant nothing.

Two gaps, not one:

1. **Commit messages were never judged.** Every merge strategy carries them: squash under `COMMIT_MESSAGES` concatenates them, a merge commit or rebase lands each one on master. The script now reads `GET /repos/{repo}/pulls/{n}/commits` (all pages) and judges each message alongside the title and body. An unreadable commit list fails closed, the same way an unreadable issue already did after the #1255 review. The job gains `pull-requests: read`; the step passes `PR_NUMBER`. Run by hand without it, the script says it is judging title and body only.

2. **Code spans were stripped.** That modelled how a PR body renders. A commit message is not rendered, and under the `PR_BODY` squash setting the body *is* a commit message. So nothing is stripped: a keyword inside code is a keyword. The failure names which source carried each one — `in body, commit e9bda80, commit 281799c` — and says how to quote a keyword on purpose: keep the word and the number apart, or drop the hash.

**Verified against the real thing, not only the unit test.** With the fix, #1255's own title, body and seven commits fail:

```
FAIL: this PR would auto-close an issue it does not own:
  "fixes 1244" in body, commit e9bda80, commit 281799c — 1244 was filed by @ramarro123
  "fixes 12" in body — 12 was filed by @belangertrading
```

(Hashes dropped from that quote so this body passes its own gate — which is the point.) `e9bda80` and `281799c` are the two lines GitHub actually scanned. The `#12` hit is one I had not seen: a live keyword in the #1255 body, inside backticks, pointing at a third person's issue. It did not fire only because the squash setting kept the body out of the commit. #1256 and #1257 pass as controls.

**Tests.** `test/check-issue-close-keywords.mjs`, 58 assertions, offline, injected issue and commit readers. The first commit message from #1255 is in the file verbatim (72-column wrapped, as squashed) and must fail; a keyword in inline code and in a fenced block must be caught; the two safe quoting forms must pass; `readCommits` is checked across a page boundary and for fail-closed on 403/404/500; `main` must name the commit in its failure and must not request commits when `PR_NUMBER` is unset.

This PR's own title, body and commit message were run through the fixed gate before opening it. `CONTRIBUTING.md` says commit messages count and how to quote safely. No `src/` change, no CHANGELOG entry, no version bump.


sprayberry-redline (approved):

Reviewed by Claude Sonnet 4.5 (gating review).

**Verdict: approve — no blocking issues found.**

## What I checked
- Full diff (`.github/workflows/ci.yml`, `CONTRIBUTING.md`, `scripts/check-issue-close-keywords.mjs`, `test/check-issue-close-keywords.mjs`), 481 lines, read in full.
- CI: all checks green (`gh pr checks 1258`), including `test` (1m2s) and `analyze`/`CodeQL`.
- Ran the PR's own offline unit test file locally (no network calls, injected fetch stubs only): 58/58 assertions pass, matching the green CI `test` job.
- Traced the new logic by hand: `closingRefs`/`closingRefsBySource` dedup-by-issue-number merge (`scripts/check-issue-close-keywords.mjs:71-86`), confirmed a keyword repeated in `title` and `body` correctly merges into one ref with `where: ['title','body']`.
- Verified `readCommits` (`scripts/check-issue-close-keywords.mjs:130-146`) paginates correctly (loops while `batch.length === 100`) and throws — never fails open — on any non-2xx response, mirroring the existing `readIssue` fail-closed contract.
- Verified `main()` (`scripts/check-issue-close-keywords.mjs:148-162`) fails closed (`return 1`) when the commit list can't be read, and that with no `PR_NUMBER` set it falls back to title/body-only judging without calling the commit reader (covered by the "no PR_NUMBER → title and body only, commits never requested" test).
- Confirmed the `pull-requests: read` permission added in `ci.yml:12-24` is the minimum needed for the new `GET /pulls/{n}/commits` call and is read-only.
- Confirmed `scripts/comment-on-keyword-close.mjs` and `.github/workflows/keyword-close-notice.yml` (mentioned in the PR body/CONTRIBUTING.md) are untouched and don't depend on the removed `stripCode` helper, so removing it doesn't silently break that half of the system.

## Findings
None blocking. One minor prose note, not worth a line-level fix request:

- `CONTRIBUTING.md:61` — "Quoting one on purpose? Keep the word and the number apart — `Fixes` … #1244 — or drop the `#`. It and `.github/workflows/keyword-close-notice.yml` comments on one that got closed anyway…" — the "It and ... comments" clause reads as a leftover from an edit (subject/verb mismatch: "It [...] comments" should probably be "It, and `keyword-close-notice.yml`, [comment]"). Cosmetic only.

## What's good
- The fix is well-motivated and directly addresses a real regression (#1255 itself closing #1244 again) with the actual offending commit message reproduced verbatim in the test file as a regression case.
- `readCommits` and the commit-judging path in `main()` both fail closed on API errors, consistent with the existing `readIssue` fail-closed design — no new fail-open path introduced.
- Test coverage for the new behavior is thorough: pagination, all-non-2xx-throws, commit-messages-judged-and-failing, no-PR_NUMBER-fallback, and the exact real-world regression string are all asserted.

sprayberry-secondread (commented):

**Verdict:** No blocking issues — the change is well-reasoned and the test suite exercises the new code paths thoroughly.

## What I checked
- `scripts/check-issue-close-keywords.mjs`: the new `readCommits` pagination loop, `closingRefsBySource`, the removal of `stripCode`, and the fail-closed error handling in `main`.
- `.github/workflows/ci.yml`: the added `pull-requests: read` permission and `PR_NUMBER` env wiring.
- `test/check-issue-close-keywords.mjs`: verified the new assertions actually cover the regression (the `#1255` commit message reproduction, pagination-across-a-page-boundary, and the 403/404/500 fail-closed cases).
- CI status: all checks pass (`gh pr checks 1258`), including `test`, `build (18/20/22)`, and `actionlint`.

## Findings

**Info — `readCommits` has no defense against GitHub's 250-commit cap on `/pulls/{n}/commits`.**
`scripts/check-issue-close-keywords.mjs:127-139`:
```js
export async function readCommits(prNumber, env, fetchFn = fetch) {
  const out = [];
  for (let page = 1; ; page++) {
    const url = `https://api.github.com/repos/${env.GITHUB_REPOSITORY}/pulls/${prNumber}/commits?per_page=100&page=${page}`;
    const res = await fetchFn(url, { headers: apiHeaders(env) });
    ...
```
GitHub's REST API for this endpoint silently caps at 250 commits regardless of pagination — it won't error, it just stops returning more. For a branch with >250 commits, `readCommits` would return an incomplete list and the gate could pass on a closing keyword hiding in a commit past #250, with no diagnostic that truncation occurred. Given this repo's PRs are small (the diff here spans a handful of commits), this is unlikely to bite in practice — flagging only because it's a real, silent gap in a function whose whole design intent is "fail closed rather than miss something."
```suggestion
if (commits.length >= 250) {
  console.error(`FAIL: PR #${prNumber} has ${commits.length}+ commits — GitHub's commit-list API caps at 250 and may be silently truncated. Judging what was returned is not safe here.`);
  return 1;
}
```

**Info — removing `stripCode` trades one false-negative for a new false-positive class.**
`scripts/check-issue-close-keywords.mjs:57-60` (the `closingRefs` diff) drops the old backtick/fence stripping entirely, matching keywords raw. This is the right call for the `#1255` regression (GitHub does scan commit messages raw), but it also means legitimate quoting of a keyword+`#N` pair together in prose or a pasted log — the exact scenario the original `stripCode` was written to protect (per the file's own history: "half of #1244's thread is pasted JSON") — will now fail CI. The PR's own body works around this by deliberately writing `"fixes 1244"` without a `#` to pass its own gate, which is a good sign the author is aware of the trade-off; it's a real behavior change for future contributors pasting logs/examples, mitigated by the clear failure message and escape hatch (`Fixes` … `#N` split, or drop the `#`), so this is informational rather than a bug.

## What's good
- The regression is reproduced concretely in the test suite using the actual `#1255` commit message text, not just a synthetic case.
- Fail-closed behavior is now consistent for both issue lookups and commit-list reads (`main`, `scripts/check-issue-close-keywords.mjs:150-158`).
- `closingRefsBySource` correctly attributes each offending reference to its source (title/body/commit sha) for an actionable failure message, and the test at `test/check-issue-close-keywords.mjs` checks that attribution across sources.
- No `src/` change and no CHANGELOG entry, consistent with this being a CI-only fix, and the PR body confirms this was checked against the repo's own gate before opening.

Issue #1260, opened 2026-09-07
Effort suffix (`model:high`) is refused on the Codex path — routing matches the model name before the suffix is stripped


askalf:
## Summary

The `model:effort` suffix from #419 works on the Claude path and is **refused on the Codex path**. The routing guard matches the model name against each provider's list *before* the effort suffix is stripped, so a Codex model with a suffix looks like a model nobody serves and is rejected `model_unroutable`.

Consequence: **there is no way to set effort for Codex traffic alone.** The only other lever, `--effort` / `DARIO_EFFORT`, is process-wide, so raising effort for a GPT-backed workload also raises it for every Claude-backed one.

## Evidence

Same proxy, same request body, three model names:

```
claude-sonnet-5        -> 200  served=claude-sonnet-5
claude-sonnet-5:high   -> 200  served=claude-sonnet-5     ← suffix stripped, works
gpt-5.6-terra:high     -> 400  model_unroutable           ← refused
```

The rejection body names the cause exactly:

```json
{"type":"invalid_request_error","message":"no provider lists model \"gpt-5.6-terra:high\"
 (consulted codex account fleet (5 listed slugs), no openai backend, claude catalog (11 bases));
 refused locally rather than forwarded to the Claude pool, which would 404 it after spending a request"}
```

The account lists `gpt-5.6-terra`. The guard is comparing `gpt-5.6-terra:high` against that list.

## Why it matters

`parseEffortSuffix` exists precisely so clients that cannot set `output_config.effort` can pick effort by model name. Codex-backed models are reasoning models where effort is the most consequential knob available, and they are the ones that cannot express it.

It is also the knob with the widest blast radius, which is the other half of why it should be per-path. Measured through dario on one identical prompt, same model, only the effort suffix changing:

| effort | output tokens | wall time |
|---|---|---|
| low | 1,982 | 18.2s |
| high | 3,947 | 34.8s |
| max | 11,160 | 99.9s |

Reasoning is billed as output, and output is priced several times input, so a process-wide `max` is not a rounding error — it is a real cost increase on every caller, plus 5.5x the latency. On an agent with a wall-clock timeout and a dozen turns per run, that alone can push runs into the timeout. A per-path setting lets one workload move without dragging every other one with it.

## Suggested fix

Strip the effort suffix **before** provider/model-list matching, and carry the parsed effort into the Codex request as `reasoning.effort`, the same way the Claude path carries it today.

Sketch:

[...]

Issue #1259, opened 2026-09-07
Codex path: the system prompt is never cached — dario sends it as Responses `instructions`, which is outside the cacheable prefix


askalf:
## Summary

dario maps an Anthropic-shape `system` prompt to the Responses API's top-level `instructions` field. **`instructions` does not participate in OpenAI's prompt caching.** The same text placed inside `input` caches normally. The result: on a Codex-served request, the system prompt is re-billed at full price on every turn of a multi-turn agent loop.

This is not the `prompt_cache_key` machinery — that part works.

## Reproduction

Against a ChatGPT-plan-backed model (`gpt-5.6-terra` here), two requests 1.5s apart, prefix well over OpenAI's 1,024-token floor:

**A — long prefix in `system`:**

```jsonc
POST /v1/messages
{ "model": "gpt-5.6-terra", "max_tokens": 16,
  "system": "<~1,400 tokens, identical in both requests>",
  "messages": [{ "role": "user", "content": "Say ONE." }] }
// req2 appends: assistant "ONE", user "Say TWO."
```

```
req1 usage: {"input_tokens":1461,"cache_read_input_tokens":0}
req2 usage: {"input_tokens":1475,"cache_read_input_tokens":0}   ← never caches
```

**B — the same text moved into the first user message:**

```
req1 usage: {"input_tokens":2418,"cache_read_input_tokens":0}
req2 usage: {"input_tokens":640,"cache_read_input_tokens":1792}  ← caches
```

**C — OpenAI chat shape, `role: "system"` message**, via `/v1/chat/completions`: `cached_tokens` 0 on both requests. So both wire shapes are affected.

## Where it comes from

`src/anthropic-responses-translate.ts` documents the mapping and then does it:

```
:35   *   - system  → top-level `instructions` (a string), NOT a role message.
:666  const instructions = flattenSystem(body.system);
:667  if (instructions.length > 0) out.instructions = instructions;
```

## Ruled out

- **`prompt_cache_key` is attached correctly** (`codex-backend.ts:948`) and survives `toCodexSupportedBody` (it is in `CODEX_SUPPORTED_FIELDS`).
- **Usage reporting is correct** — `anthropicUsageFromResponses` reads `input_tokens_details.cached_tokens` and nets it out of `input_tokens`. Test B proves the read-back path works; A genuinely returns zero.

## Impact

[...]

Issue #1236, opened 2026-09-06
route(): a model no provider lists falls through to the Claude pool verbatim and 404s upstream instead of a local routing error


askalf:
`src/provider-adapter.ts` `claudeAdapter.claimsPrimary()` returns `true` unconditionally. A request whose model is neither in the Codex account's discovered slugs (`isCodexModel`, `src/codex-backend.ts:184`, exact match) nor a Claude catalog base is forwarded to Anthropic with `model` untouched.

Repro: Codex account present but `getCodexModelSlugs` returns a list without `gpt-5.6-sol` (slug de-listed, discovery degraded to `[]`, or the cached-slugs path). `POST /v1/messages {"model":"gpt-5.6-sol"}` goes to the Claude pool and returns `404 not_found_error "model: gpt-5.6-sol"` with an Anthropic `req_011…` id, attributed to whatever seat sent it, after spending a pool request.

The fleet hit this shape twice for other causes (v5.5.87 / #1141, and #1138). Those causes are fixed; the fall-through that produced the 404 is unchanged.

Expected: when the primary claim is the default adapter and `isClaudeServableModel(model, getCachedBases(), resolver)` (`src/claude-model.ts`) is false, answer 400 locally with `x-dario-upstream-rejection: model_unroutable` naming the providers consulted, and log under `--verbose`: `no provider lists <model>; refusing` so a fleet log can tell this from a real Anthropic model error.

Not a failover bug: `buildPoolFallbackBody` rewrites `model` in both directions and `pickClaudeTarget` is catalog-validated since v6.0.1.

Evidence: 4 fleet executions 2026-08-29 22:23Z to 23:24Z on a Sonnet-configured seat, all pre-#1141; dario 6.0.25 on the box today.

Verified against master 8640fc2 (v6.0.31): claudeAdapter.claimsPrimary() at src/provider-adapter.ts:123 still returns true unconditionally; isCodexModel at src/codex-backend.ts:184 is still an exact slug match; no local model-unroutable rejection exists in src/.

PR #1261, opened 2026-09-07
an effort suffix on a Codex model routes instead of 400ing (v6.0.38)

askalf:
Addresses #1260.

## The bug

#419 let a client that cannot set `output_config.effort` choose a reasoning level by naming it in the model (`opus-4-8:high`, or Cursor's hyphen spelling `claude-opus-4-8-high`). Only the Claude half ever honoured it. Probed live against the ChatGPT plan on 2026-09-07:

| model sent | result |
|---|---|
| `claude-sonnet-5` | 200 |
| `claude-sonnet-5:high` | 200, served `claude-sonnet-5` |
| `gpt-5.6-terra:high` | **400 `model_unroutable`** |

Routing matches the model against each provider's list *before* the suffix is stripped, so `gpt-5.6-terra:high` matched nothing — the account lists `gpt-5.6-terra`.

## Why the fix is a third strip site, not a change to the existing two

The two strip sites in `proxy.ts` skip OpenAI-shaped names deliberately, and that comment is right: an openai-compat backend may serve a model whose real id ends in `-high` or `-low`, and stripping there would rewrite a legitimate name against a catalog dario cannot see.

Codex is the one provider that publishes its routable set, so the ambiguity is decidable there and nowhere else. The strip now also happens after model discovery, guarded by the rule that makes it safe:

> strip only when the name **as written** matches no slug **and** the stripped name matches one.

That is the same as-written-wins rule `resolveClaudeTarget` has applied to chain entries since #1161. It also makes this change strictly additive — the only requests whose routing changes are the ones that answer 400 `model_unroutable` today, because a name that already routes is never re-read.

## What is in here

- **A suffixed Codex model routes to the subscription.** The suffix is stripped from the outbound body (the backend 400s on a slug it does not list) and the level is sent as `reasoning.effort`.
- **A slug that genuinely ends in an effort word is routed exactly as written**, and no effort is invented from its name.
- **A model no provider lists is still refused locally** with `x-dario-upstream-rejection: model_unroutable` and no upstream request spent (#1236 intact).
- **`pickCodexFallback` learns the same rule.** A `--pool-fallback` entry naming a Codex model with a suffix matched no slug and was skipped in silence — no error, no log, just a failover the operator configured that never fired. It now carries the declared effort into the failover request, mirroring the Claude end since #1161.
- **The per-request fallback chain is selected from the stripped name.** A per-model spec is keyed on the model being routed; keying it on a name still carrying a dario-side suffix would miss the operator's own entry for that model.
- **`reasoning.effort` accepts the levels the backend actually has.** The thinking-budget mapping only ever produced `low`/`medium`/`high`; a named effort can now be `none`, `minimal`, `xhigh` or `max`. dario's own `ultracode` tier maps to `max`, the nearest upstream equivalent, rather than being dropped; `client` forces nothing.
- **Documented** in `docs/configuration.md`, where the suffix was not written down at all before.

## On `REASONING_HEADROOM`

Widening the reachable levels needed reservations for `xhigh` and `max`. They are set **above** the measured trend rather than on it, because under-reserving is the failure the table exists to prevent: reasoning eats the whole budget, the response comes back `incomplete`, and the turn arrives empty. Both values are clamped by `RESPONSES_MAX_OUTPUT_CAP` anyway.

Measured through dario on one identical prompt, same model, only the level changing:

| level | output tokens | wall time |
|---|---|---|
| low | 1,982 | 18.2s |
| high | 3,947 | 34.8s |
| max | 11,160 | 99.9s |

Worth stating plainly in the docs, and it is: reasoning is billed as output, so `max` is a real cost increase on every request that names it, and on a long agent loop it can push a run into its own timeout. This PR does not raise anyone's effort — it only makes the level nameable per request.

## Tests

`test/codex-effort-suffix.mjs` (new, 17 checks) drives the **real handler** against a Codex stub that records request bodies, and every assertion reads the body the stub actually received. That is the standard #1161's review established: picker-level assertions all pass with the propagation line deleted, so they prove nothing. Deleting the propagation fails this file.

It covers the suffixed model, the same slug without a suffix (nothing forced), the real-id-ends-in-an-effort-word trap, the still-refused unlisted model, and a fallback chain entry carrying the suffix.

`test/codex-backend.mjs` gained 8 picker checks for `pickCodexFallback`, including chain ordering and the as-written-wins case.

Full suite: **191/191 pass**. Build clean.

## Follow-up not in here

The `ticket_ops` churn and the uncached system prompt (#1259) are separate. #1259 in particular needs a measurement of verdict quality before it becomes a patch, not just a cache-hit number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.38] - 2026-09-07
+
+### Fixed
+
+- **An effort suffix on a Codex model name works, instead of answering 400 `model_unroutable` (#1260).** `#419` let a client that cannot set `output_config.effort` choose one by model name (`opus-4-8:high`, or Cursor's `claude-opus-4-8-high`). Only the Claude half ever honoured it: routing matched the model against each provider's list *before* the suffix was stripped, so `gpt-5.6-terra:high` matched nothing and was refused locally while `claude-sonnet-5:high` served fine. The two existing strip sites skip OpenAI-shaped names deliberately, because an openai-compat backend may serve a model whose real id ends in `-high` and a blind strip there would rewrite a legitimate name against a catalog dario cannot see. Codex is the one provider that publishes its routable set, so the strip now also happens after discovery, guarded by the same as-written-wins rule `resolveClaudeTarget` uses for chain entries (#1161): strip only when the name as written matches no slug and the stripped name matches one. A slug that genuinely ends in an effort word is still routed exactly as written, a model no provider lists is still refused, and the suffix never reaches the backend. The named effort is carried into the Codex request as `reasoning.effort`.
+- **A `--pool-fallback` chain entry naming a Codex model may carry an effort suffix too (#1260).** `pickCodexFallback` matched entries against the account's slugs literally, so `gpt-5.6-terra:high` matched nothing and the entry was skipped in silence — no error, no log, just a failover the operator configured that never fired. It now reads the suffix under the same guard and carries the declared effort into the failover request, which is what the Claude end of a chain has done since #1161.
+
+### Changed
+
+- **`reasoning.effort` accepts the levels the backend actually has.** The thinking-budget mapping only ever produced `low`/`medium`/`high`; a named effort can now also be `none`, `minimal`, `xhigh` or `max` (`ultra` is not a level the backend accepts). dario's own `ultracode` tier maps to `max`, the nearest upstream equivalent, rather than being dropped. `REASONING_HEADROOM` gained reservations for the new levels — measured through dario on one identical prompt with only the level changing, `low` returned 1,982 output tokens, `high` 3,947 and `max` 11,160, so the reservations sit above that trend rather than on it. Under-reserving is the failure the table exists to prevent: reasoning eats the whole budget, the response comes back `incomplete`, and the turn arrives empty.
+
 ## [6.0.37] - 2026-09-07
 
 ### Fixed
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -74,6 +74,34 @@ Off by default, each one a deliberate divergence from what real CC sends.
 | `DARIO_EFFORT` | `--effort=` | Forces a reasoning-effort level. Can flip requests to overage billing — watch `-v` logs for representative-claim changes ([`#87`](https://github.com/askalf/dario/issues/87)). |
 | `DARIO_MAX_TOKENS` | `--max-tokens=` | Anthropic enforces the per-model ceiling server-side, so too-high values return a clean 400 ([`#88`](https://github.com/askalf/dario/issues/88)). |
 
+### Per-request effort, by model name
+
+`DARIO_EFFORT` is process-wide: it applies to every caller, Claude and Codex
+alike. A client that has no way to set `output_config.effort` can instead name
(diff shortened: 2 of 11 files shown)


sprayberry-redline (changes requested):

Verdict: changes requested — the newly documented `client` model-name suffix is not parsed, so it cannot work.

**Blocking — `docs/configuration.md:89-90`**

> `Levels: \`low\`, \`medium\`, \`high\`, \`xhigh\`, \`max\`, plus dario's own \`ultracode\``
> `(which reaches a Codex backend as \`max\`) and \`client\`, which forces nothing.`

The new documentation presents `client` as a per-request model-name level, but the changed parser's suffix vocabulary omits it:

> `const SUFFIX_EFFORTS: ReadonlyArray<EffortValue> = ['ultracode', 'medium', 'xhigh', 'high', 'low', 'max'];` (`src/effort.ts:32`)

Consequently, a caller following this documentation and sending `gpt-5.6-terra:client` leaves the model unchanged; it matches no published Codex slug and is rejected as `model_unroutable`, rather than routing `gpt-5.6-terra` while forcing no effort. The same failure applies to a `--pool-fallback` entry ending in `:client`. Either support `client` in the suffix parser (and add handler/fallback coverage) or remove it from the model-name suffix documentation.

```suggestion
const SUFFIX_EFFORTS: ReadonlyArray<EffortValue> = [
  'ultracode', 'medium', 'xhigh', 'high', 'low', 'max', 'client',
];
```

I reviewed the routing, fallback, translation, and new test changes. CI is green; the issue above is confined to the newly documented but unsupported suffix.

askalf:
Good catch, and correct — the documentation was wrong, not the parser.

`client` is in `VALID_EFFORT_VALUES` (so it is a legal `--effort` / `DARIO_EFFORT` value) but deliberately absent from `SUFFIX_EFFORTS`, and I conflated the two vocabularies when writing the new docs section.

I took the documentation fix rather than the suggested parser change, for two reasons:

1. `client` means "leave the client's own choice alone", so as a suffix it is a no-op — `model:client` and `model` would be the same request.
2. Accepting it would make a model genuinely named `...-client` strippable to a name no backend lists, which is the exact hazard the as-written-wins guard in this PR exists to avoid.

Pinned with three assertions in `test/effort-flag.mjs` so the doc and the parser cannot drift apart again: `gpt-5.6-terra:client` and `some-model-client` both survive parsing intact, and `client` remains in `VALID_EFFORT_VALUES`.

Full suite 191/191. Pushed in ffc7ef0.

sprayberry-redline (approved):

**Verdict: Approve — no blocking issues found on live head `ffc7ef079eee641d2f432d2a4782d41e3483209e`.**

I reviewed the live-head diff (including the re-review commit), the routing and fallback propagation paths, and the new wire-level coverage. In particular, the changed Codex routing guard preserves an as-written listed slug and only strips a recognized suffix when the stripped slug is listed; the selected fallback's effort is propagated into the Anthropic-to-Responses translation. The re-review commit correctly documents and tests that `client` is configuration-only, not a model-name suffix.

All reported CI checks are green, including build, test, CodeQL, compat, live-test, and wire-drift. I did not run the local test suite, per review environment policy.

What's good: the added integration test asserts the stub's received Responses body, covering both the outbound bare model and `reasoning.effort`, rather than only testing the picker/parser.

Issue #1263, opened 2026-09-08
windowKey treats grid-aligned resets as one subscription — 5h resets land on a 20-minute grid, so independent seats collide


askalf:
## Problem

`windowKey()` decides that two pool seats are one subscription under two aliases when they report the same `claim@reset` pair (`src/pool.ts:164-168`). The stated justification is in the doc comment above it:

> Two seats that report the same key are one subscription under two aliases (dario#1244, "a few have the same issue"): two independent windows all but never share a reset second, and two readings of one window always do.

**That premise is false.** Anthropic aligns the 5-hour window reset to a 20-minute grid, so independent windows share reset seconds routinely.

## Evidence

From a user's `/accounts` listing on #1244 — two seats that are demonstrably *not* the same subscription (they reported 102% and 4% utilization at overlapping times):

| seat | `reset_at` (s) | UTC | `mod 1200` |
|---|---|---|---|
| matteo | 1788864000 | 2026-09-08T10:40:00Z | **0** |
| test | 1788878400 | 2026-09-08T14:40:00Z | **0** |

Both land exactly on a 20-minute boundary, not on an arbitrary second.

A 5-hour window contains `300 / 20 = 15` reset slots. So for a pool of `N` seats, collisions are **guaranteed by pigeonhole once N > 15**, and likely well before that by the birthday bound. The reporter runs 18 seats and saw a 7-alias `shares_window_with` group of accounts that are independent subscriptions.

## Impact

Diagnostic only, but actively misleading — and it misled the reporter into a wrong conclusion about his own configuration, which is the cost.

Verified consumers of the grouping, all display:

- `sharesWindowWith` on `GET /accounts` and the admin listing (`proxy.ts:2551`, `proxy.ts:2664`)
- the one-shot `seats "X" and "Y" report the same … window` log line (`proxy.ts:1683`)
- `distinctWindows` (`pool.ts:191`), which feeds the `dario accounts list --live` window count
- `doctor-core.ts:304` tells the operator that `sharesWindowWith` "says for sure" — so the wrong answer is presented as authoritative

Confirmed **not** affected: seat selection, parking, and headroom. `accountIneligibility()` reads only the seat's own snapshot, and the one mechanism that copies a reading between seats (`adoptSnapshot`) is multi-instance pool-sync, keyed by alias rather than by window.

## Why the fix is not just a finer key

[...]

Issue #1262, opened 2026-09-08
Fable routing expects `7d_fable`, but Anthropic sends `7d_oi`


dizzystuff:
### Problem

Dario expects `7d_fable` for Fable's weekly quota, but Anthropic sends `7d_oi`. As a result, Dario ignores the Fable quota when calculating account headroom.

The parser stores the reading as `perModel7d.oi`. For Fable requests, `modelFamily()` returns `fable`, so `computeHeadroom()` looks up `perModel7d.fable` instead.

I'm using two Claude Max accounts with `fill-first`. A Fable response reported 98% Fable usage and 63% shared weekly usage. Dario calculated 37% headroom, overlooking the nearly exhausted Fable quota.

### Environment

- Dario 6.0.37 from npm. We also inspected 6.0.38; it retains the same lookup.
- macOS on Apple Silicon, with the proxy running under Bun 1.4.2.
- Two Claude Max accounts, using `--pool-strategy=fill-first`.
- Requested model: `claude-fable-5-1`.

### Reproduction

These utilisation headers came from a successful Fable response:

```http
anthropic-ratelimit-unified-5h-utilization: 0.01
anthropic-ratelimit-unified-7d-utilization: 0.63
anthropic-ratelimit-unified-7d_oi-utilization: 0.98
```

The following reproduces the calculation without credentials or API requests. We ran it with Node 26.8.1.

1. Open the installed package directory with `cd "$(npm root -g)/@askalf/dario"`, or use a built Dario checkout containing `dist/pool.js`.
2. Run:

```bash
node --input-type=module <<'JS'
import { parseRateLimits, modelFamily, computeHeadroom } from './dist/pool.js';

const snapshot = parseRateLimits(new Headers({
  'anthropic-ratelimit-unified-5h-utilization': '0.01',
  'anthropic-ratelimit-unified-7d-utilization': '0.63',
  'anthropic-ratelimit-unified-7d_oi-utilization': '0.98',
}));
const family = modelFamily('claude-fable-5-1');
console.log(JSON.stringify({
  family,
  perModel7d: snapshot.perModel7d,
  headroom: computeHeadroom(snapshot, family),
}));
JS
```

Actual output:

```json
{"family":"fable","perModel7d":{"oi":0.98},"headroom":0.37}
```

Expected: approximately `0.02` headroom, because Fable routing should include the `oi` quota in the calculation.

### Existing evidence and test gap

[...]

PR #1264, opened 2026-09-08
seats know who they are; a 429 parks only on an exhausted window; 7d_oi binds Fable (v6.0.39)

askalf:
Addresses #1262, #1263 and the 2026-09-08 findings on #1244. Three defects, each verified on the wire or on a live box before anything changed.

## 1. Fable's weekly allowance never counted toward headroom (#1262)

It arrives as **`7d_oi`** — the plan's included-overage credit — not as `7d_fable`, which has never been observed. `computeHeadroom` looked up the imaginary bucket and read a seat at `7d 0.63, 7d_oi 0.98` as **37% free** when it was two points from a hard 429.

The evidence that `oi` binds Fable, from a live Max account on 2026-07-05: Fable drew `representative-claim: seven_day_overage_included` at 7d 82% with `7d_oi` at 99%, served at $0 — and at `7d_oi ≥ 1.0` Fable answered **429 `rate_limit_error`** (`7d_oi-status: rejected`, `7d_oi-surpassed-threshold: 1.0`) while Opus kept serving.

Two mechanisms:
- **`WIRE_BUCKET_BINDINGS` seeds `oi → fable`** with that evidence, so the reporter's repro is right from the first response: it now yields `0.02`.
- **A seat learns bindings from its own responses.** An `*_overage_included` claim, or a `7d_<bucket>-status: rejected`, on a response for family F binds that bucket to F on that seat. Our own fleet box's `login` seat is on claim `seven_day_overage_included` today with no Fable traffic at all — so `oi` is not Fable-only, and Opus drawing on included overage is bound by `oi` for Opus exactly when its responses say so, never by assumption.

`dario doctor` labels the bucket `Included overage credit (7d, oi)` instead of presenting it as a model family. `pool-e2e` now exercises the header Anthropic actually sends.

## 2. Any 429 parked the seat until its stated reset (#1244 follow-up)

Our fleet box parked a seat for **546 hours** on this reading:

```
rate limited (429) on account "pro1": 5h 0%, 7d 0%, claim unknown, resets in 546h 49m — parked until the window rolls
```

Nothing in that 429 said a window was over. The status code alone decided, and `reset` was honoured blindly.

Now the headers decide. A utilization at or past the 1.0 threshold on any window or bucket (the unified headers are a ratio against `surpassed-threshold: 1.0`; every live rejection has read 1.00–1.06) parks until `reset`, as before. Anything else cools for the response's own `retry-after`, or a minute, **stays probeable**, and logs what it declined:

```
429 without an exhausted window (5h 0%, 7d 0%, claim unknown; stated reset in 546h 49m not honoured) — cooling 1m, seat stays probeable
```

Snapshots from before the field (`exhausted` undefined) behave exactly as before.

## 3. "Shares a window" was inferred from a shared reset second (#1263)

The heuristic assumed independent windows all but never share a reset second. Anthropic aligns the five-hour reset to a **20-minute grid** — two demonstrably different accounts on #1244, both resetting at exactly `:40:00` — so a window has 15 possible reset seconds and a pool of 18 seats collides by pigeonhole. It told an operator seven colleagues were one subscription.

Replaced by **identity**. Every add path now reads the token's OAuth profile (`GET /api/oauth/profile`, probed live: `account.uuid`, `account.email`, `organization.uuid`, `rate_limit_tier`, `seat_tier` all present) and records the account uuid, the masked email and the tier fields on the seat. A record from before is filled in on its next token refresh — one GET, through the proxy's own `fetch`, never fatal, and kept out of the token exchange itself so sequenced refresh mocks see exactly the calls they always did.

`sharesWindowWith` / `distinctWindows` keep their names and now mean *the same account*; `sameAccountAs`, `accountId`, `accountEmail`, `rateLimitTier`, `seatTier` sit beside them on both listings. `dario accounts list --live` prints `same account as …` and `account ma***@…`; `dario doctor` gains an `Accounts` row; the proxy says it once per pair at start-up.

## Found on the way: one client identity across every seat

Every add path copied the **machine's** Claude Code identity into every alias when one was installed. On a machine running a pool of colleagues' tokens that meant 18 different OAuth accounts all presenting ONE `device_id` / `account_uuid` in `metadata.user_id` — and Anthropic ties what it sees to that identity (the identity-drift check documents the bearer being cross-validated against it; the startup banner warns billing may change without one). It fits #1244's pattern — every rejection reading ≈1.0 on a seat's first request while each colleague's own usage page said 0% — but I am **not** claiming it as the cause. The reporter's own evidence argues against that: the seat he deleted and re-added carried the same shared identity and came back reading 4%. A five-hour window cannot fall from 102% to 4% inside itself, so the rejected reading was not that user's window at all; the likelier reading is a limit above the user, at the organization, which 18 parallel seats reach and one seat does not. This change is hygiene worth doing on its own terms, and the diagnosis stays open on #1244 pending the direct-token check I asked the reporter for.

A new alias now takes the local identity only when no other alias holds it, or the holder is *proven* (same account uuid) the same account; otherwise it gets its own, exactly as a machine without Claude Code always did — which is how the fleet box has run every seat. **Existing seats are never rewritten unasked.**

```
dario accounts identity                     # per-seat report; warns on the shared case
dario accounts identity --fresh <alias>...  # or --fresh --all
```

The running proxy presents a rewritten identity on the seat's next request; `pool.add` takes a changed on-disk identity without a restart. `dario doctor` warns on the `Client identity` row.

## Tests

Five new files, and fixtures moved from imagined headers to observed ones:

| file | covers |
|---|---|
| `oi-bucket-binds-fable.mjs` | the reporter's repro verbatim → 0.02; seed does not bind other families; learning via claim and via bucket status; persistence across readings; routing |
| `pool-429-without-exhausted-window.mjs` | the fleet box's 429 → cools a minute, stays probeable; retry-after honoured; the reporter's 1.02 still parks; old snapshots unchanged; once-per-parking log contract |
| `accounts-identity-policy.mjs` | profile fetch (live shape, every failure → null); identity policy in all five cases; `--fresh` rewrites only what it should |
| `pool-window-peers.mjs` | identity-based peers; the grid collision is NOT a peer relation; a changed on-disk identity is presented without restart |
| `doctor-account-identity.mjs` | the `Accounts` and `Client identity` rows, masking included |

`pool-rate-limit-reset` had one fixture that "parked" a seat on a 429 reading 10% — the exact class this change says is not a parking; its intent (a recovered seat is reachable again) is kept under the new rule. Full suite **195/195**.

## Compatibility

- `sharesWindowWith` / `distinctWindows` change meaning (same account) but keep their names and types; `same_account_as` etc. are additive.
- A 429 that showed no exhausted window no longer parks on its reset. Genuine parkings are unchanged.
- No stored seat's client identity is rewritten; the profile back-fill only *adds* fields.


sprayberry-redline (dismissed):

Verdict: approve — no blocking issues found in the reviewed change.

I reviewed the rate-limit parsing, family/bucket binding and routing changes; the non-window-429 cooldown/eligibility path; OAuth profile persistence and client-identity selection; and their added coverage. In particular, the changed rejection path distinguishes only observed exhaustion (`src/pool.ts:631-637`) and applies the short cooldown rather than the advertised reset for other 429s (`src/pool.ts:1119-1128`), while eligibility expires that cooldown (`src/pool.ts:377-413`). The profile fetch is non-fatal and validates the account UUID before persistence (`src/accounts.ts:204-237`), and client-identity reuse is limited to an unheld identity or a profile-proven same account (`src/accounts.ts:269-281`).

All reported CI checks are green, including test, compatibility, CodeQL, wire-drift, and live-test. I did not run the local test suite.

What's good: the diff keeps the new wire-derived assumptions narrow, retains compatibility fields for account peers, and adds focused regression coverage for each changed behavior.

sprayberry-redline (changes requested):

**Verdict: request changes — the new non-window 429 cooldown can be bypassed within the same client request.**

### High — `src/pool.ts:430-433`, `src/pool.ts:1085`

> `return rl.status === 'rejected' && rl.exhausted !== false && rl.reset > 0 && rl.reset * 1000 > now;`
>
> `const probeable = candidates.filter(a => !isInAuthCooldown(a, now) && !isParkedInLiveWindow(a, now));`

A 429 with no exhausted utilization now sets `exhausted: false` and `cooldownUntil`, but the changed `isParkedInLiveWindow` explicitly makes that seat not parked and the failover filter does not check `cooldownUntil`. In a pool where the only alternate seat is already rate-limited, a request sent to seat A that receives such a 429 will immediately fail over to A again (all other candidates are filtered as parked). This repeats the rejected request rather than honoring the upstream `retry-after`/one-minute cooldown; with a single-seat pool it is the same issue if this fallback path is reached. The nearby comment even says “Cool-downs are skipped,” but the predicate does not implement it.

```suggestion
const probeable = candidates.filter(a =>
  !isInAuthCooldown(a, now)
  && !isParkedInLiveWindow(a, now)
  && !(a.rateLimit.exhausted === false
    && a.rateLimit.cooldownUntil !== undefined
    && now < a.rateLimit.cooldownUntil),
);
```

I reviewed the changed TypeScript behavior and its focused tests, including the 429 classification/routing, bucket binding, account-profile identity propagation, and the API/listing changes. CI is green, but the focused 429 test coverage does not exercise the all-other-seats-parked failover path above.

askalf:
Confirmed and fixed in 7e10fe3 — good catch, and it was worse than the quoted site.

`isParkedInLiveWindow` reports false for a cooling seat by design (the reset that 429 stated was never that seat's window), but **both** fallback filters tested only that predicate:

- `select()` all-exhausted branch — `pool.ts:932`
- `selectExcluding()` mid-flight failover — `pool.ts:1085` (the one you quoted)

So the bypass existed from two entry points, not one. Rather than take the inline suggestion at the quoted site, I gave the condition a single definition — `isProbeable` = not auth-cooling, not parked in a live window, not cooling after a rejection — and both filters go through it. The bug was two independent copies of a condition that a new state got added to neither; a second copy would have preserved that shape.

Two consequences I also fixed, since they follow from the same state being invisible:

- `parkedUntil()` and `parkedCount()` now count a cooling seat. A pool that is entirely parked-or-cooling previously reported "not parked", fell past the local-429 answer, and spent a probe that could only 429 again. It now answers honestly with `retry-after` at the earliest return.
- An auth cool-down still does **not** count as a rate limit — that contract is explicit in `parkedUntil`'s comment, and there is now a test pinning it, since my change moved code near it.

Nine checks added to `test/pool-429-without-exhausted-window.mjs`: your exact scenario from both entry points, the mixed parked+cooling pool and its `retry-after`, and the return to service once the cool-down passes. Full suite 195/195.

sprayberry-redline (approved):

**Verdict: approve — no blocking issues found in the live-head follow-up.**

I reviewed the changes since the prior `CHANGES_REQUESTED` verdict, specifically the new shared `isProbeable` predicate and both fallback selection paths in `src/pool.ts`, plus the regression coverage in `test/pool-429-without-exhausted-window.mjs`. I also checked the PR's CI signal; all reported checks are green. I did not run the local test suite.

What's good: `isCoolingAfterRejection` at `src/pool.ts:444-448` precisely distinguishes a non-window 429 cool-down from a live-window parking, and `isProbeable` centralizes the three transient exclusions. Both `select()` (`src/pool.ts:968`) and `selectExcluding()` (`src/pool.ts:1130`) now use it, eliminating the previously missed failover path. The added regression cases exercise both entry points, mixed parked/cooling pools, expiry of the cool-down, and preservation of the auth-cool-down contract.

PR #1273, opened 2026-09-09
a utilization reading expires with its own window (v6.0.44)

askalf:
Addresses #1244.

## The bug

#1232 let a *rejection* expire against `anthropic-ratelimit-unified-reset`. The utilization on the same snapshot never got the same treatment, and that asymmetry strands a seat that never 429'd at all.

A seat whose last response read `5h 99%` computes headroom `0.01`, under `POOL_HEADROOM_FLOOR` (`0.02`). Below the floor:

- the selector skips it
- `pickFillFirst` won't take it
- sticky bindings rebind away from it
- `drainQueue`'s probe loop **breaks** on it

So nothing sends it a request, so `updateRateLimits` never runs, so the reading never refreshes. A rejection at least has the all-exhausted fallback in `select()`; this seat has nothing, because it was never *ineligible* — only permanently unattractive.

## The report

Nine-seat pool, every seat reading `0.98`–`1.03`, so every seat at or under the floor and `waitForAccount` queueing until it timed out. The whole pool reported exhausted while two of its seats carried `reset_in_ms: 0` and were provably free — the operator confirmed one at 0% on the dashboard. One of those two had served exactly **one** request, 3.7 hours earlier, and had been frozen out ever since.

`rejected_count` was `0` on every seat. That is why none of the rejection-side fixes reached this: #1232, #1254 and the 6.0.39 cool-down each require a 429 to have happened.

## The fix

`computeHeadroom` retires a reading whose stated reset has passed, dropping **only the bucket the reading's own `claim` names**:

- `five_hour` → `util5h`
- `seven_day` (including `seven_day_overage_included`) → `util7d`
- anything else → untouched

A five-hour rollover must not clear a seven-day reading, or traffic lands on a seat whose weekly quota really is spent. A `claim` naming neither window is left alone rather than guessed at — guessing is how a genuinely throttled seat gets pushed back into rotation.

Two deliberate choices worth review:

**Not `rateLimitWindowPassed`.** That predicate answers "is this rejection over", and for a 429 that named no exhausted window it answers yes as soon as the cool-down elapses, whatever the window says. Reusing it would zero a true `5h 99%` reading a minute after a non-exhausted 429 and send traffic the seat cannot serve. The check here is the stated rollover itself. There is a test pinning this.

**Zeroing, not flagging.** The seat becomes an attractive candidate, is picked once, and its own response refreshes the reading — the same "asking is the only way back" rule `isParkedInLiveWindow` documents, and it costs no synthetic probe. It also answers the reporter's "should it send a hello to get new data?" without adding a request.

## Tests

New `test/pool-elapsed-window-utilization.mjs`, 23 assertions. Verified to actually catch the bug: with `expireElapsedWindow` neutered to a no-op, **9 behavioural assertions fail** — including the two that reproduce the outage — while the safety guards still pass.

Covered: the seconds-vs-milliseconds unit trap in both directions; each `claim` mapping and the cross-bucket guards; purity; the non-exhausted-429 cool-down regression; a per-model `7d` bucket still suppressing headroom for its family after a 5h rollover; the nine-seat pool returning a seat again; and the inverse — with windows still live the same pool stays exhausted, so there is no false rescue.

196/196 suites green.


sprayberry-redline (changes requested):

**Verdict: changes requested** — the release version is inconsistent, and the required preflight check is failing.

### Blocking — release artifact consistency
`package.json:3`
> `"version": "6.0.44",`

This version bump leaves the committed lockfile at `6.0.43`. On this PR's head, the required `validate-package-json` check fails with `package-lock.json: version 6.0.43 ≠ package.json 6.0.44`, so the release artifact cannot pass its own consistency preflight or merge through the normal required-check path. Update the lockfile's root/package version together with the manifest.

```suggestion
# Regenerate or update package-lock.json so its root version matches package.json:
"version": "6.0.44"
```

What's good: I reviewed the pool change and its new behavioral coverage. The elapsed-window check correctly compares the seconds-valued reset after conversion to milliseconds, only retires the bucket identified by the representative claim, and preserves per-model weekly saturation. CodeQL and analysis are green; live-test and compat were pending when reviewed.

I did not run the local test suite, per review environment policy; CI is the test signal.

sprayberry-secondread (commented):

**Verdict:** Solid, well-tested fix for the stated bug, but the new expiry logic has a gap on the per-model-bucket path it doesn't cover — that path can still strand a seat the same way #1244 did, just via `perModel7d` instead of `util7d`.

## Findings

### Medium — `expireElapsedWindow` never retires a stale `perModel7d` bucket, so `computeHeadroom(snapshot, family)` can still floor a seat whose window has rolled

`src/pool.ts:783-788`:
```ts
if (!(snapshot.reset > 0 && snapshot.reset * 1000 <= now)) return snapshot;
if (snapshot.claim === 'five_hour') return { ...snapshot, util5h: 0 };
// `seven_day`, plus the overage variants that carry the same weekly window
// (`seven_day_overage_included`, see the overage note above).
if (snapshot.claim.startsWith('seven_day')) return { ...snapshot, util7d: 0 };
return snapshot;
```
This only ever zeroes `util5h` or `util7d`. It never touches `snapshot.perModel7d`, which `computeHeadroom` also folds into its max via `bucketsBindingFamily` (`src/pool.ts:723-729`):
```ts
const rl = expireElapsedWindow(snapshot, now);
const utils = [rl.util5h, rl.util7d];
if (family) {
  for (const bucket of bucketsBindingFamily(rl, family)) {
    const util = rl.perModel7d[bucket];
    if (util !== undefined) utils.push(util);
  }
}
```
Failure scenario: this PR's own CHANGELOG entry for #1262 (`CHANGELOG.md:41`) establishes that Fable's weekly allowance arrives as `7d_oi`, not `7d_fable`, and that `7d_oi` shares the same weekly window as `representative-claim: seven_day_overage_included`. Give a Fable-bound seat a snapshot with `claim: 'seven_day_overage_included'`, `reset` in the past, `util7d: 0.30`, and `perModel7d: { oi: 0.99 }` (a live-observed shape per the #1262 note: "`7d 0.63, 7d_oi 0.98`"). `expireElapsedWindow` zeroes `util7d` (claim starts with `seven_day`) but leaves `perModel7d.oi` at 0.99 untouched. `computeHeadroom(snapshot, 'fable', now)` still includes `rl.perModel7d['oi'] = 0.99` via `bucketsBindingFamily`, so headroom is still `1 - 0.99 = 0.01`, under `POOL_HEADROOM_FLOOR` (0.02) — the exact stranding this PR exists to fix, just reachable through the per-model bucket the fix doesn't zero. The seat never 429'd (same precondition as #1244's report), so it never re-probes and never refreshes on its own, same as the bug being fixed.

Suggested fix:
```ts
if (snapshot.claim.startsWith('seven_day')) {
  const perModel7d = Object.fromEntries(
    Object.entries(snapshot.perModel7d).map(([k]) => [k, 0]),
  );
  return { ...snapshot, util7d: 0, perModel7d };
}
```
(Or, if per-model buckets can carry independent resets in the future, gate this more narrowly — but today `parseRateLimits` doesn't capture a per-bucket reset at all, so there's no finer signal available than the unified `reset`/`claim` this function already uses.)

### Missing test coverage
`test/pool-elapsed-window-utilization.mjs` covers the per-model-bucket case only for the case where the bucket is a *separate, still-open* window (`'a spent per-model 7d bucket still suppresses headroom for that family'`, line ~145) — it never exercises a per-model bucket that shares the *same, rolled-over* weekly window (the `oi`/Fable case from #1262). Given the fix explicitly reasons about `seven_day_overage_included` sharing the weekly window, a test pinning the per-model side of that sharing would have caught the gap above.

## What's good
- The core fix (unified `util5h`/`util7d`) is correct and directly addresses the reported outage with a clear, narrow blast radius: only the bucket the `claim` names is touched, and the seconds-vs-ms unit trap is explicitly guarded and tested.
- The decision to check the stated rollover directly rather than reuse `rateLimitWindowPassed` is well-reasoned and the PR body's justification checks out against the code (`rateLimitWindowPassed` at `src/pool.ts:381` would indeed fire on cool-down elapse regardless of window state).
- New test file is thorough for what it covers: unit-trap regression, cross-bucket isolation (5h rollover doesn't clear 7d and vice versa), purity, and an end-to-end nine-seat pool reproduction of the reported outage. All CI checks pass.
- README/CHANGELOG/version bump housekeeping in this PR is straightforward and consistent with the stated line-count guard.

sprayberry-redline (dismissed):

**Verdict: Approve — no blocking issues found in the live-head diff.**

I reviewed the live head `1a682ba4878d2dcb24626f7f07c2936fe20694c1`, concentrating on the changed rate-limit expiry and its regression coverage. `expireElapsedWindow` compares the seconds-based reset timestamp against the millisecond clock before clearing only the claim-selected utilization bucket; `computeHeadroom` consistently uses the returned snapshot, including family-bound weekly buckets. The added coverage exercises elapsed versus live windows, unit conversion, bucket isolation, cooldown behavior, purity, selection, and the all-stale-pool regression. CI is green, including build, test, CodeQL, compatibility, and live-test checks.

What's good: the explicit injected clock makes the rollover boundary deterministic, and the tests preserve the important safety invariant that a five-hour rollover cannot erase weekly or per-model weekly utilization.

I did not run the repository test suite locally, per review environment policy; I relied on the completed CI checks.

sprayberry-redline (approved):

**Verdict: Approve — no blocking issues found in the live-head diff.**

I reviewed the rate-limit expiry path and its new behavioral coverage. In particular, `src/pool.ts:723-731` applies expiry before calculating both unified and family-bound headroom; `src/pool.ts:783-788` correctly compares the epoch-seconds reset to a millisecond clock and only clears the claimed 5-hour or 7-day unified bucket. The added tests cover rollover boundaries, units, non-mutation, cross-bucket preservation, the non-window-429 cooldown case, model-family headroom, and the reported selection failure.

What's good: the change keeps a live 99% reading intact despite an elapsed non-window rejection cooldown, avoiding the unsafe reuse of rejection-expiry logic. CI build/validation checks are green; the remaining analysis, compatibility, smoke, live-test, and test checks were pending when reviewed.

PR #1277, opened 2026-09-10
gpt-5.6-luna reads as a tier instead of the first-written one (v6.0.47)

askalf:
Addresses #1272.

## What actually drifted

`codex-drift-watch` fires on "wire-contract validation **or** the account-visible model list drifted", so the issue title doesn't say which. It was the list. The run log carries `Codex wire contract passed for gpt-5.6-sol.` — the contract is fine. The account now lists:

```
gpt-5.5   gpt-5.6-luna   gpt-5.6-sol   gpt-5.6-terra   gpt-reserve
```

Two things follow from that, one real and one cosmetic.

## The real one: luna has no tier

The Codex rungs are a size ladder — `sol` (sun) > `terra` (earth) > `luna` (moon) — and `selectPoolFallbackModels` classifies by model **name**, because the backend publishes a routable set and no tier metadata. `/v1/models` returns `gpt-5.6-luna` with nothing but `id`, `object`, `created`, `owned_by`.

`luna` matched none of `haiku|mini|small`, `opus|sol|large`, `sonnet|terra|medium`. So it fell to `'default'`, and with no `default:` entry in the operator's map, to:

```js
tiers.values().next().value   // the first entry — whichever tier was typed first
```

Map order is not meaningful to anyone writing one. On an **opus-first** map that silently sent luna traffic to the most expensive rung. Nothing logged it.

`luna` now joins the haiku test. The first-entry fallback for genuinely unclassifiable slugs (`gpt-reserve`) is documented behaviour on line 4 of that file and is unchanged.

⚠️ **The tier is inferred from the naming ladder, not from anything the backend states.** If luna turns out not to be the small rung, this is a one-word change — flagging it rather than burying it.

## The cosmetic one

The doc example and test fixture used `haiku:gpt-5.4-mini`, which is no longer on the account-visible list. Refreshed to `gpt-5.6-luna`.

To be clear about what this does **not** touch: there is no hardcoded default tier map — `selectPoolFallbackModels` parses an operator-supplied spec, so nothing was silently broken by the list changing. And the `gpt-5.4-mini` in `proxy.ts:847` is the unrelated legacy `OPENAI_MODEL_MAP` (OpenAI name → Anthropic model, for clients sending GPT names), left alone.

## Tests

Five new assertions in `test/pool-fallback-tier.mjs`.

The map in them is deliberately written **opus-first**, and that is the point. Written haiku-first — the natural way — an unclassified luna falls through to the haiku target *anyway*, so the assertion passes against the very bug it exists to catch. I found that by neutering the fix: with a haiku-first map only 1 of 4 luna assertions failed. Opus-first, the primary assertion fails loudly with `["gpt-5.6-sol"] !== ["gpt-5.6-luna"]`, which is also the real-world symptom.

Verified both directions: passes with the rung, fails without it.

196/196 suites green. `preflight` clean, README line-count guard ok at ~33k.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,16 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.47] - 2026-09-10
+
+### Fixed
+
+- **`gpt-5.6-luna` reads as a tier instead of falling through to whichever one is written first (#1272).** codex-drift-watch reported the account-visible model list as `gpt-5.5`, `gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-reserve`. The Codex rungs are a size ladder — `sol` (sun) > `terra` (earth) > `luna` (moon) — and `selectPoolFallbackModels` reads model *names* because the backend publishes a routable set and no tier metadata. `luna` matched none of the three tier tests, so a request naming it fell to `default` and, with no `default:` entry in the operator's map, to `tiers.values().next().value` — the first entry, i.e. whichever tier happened to be typed first. On an opus-first map that silently routed luna traffic to the **most expensive** rung. `luna` now joins `haiku|mini|small`. A genuinely unclassifiable slug (`gpt-reserve`) keeps the documented first-entry behaviour, which is unchanged. The wire contract itself did not drift — `codex-drift-watch` reported it passing for `gpt-5.6-sol`; only the model list moved.
+
+### Changed
+
+- **The tier-map doc example no longer names a model the account cannot serve.** `haiku:gpt-5.4-mini` → `haiku:gpt-5.6-luna`; `gpt-5.4-mini` is not on the account-visible list any more. This is a comment and a test fixture only — there is no hardcoded default tier map, and the `gpt-5.4-mini` entry in `proxy.ts`'s `OPENAI_MODEL_MAP` is the unrelated legacy OpenAI-name → Anthropic-model translation, which is untouched.
+
 
 ## [6.0.46] - 2026-09-10
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.46",
+  "version": "6.0.47",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.0.46",
+      "version": "6.0.47",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.0.46",
+  "version": "6.0.47",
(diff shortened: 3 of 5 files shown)


sprayberry-redline (approved):

**Verdict: Approve — no blocking issues found in the live-head diff.**

The `luna` classification is added at the existing tier-selection boundary, preserves the subsequent `default` and first-configured-tier fallback for unknown models, and the new opus-first fixture exercises the prior order-dependent failure case. I reviewed all five changed files and the complete PR diff; I did not run the local test suite. CI is green, including `test`, `build`, CodeQL, compatibility, and live-test checks.

What's good: the regression coverage deliberately avoids a haiku-first map, so it would fail against the pre-change fallback behavior rather than passing accidentally because of map insertion order.

sprayberry-secondread (commented):

**Verdict:** No blocking issues — the fix is correct and the new tests actually exercise the bug they claim to.

## What I checked

- `src/pool-fallback-tier.ts:23` — the added `luna` alternative:
  ```
  const tier = /haiku|mini|small|luna/.test(model) ? 'haiku'
    : /opus|sol|large/.test(model) ? 'opus'
    : /sonnet|terra|medium/.test(model) ? 'sonnet'
    : 'default';
  ```
  Traced all five account-visible slugs from the PR body against this ternary chain: `gpt-5.6-luna` → haiku (new), `gpt-5.6-sol` → opus, `gpt-5.6-terra` → sonnet, `gpt-5.5`/`gpt-reserve` → `default` → falls through to `tiers.values().next().value` (unchanged, pre-existing behavior, documented in the file header). No overlap between the `luna` alternative and the other three branches, and haiku is checked first in the chain so a `luna` match can't fall through to `opus`/`sonnet` — matches the intended size-ladder semantics described in the PR body.
- `test/pool-fallback-tier.mjs:22-41` — verified the four new assertions by hand-tracing `selectPoolFallbackModels` against the opus-first ladder `opus:gpt-5.6-sol,sonnet:gpt-5.6-terra,haiku:gpt-5.6-luna`: `luna`→`gpt-5.6-luna`, `terra`→`gpt-5.6-terra`, `sol`→`gpt-5.6-sol`, `gpt-reserve`→`gpt-5.6-sol` (first-entry fallback). All four match the map ordering chosen specifically to fail loudly if `luna` were still unclassified — the PR body's claim that a haiku-first map would mask the bug is consistent with the fallback logic (`tiers.values().next().value` returns whichever tier was inserted first).
- Version/changelog bump (`package.json`, `package-lock.json`, `CHANGELOG.md`) is consistent (6.0.46 → 6.0.47) and matches the source change.
- `gh pr checks 1277` — all 15 checks pass (test, build ×3, compat, docker-cap-drop-smoke, live-test, CodeQL, etc.), nothing pending or failing.

## Minor, non-blocking observation

The `default` bucket now silently absorbs both `gpt-5.5` and `gpt-reserve` (two structurally different cases: one is a real, currently-unclassified size on the roster; the other is explicitly called out as "genuinely unclassifiable"). That's pre-existing behavior this PR doesn't change or regress, so it's not a finding against this diff — just worth someone eyeballing if `gpt-5.5` ever needs its own rung.

No missing-test gaps for the change as scoped: the regression this PR fixes (silent opus-first misroute) is directly covered, and the fallthrough-fallback path was deliberately re-verified in both directions per the PR body.

Issue #1282, opened 2026-09-11
Pool-parked 429: no opt-in path to extra usage once every seat's 5h window is spent


robincle:
**Version:** 6.0.50 (Bun, macOS, pool mode)

### What happened

With every seat inside a live 5h window, the proxy answers locally:

```
API Error: Request rejected (429)
All 5 pool seats are over their rate-limit windows; the earliest resets in 2427s.
Nothing was sent upstream.
```

The windows were genuinely exhausted, so the 429 itself is correct — this is
not a mis-parked-seat report (that is #1244, where seats are parked at ~1%
utilization). My question is about what options an operator has *after* the
verdict is correct.

### The gap

The account has extra usage available. There appears to be no supported way to
let it absorb traffic once the subscription windows are spent, and the two
relevant mechanisms are independent:

- `POOL_PARKED` (`proxy.js:2738`) decides **before** any upstream call, purely
  from the per-seat `anthropic-ratelimit-unified-reset` verdict.
- The overage-guard (`overage-guard.js`) is **reactive** — it trips only after a
  response carries `representative-claim: overage` / `api` / another credit
  bucket, then halts the whole proxy with a 503.

So `DARIO_OVERAGE_GUARD=off` and `--overage-behavior=warn` disarm the second
mechanism, but as far as I can tell from the code they do not affect the first:
a parked pool still short-circuits to 429 without a request leaving the process.

### Ask

Either of these would resolve it:

1. **Document** that disabling the overage-guard does not change pool-parked
   behaviour, and that extra usage is unreachable while seats are parked. The
   guard docs read as the knob for "let per-token billing happen", which makes
   the current split non-obvious.
2. **An explicit opt-in** for the parked case — e.g.
   `--pool-exhausted=<429|extra-usage>`, defaulting to today's `429` — so an
   operator who knowingly wants to spend extra usage can let one request through
   and find out from the real response rather than a locally synthesized verdict.

I appreciate that the guard exists precisely because silent reclassification to
per-token billing is the failure mode nobody wants (#288, #38, #23). This is a
request for a deliberate, opt-in escape hatch, not for weakening the default.

### Happy to test

I can run either behaviour against a live 5-seat pool and report what Anthropic
actually returns for a request issued into a spent 5h window with extra usage
enabled — that data point seems missing from the thread history.

askalf:
Your reading of the code is right on all three points, and I checked each against 6.0.50 rather than from memory:

[...]

PR #1283, opened 2026-09-11
Document the way out of a parked pool, and what it costs (v6.0.51)

askalf:
Closes the documentation half of #1282. **Docs and a test only — no behaviour changed.**

## What the report got right

Verified against 6.0.50 source, all three claims hold:

- `writePoolParked` (`proxy.ts:3097`, decided at `:3241`) answers from `pool.parkedUntil()` alone — nothing upstream.
- `overage-guard.ts` reads the `representative-claim` on a **completed** response record, so it structurally cannot fire pre-flight.
- The parked decision block contains no reference to the guard. The knobs are parsed in `cli.ts` and reach only `OverageGuard`.

So `DARIO_OVERAGE_GUARD=off` genuinely does not change the parked answer, and the docs do read as though it should.

## What it could not see from the code

**The escape it asks for already exists.** `poolAccount = pinnedAccount` lands at `proxy.ts:3230`, *before* `parkedUntil` is consulted at `:3241` — so a seat pin (`x-dario-account`) never reaches the short-circuit. `docs/admin-api.md` already describes the primitive ("no failover… the upstream status comes back as-is"); nothing connected it to the parked case.

## Two consequences, both measured

The new test asserts them so they cannot regress into a surprise:

| | |
|---|---|
| **The default guard halts the proxy behind it** | The moment the pinned response bills to anything but the subscription, every request after it — pinned or not — gets `503 dario_overage_guard`. The pin is only usable with `--overage-behavior=warn` or the guard off. |
| **A successful pinned response un-parks that seat** | Its snapshot replaces the rejection, so the seat leaves `rejected` and ordinary traffic lands on it again until the next 429. The pin opens the door; it does not hold it open for one request. |

That second one is the real subtlety: the two mechanisms are independent in the *parking* decision but joined in the *outcome*, which is a sharper statement than "the guard does not affect parking".

`--pool-fallback` is named in the docs as the different answer it is — it keeps the gateway serving by going to another provider, so it never spends Anthropic extra usage.

## Deliberately not done here

The report's second ask — a first-class `--pool-exhausted=<429|extra-usage>` — is a fair shape and is **not** in this PR. The primitive exists; what is missing is a non-admin, proxy-level flag, plus a decision on whether "one probe" should mean one request or, as today, the door left open. The reporter has offered to run a pinned request against a live five-seat pool with extra usage enabled and paste the real headers. That is the one input neither the code nor a synthetic upstream can supply, and it should settle the flag's semantics rather than have them guessed at.

## Verification

- `test/pool-parked-pinned-escape.mjs` — **21/21**, in-process against an injected upstream.
- Full suite **197/197**.
- `check:readme` (93 links) and `lint:pkg` clean.
- The synthetic 200 carries no rate-limit headers, so it replaces the seat snapshot with an empty one. A real extra-usage 200 may carry overage-bucket headers and leave a different state — called out in the issue thread rather than asserted here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,31 @@ checklist.
 
 ## [Unreleased]
 
+## [6.0.51] - 2026-09-11
+
+### Documentation
+
+- **How to reach extra usage once every seat is parked, and what it costs (#1282).** The parked
+  429 is decided from the seats’ own rate-limit verdicts before anything is sent, and the
+  overage-guard is reactive — it reads the `representative-claim` on a response that already came
+  back. Neither knob unlocks the other, so `DARIO_OVERAGE_GUARD=off` does not change the parked
+  answer; the reporter read that correctly off the source. What was missing is that the escape
+  already exists: a **seat pin** (`x-dario-account`) is assigned before the parked check runs, so a
+  pinned request goes upstream and the real status comes back. `multi-account-pool.md` now says so,
+  next to the two consequences that make it a deliberate choice — the default guard **halts the
+  proxy** behind a response that bills to anything but the subscription, and a successful pinned
+  response **un-parks that seat** for ordinary traffic. `--pool-fallback` is named as the different
+  answer it is: it keeps the gateway serving by going to another provider, never spending
+  Anthropic extra usage.
+
+### Added
+
+- `test/pool-parked-pinned-escape.mjs` — 21 assertions over the whole path: an ordinary request
+  answered locally with nothing upstream, a pinned request reaching upstream on its seat and
+  returning the real status, the seat leaving `rejected` afterwards, the guard halting behind it by
+  default, and the same probe staying survivable under `warn` and with the guard off.
+
+
 ## [6.0.50] - 2026-09-10
 
 ### Fixed
--- a/docs/multi-account-pool.md
+++ b/docs/multi-account-pool.md
@@ -106,6 +106,36 @@ curl http://localhost:3456/analytics    # per-account / per-model stats, burn ra
 
 **When every seat is parked.** A pool whose seats are all `rejected` inside live windows does not probe them again: dario answers the request itself with `429`, `retry-after` set to the earliest reset, `x-dario-upstream-rejection: pool_parked`, and nothing sent upstream. One log line marks the transition (`pool parked: all 6 seats are over their rate-limit windows, earliest resets in 21m`). Before 6.0.35 every such request re-probed the earliest-reset seat, so `rejectedCount` on that seat grew by one per request — a seat reading `rejected_count: 500` next to `request_count: 1` was that, not a seat that needed a re-login. With a `--pool-fallback` armed, the request goes to the fallback instead, as before.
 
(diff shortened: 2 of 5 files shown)


sprayberry-redline (approved):

**Verdict: Approve — no blocking issues found.**

I reviewed the complete 261-line change: the release metadata, the new parked-pool guidance, and its targeted integration test. The documented order is grounded in the implementation: pin assignment precedes `parkedUntil`, and the test covers the locally answered unpinned request, the authenticated pinned upstream request, resulting seat state, and guard halt/warn/off outcomes. I also checked the current CI signal: all reported required checks are passing.

What's good: the documentation clearly distinguishes the pin's bypass of the parking short-circuit from the independent reactive overage guard, including the operational consequences and fallback alternative.

sprayberry-secondread (commented):

**Verdict:** No blocking issues — docs and test only, and the technical claims hold up against the source.

## What I checked

- Confirmed this PR touches no runtime source: `CHANGELOG.md`, `docs/multi-account-pool.md`, `package.json`/`package-lock.json` (version bump 6.0.50 → 6.0.51), and a new test file `test/pool-parked-pinned-escape.mjs`.
- Verified the central claim in `docs/multi-account-pool.md:109-113` against `src/proxy.ts`: `poolAccount = pinnedAccount` (around `src/proxy.ts:3230`) is assigned before `pool.parkedUntil()` is consulted (`src/proxy.ts:3241`), so a pinned request (`x-dario-account`) does bypass the parked short-circuit as the docs describe.
- Verified the overage-guard is reactive: `src/overage-guard.ts:10-11` — "The guard subscribes to the Analytics record stream — every completed request emits a record carrying its `claim`" — confirming it cannot fire pre-flight, consistent with the doc's claim that `DARIO_OVERAGE_GUARD=off` does not change the parked answer.
- Verified the seat-pin admin gate (`src/seat-pin.ts`) is pre-existing, unchanged by this PR, and uses `timingSafeEqual` for the admin token — the new docs describe existing, already-safe behavior rather than introducing new attack surface.
- New test file `test/pool-parked-pinned-escape.mjs` is auto-discovered by `test/all.test.mjs`'s `readdirSync`-based driver (confirmed no explicit registration needed), and each `startProxy()` instance's lack of an explicit `.close()` matches the existing convention in the other `*.mjs` test files (process exits via `process.exit()` at end of file).
- `gh pr checks 1283` — all checks pass (build, test, live-test, compat, CodeQL, lint, etc.).

## Findings

None that meet the bar for this review. The PR is narrowly scoped, the doc claims are traceable to specific source lines and check out, and the new test (21 assertions across 3 scenarios: default guard, `warn`, and guard off) exercises the documented escape path and both stated consequences (guard halts on overage response; pin un-parks the seat for ordinary traffic afterward) against a synthetic upstream.

## What's good

- The PR body is unusually precise about what it does and does not claim to fix, explicitly deferring the `--pool-exhausted` flag design to a follow-up pending real-world header data — a reasonable scope boundary.
- The new test's synthetic-200-has-no-rate-limit-headers caveat is called out in the PR body rather than silently asserted as general truth, which is honest about the limits of what an in-process test can prove.

PR #1286, opened 2026-09-11
Finish a streamed answer that dies mid-way from the other subscription (v6.1.0)

askalf:
## The answer does not stop when the plan does

A streamed `/v1/messages` or `/v1/chat/completions` answer that broke part-way through ended where it broke: no `message_stop`, no `[DONE]`, an SDK throwing "stream ended without producing a Message", and no failover because bytes were already on the wire. This finishes the **same client stream** from the other subscription.

```
client ──▶ dario ──▶ Claude pool   ✗ dies after 1,240 chars
                 └─▶ ChatGPT plan  ✓ resumes at char 1,241, same stream
```

Live, on the box, through this branch's build against the real subscriptions (Opus 5 cut → GPT-5.6-terra; GPT-5.6-sol cut → Sonnet 5; prose and code), the seams verbatim:

```
…each side must learn the other's starting point in order to det<CUT>ermine which bytes are new…
…an optional hours component, an optional minutes
    component, and an optional<CUT> seconds component. Components must appear in
…This tells the server that the client wants to<CUT> open a connection and lets the server…
    Raises ValueError if the<CUT> string is empty or does not match the expected format.
```

Left of the cut is one model, right of it the other. 4/4 runs, anchor matched exactly in all four, every stream one grammatically valid message.

## How

`src/midstream.ts` sits between the request handler and the client socket at both dispatch sites (the Claude read loop in `proxy.ts`, and `forwardToCodex`):

1. every client-bound frame passes through a guard that tracks what the client has seen and **withholds** a terminal error frame (and, once `forwardToCodex` flags `response.failed`, the polite `message_delta`/`message_stop` the translator emits for a failed turn);
2. `finish()` replaces `res.end()` on the streaming exits. Clean stream → ends as before. Dead-but-unfinished stream with content on the wire → the **client's own request** is re-issued through dario's front door (a loopback `POST`, so the pool, the codex translator, cch and every request rule apply to the resume) at the other provider's entry in `--pool-fallback`, with the partial answer appended as the assistant turn and a resume request as the user turn;
3. the resume's first text block continues the client's open block (no new `content_block_start`), its thinking is dropped, later blocks are renumbered, and it closes the message. An SSE comment (`: dario continuation gpt-5.6-terra (codex fleet) after 1240 chars`) marks the seam.

Two decisions that were settled empirically and are easy to get wrong again:

- **No assistant prefill.** Claude 4.6+ answers a trailing assistant turn with a 400; the Responses API never had it. The resume is instruction-driven on both providers.
- **The seam is a whitespace problem.** Told merely to "continue", Claude-as-continuer dropped the boundary space 2/3 times in the spike (`replies<cut>with`, a 3-vs-4-space indent). The user turn asks the model to repeat the last ~40 characters verbatim; dario holds the head of the resume, finds the repeat with a whitespace/quote-normalized match, cuts it, and streams the rest. The model renders the join inside its own token stream; dario only trims.
- **The resume turn is written as the user asking for the rest, not as a transport notice.** The first live run caught `claude-sonnet-5` answering `[transport notice] … resume now` with *"Note: that 'transport notice' isn't an actual system message — it's just text in your prompt"* and stopping — exactly the injection-awareness it should have. "My connection dropped while you were writing that reply, so I only received it up to this point: «…». Please pick up exactly where you left off…" is an ordinary request and gets the ordinary answer (second commit).

Target selection reuses the chain a mid-flight 429 uses: `pickCodexFallback` over `selectPoolFallbackForBody` for a Claude stream, the already-computed `claudeTarget` for a codex stream. The request's queue slot is released before the loopback so a `--max-concurrent=1` proxy resumes instead of waiting on itself (asserted in the wiring test). The loopback carries `x-dario-continuation`; a request carrying it is never continued itself.

**Not continued, on purpose:** a cut inside a `tool_use` block or an OpenAI `tool_calls` delta (half a JSON argument is not resumable), non-streaming requests (nothing on the wire), the api-key OpenAI backend, `/v1/complete`. Those end exactly as before. No chain entry for the other provider → ends as before, one log line says why.

**Switches:** on by default (it only acts where the alternative is a broken stream). `--no-midstream-continue` / `DARIO_MIDSTREAM_CONTINUE=0`.

## Evidence

- `test/midstream.mjs` — 81 assertions over the pure pieces (frame splitter, client-stream state, anchor/trim/seam rules, resume body, splicer on both wire shapes, guard against a fake loopback).
- `test/midstream-continuation-wiring.mjs` — 32 assertions against a real `startProxy` (fake Anthropic upstream via `fetchImpl`, codex stub via `DARIO_CODEX_BASE_URL`, temp HOME): Claude stream dies → resumed on codex on both wire shapes; codex `response.failed` → resumed on the Claude pool; in-band `overloaded_error` withheld and resumed; `tool_use` cut left alone; `--no-midstream-continue`; no nesting; `--max-concurrent=1` slot accounting. Every continued stream replayed through a strict grammar check of the wire shape.
- Full suite 199/199.
- Spike before any of this (outside the proxy, prod 6.0.51, real Opus 5 + real ChatGPT Plus, both directions, prose + code): 10 runs, zero restarts, zero preamble, zero repetition; the official `@anthropic-ai/sdk` `MessageStream` accepted every spliced stream as one message.
- Live on the box through this branch's build (second instance on :3999 sharing the store, chaos taps cutting the real Anthropic stream and a relay cutting the real codex stream): 4/4 above.

## Trade-offs and unknowns for review

- **`message_start.model` stays the primary's.** A header can't be added after streaming started; the SSE comment is the only served-by signal. If a served-by marker in the wire itself is wanted (a custom event, a `message_delta` field), that is a follow-up — I did not invent a non-standard frame.
- **Resume TTFT is the stall the user sees.** 3–9 s in the live runs; the spike saw two outliers (24 s, 79 s) it could not attribute (queue wait vs upstream). The new `continuation done: +N chars in Nms` log line makes it observable from now on.
- **One extra request on the other subscription**, carrying the whole conversation plus the partial. The original is logged/counted as it was; the resume is logged as its own request.
- **Backpressure during the resume is not honoured** — the spliced frames are written without `waitForDrain`. A resume is a few KB; flagging it because the read loop is careful about this.
- **A second JSON.parse per streamed frame** (the guard beside the analytics parser) on every guarded stream. Small; could share a parse later.
- **Codex Anthropic-shape failures were already invisible to streaming clients** (`response.failed` → `message_delta` + `message_stop`, pre-existing). With the guard present that turn is now continued instead; with `--no-midstream-continue` behaviour is unchanged.
- Bumped to **6.1.0** in-PR (new default-on behaviour); `[Unreleased]` promoted, notes extract cleanly.


sprayberry-redline (changes requested):

Verdict: request changes — the continuation path can report a truncated second stream as a successful completion.

### Blocking — correctness: `src/midstream.ts:819-820`
> `if (!sawContent) return false;`
> `for (const out of splicer.end()) this.o.write(out);`

If the fallback provider sends some content and then its socket resets (or otherwise ends its SSE response) without an `error` frame and without its terminal event, `sawContent` is true and this branch calls `Splicer.end()`. That method manufactures `message_delta`/`message_stop` (or `finish_reason: "stop"` and `[DONE]`) when no terminal was observed. The client consequently accepts a partial fallback answer as a clean completed answer, contrary to the intended failure semantics; this is exactly a realistic upstream failure mode the feature is meant to handle.

```suggestion
// Track whether the resume supplied its wire terminal event, and only splice/close
// normally when it did. If its body ends without one, preserve the unfinished
// stream (or forward an available failure frame) rather than synthesizing a
// successful end_turn/stop.
```

I reviewed the changed continuation state/splicer and both proxy/codex wiring paths, plus the new tests and documentation. CI is green; I did not run the local suite.

askalf:
Fixed in 82c9ad1. `Splicer.end()` is gone; the splicer records whether the resume delivered its own terminal (`message_stop` / `[DONE]`). If it did, nothing is added. If it did not, `abandon()` releases only the text still held for the anchor check and the stream is left unfinished — no `content_block_stop`, no stop reason, no `[DONE]` — so the client sees the truncation exactly as it would have on the primary. `finish()` reports `continued-unfinished`; the log line reads `continuation ended without its terminal event after +N chars — stream left unfinished`.

Covered three ways: `test/midstream.mjs` (splicer `abandon()` synthesizes nothing; guard end-to-end against a fake loopback whose body closes without a terminal), and `test/midstream-continuation-wiring.mjs` scenario H (real `startProxy`, codex stub drops the socket mid-resume → the second provider's text is on the wire, no `message_stop`, no `message_delta`). Docs and CHANGELOG say the same.

sprayberry-redline (dismissed):

I reviewed the live head `82c9ad14d28fab410b61422593581833c7525005`, focusing on the re-review fix and the mid-stream guard/splicer integration in `src/midstream.ts`, `src/proxy.ts`, and `src/codex-backend.ts`. In particular, the changed terminal-event handling now releases held continuation text without synthesizing `message_stop`/`[DONE]` after a second truncation, while retaining normal clean termination when the continuation supplies its own terminal event. I also checked the queue-slot release before the loopback continuation and the loopback nesting guard.

What's good: the fix directly addresses the prior false-completion path, keeps the downstream stream deliberately unfinished on a second failed continuation, and the added unit/wiring coverage described in the PR is appropriately targeted. The reported `advise` CI check is passing; no other required check was reported.

sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the reviewed streaming-continuation implementation.

I reviewed the complete changed production surface in `src/midstream.ts`, the proxy and Codex-backend wiring, CLI switch plumbing, and the new unit/wiring coverage. In particular, I checked terminal-frame withholding/release, prevention of nested continuation requests, queue-slot release before the loopback request, client authentication propagation, continuation target selection, and the incomplete-second-stream path. The completed CI checks are green; `live-test` was still pending at review time.

What's good: the change explicitly keeps a second interrupted continuation unfinished rather than synthesizing a successful terminal frame, and the hermetic end-to-end wiring test exercises both Anthropic and OpenAI wire shapes plus the one-slot queue case.

PR #1288, opened 2026-09-11
the ChatGPT subscription is a pool, not a single seat (v6.0.54)

askalf:
Addresses #1244 (the pooling half) and directly prevents a repeat of the 2026-09-07 outage.

## The bug, in one line

```js
export async function selectCodexAccount(preferredAlias?: string) {
  ...
  return [...all].sort((a, b) => a.alias.localeCompare(b.alias))[0];
}
```

`selectCodexAccount` returned the **alphabetically first** account, every time. `dario add altman` has always stored as many ChatGPT seats as you like — and dario used exactly one of them.

That is why the account-wide 429 on 2026-09-07 took the whole GPT lane down. A healthy second seat sat there, unreachable, while every request failed over to Claude and the two-family review guarantee quietly became Claude reviewing Claude.

## What changes

A seat that answers 429 is cooled for as long as the upstream asked — `retry-after`, else a minute, capped at fifteen — and the next request picks a peer. A success clears it; a recovered seat returns on its own.

When **every** seat is cooling, the pool says so locally rather than spending a request that can only 429 again (`allCodexAccountsCooled`, `codexPoolRetryAfterMs`).

Reuses `ProviderCooldowns` keyed by **alias** instead of provider name. It was already the right shape — arbitrary string key, injectable clock, entries dropped on read — so this adds no second cool-down implementation.

## Rotation is per-conversation, and that is the whole design

The Codex prompt cache is **scoped to the serving account**. A conversation that builds a prefix on seat A reads nothing from it on seat B, and this lane measures **59% cache share in production against a 73% controlled ceiling**. Rotating per request would have traded a rate-limit problem for a cache problem and come out behind.

So a conversation binds to a seat and stays there until that seat actually declines, at which point the binding follows it to a peer. Bindings are bounded at 500, oldest evicted first — losing one costs a cache miss, never correctness.

## Two deliberate non-choices

**Not headroom routing like the Claude pool.** Claude reports `anthropic-ratelimit-*` on every response, so that pool reads utilisation *before* it picks. The Codex backend states nothing until it 429s — the only signal is the decline itself plus its `retry-after`. Fill-first with cool-down eviction is what the available signal supports. If the backend ever reports utilisation, that is where headroom goes.

**An explicit pin stays an instruction.** `x-dario-account` / `DARIO_CODEX_ACCOUNT` is honoured even while that seat is cooling: the caller asked *that* account a question and is entitled to its answer, 429 included. Silently answering from a different seat would misattribute the reply.

## Tests

24 assertions in `test/codex-pool.mjs`, running against a sandboxed `HOME` so it never reads the operator's real seats.

Verified to discriminate: against the old always-first selection it fails **6/6** on the cases that matter — rotation after a decline, the all-cooled fail-fast, and both stickiness-follows-the-decline cases.

Covers the 2026-09-07 scenario end to end: three seats, decline each in turn, confirm it walks to the peer, confirm the pool answers locally once all three are cooling, confirm a cleared seat comes back.

198/198 suites green. preflight clean, README line-count guard ok.

## Not in this PR

Mid-flight failover — retrying a *different* seat inside the same request after a 429 — is the next increment. Today the first request after a decline is the one that pays; with this merged, it is one failed request instead of an entire lane.


sprayberry-redline (changes requested):

Verdict: request changes — the new per-seat cooldown is bypassed by the existing provider-wide cooldown, so a 429 from one seat still takes the entire Codex lane out of rotation.

### Blocking — `src/proxy.ts:3715-3717`
> ```ts
> if (d.status !== 429) return;
> providerCooldowns.note('codex', d.retryAfterMs);
> noteCodexDecline(d.alias, d.retryAfterMs);
> ```

With `alpha` and `bravo` configured, a request served by `alpha` that receives a 429 executes both calls. The new `noteCodexDecline` correctly cools only `alpha`, but the simultaneously-written `providerCooldowns.note('codex', ...)` marks the whole Codex provider as cooling. The normal routing gate then treats Codex as unavailable (`canAttempt('codex', ..., providerCooldowns)`), so the next request never reaches `selectCodexAccount` to choose healthy `bravo`; it falls through to Claude or returns the all-providers-limited response. This preserves the single-seat outage behavior this change is intended to remove. The added unit test calls `noteCodexDecline` and `selectCodexAccount` directly, so it does not exercise this proxy path.

```suggestion
// For a multi-seat Codex pool, a 429 is a seat-level condition.  Keep
// provider-wide cooldown only for an upstream condition that truly applies
// to every seat, and let normal routing select a non-cooled account.
noteCodexDecline(d.alias, d.retryAfterMs);
```

Please also add a proxy-level regression test that sends a 429 from one configured Codex alias and verifies the following request is dispatched to a different healthy alias.

What's good: the account-selection helpers correctly exclude cooled aliases, honor explicit pins, and the focused pool unit coverage covers the all-seats-cooled bookkeeping. CI was green for the completed checks; `wire-drift` was pending when reviewed.

askalf:
Good catch — you were right, and chasing it with the proxy-level test you asked for surfaced **two more** defects. All three were invisible to the unit suite, which calls `selectCodexAccount` directly and never sees the routing gate.

**1. The one you found.** Cooling the provider on any single 429 defeats the pool exactly as you described.

I did not take the suggested edit verbatim, though — dropping provider cooling for 429s breaks the other direction. On a **single-seat** deployment (what the fleet box actually runs) nothing would fail fast, and every request would re-hammer a seat already known to be limited instead of falling through to Claude. The condition the provider cool-down was always trying to express is **all seats cooling**, so that is what it now keys on: the seat cools unconditionally, the provider only once every seat is cooled. Both cases hold.

**2. `onDecline` fired only inside the `deferOnUnavailable` branch.** With no `--pool-fallback` configured, a 429 went straight to the client and the seat was never cooled at all — so selection returned the same limited account forever. The seat said no; that is true whether or not somewhere else exists to send the request. Moved out of the branch.

**3. The clear was keyed on `forwardToCodex` returning `true`.** That return means *"I wrote a response"*, which is equally true when what it wrote was the upstream's 429. So the cool-down was erased one line after being recorded. This is the one that actually kept the pool inert, and it only showed up under the integration test — every unit assertion passed the entire time. Now keyed on a 2xx from the outcome, which already carries the status and the alias.

**`test/codex-pool-seat-rotation.mjs`** asserts on observed traffic rather than internal state: which seat's bearer token the stub upstream actually saw. Two seats, 429 the first, prove request three carries `bravo`'s token and that `alpha` is not re-probed while cooling; then 429 both and prove the lane stops spending upstream requests entirely.

199/199 green, preflight clean.

sprayberry-secondread (commented):

**Verdict:** The seat-level cool-down/rotation mechanism (the fix for the 2026-09-07 outage) is correct and well tested, but the per-conversation stickiness half of `selectCodexAccount` is dead code on the actual request path — it is only exercised by its own unit test.

## Findings

### Medium — Codex conversation stickiness is never invoked from the request path

`src/codex-accounts.ts:170-172` adds:
```ts
export async function selectCodexAccount(
  preferredAlias?: string,
  opts?: { stickyKey?: string | null },
): Promise<CodexAccountCredentials | null> {
```
and the CHANGELOG/PR body both claim "a conversation binds to a seat and stays there until that seat actually declines." But every production call site of `selectCodexAccount` in `src/proxy.ts` (lines 2383, 2977, 3389, 3634) and `src/compare.ts:187` calls it with zero arguments — `opts` is always `undefined`, so `key` is always `null` and the sticky branch (`src/codex-accounts.ts:184-194`) never runs. Claude's pool has real stickiness wiring (`stickyKey`, `pool.selectSticky`, `pool.rebindSticky` in `src/proxy.ts:4025-4131` and around 4813/4916/4950), but nothing in `proxy.ts` calls `selectCodexAccount(alias, { stickyKey })` or reads `codexStickyAliasFor` outside of `test/codex-pool.mjs`.

**Failure scenario:** a conversation makes turn 1 on seat `alpha`, builds a prompt-cache prefix there. Turn 2 arrives — `selectCodexAccount()` is called with no sticky key, so it just returns "the first seat alphabetically that is not cooling," which happens to be `alpha` again only by alphabetical coincidence, not by binding. As soon as a lower-alias seat frees up mid-conversation (e.g. `alpha` cools then recovers while another seat was serving), or once there are enough seats that turn 2 doesn't land back on the same one deterministically in all cases (e.g. a decline elsewhere reshuffles the "first free" pick), the conversation's cache is silently dropped — exactly the cache regression the PR body says this design exists to avoid ("Rotating per request would have traded a rate-limit problem for a cache problem and come out behind"). The 24 assertions in `test/codex-pool.mjs` prove `selectCodexAccount(undefined, { stickyKey })` works in isolation; `test/codex-pool-seat-rotation.mjs` proves rotation end-to-end but never asserts stickiness end-to-end (it has no second conversation/turn scenario), so CI is green while the integration gap is unexercised.

```suggestion
// in proxy.ts wherever a stickyKey is already computed for the Claude pool,
// pass the same key through to Codex selection, e.g.:
const stored = await selectCodexAccount(pinnedAlias, { stickyKey });
```

### Info — `allCodexAccountsCooled()` re-lists accounts on every 429, not just when it matters

`src/codex-accounts.ts:203-206`:
```ts
export async function allCodexAccountsCooled(): Promise<boolean> {
  const aliases = await listCodexAccountAliases();
  return aliases.length > 0 && aliases.every((a) => codexCooldowns.isCooled(a));
}
```
This is awaited from inside every codex-decline handler (`src/proxy.ts:2437-2440`, `:3836` area), so a burst of 429s from a still-partially-healthy pool triggers a filesystem/account listing read per decline just to decide whether to cool the provider. Given cool-downs already carry `retryAfterMs` and the check is cheap relative to a network round trip, this is unlikely to matter in practice — flagging only because it re-does work that could be scoped to "did the cooled-count change" rather than a full re-list on every decline.

## What's good

- The seat/provider cool-down split (`src/codex-backend.ts:150-153`, `src/proxy.ts:2418-2441` and its proxy.ts:3810s mirror) is correct: recording the decline unconditionally (`src/codex-backend.ts:1068-1075`, moving `onDecline` outside the `deferOnUnavailable` branch) fixes a real bug where a 429 with no `--pool-fallback` configured never cooled the seat at all.
- Clearing the cool-down on `o.status >= 200 && o.status < 300` rather than on `forwardToCodex` returning `true` (`src/proxy.ts:3786`ish) is a real, well-explained fix — the old code cleared the cool-down on a served-429 response, one line after recording it.
- `test/codex-pool-seat-rotation.mjs` is genuinely useful: it is the only test that observes which seat's bearer token the upstream actually saw, which is exactly the signal needed to catch the provider-cooldown-short-circuit bug the PR itself found and fixed mid-review.
- I checked `gh pr checks` — all CI (build/test/compat/wire-drift) is green, consistent with these being real but narrow gaps rather than build breaks.

I read the full diff (`CHANGELOG.md`, `package.json`/lock, `src/codex-accounts.ts`, `src/codex-backend.ts`, `src/proxy.ts` hunks, both new test files) and cross-checked call sites against a shallow clone of `feat/codex-pool` at the reviewed head. I did not run the test suite locally; CI is green per `gh pr checks`.

sprayberry-redline (changes requested):

Verdict: request changes — model-incompatible first peers prevent retrying a later compatible seat.

### Blocking — `src/proxy.ts:3789-3800`
> `let codexPeer = await selectCodexAccountExcluding(codexTried).catch(() => null);`
>
> `if (peerSlugs && !isCodexModel(rawModel, peerSlugs)) codexPeer = null;`
>
> `fetch, canDefer || codexPeer !== null,`

With three available seats, suppose `alpha` is selected and returns 429, `bravo` sorts next but its cached model list does not contain `rawModel`, and `charlie` does contain it. The changed code obtains only `bravo`; after detecting that it cannot serve the model, it replaces the peer with `null` rather than continuing the selection past `bravo`. If there is no Claude fallback, `deferOnUnavailable` is then false, so `forwardToCodex` writes alpha's 429 to the client. Even when a Claude fallback exists, the request falls to Claude rather than the usable `charlie` Codex seat. This contradicts the stated peer retry behavior and makes heterogeneous account model availability fail unnecessarily.

```suggestion
// Keep excluding candidates known not to support rawModel until finding a
// compatible (or unknown-capability) peer.
let codexPeer: CodexAccountCredentials | null = null;
const peerTried = new Set(codexTried);
while (true) {
  const candidate = await selectCodexAccountExcluding(peerTried).catch(() => null);
  if (!candidate) break;
  const peerSlugs = rawModel ? peekCodexModelSlugs(candidate.alias) : null;
  if (!peerSlugs || isCodexModel(rawModel!, peerSlugs)) {
    codexPeer = candidate;
    break;
  }
  peerTried.add(candidate.alias);
}
```

What's good: I checked the live re-review head, the new account cooldown and sticky-selection paths, the proxy retry path, and the added pool tests. CI is green (including `test`, build matrix, and CodeQL); I did not run the suite locally.

sprayberry-redline (changes requested):

Verdict: request changes — the advertised per-conversation binding is not connected to request routing.

**Blocking — `src/codex-accounts.ts:446-473`**

> `opts?: { stickyKey?: string | null },`
>
> `const key = opts?.stickyKey ?? null;`
>
> `if (key) bindCodexSticky(key, free.alias);`

The new selector only reads or records a sticky binding when its new optional `stickyKey` argument is supplied. The proxy's production account selection remains `selectCodexAccount()` (without that argument), so no incoming conversation ever supplies a key or reaches `bindCodexSticky`. Consequently, two requests for the same conversation are independently assigned the first non-cooled alias; if that alias changes because another request cools it, the conversation loses its intended binding even though its own seat has not declined. This defeats the PR's cache-preservation invariant and contradicts the documented per-conversation rotation behavior. Please derive a stable conversation identifier at the proxy call site and pass it as `stickyKey`, with a proxy-level test that sends repeated requests for the same conversation while another account's state changes.

```suggestion
const stickyKey = conversationKeyFromRequest(parsedBody, req);
const stored = await selectCodexAccount(undefined, { stickyKey });
```

What's good: I reviewed the live re-review diff, including the all-seats cooldown gate and the peer-model scan/failover loop. The added proxy-level coverage exercises the 429 peer retry and heterogeneous model catalog path; all reported CI checks are green.

sprayberry-redline (changes requested):

Verdict: request changes — all-seats-cooled handling is asynchronous, so an exhausted Codex-only pool can return a generic 503 rather than the promised local rate-limit response.

### Blocking — `src/proxy.ts:597-600`
> `noteCodexDecline(d.alias, d.retryAfterMs);`
>
> `void allCodexAccountsCooled().then((all) => {`
>
> `  if (all) providerCooldowns.note('codex', d.retryAfterMs);`
>
> `}).catch(() => { /* a status read must never fail a request */ });`

When the last usable seat answers 429 in the mid-flight loop, `noteCodexDecline` makes that seat cooling synchronously, but the provider cooldown is deferred behind `listCodexAccountAliases()`. The loop can immediately finish with no peer (`if (served || !codexPeer) break` at `src/proxy.ts:608`) before this detached promise updates `providerCooldowns`. With no configured Claude target, the following fallback construction takes the generic `upstream_unavailable`/503 path rather than `writeAllProvidersRateLimited(['codex'])`; it also cannot use the imported `codexPoolRetryAfterMs` to describe the pool cooldown. Thus an all-cooled pool does not reliably fail locally as a rate limit, contrary to the new behavior described by this PR. The existing test only verifies that a subsequent request reaches no upstream; it does not assert the immediate last-429 response's status or retry information.

```suggestion
noteCodexDecline(d.alias, d.retryAfterMs);
if (await allCodexAccountsCooled()) {
  providerCooldowns.note('codex', await codexPoolRetryAfterMs());
}
```

Apply this synchronously in the decline path (or explicitly handle `allCodexAccountsCooled()` before leaving the loop), then add a proxy-level assertion that the request which exhausts the final seat returns the defined local 429 response and retry-after rather than a 503.

What's good: the re-review head adds an end-to-end peer-model scan and a same-request rotation test, which cover the previously missing heterogeneous-seat and routing-gate cases. CI's `advise` check is green; `auto-merge` is skipped.

askalf:
All three review points are addressed on the current head. Verified against the branch rather than from the diff, since the review was written at 23:09 and the branch moved at 01:36.

## Blocking — provider-wide cooldown defeating the per-seat one

Fixed, at **both** sites — `src/proxy.ts:2475` and `src/proxy.ts:3927`:

```ts
noteCodexDecline(d.alias, d.retryAfterMs);
void allCodexAccountsCooled().then((all) => {
  if (all) providerCooldowns.note('codex', d.retryAfterMs);
}).catch(() => { /* a status read must never fail a request */ });
```

The suggestion was to drop the provider-wide call entirely. That was not taken, deliberately, and the reasoning is in the code comment: on a **single-seat** deployment, removing provider cooling means nothing fails fast and every request re-hammers a seat already known to be limited instead of falling through to Claude. `allCodexAccountsCooled()` is the condition the provider cool-down was always reaching for — it now says what it meant.

Your diagnosis was exactly right; only the remedy differs, and in the direction of keeping a behaviour your suggestion would have removed.

## Blocking — "add a proxy-level regression test"

`test/codex-pool-seat-rotation.mjs` does this, and it drives the real proxy rather than the helpers:

```
line  98: const { startProxy } = await import('../dist/proxy.js');
line 112: const server = await startProxy({ ... fetchImpl: anthropicFetch ... })
line 118: fetch(`http://127.0.0.1:${PROXY_PORT}/v1/messages`, ...)
```

Two configured Codex seats, a stub ChatGPT backend, Anthropic through `fetchImpl`. A 429 from one alias, and the following request lands on the healthy peer. **PASS.**

That answers the specific concern that the earlier unit test called `noteCodexDecline` and `selectCodexAccount` directly and so could not have caught this.

## Second opinion — Codex stickiness was dead on the request path

Correct as filed, and now wired. `selectCodexAccount` is called with a sticky key on all three real request paths:

| site | |
|---|---|
| `proxy.ts:2421` | `{ stickyKey: codexStickyKeyForBody(body) }` |
| `proxy.ts:3442` | `{ stickyKey: codexStickyKeyForBody(clientBodyBytes) }` |
| `proxy.ts:3696` | `{ stickyKey: codexStickyKey }` |

Two bare calls remain and are intentional: `proxy.ts:3015` is `/v1/models` discovery, which needs *any* account to list slugs and has no conversation to bind; `compare.ts:187` is the comparison tool, not a conversation path. Binding either would be stickiness with nothing to be sticky about.

## State

- `npm test` — **202/202**
- All eight checks green, including **Wire drift**, which was pending when this was reviewed and is the one thing the review flagged as unfinished.

Re-requesting review so the gate can re-read the current head.

sprayberry-redline (changes requested):

Verdict: changes requested — the asynchronous all-seats check can install a stale provider-wide cooldown after a healthy seat has recovered.

**Blocking — correctness — `src/proxy.ts:570-573`**

> `noteCodexDecline(d.alias, d.retryAfterMs);`
>
> `void allCodexAccountsCooled().then((all) => {`
>
> `  if (all) providerCooldowns.note('codex', d.retryAfterMs);`
>
> `}).catch(() => { /* a status read must never fail a request */ });`

The condition is evaluated asynchronously, but the resulting provider cooldown is written later without verifying that it is still true. With concurrent requests, the last-limited-seat request can observe every seat cooling; before its continuation calls `providerCooldowns.note`, another in-flight request can succeed on a just-recovered seat and run `clearCodexDecline` / `providerCooldowns.clear`. The delayed continuation then re-cools the entire `codex` provider. `canAttempt('codex', ...)` will subsequently short-circuit, so that healthy seat is not selected until the stale retry window expires — the pool again unnecessarily fails over despite an available seat.

Make the all-seats decision and provider state update atomic with respect to Codex cooldown mutations, or re-check the pool state immediately when applying the provider cooldown (and avoid a stale check writing after a success), for example:

```suggestion
noteCodexDecline(d.alias, d.retryAfterMs);
void allCodexAccountsCooled()
  .then(async (all) => {
    if (all && await allCodexAccountsCooled()) {
      providerCooldowns.note('codex', d.retryAfterMs);
    }
  })
  .catch(() => { /* a status read must never fail a request */ });
```

A shared serialization/version check around both `noteCodexDecline` and `clearCodexDecline` would be stronger than the illustrative re-check and closes the remaining interleaving.

What's good: I reviewed the changed seat selection, sticky binding, 429 decline propagation, mid-flight peer scan, and the new hermetic pool tests. The peer scan and the proxy-level rotation coverage directly exercise important routing paths. CI is green (including CodeQL, test, live-test, and Node build matrices). No local test suite was run.

askalf:
## Correction first

My previous comment said "all three review points are addressed on the current head." That was wrong, and I want to be precise about how.

I read the 23:09 review and answered it. I did not notice that the gate had re-reviewed five times since, most recently at 02:00 against `e123b2a`, and that the 02:00 verdict was still blocking on a defect the earlier review had not raised. So I declared the PR clear against a stale reading of its own review history. The blocking finding was real and unaddressed at the time I said it was addressed.

## The actual defect

From the 02:00 review:

> The condition is evaluated asynchronously, but the resulting provider cooldown is written later without verifying that it is still true.

Concretely, the shape I had shipped:

```ts
noteCodexDecline(d.alias, d.retryAfterMs);
void allCodexAccountsCooled().then((all) => {
  if (all) providerCooldowns.note('codex', d.retryAfterMs);
});
```

`every(isCooled)` decides, the promise resolves, and the continuation writes a tick later. In that gap another in-flight request can succeed on a just-recovered seat and call `clearCodexDecline`. The late write then cools the **entire** codex provider against a pool that is healthy again, `canAttempt('codex')` short-circuits, and the healthy seat is skipped until the stale window expires.

That is the single-seat outage this PR exists to prevent, reintroduced by the bookkeeping meant to prevent it. The finding is correct and it is the more serious of the two, because it converts a partial outage into a full one.

## The fix — `4c9d72c`

The review suggested re-checking inside the continuation. I did something strictly stronger, because a re-check narrows the window without closing it: the re-check is itself another `await`, with the same gap behind it.

Instead, take the alias list first, then decide **and** write with no suspension point between them:

```ts
noteCodexDecline(d.alias, d.retryAfterMs);
void listCodexAccountAliases().then((aliases) => {
  if (allAliasesCooled(aliases)) providerCooldowns.note('codex', d.retryAfterMs);
}).catch(() => { /* a status read must never fail a request */ });
```

`allAliasesCooled` is synchronous, so that callback runs as one unit and nothing can interleave. A tick-stale alias list is harmless: a seat added in that window is not cooling, so "all cooled" is false on the next decline anyway.

Applied at both decline sites — `proxy.ts:2107` and `proxy.ts:3579`.

## Proving it, rather than asserting it

`test/codex-pool-cooldown-race.mjs` (new, 16 assertions). Two things worth calling out, because my first two attempts at this test were worthless and I want the failure modes on the record:

**It stages the race deterministically, with promise ordering rather than timers.** Continuations on one promise run in registration order, so registering the recovery *first* places it exactly in the gap — after the decision, before the write. No sleeps, no flake.

**It asserts the two shapes DISAGREE.** A test that passes against the bug it describes has proven nothing. My first version staged the recovery before the check resolved, so the check saw it and the bug never reproduced — 12 passed, and it was measuring nothing. The `the two shapes disagree` assertion is what makes that failure visible instead of silent.

**It greps the compiled call sites.** The section above proves the two *shapes* differ in the abstract; without this, the bug could be reintroduced at either call site and every other assertion would still pass. Verified by reverting both sites in `dist/proxy.js` and confirming those two assertions fail:

```
=== the shipped call sites use the safe shape ===
  FAIL no call site decides across an await :: 2 found
  FAIL both decline paths take the alias list first :: 0 found
```

Then restored byte-identical.

## On the earlier disagreement

The 23:09 review suggested dropping the provider-wide cool-down entirely. I still haven't taken that, deliberately: on a single-seat deployment, removing it means nothing fails fast and every request re-hammers a seat already known to be limited instead of falling through to Claude. `single-seat pool cooling IS "all"` pins that behaviour so it can't be quietly removed later.

That disagreement stands on its own merits. It has nothing to do with the TOCTOU finding, which I should have addressed four reviews earlier.

## State

- `npm test` — **203/203** (202 + the new file)
- `npm run build` clean
- Version slot 6.2.2; `origin/master` is 6.2.1, so no collision
- CHANGELOG 6.2.2 entry extended with the same-tick guarantee

Re-requesting review against `4c9d72c`.

sprayberry-redline (dismissed):

I reviewed the pool selection, per-seat decline tracking, all-seats provider cooldown gate, sticky conversation binding, and the mid-flight peer retry/model scan in `src/codex-accounts.ts`, `src/codex-backend.ts`, and `src/proxy.ts`. I also reviewed the added focused regression coverage. CI is green (including test, build, CodeQL, and live-test); I did not run the test suite locally.

What's good: the revised retry loop excludes attempted/cooling seats, scans past peers with known-incompatible model lists, records 429s even without a configured fallback, and only clears a seat cooldown after a successful 2xx response. The all-seat test-and-provider-cooldown write is kept synchronous once aliases are read, avoiding the previously identified stale all-cooled race.

sprayberry-redline (changes requested):

Verdict: changes requested — Codex mid-flight seat failover is bypassed on the Claude-to-Codex fallback route.

- **blocking — `src/proxy.ts:2447-2451`**
  > `const hasNextOption = openaiBackend !== null && shape === 'openai';`
  > `... fetch, hasNextOption,`

  The newly added peer retry loop is only used by the primary Codex route. This changed fallback call still permits `forwardToCodex` to defer only when an API-key backend follows it, rather than when another Codex seat is available. Reproduction: configure a Claude→Codex pool fallback, two Codex accounts (`alpha`, `bravo`), no `openaiBackend`, and make `alpha` return 429 while `bravo` is healthy. The fallback selects `alpha`; `hasNextOption` is false, so `forwardToCodex` writes alpha's 429 and returns served. `noteCodexDecline` cools alpha for later requests, but this request never attempts bravo. That contradicts the PR's mid-flight peer failover and leaves a healthy seat unused for this supported fallback direction.

  Route this call through the same excluding-seat retry logic (or make its defer decision account for an eligible Codex peer) and add a proxy-level regression test for the Claude-to-Codex fallback path.

  ```suggestion
  // Determine an eligible untried Codex peer before forwarding, then defer
  // when either the API-key backend or that peer can serve the request; retry
  // the peer on a false return just as the primary Codex route does.
  const deferOnUnavailable = hasNextOption || codexPeer !== null;
  ```

What's good: the primary Codex dispatch correctly cools individual aliases, avoids provider-wide cooling until all seats are cooled, and tests the primary multi-seat rotation/peer-model cases. CI had its completed checks green; several checks were still pending at review time.

sprayberry-redline (dismissed):

Verdict: APPROVED — no blocking issues found in the live-head re-review.

I reviewed the changed Codex seat-selection, cooldown, sticky-binding, and same-request peer-failover paths, including the 429 decline hook and cached peer-model scan. The retry loop excludes already-tried aliases, skips peers whose cached model lists cannot serve the requested model, records 429s before retrying, and only clears a seat cooldown after a 2xx response. I also checked the added pool/race/model-scan tests and CI: completed checks are green; `compat` and `live-test` were still pending when reviewed.

What's good: the re-review changes preserve the important distinction between seat-level cooldowns and provider-level all-seats-cooled fail-fast, while making peer retry bounded and model-aware.

askalf:
Blocking finding confirmed and fixed in `a618a04`. This one was real and I had missed it in the same way the last one was missed: I put the peer retry into the primary Codex route and never asked whether the *other* route needed it.

## Verified before fixing

`tryCodexPoolFallback` selected one seat, forwarded once, and computed its defer decision as:

```ts
const hasNextOption = openaiBackend !== null && shape === 'openai';
```

With a Claude→Codex fallback on the anthropic wire shape and no api-key backend, **both halves are false**. So the reproduction in the review is exact: the fallback picks `alpha`, `alpha` 429s, `forwardToCodex` writes that 429 to the client and reports served, `alpha` is cooled for *later* requests — and this request never tries `bravo`. A healthy seat unused, on the one route that exists for when the other provider has already given up.

## The fix

That route now runs the same excluding-seat loop as the primary one, and defers when **either** an api-key backend or an eligible Codex peer can take the request.

One thing your suggestion didn't cover, which I hit while implementing it: the fallback route picks its *model* from the seat's slugs (`pickCodexFallback(fallbackModels, slugs)`). Reusing the first seat's pick on a peer would send a model that peer may not list, trading a 429 for a 400. So the model is re-picked per seat, and a seat listing none of the fallback models no longer ends the attempt — a peer may still list one.

I also kept `attempted.add('codex')` on the attempt itself rather than hoisting it above the guards. Moving it up would have told the rest of the request that codex had been tried when it hadn't, suppressing a later legitimate attempt — a regression I introduced mid-edit and caught before committing.

## The regression test you asked for

`test/codex-pool-fallback-seat-rotation.mjs` — proxy-level, configured as the reproduction describes (anthropic shape, no api-key backend), asserting on **observed traffic**: which seat's bearer the stub upstream saw, inside which client request.

Verified discriminating by reverting only the defer widening in `dist/proxy.js` and confirming it fails exactly where it should:

```
=== MID-FLIGHT on the FALLBACK route: alpha 429s, bravo rescues it ===
  FAIL the client still got a 200 :: 429
  FAIL two seats were touched in ONE request :: alpha
  OK   the limited seat was tried first
  FAIL the healthy peer served it :: alpha
```

Then restored byte-identical. With the fix: `200`, `alpha` then `bravo`, one client request.

It pins the other direction too — with no peer and no api-key backend the client still gets the real upstream 429, not a manufactured 503. Widening a defer condition is exactly the kind of change that quietly turns honest errors into generic ones, so that assertion is there on purpose.

## State

- `npm test` — **205/205**
- Merged `origin/master` (#1292 wire-drift feed) and the branch's #1293 workflow pin; the CHANGELOG collision from #1292 is resolved by folding its note into the 6.2.2 section, since that code ships in 6.2.2 rather than staying "unreleased"
- Version slot 6.2.2; `origin/master` is 6.2.1

Head is `090f71c`.

sprayberry-redline (changes requested):

Verdict: request changes — all-cooled Codex pools can be misrouted to Claude rather than fail fast or use the configured fallback.

### Blocking — `src/proxy.ts:3697-3699`
> `const codexStickyKey = parsedBody ? computeStickyKey(extractFirstUserMessage(parsedBody)) : null;`
> `if (await hasAnyCodexAccount()) {`
> `  const stored = await selectCodexAccount(undefined, { stickyKey: codexStickyKey });`

`selectCodexAccount` now deliberately returns `null` when every stored seat is cooling, but this call treats that result exactly like there are no Codex credentials: `codexCreds` remains null and the subsequently constructed routing context reports no usable Codex account. Therefore a request for a Codex-listed (or `codex:`-forced) model after every seat has received a 429 falls through to the Claude default path, bypassing the intended provider cooldown/all-providers-rate-limited handling. This can send a GPT request to a semantically different Claude model rather than returning the local pool-exhausted result (and is especially incorrect when no compatible Claude fallback is configured).

Handle the all-cooled result separately from an empty account set before routing, returning the established rate-limited response or deliberately invoking only a configured compatible fallback. Add a proxy-level test for a request after all seats are cooled.

```suggestion
const stored = await selectCodexAccount(undefined, { stickyKey: codexStickyKey });
if (!stored) {
  if (await allCodexAccountsCooled()) {
    writeAllProvidersRateLimited(['codex']);
    return;
  }
  // Continue only for the genuinely unavailable/no-account case.
}
```

What's good: the changed forwarding paths correctly record the declining alias and exclude already-tried seats during mid-flight retries; CI's `advise` check is green. I reviewed the changed pool selection, decline handling, routing loops, and new test surface; I did not run the local suite per review environment policy.

askalf:
Merged `master` (#1291, the Responses API / v6.3.0). Four conflicts, none mechanical — flagging the two that changed behaviour rather than text.

**`src/proxy.ts`.** #1291 extracted the codex reporting callback into a named `codexOnDone` and added a Responses-shape passthrough branch; this branch added the seat-rotation loop and the per-seat cool-down hooks. Both are kept: `codexOnDone` is now the shared reporter for both shapes, and the translated Messages path keeps the excluding-seat retry loop.

Taking master's reporting body verbatim **dropped this branch's "a 2xx clears the seat cool-down" line** — one of the behaviours under review here. Restored into `codexOnDone`, where it now covers both shapes instead of only the Messages path. Worth calling out because a merge that compiles and passes is exactly how that line disappears quietly.

**The Responses passthrough deliberately does not get seat rotation.** `forwardResponsesToCodex` has no decline hook and no defer contract, so there is nothing to defer on and no `retry-after` to read. Wiring rotation into it means changing #1291's surface, which belongs in its own change rather than in a merge resolution. It is commented at the branch so it is visible rather than rediscovered.

**Version.** master shipped 6.3.0 while this PR sat at 6.2.2, so the slot was stale. Re-slotted to **6.4.0** — the change is a feature, so it takes the next minor above master rather than a patch on a version that already shipped. Exactly the two root version fields in the lockfile.

**CHANGELOG.** master *released* as 6.3.0 the section this branch had renamed to 6.2.2, so the headings collided. This branch's pool bullets now have their own `[6.4.0]` section above the released `[6.3.0]`, rather than being folded into a release that has shipped. The wire-drift note this branch had adopted while it was still unreleased now lives only in 6.3.0 where master released it — one feature, one release.

- `npm test` — **207/207** (205 here + #1291's two new files)
- All 16 checks green on `981b421`

Head is `981b421`.

sprayberry-redline (changes requested):

Verdict: changes requested — the new pool still repeatedly selects a rate-limited seat for native Responses requests.

**Blocking — `src/proxy.ts:4008`**
> ```ts
> // No seat rotation on this path: forwardResponsesToCodex is a byte
> // passthrough with no decline contract, so there is nothing to defer
> // on and no retry-after to read. A 429 here is written to the client
> // as the backend sent it.
> served = await forwardResponsesToCodex(
> ```

A native `/responses` client using a subscription model takes this changed branch. If its selected seat receives a 429, `forwardResponsesToCodex` writes that response rather than reporting a decline; consequently `noteCodexDecline` is not called and the sticky/first-free selector continues to choose the same seat on the next request. A healthy peer is never used, reproducing the single-seat outage for the native Responses API even though the PR advertises account pooling generally. The current new tests exercise translated/fallback flows, but do not cover this branch.

```suggestion
// Give forwardResponsesToCodex the same decline callback contract as
// forwardToCodex, then use the existing tried-seat retry loop (including
// retry-after seat cooling) for the native Responses branch.
```

What's good: the translated and Claude-to-Codex fallback paths now use bounded peer selection, per-seat cooldowns, and the CI checks are green.

askalf:
Blocking finding confirmed and fixed in `c9eb272`. You were right and my comment was the wrong call — I want to be explicit about that, because I wrote the comment you quoted.

I said the native Responses branch couldn't rotate because `forwardResponsesToCodex` is a byte passthrough with no decline contract, and filed it as #1291's surface. The premise was true; the conclusion wasn't. A 429 on that branch called no decline hook, so `noteCodexDecline` never ran, so selection handed the **same** rate-limited seat back on every following request. A healthy peer was never reached — the single-seat outage this whole change exists to remove, surviving on the one wire shape Codex CLI actually speaks. "No decline contract" was a reason to *give* it one, not a reason to leave it out of the pool.

## The fix

`forwardResponsesToCodex` now takes `onDecline` and `deferOnUnavailable` with the same semantics `forwardToCodex` has: a 429 or 5xx is the seat declining, reported whether or not anything is waiting to take over, and deferring returns `false` without writing so the caller can retry on a peer. The branch moved inside the loop, on the same seat sequence and the same defer condition as the translated path.

**The decline handler is now defined once and passed to all three forwards** — the Claude-to-Codex fallback, the translated Messages path, and the native Responses passthrough. It had been hand-copied per call site, and that is exactly how one path ended up cooling nothing while its neighbour cooled correctly. A fourth call site now inherits it by construction. I verified the two copies were identical logic before collapsing them, rather than assuming.

## Tests

`test/codex-pool-responses-seat-rotation.mjs` drives real `/v1/responses` requests and asserts on observed traffic. It covers both halves, because the second is the one that was silently broken:

1. mid-flight — a 429 on seat one is rescued by seat two inside the **same** request;
2. the cool-down actually lands — the **next** request skips the limited seat.

Verified discriminating by reverting the decline wiring in `dist` — 7 of 12 assertions fail, in exactly the pattern you described:

```
=== MID-FLIGHT on the NATIVE path: alpha 429s, bravo rescues it ===
  FAIL the client still got a 200 :: 429
  FAIL two seats were touched in ONE request :: alpha
=== the 429 actually COOLED the seat — the next request skips it ===
  FAIL alpha was not re-probed while cooling :: alpha
  FAIL bravo served it directly, no wasted attempt :: alpha
```

Then restored byte-identical.

The cooldown-race guard now pins the stronger invariant this created: one decision site, one handler defined once, three call sites passing it. Reintroducing a per-site copy fails the test rather than merely differing in the abstract.

## Also in this push

Merged master (#1294, web search, v6.4.0). Worth flagging one thing, since it would have passed review silently: master released **6.4.0**, the slot this branch re-slotted onto one merge earlier. Both sides wrote the same `## [6.4.0]` heading, so git auto-merged the *heading* and conflicted only on the bullets under it — either resolution would have filed this branch's bullets inside a release that has already shipped, under a heading that looked right. The version in `package.json` auto-merged to master's with no conflict at all, which is the same problem without the marker.

Re-slotted to **6.5.0**, and the released `[6.4.0]` heading handed back to master's web-search bullets. Verified by section, not by eye: pool and responses bullets under 6.5.0, web search under 6.4.0.

- `npm test` — **209/209**
- Head is `e92254d`

sprayberry-redline (approved):

Verdict: **Approved** — no blocking issues found in the live-head re-review.

I reviewed the live-head diff, concentrating on the newly added per-seat cooldown, sticky-seat selection, all-seats provider cooldown transition, and retry loops for both Messages and Responses wire shapes. I also reviewed the added proxy-level regression coverage. CI is fully green (including test, build, CodeQL, analyze, compat, and live-test); I did not run the local test suite per review policy.

What's good: the shared `codexOnDecline` handler correctly records the declining alias before the asynchronous alias-list lookup, and `allAliasesCooled` makes the subsequent all-seat decision and provider-cooldown write synchronously. The retry loops exclude attempted aliases and probe cached model availability before selecting a peer, which avoids both retry loops and known-incompatible peer attempts. The new tests cover the cooldown interleaving plus the fallback and native Responses paths.

No blocking findings.

PR #1289, opened 2026-09-11
Resume on the same model first; the other provider is the second hop (v6.2.0)

askalf:
## Resume on the same model first; the other provider is the second hop

Follow-up to #1286. 6.1.0 knew one continuation target — the other provider's `--pool-fallback` entry — so a user with one plan and no chain got nothing, and a user with both got a model swap for what is usually a transient reset.

**First choice: the model the client asked for**, as a fresh request through dario's own front door. The pool picks a seat (sticky binding keeps the prompt cache warm), and if the provider cannot take the request at all the existing pre-byte failover already hands it to the other one. This is what most dying streams get, and it works with one plan.

**Second choice: the other provider's chain entry**, taken when the first delivers nothing (refused, unreachable, dead before its first byte) or when the resume itself dies mid-way. In the second case the resume's own guard makes the hop, so the client stream carries two seam comments:

```
: dario continuation claude-opus-5 (same model) after 309 chars
: dario continuation gpt-5.6-terra (codex fleet) after 300 chars
```

`x-dario-continuation` now carries the depth; depth 2 is never continued — two hops, never three. A guard on a depth-1 request skips choice 1 (that model just failed twice). The log names who served each continuation (`continuation done by claude-opus-5: +1451 chars …`).

## Evidence

- `test/midstream.mjs` 93 (choices in order; depth-1 guard starts at choice 2; every choice refused → `resume-failed`; inner seam comment forwarded; `continuationDepth`).
- `test/midstream-continuation-wiring.mjs` 43 — rewritten around per-request plans so each scenario states what the primary, the same-model resume, and the other provider do: same-model resume on both wire shapes; same-model refused → other provider; same-model dies → second hop inside the resume, both seams on the client stream; codex `response.failed` → codex again; codex fails twice → Claude pool; nobody left for the second hop → left unfinished; `tool_use` cut, flag off, depth 2. Requests dario makes on its own (probes) no longer consume a plan entry — the earlier version was order-sensitive to them.
- Full suite 199/199.
- **Live on the box** through this build (second instance sharing the store, chaos taps on the real Anthropic stream and a relay on the real codex stream): same-model resumes **4/4** — Opus 5 → Opus 5, GPT-5.6-sol → GPT-5.6-sol, prose and code, anchor exact:
  ```
  …the other end acknowledges what<CUT> it has received. The catch is that both directions…
  …confirm that it has received the other side's<CUT> sequence number. This establishes…
  ```
  and the forced two-hop **2/2** — Opus dies, Opus resume dies, GPT-5.6-terra finishes it; GPT dies, GPT resume dies, Sonnet 5 finishes it — each one valid message with two seams.

## Trade-offs

- A same-model resume usually lands on the seat that just died (sticky binding). If that seat is the problem rather than the connection, the cost is one wasted hop before the other provider; the pool learns nothing from a mid-stream death because no headers come back with it.
- Two hops means up to two extra requests; the live two-hop cases stalled the client ~13–22 s total.
- Bumped to 6.2.0 in-PR.


sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the live head.

I reviewed the continuation-depth and target-selection changes in `src/midstream.ts` and both proxy integration paths in `src/proxy.ts`, plus the updated unit/wiring coverage. In particular, I checked the depth-1 path starts at the cross-provider choice, depth-2 requests cannot create another guard, failed pre-content loopbacks advance to the next choice, and nested seam comments are forwarded. CI is green, including `test`, `build`, `CodeQL`, `compat`, and `live-test`.

What's good: the bounded depth protocol is explicit (`x-dario-continuation` carries depth rather than merely acting as a boolean), and the hermetic wiring cases cover both wire shapes, refused first choices, a dying same-model resume followed by the second hop, and exhaustion without manufacturing a terminal completion.

PR #1290, opened 2026-09-12
Doctor reports continuation; a chaos tap to watch it happen (v6.2.1)

askalf:
## Doctor reports continuation; a chaos tap to watch it happen

Two small things that make the feature from #1286/#1289 visible.

**`dario doctor` — a `Continuation` row next to `Failover`**, saying which hops a dying stream can take on this host:

```
[ OK ]  Continuation  on: a dying stream resumes on the same model, then on gpt-5.6-sol → claude-sonnet-5 (two hops)
[ OK ]  Continuation  on: a dying stream resumes on the same model only — add --pool-fallback for a second hop on the other subscription
[INFO]  Continuation  off — a stream that dies mid-answer ends truncated (unset DARIO_MIDSTREAM_CONTINUE / drop --no-midstream-continue)
```

Configuration only, like Failover: it does not claim a hop works. `continuationReadiness()` is pure; every branch is in `test/doctor-formatter.mjs`.

**`DARIO_CHAOS_CUT_AFTER=<chars>`** (`DARIO_CHAOS_CUT_STREAMS=<n>`) kills the first streamed answer after that many characters — from dario's side, the way a real reset does — so anyone can watch the continuation with any client:

```bash
DARIO_CHAOS_CUT_AFTER=300 dario proxy
```

`chaosCutFetch` wraps the upstream fetch for both the Claude leg and the codex leg, recognises both providers' text framing, spares resumes (so what you see is the first hop), and dario warns loudly at startup while it is set. It is the in-process form of the rig the live tests used on the box; it is a demo and test affordance, never a default.

## Evidence

- `test/midstream.mjs` 98 (`chaosCutFetch`: cut after N, logged, next stream untouched, resumes spared, non-stream paths pass through).
- `test/midstream-continuation-wiring.mjs` 48 (scenario L: a third proxy with the tap set, a healthy upstream, one cut, one valid message with a same-model seam, the following stream untouched).
- `test/doctor-formatter.mjs` 52.
- Full suite 199/199.
- Bumped to 6.2.1 in-PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,19 @@ checklist.
 
 ## [Unreleased]
 
+## [6.2.1] - 2026-09-12
+
+### Added
+
+- **`dario doctor` reports mid-stream continuation.** A `Continuation` row next to `Failover` says
+  which hops a dying stream can take on this host: same model then the chain (two hops), same
+  model only (no chain, or a chain with nowhere to go — pointing at the Failover row), or off.
+  Configuration only, like Failover; `continuationReadiness()` is pure and every branch is tested.
+- **`DARIO_CHAOS_CUT_AFTER=<chars>`** kills the first streamed answer after that many characters
+  (`DARIO_CHAOS_CUT_STREAMS=<n>` for the first n), from dario's side, the way a real reset does, so
+  the continuation can be watched on demand with any client. Spares resumes; applied to the Claude
+  and the codex leg alike; warns loudly at startup. A demo and test affordance, never a default.
+
 ## [6.2.0] - 2026-09-11
 
 ### Changed
--- a/docs/midstream-continuation.md
+++ b/docs/midstream-continuation.md
@@ -134,6 +134,32 @@ is never closed with a synthetic `end_turn`; only the resume's own
 | `--pool-fallback=…` | where the second hop goes; without an entry for the other provider a stream gets the same-model resume only |
 
 On by default: it only ever acts where the alternative is a broken stream.
+`dario doctor` reports which hops this host can take:
+
+```
+[ OK ]  Continuation  on: a dying stream resumes on the same model, then on gpt-5.6-sol → claude-sonnet-5 (two hops)
+[ OK ]  Continuation  on: a dying stream resumes on the same model only — add --pool-fallback for a second hop on the other subscription
+[INFO]  Continuation  off — a stream that dies mid-answer ends truncated (unset DARIO_MIDSTREAM_CONTINUE / drop --no-midstream-continue)
+```
+
+## Seeing it happen
+
+Nothing about a healthy stream shows the feature, so there is a tap that
+kills one on purpose:
+
+```bash
+DARIO_CHAOS_CUT_AFTER=300 dario proxy
+```
+
+The first streamed answer dies after 300 characters — the upstream socket is
+cut from dario's side, exactly the failure a real reset produces — and the
+continuation finishes it. Point any client at the proxy, ask for something
+long, and watch the answer keep going past the cut; a raw `curl -N` shows the
+seam comment. `DARIO_CHAOS_CUT_STREAMS=3` cuts the first three instead of one.
+The tap spares resumes, so it shows the first hop — the same model finishing
+its own answer; the other subscription takes over only when that model cannot
+serve the resume. dario warns loudly at startup while the tap is set; it is a
+demo and test affordance, never a default.
 
 ## How it was proven
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@askalf/dario",
-  "version": "6.2.0",
+  "version": "6.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@askalf/dario",
-      "version": "6.2.0",
+      "version": "6.2.1",
       "license": "MIT",
       "bin": {
         "dario": "dist/cli.js"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@askalf/dario",
-  "version": "6.2.0",
+  "version": "6.2.1",
   "description": "Use your Claude and ChatGPT subscriptions in Cursor, Cline, Aider, Claude Code and the Agent SDK — at subscription pricing, not per-token API bills. One local Anthropic + OpenAI-compatible endpoint: either plan answers either wire shape, with automatic failover when one hits its limit.",
   "type": "module",
   "bin": {
--- a/src/doctor-core.ts
+++ b/src/doctor-core.ts
@@ -69,6 +69,34 @@ export interface Check {
  * live request, and this release was built on the lesson that a green config is
  * not a working path.
  */
+/**
+ * Mid-stream continuation readiness (v6.2.1). Configuration only, like
+ * failoverReadiness: it says which of the two hops a dying stream can take on
+ * this host, not that either works. The first hop (the same model again) needs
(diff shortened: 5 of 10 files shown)


sprayberry-redline (changes requested):

Verdict: changes requested — the chaos-stream limit is applied independently per provider rather than globally.

### Blocking — `src/proxy.ts:1456-1457`
> `const upstreamFetch: typeof fetch = chaosCut ? chaosCutFetch(opts.fetchImpl ?? fetch, chaosCut) : (opts.fetchImpl ?? fetch);`
> `const codexFetch: typeof fetch = chaosCut ? chaosCutFetch(fetch, chaosCut) : fetch;`

Each call to `chaosCutFetch` creates its own `left` counter (`src/midstream.ts:928`). Consequently, with `DARIO_CHAOS_CUT_STREAMS=1`, one Claude request and one Codex request are both deliberately cut: the Claude wrapper decrements its own counter and the Codex wrapper still has an untouched counter. This contradicts the startup promise that the *first 1 streamed answer* will be cut and makes a demo/test configuration cut up to twice the requested number of client streams (up to `2 * n` generally). The current unit test exercises only one wrapper, so it cannot catch the cross-provider case.

```suggestion
// Share one stateful wrapper/counter for every upstream leg.
const rawFetch: typeof fetch = opts.fetchImpl ?? fetch;
const cutFetch = chaosCut ? chaosCutFetch(rawFetch, chaosCut) : rawFetch;
const upstreamFetch = cutFetch;
const codexFetch = cutFetch;
```

If Codex must intentionally remain on the global built-in fetch rather than `opts.fetchImpl`, move the remaining-count state out of `chaosCutFetch` and pass that single shared state to wrappers around the two fetch implementations; the essential requirement is one counter for the proxy, not one per provider.

askalf:
Fixed in 440eb3d. The remaining-cuts counter now lives outside the wrappers (`chaosCutState`) and both legs share it — the codex leg has to stay on the global fetch (the Claude leg honours `opts.fetchImpl`, which tests use as the fake Anthropic upstream), so it is one state passed to two wrappers rather than one wrapper. `test/midstream.mjs` drives the Claude and codex wrappers off a single budget of 1 and sees the second leg find it spent.

sprayberry-redline (approved):

Verdict: approved — no blocking issues found in the live re-review.

The prior cross-provider cut-budget issue is addressed: `src/proxy.ts:1451-1453` creates one `chaosState` and supplies it to both wrappers, while `src/midstream.ts:947-948` checks and decrements that shared state. The added unit coverage exercises the two-wrapper budget case, and the integration coverage verifies a cut stream resumes and the following stream is untouched. I also reviewed the doctor readiness branches and the configuration parsing. CI is green except for the currently pending `compat` check; I did not run the local suite.

What's good: the state is explicitly scoped to the proxy rather than a provider wrapper, and the regression test directly covers the failure mode from the previous review.

sprayberry-secondread (commented):

**Verdict:** Not blocking, but `chaosCutFetch` has a URL-only stream detector that will corrupt a non-streaming response if the chaos tap is ever left on for a mixed workload — worth a one-line guard before this ships as a "demo affordance" people actually run against live traffic.

## Findings

### Medium — `chaosCutFetch`'s stream check is path-only, not content-type-aware; a non-streaming request on `/v1/messages` gets truncated and errors instead of passing through

`src/midstream.ts:933`
```ts
const isStream = /\/v1\/messages|\/responses/.test(url);
```
This only inspects the *request URL*, never `body.stream` on the outgoing request or `res.headers.get('content-type')` on the upstream reply. `/v1/messages` serves both streaming and non-streaming (`stream:false`) requests. If chaos mode is active and a client sends a non-streaming request that upstream answers with a large plain-JSON body (any single message over `afterChars` characters of `text`/`delta`-shaped JSON, which a `content` field can easily match), the wrapper still treats the JSON body as SSE, matches the regex against the raw JSON text, hits the char threshold, and calls `reader.cancel()` + `c.error(new Error('chaos: read ECONNRESET'))`.

Downstream, the non-streaming path in `src/proxy.ts:5260` does `await upstream.text()` with no try/catch around it — an error here is unhandled and will crash the request handler (or in Node's fetch/undici, reject the whole response promise chain), rather than emit the loud-but-graceful "cut a stream" behavior the feature intends only for actual streamed answers.

I verified this in isolation by feeding `chaosCutFetch` a non-streaming `Response` (`content-type: application/json`, `stream:false` in the request body, a 400-char `text` value): it logs `CHAOS: cutting this stream after 400 chars` and the resulting `res.text()` throws `chaos: read ECONNRESET` — exactly the failure a real reset would look like, but on a response `chaosCutFetch` was never supposed to touch.

```suggested fix
const isStream = /\/v1\/messages|\/responses/.test(url)
  && (res.headers.get('content-type') ?? '').includes('text/event-stream');
```

This is a demo/test-only affordance (never a default, per the PR description), so blast radius is limited to hosts that deliberately set `DARIO_CHAOS_CUT_AFTER`. Still, `doctor`'s own docs invite exactly this ("point any client at the proxy, ask for something long") and a client mixing streaming and non-streaming requests during that demo would hit it.

### Low — `chaosCutFetch` is instantiated twice with independent counters, so `DARIO_CHAOS_CUT_STREAMS=1` cuts up to 2 streams total, not 1

`src/proxy.ts:1456-1457`
```ts
const upstreamFetch: typeof fetch = chaosCut ? chaosCutFetch(opts.fetchImpl ?? fetch, chaosCut) : (opts.fetchImpl ?? fetch);
const codexFetch: typeof fetch = chaosCut ? chaosCutFetch(fetch, chaosCut) : fetch;
```
Each call to `chaosCutFetch(..., chaosCut)` creates its own `left` closure variable seeded from the same `chaosCut.streams`. Since the Claude leg (`upstreamFetch`) and the Codex leg (`codexFetch`) are two separate wrapped functions, a request that touches both legs (e.g. Claude 429s and pool-fallback tries Codex) can be cut once per leg — up to `streams` cuts on Claude *and* `streams` cuts on Codex, not `streams` cuts total as the flag name and the startup log (`the first ${chaosCut.streams} streamed answer(s) will be cut`) imply. Confirmed with a standalone re-implementation of the closure: two independently-constructed wrappers sharing the same `{streams:1}` object each report `left=0` after one call apiece.

```suggested fix
const chaosState = chaosCut ? { left: chaosCut.streams } : null;
// pass chaosState into chaosCutFetch and decrement the shared counter there,
// or construct a single chaosCutFetch wrapping a dispatch fn used by both legs.
```
Low severity because it is a demo/test-only counter used to script a scenario, not a safety-relevant limit, and CUT_STREAMS defaults to 1 either way; the counter just under-caps the total in a mixed Claude+Codex demo run.

## What's good
- `continuationReadiness` (`src/doctor-core.ts:79`) is pure, and every branch (off / two-hop / same-model-only-no-chain / chain-but-inert) is covered in `test/doctor-formatter.mjs`.
- The resume-sparing check (`bodyText.includes(ANCHOR_OPEN)`) correctly prevents the tap from cutting the very continuation it just triggered, and this is tested (`test/midstream.mjs`, "a resume is never cut").
- The startup warning for `DARIO_CHAOS_CUT_AFTER` is loud and unconditional, matching the PR's stated intent that this never runs silently.
- CI is green across all 15 checks (build, test, live-test, CodeQL, wire-drift, etc.).

I read the full diff (`CHANGELOG.md`, `docs/midstream-continuation.md`, `src/doctor-core.ts`, `src/midstream.ts`, `src/proxy.ts`, and the three test files) and cross-checked the two `chaosCutFetch` call sites and the non-streaming response path in `src/proxy.ts` directly against a shallow clone of the PR branch. I did not run the project's own test suite.

PR #1291, opened 2026-09-12
The OpenAI Responses API on dario: Codex CLI runs on a Claude plan (v6.3.0)

askalf:
## The OpenAI Responses API on dario — Codex CLI runs on a Claude plan

The other half of the chiasmus. Claude Code on a ChatGPT plan has worked since 5.5.89; this makes OpenAI's Codex CLI run on a Claude plan, tools and all, through the same proxy.

**Why a new endpoint:** Codex CLI 0.154 removed `wire_api = "chat"` for custom providers ([openai/codex#7782](https://github.com/openai/codex/discussions/7782)) — a provider speaks the Responses API or it cannot be used. The OpenAI Agents SDK speaks the same shape.

**How:** `POST /v1/responses` is translated once at the front door into the Messages body every other path already serves (`src/responses-inbound.ts`), then runs as an ordinary Anthropic-shape request — pool, template, failover, mid-stream continuation. Everything written back passes through a write boundary (a `Proxy` over the `ServerResponse`, installed before the body is read) that turns Anthropic SSE into the Responses event sequence, a buffered message into a response object, and an Anthropic error into the OpenAI envelope. Nothing downstream knows the client's shape. A ChatGPT-subscription model on this route is **passed through** to the codex backend untouched (`forwardResponsesToCodex`) — that is what keeps Codex's newest request features working there (`additional_tools` input items, `custom` tools, `reasoning.context`, `include`; none of which survive a Messages round trip), and the response ends on the terminal event rather than the backend's EOF because Codex closes its side the moment it has `response.completed`.

```toml
# ~/.codex/config.toml
model = "claude-opus-5"
model_provider = "dario"
[model_providers.dario]
name = "dario"
base_url = "http://127.0.0.1:3456/v1"
env_key = "DARIO_API_KEY"
wire_api = "responses"
```

## Evidence

- **Live with Codex CLI 0.154.0 on the box**, this build as a second instance on the shared store: a plain turn (`PONG CHIASMUS`); the full agent loop on `claude-sonnet-5` and `claude-opus-5` — Claude calls `exec_command`, Codex runs `ls -a` / `wc -c`, returns `function_call_output`, Claude answers from it — with Codex's 17 KB system prompt cached on the Claude side (98–99% cache reads from the second turn); the same loop on `gpt-5.6-sol` through the passthrough, no aborts.
- `test/responses-inbound.mjs` 50 (request translation on the captured Codex first turn and a tool round trip; buffered and streamed response translation; the write boundary's SSE/JSON decision).
- `test/responses-inbound-wiring.mjs` 25 against a real `startProxy`: Codex-shaped request on the Claude pool translated both ways; the tool round trip landing as tool_use/tool_result; the codex passthrough with `additional_tools` and the response ending on the terminal event (asserted < 250 ms against a backend that holds the socket 300 ms); non-stream on both legs; invalid JSON and `previous_response_id` in the OpenAI envelope.
- Full suite 201/201. Bumped to 6.3.0 in-PR.

## Trade-offs and what is not built

- **Dropped on the Claude pool, one `--verbose` line each:** hosted tool types (`web_search`, `file_search`, `mcp`…), `custom` freeform tools (Codex's JS `exec` orchestration tool for GPT profiles), inbound `reasoning` items (OpenAI's encrypted content), `text.format`. `previous_response_id` is a 400 — dario is stateless; Codex sends none.
- **No buffered response from a ChatGPT-subscription model on this route** (the backend streams; folding a Responses stream into a response object is not written) — a 400 naming `stream`.
- **Codex sends two request profiles**: for models it has metadata for (gpt-*) the newest one (`additional_tools`, `custom` tools, no top-level `instructions`); for unknown models (`claude-*`) the classic one. The translation handles both; the passthrough needs neither.
- Mid-stream continuation covers this route on the Claude pool (the translated request is ordinary) and not on the passthrough.
- The chat-shape codex path (`forwardToCodex` with shape `anthropic`) still waits for the backend EOF after the terminal event; a Responses client no longer takes that path, so I left it alone here.


sprayberry-redline (changes requested):

Verdict: request changes — the new ChatGPT Responses passthrough rejects a supported Responses API continuation before it reaches the native backend.

### Blocking — `src/proxy.ts:3476-3483`
> `if (isResponses && parsedBody !== null) {`
> `  ...`
> `  const t = responsesRequestToAnthropic(parsedBody);`
> `  ...`
> `  parsedBody = t.body;`

The unconditional translation runs before routing decides whether the request will use the native Codex/ChatGPT passthrough. `responsesRequestToAnthropic` rejects any `previous_response_id` (see its changed `src/responses-inbound.ts:169-171`), so a normal streamed `gpt-*` Responses follow-up such as `{ model: "gpt-5.6-sol", stream: true, previous_response_id: "resp_...", input: "continue" }` receives dario's 400 instead of being forwarded to the backend that natively supports that stateful API. This contradicts the passthrough's purpose of preserving request features that cannot survive a Messages round trip, and breaks multi-turn Responses clients on the ChatGPT plan. The wiring coverage only tests a first turn and therefore misses this path.

```suggestion
// Determine whether the selected provider is the native Codex Responses path
// before applying the Responses → Messages conversion. Preserve the parsed
// original body for that path, and call responsesRequestToAnthropic only for
// requests that will be served through the Claude/Messages route.
if (isResponses && parsedBody !== null && !willUseCodexResponsesPassthrough) {
  const t = responsesRequestToAnthropic(parsedBody);
  parsedBody = t.body;
  body = Buffer.from(JSON.stringify(t.body));
}
```

The request/response translators and the end-to-end first-turn/tool-round-trip coverage are otherwise thoughtfully structured. I reviewed the changed implementation and tests; CI still has pending `test`, `analyze`, `live-test`, and `wire-drift` checks, so I did not treat the pending suite as a passing signal.

askalf:
Fixed in 8f7b34c. The translator now reports `previous_response_id` as unsupported instead of refusing; the route decides: the codex passthrough forwards the original body untouched (`store` as the client sent it — a stateful client needs it on), and the Claude pool answers a 400 naming the field *after* routing, with the way out in the message. Wiring test covers both: the same field forwarded on `gpt-5.6-sol` and refused on `claude-sonnet-5` with no upstream request. Also addressed the CodeQL finding on the same block: a translation failure's 400 no longer echoes an arbitrary error's message — only the translator's own verdicts reach the client. (c0bb4e7 removes a stray file from another branch that rode along in the previous push.)

sprayberry-redline (changes requested):

Verdict: request changes — the ChatGPT Responses passthrough still rejects supported stateful follow-up requests before routing can forward them.

### Blocking — `src/proxy.ts:3481-3488`
> `if (isResponses && parsedBody !== null) {`
> `  responsesBodyRaw = parsedBody;`
> `  const t = responsesRequestToAnthropic(parsedBody);`
> `  ...`
> `  parsedBody = t.body;`

A streamed ChatGPT-model continuation such as `{ "model": "gpt-5.6-sol", "stream": true, "previous_response_id": "resp_...", "input": "continue" }` enters this unconditional translation before the later native-Codex passthrough selection. The changed translator explicitly records `previous_response_id` as unsupported, so this path returns a local 400 rather than reaching `forwardResponsesToCodex`, despite that backend natively supports stateful Responses requests. That breaks multi-turn Responses clients on the ChatGPT plan.

```suggestion
// Decide whether this is a native Codex Responses request before translating.
// Preserve the parsed client body for that path; translate only requests that
// will be served through the Claude/Messages route.
if (isResponses && parsedBody !== null && !willUseCodexResponsesPassthrough) {
  const t = responsesRequestToAnthropic(parsedBody);
  parsedBody = t.body;
  body = Buffer.from(JSON.stringify(t.body));
}
```

The first-turn/tool-round-trip translation and its test coverage are thoughtfully structured. I reviewed the changed request routing, translators, passthrough, and added tests. CI is green except `compat` and `wire-drift`, which are still pending; I did not run the local suite.

askalf:
The second review reads the translation site as refusing — it does not, and did not after 8f7b34c: `responsesRequestToAnthropic` only *records* `previous_response_id` in `t.unsupported`; no response is written there. The refusal lives after the codex branch (`if (isResponses && responsesUnsupported.length > 0)`, just before "Claude's turn"), which the codex passthrough never reaches — it returns after `forwardResponsesToCodex(rawRes, { ...responsesBodyRaw, model: rawModel }, …)` with the original body. Both halves are asserted end to end in `test/responses-inbound-wiring.mjs` scenario D: `{ model: gpt-5.6-sol, stream: true, store: true, previous_response_id: 'resp_prev' }` reaches the codex stub with `previous_response_id` and `store` intact and answers 200; the same field on `claude-sonnet-5` is a 400 naming it with no upstream request. 0c77343 says this at the translation site so the next reader lands on it. (67fa62f drops files from another branch that rode along in the worktree.)

sprayberry-redline (dismissed):

The prior concern is resolved: the route retains the original Responses payload and defers rejection of unsupported translated features until after the native Codex passthrough has had the opportunity to serve the request. I reviewed the live-head routing clarification at `src/proxy.ts:3476-3484` in context with the raw-body capture, native passthrough, and Claude-only rejection paths at `src/proxy.ts:3486-3494`, `src/proxy.ts:3857-3882`, and `src/proxy.ts:3972-3980`. Thus a ChatGPT Responses request with `previous_response_id` reaches the backend unchanged, while the Claude-pool path still returns the documented field-specific 400.

What's good: the change makes the routing invariant explicit at the point where the raw and translated bodies diverge, and the stated regression coverage covers the stateful passthrough case. CI is green except for the still-pending `compat` check; I did not run the local suite.

sprayberry-redline (changes requested):

**Verdict: request changes — the required `test` CI check is failing on this head, so this change is not presently merge-ready.**

### Blocking — CI failure
The required `test` job for live head `6bbb9a236ab6b6a006ef6df4a5d959944d20dcb8` fails: https://github.com/askalf/dario/actions/runs/34667300175/job/103481726195. Its log reports `codex-refresh-failure-ttl.mjs` failing with `listen EADDRINUSE: address already in use 127.0.0.1:41043` and concludes `pass 201`, `fail 1`. Please make the test run reliably (or rerun it if this was confirmed infrastructure-only) and obtain a green required check before merge.

I reviewed the new `/v1/responses` request translation, response write boundary, Codex Responses passthrough, routing order, and the added focused tests. I did not run the local suite, per review-environment policy. The other completed CI checks are green; `compat`, `live-test`, and `wire-drift` remain pending.

What's good: the route preserves the raw body until provider routing, so Responses-only fields can reach the native Codex backend while the Claude route can reject unsupported stateful requests explicitly; the streaming passthrough also terminates on the Responses terminal event rather than waiting for backend EOF.

askalf:
The `test` failure on 6bbb9a2 is the free-port race the suite already documents (`test/helpers/free-port.mjs`: the port is released before it is returned, so two of the eight parallel test files can draw the same number — `codex-refresh-failure-ttl.mjs` already uses `freePort()` for all three of its listeners). Infrastructure-only; the failed job is re-running. Nothing in this PR touches that test.

askalf:
bd07953 makes the suite reliable against that race rather than relying on reruns: the driver (`test/all.test.mjs`) now runs a file once more when its only failure output is `EADDRINUSE` — the free-port collision between parallel files — and says so in the log; every other failure is reported exactly as before. The rerun of the flaky job on 6bbb9a2 was green as well. No change to what any test asserts.

sprayberry-redline (changes requested):

Verdict: changes requested — the Responses passthrough changes the API's default statefulness.

### Blocking — `src/codex-backend.ts:943`
> `const upstreamBody = { ...body, stream: true, store: body.store === true };`

A normal Responses request that omits `store` is supposed to retain the API default (stored response/state enabled), but this line always sends `store: false`. For example, an Agents SDK client can make an initial streaming request without explicitly setting `store`, receive a response ID, then send its next turn with `previous_response_id`; the initial response was not stored, so the backend cannot resolve that ID. This contradicts the route's stated support for stateful `previous_response_id` passthrough and changes the semantics of every caller that relies on the default.

```suggestion
const upstreamBody = { ...body, stream: true };
```

Only override `store` if the proxy has a separate, documented reason to do so; otherwise preserve omission so the backend applies its own default.

I reviewed the live-head Responses request/response translation, proxy routing, and Codex passthrough. The ordinary test/build/CodeQL checks are green; `compat` and `live-test` are still pending.

What's good: routing unsupported Claude-pool fields after provider selection, while retaining the original body for native Codex passthrough, is a sound separation.

askalf:
Fixed in 50f56b5: the passthrough no longer touches `store` — sent as the client sent it, omitted when omitted, so the backend applies its own default and a stateful client's next `previous_response_id` resolves. Only `stream` is forced. Wiring now covers all three: `store: true` kept, `store: false` kept, omitted stays omitted.

sprayberry-redline (approved):

Verdict: **Approved** — no blocking issues found in the live-head re-review.

I reviewed the live-head follow-up to the previous requested change, including `src/codex-backend.ts` and the corresponding wiring coverage. The passthrough now preserves an omitted `store` field while still forcing `stream`, so the backend can apply its default and stateful clients can use a later `previous_response_id`. The new wiring assertion covers precisely that omitted-field path. CI is green apart from the currently pending `live-test` check.

What's good: the corrective change is narrowly scoped, documents the statefulness rationale next to the behavior, and tests the regression scenario.

PR #1292, opened 2026-09-12
The Claude Code wire-drift feed: a page, RSS and JSON Feed from git history

askalf:
## The Claude Code wire-drift feed

dario's template watcher already knows, every 30 minutes, whether what Claude Code sends on the wire has moved. This publishes that knowledge: **every change the watcher has ever observed, as a page with RSS and JSON Feed**, rebuilt from git history on every template change.

`https://askalf.github.io/dario/drift-feed/` (live once this merges and the first deploy runs)

`scripts/drift-feed.mjs` walks the commits that touched `src/cc-template-data.json` and diffs consecutive snapshots into the terms a reader cares about: beta flags added/removed by name, tools added/removed and schemas changed by name, header values added/removed/changed, body field order, the system prompt and each variant with its char delta and a bracketed «excerpt» of the changed span (whitespace-only edits called out). Two rules keep it honest: a release number inside a header (`user-agent: claude-cli/2.1.268 → .269`) is a version label, not a wire change; and a release that changed nothing on the wire still gets a line saying so — that is information too.

Today: **120 entries, 62 wire changes, since 2026-04-12.** Sample of what it reads like:

> **Claude Code 2.1.265: beta flags, tools, system prompt changed on the wire** — beta flag added: `mid-conversation-tool-changes-2026-07-01` · 1 tool schema changed: `Bash` · system prompt changed (−1 chars) *(whitespace only)* · …
> **Claude Code 2.1.235: system prompt changed on the wire** — system prompt changed (−51 chars): `…hand off mid-task.⏎«⏎<total_tokens>15000000 tokens left</total_tokens>⏎»` → `…hand off mid-task.⏎«»`

`.github/workflows/drift-feed.yml` builds it (fetch-depth 0) and deploys to GitHub Pages on every master push touching the template or the generator, daily as a backstop, and on demand. Nothing is committed — the feed *is* the history. `actions/configure-pages` with `enablement: true` creates the Pages site on the first run (a personal token could not: 403 on `POST /pages`); if org policy blocks that too, enabling Pages by hand (Settings → Pages → Source: GitHub Actions) once is the whole setup.

## Evidence

- `test/drift-feed.mjs` 17: a version-only release, a wire change with every kind, the whitespace-only case, the first observation (one line, not a wall of "added"), the three renderers, and HTML escaping of template-sourced text.
- Generated locally from this checkout and rendered in a fresh headless Chrome: light/dark, quiet version entries dimmed, wire changes with their change lists and expandable excerpts.
- No `src/` change; nothing releases. New actions in tag form per CLAUDE.md (Dependabot pins them Monday).

## Trade-offs

- The excerpt shows the *first* changed span only; a prompt edit in two places shows one. The diff link goes to the rebake PR for the rest.
- `_captured` on a version-label refresh is the previous bake's stamp; the page hides it there rather than show a stale "captured" time.
- The 2.1.235 entry exposes a capture artefact (`<total_tokens>` injected into the baked prompt that day) — that is what the watcher committed, and the feed reports history as it is.

### DIFF
--- a/.github/workflows/drift-feed.yml
+++ b/.github/workflows/drift-feed.yml
@@ -1,0 +1,73 @@
+name: Claude Code wire-drift feed
+
+# Publishes docs/drift-feed (page + RSS + JSON Feed) to GitHub Pages: every
+# change to what Claude Code sends on the wire, as the template watcher
+# observed it. Built from git history alone (scripts/drift-feed.mjs walks the
+# commits that touched src/cc-template-data.json), so there is nothing to
+# commit — a template rebake lands on master, this rebuilds and redeploys.
+#
+# Runs on every master push that touches the template or the generator, once
+# a day as a backstop, and on demand. Never from a fork (see #1039).
+
+on:
+  push:
+    branches: [master]
+    paths:
+      - src/cc-template-data.json
+      - scripts/drift-feed.mjs
+      - .github/workflows/drift-feed.yml
+  schedule:
+    - cron: '23 6 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: drift-feed-pages
+  cancel-in-progress: true
+
+jobs:
+  build:
+    if: github.repository == 'askalf/dario'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: askalf/checkout-with-retry@115a6407547e9711edbc2e915838d495cad9583f  # v1.1.0
+        with:
+          # The feed IS the history: every commit that touched the template.
+          fetch-depth: 0
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 22
+      - name: Build the feed from git history
+        run: |
+          node scripts/drift-feed.mjs --out dist-feed/drift-feed --site "https://askalf.github.io/dario/drift-feed"
+          # The Pages root: nothing lives there yet but this feed.
+          echo '<!doctype html><meta charset="utf-8"><meta http-equiv="refresh" content="0; url=drift-feed/"><title>dario</title><a href="drift-feed/">Claude Code wire drift</a>' > dist-feed/index.html
+          touch dist-feed/.nojekyll
+      - uses: actions/upload-pages-artifact@v4
+        with:
+          path: dist-feed
+
+  deploy:
+    needs: build
+    if: github.repository == 'askalf/dario'
+    runs-on: ubuntu-latest
+    permissions:
+      pages: write
+      id-token: write
+      # configure-pages' enablement reads the repo to create the site.
+      contents: read
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      # Creates the Pages site on first run (source: GitHub Actions) — the
+      # workflow token can, a personal token could not.
+      - uses: actions/configure-pages@v5
+        with:
+          enablement: true
+      - id: deployment
+        uses: actions/deploy-pages@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ checklist.
 
 ## [Unreleased]
 
+### Added
+
+- **The Claude Code wire-drift feed** — <https://askalf.github.io/dario/drift-feed/>: every change to what
+  Claude Code sends on the wire, as the template watcher observed it, as a page + RSS + JSON Feed.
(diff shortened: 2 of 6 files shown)


sprayberry-redline (approved):

Verdict: **Approved** — no blocking issues found in the live head.

I reviewed the full 466-line change: the history walker and snapshot differ, HTML/RSS/JSON renderers and escaping, the Pages workflow's scoped permissions/concurrency/deployment flow, and the new unit-test coverage. The repository CI checks are green; `live-test` is still pending. I did not run the test suite locally.

What's good: the feed generator keeps its diff operation pure and covers version-only, structural wire-change, whitespace-only, first-observation, renderer, and HTML-escaping cases. The deployment workflow is limited to the canonical repository and uses minimal job-specific permissions.

No blocking findings.

sprayberry-secondread (commented):

**Verdict:** No blocking issues; one real mislabeling bug worth fixing before the feed builds up history, plus a minor coverage gap.

## Findings

**1. Medium — `scripts/drift-feed.mjs:166,190-193` — "tool schema changed" also fires on description-only text edits**

```js
const hash = (v) => createHash('sha256').update(JSON.stringify(v ?? null)).digest('hex').slice(0, 10);
...
const schemaB = new Map((Array.isArray(b.tools) ? b.tools : []).map((t) => [t?.name, hash(t)]));
const schemaA = new Map((Array.isArray(a.tools) ? a.tools : []).map((t) => [t?.name, hash(t)]));
const reschemaed = [...schemaA.keys()].filter((n) => schemaB.has(n) && schemaB.get(n) !== schemaA.get(n));
if (reschemaed.length > 0) push('tool', `${reschemaed.length} tool schema${reschemaed.length === 1 ? '' : 's'} changed: ${reschemaed.map((n) => `\`${n}\``).join(', ')}`);
```

`hash(t)` hashes the whole tool object (`name`, `description`, `input_schema`), not just `input_schema`. Reproduced against the fetched head (`725b8415e`): feeding `diffTemplates` two templates whose `Read` tool has an *identical* `input_schema` but a wording-only `description` change (`"Reads a file from the local filesystem."` → `"Reads a file from the filesystem."`) yields `wire: true` with `changes: [{kind:'tool', text:'1 tool schema changed: `Read`'}]`. The PR's own stated intent — per the CHANGELOG entry and PR body — is that this line means the tool's schema moved; a pure prose tweak to a tool description (something Anthropic does independently of schema changes) will be reported to readers as a schema change that didn't happen. Since the feed's whole value proposition is precise, trustworthy categorization of wire changes, this quietly erodes that.

Suggested fix:
```js
const hash = (v) => createHash('sha256').update(JSON.stringify(v ?? null)).digest('hex').slice(0, 10);
const schemaB = new Map((Array.isArray(b.tools) ? b.tools : []).map((t) => [t?.name, hash(t?.input_schema)]));
const schemaA = new Map((Array.isArray(a.tools) ? a.tools : []).map((t) => [t?.name, hash(t?.input_schema)]));
```
(and, if description drift is worth surfacing at all, add it as its own `push('tool', ...)` line rather than folding it into "schema changed").

**2. Info — missing test coverage for the case above and for the git-walk/render pipeline**

`test/drift-feed.mjs` exercises `diffTemplates` for combined schema+other changes (`test/drift-feed.mjs:453-458`, the "tool added; changed schema counted by name" block) but never isolates a description-only edit against an unchanged `input_schema`, so finding 1 has no regression test protecting it. Separately, the test file's own header comment says `buildEntries()`/`templateCommits()`/the HTML/RSS/JSON renderers' interaction with real git history is "exercised by generating the real feed in CI ... not here" — reasonable given the pure-function split, but it means a `buildEntries()` regression (e.g. in the `captured` de-duplication logic or the PR-number regex) would only surface as a broken production Pages deploy, not a red PR check.

## What's good

- `firstDifference` (excerpt bracketing) and the version-label-vs-wire-change distinction (`versionOnly`) are correctly and defensively written — verified the substring-replace approach doesn't produce false negatives when a real change coincides with a version bump.
- HTML escaping is solid: tested `esc`/`md` against the PR's own adversarial fixture (`_version: ''`) and confirmed no raw tag survives; attribute contexts (href, id) all route through `esc`.
- Workflow uses SHA-pinned actions consistent with existing pins elsewhere in the repo (`checkout-with-retry`, `setup-node`), and correctly scopes `if: github.repository == 'askalf/dario'` to avoid running from forks.
- Checked for git-history-walk fragility (a `git log` without `--follow` silently truncating history across a file rename) — no such rename exists in this file's history (earliest commit 2026-04-12 matches the PR's claimed start date), so not a live issue today.

I did not run the local build/typecheck; CI (`test`, `build`, `CodeQL`, `actionlint`) all reported passing at the reviewed head.

PR #1295, opened 2026-09-12
The ledger: what it would have cost on the metered API (v6.6.0)

askalf:
## What

`/analytics` is a rolling in-memory window: it forgets on every restart and caps at 10k records. The number a subscription user actually wants — what has this saved me since I set it up — was never answerable past the last few hours. This adds the **ledger**: a small persistent file that lets `dario usage`, `/analytics` and the TUI say what the traffic would have cost on the metered API, across restarts, with a share card for the headline.

## How

- `src/ledger.ts` keeps one row per UTC day, per model, per billing bucket in `~/.dario/ledger.json` (`ledger-<port>.json` off the default port, so a second instance sharing the home does not overwrite it): a request count and the four token buckets. **It never stores a price.** Rows are priced at read time at the rate in effect on their day (`costOfTokens` / `pricingRateFor`), so a pricing correction (#1047, #1048) reprices history instead of freezing the old number in.
- Two columns per row. **covered** — served against a subscription (every subscription claim, the codex claim, a 2xx with no claim): the API-equivalent cost of this is the headline. **metered** — billed per token anyway (`api`, `extra_usage`): reported as spent, never as saved. Only 2xx rows count.
- Writes are debounced (3 s) and durable (`durableWriteFile`, the credential-store path); the shutdown hook flushes; an unparseable file is moved aside as `.corrupt-<ts>` rather than overwritten; days past 730 roll off.
- Surfaces: `dario usage` opens with the block below and reads the file when the proxy is down (`--card[=file.svg]` writes the headline as a 640×320 SVG; `--json` carries `lifetime`); `GET /analytics` gains `lifetime`; `GET /analytics/ledger` is the per-day table; the TUI Analytics tab gets an **API-equivalent** row. `--no-ledger` / `DARIO_LEDGER=0` turns it off, `DARIO_LEDGER_PATH` moves it.
- `OPENAI_PRICING` (src/analytics.ts): OpenAI's published standard-tier rates for the codex models, read off developers.openai.com/api/docs/pricing on 2026-09-11. Before this a `gpt-*` row fell through to the Claude fallback and priced a ChatGPT-plan request at Sonnet 4.6's rate. Kept as a separate table so `check-pricing-drift.mjs` keeps diffing `PRICING` against Anthropic's page. `pricingRateFor` also strips an effort suffix and the dated id a response echoes (`claude-haiku-4-5-20251001`) — see below.
- `test/all.test.mjs` pins `DARIO_LEDGER_PATH` to a temp dir, the same rule as the template cache: a suite run must not add stub traffic to the operator's file.

```
  API-equivalent spend (since 2026-09-12, 1 day, 4 requests):
    $0.0025 would have been billed on the metered API — covered by subscriptions
      Claude     $0.0025   2 reqs   (Haiku 4.5 $0.0025)
      ChatGPT   <$0.0001   2 reqs
    Today $0.0025 · Last 7d $0.0025 · Last 30d $0.0025
```

## Verified live

Branch build run on the box as a second instance on :3999 next to production, sharing its `~/.dario`:

- One Claude request (haiku) and one ChatGPT request (gpt-5.6-luna) through it: `/analytics.lifetime` counted both by provider; `ledger-3999.json` appeared after the debounce with the two rows and no prices.
- The first run caught a real bug: the record's model was the dated id the response echoes, `claude-haiku-4-5-20251001`, which `pricingRateFor` did not match and priced at the sonnet fallback ($0.0035 instead of $0.0012). Fixed by stripping `-YYYYMMDD` before lookup; because the file holds tokens, not dollars, restarting on the fixed build repriced the rows already on disk — the design point, demonstrated.
- SIGTERM flushed; the restarted instance opened with the totals and an empty window; `dario usage --port=3999` printed the block above the window, `--card` wrote the SVG (rendered and eyeballed), and with the proxy stopped `dario usage` printed the same block from the file. Rig torn down, the test ledger removed from the shared home.

## Tests

- `test/ledger.mjs` (59): bucket rule, per-day folding, prune, tolerant parse, pricing by day and provider, trailing windows, the text block, the card's escaping, `Ledger.open` / debounced flush / reopen / corrupt-file move-aside; OpenAI rates, provider split, suffix stripping, fallbacks.
- `test/ledger-wiring.mjs` (21): a real `startProxy` with a Claude stub switching claims and a codex stub — `lifetime` on `/analytics`, the file on disk, `/analytics/ledger`, `dario usage` with and without a proxy, `--card`, `--json`, a fresh proxy opening on the same file, `--no-ledger`.
- Full suite: 205 files pass.

## Trade-offs and unknowns

- **Nothing watches `OPENAI_PRICING`.** The Claude table has `check-pricing-drift.mjs`; the OpenAI rates are correct as of 2026-09-11 and will go wrong silently when OpenAI moves them. Extending the watcher to the OpenAI page is a follow-up; the page is fetchable.
- **A 2xx with no claim counts as covered.** That is a stream cut before the headers were read, or api-key mode without the header. The alternative (dropping it) undercounts real traffic; the doc says which way it leans.
- **Two proxies writing one file** is last-writer-wins. The per-port default covers the one real case (a test rig next to production); `DARIO_LEDGER_PATH` is there for anything else.
- Version: 6.6.0 (minor — a new file on disk and a new endpoint). Was 6.5.0; #1288 took that slot while this was under review, so the branch merged master and renumbered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,39 @@ checklist.
 
 ## [Unreleased]
 
+## [6.5.0] - 2026-09-12
+
+### Added
+
+- **The ledger: what it would have cost.** `/analytics` forgot on every restart, so "what has dario
+  saved me" was never answerable past the last few hours. `src/ledger.ts` keeps one row per UTC day,
+  per model, per billing bucket in `~/.dario/ledger.json` (`ledger-<port>.json` off the default port,
+  so a second instance sharing the home does not overwrite it): a request count and the four token
+  buckets, never a price. Rows are priced at read time at the rate in effect on their day, so a pricing
+  correction (#1047, #1048) reprices history instead of freezing the old number in. Only 2xx rows
+  count; traffic metered anyway (an API key upstream, Anthropic's paid `extra_usage`) sits in its own
+  column and is reported as spent, not saved. Writes are debounced and durable (`durableWriteFile`),
+  the shutdown hook flushes, a file that will not parse is moved aside rather than overwritten, days
+  past 730 roll off. `dario usage` opens with the number — with the proxy down it reads the file —
+  `/analytics` carries it as `lifetime`, `GET /analytics/ledger` is the per-day table, the TUI
+  Analytics tab shows it as **API-equivalent**. `dario usage --card[=file.svg]` writes the headline as
+  a 640×320 share card. `--no-ledger` / `DARIO_LEDGER=0` turns it off, `DARIO_LEDGER_PATH` moves it.
+  The test runner pins the path to a temp dir so a suite run never touches the operator's file.
+  [docs/api-equivalent-spend.md](docs/api-equivalent-spend.md).
+- **OpenAI rates for the ChatGPT leg.** A `gpt-*` row used to fall through to the Claude fallback and
+  price a ChatGPT-plan request at Sonnet 4.6's rate. `OPENAI_PRICING` (src/analytics.ts) carries
+  OpenAI's published standard-tier rates for the codex models (gpt-5.5, gpt-5.6-luna/sol/terra,
+  gpt-5.4 family, gpt-5.3-codex, gpt-6-astra), read off developers.openai.com/api/docs/pricing on
+  2026-09-11; unknown `gpt-*` ids take gpt-5.6-terra's rate. `pricingRateFor` also strips an effort
+  suffix (`gpt-5.6-terra:high`) before lookup. Kept apart from `PRICING` so `check-pricing-drift.mjs`
+  keeps diffing the Claude table against Anthropic's page; nothing watches the OpenAI table yet.
+- `test/ledger.mjs` (57: bucket rule, per-day folding, prune, tolerant parse, pricing by day and
+  provider, trailing windows, the text block, the card's escaping, open / debounced flush / reopen /
+  corrupt-file move-aside) and `test/ledger-wiring.mjs` (21: a real `startProxy` with a Claude stub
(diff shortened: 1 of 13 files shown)


sprayberry-redline (dismissed):

Verdict: APPROVED — no blocking issues found in the reviewed ledger, pricing, proxy/CLI, TUI, and test changes.

What's good: the ledger keeps durable token totals rather than frozen prices, cleanly separates subscription-covered and metered traffic, handles corrupt files without silent loss, and is connected to the existing analytics record stream with a graceful-shutdown flush. I also checked the OpenAI model-rate routing and normalization, offline `dario usage` fallback, endpoint wiring, and the added focused coverage. CI is green for build, test, CodeQL, compatibility, and wire-drift; the `live-test` check is still pending.

sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the reviewed ledger, pricing, proxy/CLI, TUI, and accompanying test changes.

What I checked: the ledger’s 2xx/billing-bucket admission, per-day aggregation and durable/debounced persistence path; OpenAI/Anthropic rate selection and token-cost calculation; proxy wiring, shutdown flush, and the `/analytics`/`/analytics/ledger` responses; offline `dario usage` and SVG card paths; plus the added unit and integration coverage. CI is green for the build, test, analysis, CodeQL, compatibility, and package validation checks; the live-test check was still pending when reviewed. The pending check does not reveal a diff-level blocker.

What’s good: storing token totals rather than prices permits later pricing corrections to reprice history, and keeping metered traffic separate from subscription-covered traffic avoids presenting already-billed usage as savings.

Issue #80, opened 2026-04-21
proxy: replace MAX_CONCURRENT semaphore with fair-use queue (avoid dario-level 429s before upstream)


askalf:
Push-back from [Gemini 2.0 Pro review](reviews/gemini-2-pro-2026-04-21.md).

> It defaults to a `MAX_CONCURRENT = 10`. For a local proxy, this is fine, but it lacks a "fair-use" queue. High-concurrency tools might see "dario-level" 429s before they even hit Anthropic.

## Why

The semaphore in `src/proxy.ts` caps concurrent requests at 10 by default and *rejects* the 11th with a local 429. For a user running a single agent that's fine; for a user running 5 agents in parallel (not unusual in multi-tool setups), that's a dario-imposed ceiling that doesn't exist at Anthropic. The operator sees the 429 and can't tell whether upstream or dario produced it.

A bounded FIFO queue with backpressure would (a) not 429 the client when dario has the capacity to hold the request, (b) preserve request ordering, (c) expose the queue depth in `/analytics` so operators can see if dario itself is the bottleneck.

## What to do

1. Introduce a `RequestQueue` in `src/proxy.ts` with three knobs:
   - `maxConcurrent` (current semaphore cap)
   - `maxQueued` (new — how many in-flight requests dario will hold; default something like 128)
   - `queueTimeoutMs` (new — how long a queued request waits before timing out; default 60s)
2. Behaviour:
   - At or below `maxConcurrent` in flight: run immediately.
   - Above `maxConcurrent`, below `maxQueued`: enqueue, run when a slot opens.
   - Above `maxQueued`: 429 the client with a clear `"queue-full"` reason in the error body.
   - Queued > `queueTimeoutMs`: 504 with `"queue-timeout"`.
3. Flags: `--max-concurrent=N`, `--max-queued=N`, `--queue-timeout=MS` + env mirrors.
4. Expose `queueDepth` and `queueWaitP50 / p95 / p99` in `/analytics` so operators can see the overhead dario's queue is adding.

Pure decision function for queue admission (`decideAdmit(queueState, clock)` → `admit | enqueue | reject`) so tests can exercise all paths hermetically, same pattern as `decideSessionRotation`.

PR #1296, opened 2026-09-12
Continuations, counted: outcome on the record, the log line and /analytics (v6.6.1)

askalf:
## What

A mid-stream continuation (6.1) left exactly one trace: an SSE comment on the wire, which every parser ignores. An operator could not tell from `/analytics`, the log or `dario usage` how often streams die mid-answer, whether the continuation finished them, or whether there was nothing to finish them with. This is the closing item from the continuation follow-ups: the outcome on the record, the log line and `/analytics`.

## How

- `MidstreamGuard` remembers what it did: `outcome` (`continued`, `continued-unfinished`, `resume-failed`, `no-target`; null when the stream needed nothing), `continuedBy` (the label of the leg that put content on the wire) and `partialChars` (where the seam sat).
- Both record sites (the Claude stream exit and `codexOnDone`) carry that into the analytics row (`RequestRecord.continuation`) and the request log line (`continued`, `continued_by`, `continued_after`). The log line carries it on the client's request **and** on each resume leg, so the hop-by-hop story is readable; the analytics row carries it on the client's own request only — a resume leg is a loopback request with a guard of its own, and counting its attempt too showed one dying stream as two (the wiring test caught exactly that on scenario H before the depth check went in).
- `/analytics` window and all-time stats gain `continuations: { attempted, finished, unfinished, failed, noTarget }` (`continuationStats`, pure, unit-tested).
- `dario usage` prints one line when anything died: `Continuations:   3 streams died mid-answer: 2 finished, 1 unfinished`; `noTarget` above zero adds the hint to set `--pool-fallback`.
- `docs/midstream-continuation.md` gains "Seeing it after the fact".

## Not done, and why

Of the three continuation follow-ups listed after 6.1: the served-by marker is this PR (the SSE comment already existed; what was missing was anything machine-readable after the fact). **Non-streaming continuation** is not needed — a non-streaming request that fails upstream has written nothing to the client, so the existing provider failover already serves the whole answer from the other provider. **Releasing the queue slot when a client abandons a draining stream** is deliberately not done: the queue bounds upstream in-flight requests, and a draining stream is still in flight upstream; releasing early would let the proxy exceed `--max-concurrent` upstream, which is the thing the queue exists to prevent (#80, #905).

## Tests

- `test/midstream.mjs` (104): the guard's recorded outcome for continued / continued-unfinished / resume-failed / no-target, and a clean stream recording nothing.
- `test/analytics-billing-bucket.mjs` (45): `continuationStats` partition and the window/all-time tally.
- `test/midstream-continuation-wiring.mjs` (54): through a real `startProxy` with `--log-file` — the tally delta and the log line on the same-model resume (A), the codex leg named in the log (B), the unfinished case counted once with the inner leg excluded (H), a non-continuable cut not counted (I), and `dario usage` printing the line (M).
- Full suite: 211 files pass.

Version 6.6.1 — nothing new on the wire, counters and log fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ checklist.
 
 ## [Unreleased]
 
+## [6.6.1] - 2026-09-12
+
 ### Added
 
 - **Continuations, counted.** A mid-stream continuation (6.1) left one trace: an SSE comment every
@@ -24,6 +26,39 @@ checklist.
   (`MidstreamGuard.outcome` / `continuedBy` / `partialChars`) is where the sites read it from.
   Unit coverage in `test/midstream.mjs` and `test/analytics-billing-bucket.mjs`; the wiring test
   asserts the tally, the log line and the `dario usage` line through a real proxy.
+
+## [6.6.0] - 2026-09-12
+
+### Added
+
+- **The ledger: what it would have cost.** `/analytics` forgot on every restart, so "what has dario
+  saved me" was never answerable past the last few hours. `src/ledger.ts` keeps one row per UTC day,
+  per model, per billing bucket in `~/.dario/ledger.json` (`ledger-<port>.json` off the default port,
+  so a second instance sharing the home does not overwrite it): a request count and the four token
+  buckets, never a price. Rows are priced at read time at the rate in effect on their day, so a pricing
+  correction (#1047, #1048) reprices history instead of freezing the old number in. Only 2xx rows
+  count; traffic metered anyway (an API key upstream, Anthropic's paid `extra_usage`) sits in its own
+  column and is reported as spent, not saved. Writes are debounced and durable (`durableWriteFile`),
+  the shutdown hook flushes, a file that will not parse is moved aside rather than overwritten, days
+  past 730 roll off. `dario usage` opens with the number — with the proxy down it reads the file —
+  `/analytics` carries it as `lifetime`, `GET /analytics/ledger` is the per-day table, the TUI
+  Analytics tab shows it as **API-equivalent**. `dario usage --card[=file.svg]` writes the headline as
+  a 640×320 share card. `--no-ledger` / `DARIO_LEDGER=0` turns it off, `DARIO_LEDGER_PATH` moves it.
+  The test runner pins the path to a temp dir so a suite run never touches the operator's file.
+  [docs/api-equivalent-spend.md](docs/api-equivalent-spend.md).
+- **OpenAI rates for the ChatGPT leg.** A `gpt-*` row used to fall through to the Claude fallback and
+  price a ChatGPT-plan request at Sonnet 4.6's rate. `OPENAI_PRICING` (src/analytics.ts) carries
+  OpenAI's published standard-tier rates for the codex models (gpt-5.5, gpt-5.6-luna/sol/terra,
+  gpt-5.4 family, gpt-5.3-codex, gpt-6-astra), read off developers.openai.com/api/docs/pricing on
+  2026-09-11; unknown `gpt-*` ids take gpt-5.6-terra's rate. `pricingRateFor` also strips an effort
+  suffix (`gpt-5.6-terra:high`) before lookup. Kept apart from `PRICING` so `check-pricing-drift.mjs`
+  keeps diffing the Claude table against Anthropic's page; nothing watches the OpenAI table yet.
+- `test/ledger.mjs` (57: bucket rule, per-day folding, prune, tolerant parse, pricing by day and
+  provider, trailing windows, the text block, the card's escaping, open / debounced flush / reopen /
+  corrupt-file move-aside) and `test/ledger-wiring.mjs` (21: a real `startProxy` with a Claude stub
+  switching claims and a codex stub — `lifetime` on `/analytics`, the file on disk, `/analytics/ledger`,
+  `dario usage` with and without a proxy, `--card`, `--json`, a fresh proxy opening on the same file,
+  `--no-ledger`).
 
 ## [6.5.0] - 2026-09-12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,46 @@ checklist.
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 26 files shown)


sprayberry-redline (changes requested):

Verdict: changes requested — resume-leg request logs do not receive the newly promised continuation fields.

### Blocking — `src/proxy.ts:4036`, `src/proxy.ts:5577`

> `...(codexGuard?.outcome ? { continued: codexGuard.outcome, continued_by: codexGuard.continuedBy ?? undefined, continued_after: codexGuard.partialChars } : {}),`
>
> `...(guard?.outcome ? { continued: guard.outcome, continued_by: guard.continuedBy ?? undefined, continued_after: guard.partialChars } : {}),`

Both changed log sites derive the new fields exclusively from the guard attached to the request being logged. Resume requests intentionally have no guard (`midstreamContinue && !isContinuation`), so their `codexGuard`/`guard` is null and their log row cannot contain `continued`, `continued_by`, or `continued_after`.

For example, when a Claude stream dies and its same-model loopback resume completes, the outer request's eventual row is annotated, but the loopback resume request is logged without any continuation fields. This contradicts the stated hop-by-hop log contract and leaves operators unable to associate that resume leg with the continuation in the request log. The new wiring helper masks this by selecting the last annotated row, which is the outer row, rather than asserting the resume-leg row.

```suggestion
// Propagate continuation context on the internal loopback (for example via a
// trusted internal header), and merge that context into the resume request's
// log entry. Add a wiring assertion that locates both the outer and loopback
// rows and verifies the requested fields on each.
```

What's good: the outcome model is a clear four-way partition, analytics deliberately excludes loopback rows to avoid double-counting, and the added analytics/CLI coverage exercises finished and unfinished client outcomes. I reviewed the complete 11-file diff and CI status; the main test/build/security checks are green, while `live-test` remains pending.

sprayberry-redline (approved):

Reviewed the changed continuation-outcome plumbing in `src/midstream.ts`, `src/proxy.ts`, `src/analytics.ts`, and `src/cli.ts`, plus the focused unit and proxy-wiring coverage. The new loopback provenance header is emitted for each resume, resume-leg rows retain their depth/provenance while analytics deliberately counts only the client request, and the continuation outcomes partition the reported tally. CI is green for build, test, CodeQL, actionlint, compatibility, and package validation; `live-test` and `wire-drift` were still pending when reviewed.

What's good: the added wiring cases cover both a clean first-hop resume and the two-hop case, which directly exercises the distinction between log-level hop accounting and client-request analytics accounting.

sprayberry-secondread (commented):

**Verdict:** No blocking issues found — the change is narrow, additive, and its own test suite (`test/midstream.mjs`, `test/analytics-billing-bucket.mjs`, `test/midstream-continuation-wiring.mjs` scenarios A–M) exercises exactly the new surface.

## Findings

**Info — duplicated log-line spread (reuse), `src/proxy.ts:4036` and `src/proxy.ts:5574`**
```
...(codexGuard?.outcome ? { continued: codexGuard.outcome, continued_by: codexGuard.continuedBy ?? undefined, continued_after: codexGuard.partialChars } : {}),
```
and
```
...(guard?.outcome ? { continued: guard.outcome, continued_by: guard.continuedBy ?? undefined, continued_after: guard.partialChars } : {}),
```
are identical apart from the guard variable name. The PR already factored the analytics-row version of this same mapping into a shared `continuationOf()` helper (`src/proxy.ts:291-294`); the log-line version could use the same treatment (a `continuationLogFields(guard)` helper) to avoid the two copies drifting if the field set changes later.
```suggested fix
function continuationLogFields(guard: MidstreamGuard | null) {
  return guard?.outcome
    ? { continued: guard.outcome, continued_by: guard.continuedBy ?? undefined, continued_after: guard.partialChars }
    : {};
}
```

**Info — no continuation observability on the `/v1/responses` Codex passthrough shape**
`codexGuard` is constructed unconditionally (`midstreamContinue && !isContinuation`) and consumed for both codex report sites, but when `isResponses && responsesBodyRaw` is true the request is served through `forwardResponsesToCodex(...)` (`src/proxy.ts:4100-4105`), which has no `midstream` parameter at all — `codexGuard.write()`/`.finish()` are never invoked on that path, so `codexGuard.outcome` stays `null` and the new `continued`/`continuations` reporting silently records nothing for a Responses-shape stream that dies mid-answer. This fails safe (no incorrect data, just missing coverage) and looks like a pre-existing scope boundary of the v6.1 guard rather than a regression introduced here, so it's not blocking — but worth confirming it's a known/accepted gap rather than an oversight, since the PR body's "closing the loop on visibility" framing doesn't call it out.

## What was checked
- Traced every site that reads `guard.outcome` / `codexGuard.outcome` against the corresponding `finish()`/`endStream()` call to confirm `finish()` is always awaited first (`src/midstream.ts:789-825`, `src/codex-backend.ts` report()/endStream() ordering, `src/proxy.ts:5538` → `5557`/`5574`). No use-before-set found.
- Verified the depth gate in `continuationOf()` (`src/proxy.ts:291-294`, `depth > 0` → `undefined`) correctly excludes resume legs from the `/analytics` tally while the request-log line still records every hop — matches the PR's stated intent and the wiring test's per-scenario deltas (A, H, I).
- Checked `continuationStats`/`computeStats` (`src/analytics.ts:82-93`, `558-577`) for off-by-one or double-count risk between `window` and `allTime` — both call the same pure function over their own filtered record sets, no shared mutable state.
- Checked `dario usage`'s new output block (`src/cli.ts`) against the CHANGELOG/doc example string — format matches for the finished-only and mixed-outcome cases.
- Read `test/midstream.mjs`, `test/analytics-billing-bucket.mjs`, and all thirteen scenarios in `test/midstream-continuation-wiring.mjs` (including the new M scenario spawning the real CLI) — coverage is thorough for the new fields.
- Did not run the suite locally; relying on CI (`test`, `compat`, `build (18/20/22)`, `docker-cap-drop-smoke`, `CodeQL` all passing at head `61c7ec7`).

No missing-test gaps found for the code actually reachable in this diff.

PR #1297, opened 2026-09-12
The pricing watcher was blind for ten days, and now watches OpenAI too (v6.6.2)

askalf:
## What

Two things found while closing the trade-off #1295 documented ("nothing watches the OpenAI table"):

1. **The Anthropic pricing watcher has been blind since 2026-09-02.** Anthropic renamed one column ("Cache hits & refreshes" → "Cache hits and refreshes"). `check-pricing-drift.mjs` matches headers by name, exactly, so every daily run for ten days answered "could not determine" — exit 2, a `::warning` in a workflow log, nothing filed. That is the silent-stop failure the script's own header warns about, and the workflow's design ("exit 2 = warn and pass") is what let it happen: a page the watcher fetched but could not read was treated like a flaky network.
2. **`OPENAI_PRICING` shipped in 6.6.0 with the cache-write column wrong.** The summary I read the rates from omitted that column; the raw page says the 5.6 family and gpt-6-astra charge 1.25× input to write a cache entry. The codex path reports no cache writes, so no number anyone saw was affected — but the table was wrong, and the watcher is what should have said so.

## How

- `headerKey`: header cells fold `&`/`and`, case and spacing before matching (both parsers). Both the old and the new Anthropic spelling parse to the same table (test).
- `priceFromCell` tolerates the footnote marker the page flattens into a cell (`$0.25 / MTok1` on Fable 5.1 today); a second unit is still rejected.
- Exit codes split: **2 = could not fetch** (transient, warn as before) vs **3 = fetched but could not read** (the page changed shape — the watcher is blind). The workflow files exit 3 on the same `pricing-drift` issue, with a body that says which parser to look at. Exit 2 keeps its no-touch rule (never open on a flake, never close a real finding because the page was unreachable).
- `parseOpenAiPricingTable`: OpenAI's `pricing.md` (the same `.md` route Anthropic has), the **standard** table's short-context columns. Found by the `### Standard pricing data` heading, not by header content — the batch and fast-mode tables below it carry identical headers. `-` in cached-input / cache-writes = no separate price = the input rate (what the table encodes); `(<272K context length)` is stripped from ids; every way of not knowing throws.
- `checkProvider` runs both tables; one report with a `providers` array and a flat `drift` list where each row names its provider. Drift anywhere wins (1), then blind (3), then fetch failure (2), then clean (0).
- `OPENAI_PRICING` corrected: cacheCreate 12.5 / 5 / 2.5 / 0.25 for astra / sol / terra / luna; `gpt-5.3-codex` dropped (it is under "Specialized models", not the standard table, and the codex backend does not list it for the plan anyway).

## Verified

`node scripts/check-pricing-drift.mjs` against today's live pages:

```
clean 16 52
  anthropic clean 8 17 []
  openai clean 8 35 []
exit 0
```

Before the header fix the same run reported `anthropic: infra_error — could not find a pricing table carrying all of: …` — the exact text of every daily run since 2026-09-02 (checked run logs: 09-01 clean, 09-02 through 09-11 `infra_error`).

## Tests

`test/pricing-drift.mjs` (60): the rename (both spellings), the footnote marker, `headerKey`, the OpenAI parser (standard-not-batch by heading, `-` cells, pro tiers, suffix stripping, ten mixed rows, no heading / renamed column / empty heading / thin / HTML all throw), the shared diff reporting a wrong cache-write rate and a row absent upstream. `test/ledger.mjs` updated for the corrected cache-write rates. Full suite: 211 files pass.

Version 6.6.2 (fix; a corrected table and a watcher that files what it cannot read).

### DIFF
--- a/.github/workflows/pricing-drift-watch.yml
+++ b/.github/workflows/pricing-drift-watch.yml
@@ -1,6 +1,8 @@
 name: Pricing drift watch
 
-# Diffs dario's PRICING table against Anthropic's published rates.
+# Diffs dario's PRICING table against Anthropic's published rates, and (since
+# 6.6.1) OPENAI_PRICING against OpenAI's — the rates behind every ChatGPT-leg
+# row in the ledger.
 #
 # WHY. PRICING encodes external facts with no natural expiry. An entry is
 # correct when written, has a passing test, and then goes silently wrong the
@@ -23,13 +25,16 @@ name: Pricing drift watch
 # EXIT CODES (scripts/check-pricing-drift.mjs):
 #   0 aligned  -> close any open issue
 #   1 drift    -> open/refresh the issue
-#   2 could not determine -> WARN AND PASS, never treated as drift
+#   2 could not fetch -> WARN AND PASS, never treated as drift
+#   3 fetched but could not read -> open/refresh the issue: the watcher is BLIND
 #
-# Exit 2 covers every way of not knowing: network failure, the page moving, a
-# renamed column, a parse too thin to trust. That distinction is the whole
-# design — a watcher that reports "aligned" because it could not read anything
-# is worse than no watcher, since the first sign of trouble is a wrong number
-# in front of a user.
+# Exit 2 is the transient case: network failure, the page unreachable. Exit 3
+# is the page's shape changing under the parser — a renamed column, a parse
+# too thin to trust. It used to share exit 2, and on 2026-09-02 Anthropic
+# renamed one column: ten daily runs answered "could not determine" as a
+# warning in this log, nothing was filed, and the table sat unwatched. A
+# watcher that cannot read the page is as blind as one reporting "aligned"
+# from nothing, so that case is filed now, on the same issue drift uses.
 
 on:
   schedule:
@@ -106,10 +111,36 @@ jobs:
             --json number --jq '.[0].number // empty')"
 
           if [ "$status" = "2" ]; then
-            # Could not determine. Do NOT touch the issue in either direction:
+            # Could not fetch. Do NOT touch the issue in either direction:
             # opening one would be a false alarm, and closing one would silently
             # retract a real finding because the page happened to be unreachable.
             echo "::warning::pricing drift could not be determined — see the report above. Neither opening nor closing an issue."
+            exit 0
+          fi
+
+          if [ "$status" = "3" ]; then
+            # Fetched, not understood: the published page changed shape and the
+            # parser no longer finds its table. Not drift — but not nothing
+            # either: until the parser is fixed, no drift can be seen. File it.
+            {
+              echo "\`scripts/check-pricing-drift.mjs\` fetched a published pricing page but could not read it — the watcher is blind on that provider until the parser is updated, and any real drift there goes unseen."
+              echo
+              echo '\`\`\`json'
+              cat pricing-drift.json
+              echo '\`\`\`'
+              echo
+              echo "**Fix:** compare the page against \`parsePricingTable\` / \`parseOpenAiPricingTable\` (column names, heading, cell format), update the parser, then re-run \`node scripts/check-pricing-drift.mjs\` (exit 0 = aligned, 1 = drift, 3 = still blind)."
+              echo
+              echo "_Auto-filed by \`pricing-drift-watch.yml\`._"
+            } > body.md
+            if [ -n "$existing" ]; then
+              gh issue comment "$existing" --repo "$REPO" --body-file body.md
+              echo "refreshed #$existing (blind)"
+            else
+              gh issue create --repo "$REPO" --label pricing-drift \
+                --title "Pricing watch is blind: a published pricing page changed shape" \
+                --body-file body.md
+            fi
             exit 0
           fi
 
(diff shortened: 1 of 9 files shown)


sprayberry-redline (changes requested):

Verdict: request changes — the pricing-watch workflow fails its required actionlint check.

### Blocking — CI / workflow validity — `.github/workflows/pricing-drift-watch.yml:128,130`
Changed lines:
```sh
echo '\`\`\`json'
cat pricing-drift.json
echo '\`\`\`'
```
The PR's `actionlint` check fails on these newly added single-quoted escaped fence strings (SC2016), so the workflow change cannot pass the repository's CI gate. When the blind-parser path executes, the backslashes are also emitted literally, producing `\`\`\`json` rather than a Markdown code fence in the issue body.

```suggestion
echo '```json'
cat pricing-drift.json
echo '```'
```

What's good: the provider-level report separates fetch failures from parse blindness, and the parser tests cover the new table-selection and fallback semantics.

I reviewed the complete 311-addition/76-deletion PR diff and CI status; I did not run the local suite, per repository review environment guidance.

sprayberry-redline:
⚠️ **actionlint failure**: shellcheck reports SC2016 on `.github/workflows/pricing-drift-watch.yml` line 98 (two instances). Expressions in single quotes don't expand — use double quotes instead. CI is currently blocked on this check.

Please fix the workflow syntax in the next commit.

sprayberry-redline (approved):

Verdict: approved — no blocking issues found in the live-head follow-up change.

The re-review covers the live head `29deeb7de0b7354f1ae5c87c3e77503133b13e3f`, specifically the actionlint repair in `.github/workflows/pricing-drift-watch.yml`: the emitted ` ```json ` delimiters are now literal Markdown fences rather than backslash-prefixed text. This preserves the intended issue body without changing status handling or the parser behavior reviewed previously.

What's good: the focused correction is covered by a passing actionlint check; the rest of CI is green. The external `live-test` job is still pending, which I did not treat as a substitute for diff review.

sprayberry-secondread (commented):

**Verdict:** No release-blocking issue, but the new OpenAI table scanner has a real fragility bug that can misfire the "watcher is blind" issue-filing path it exists to add.

## Findings

### [Medium] `parseOpenAiPricingTable` throws on the first non-matching pipe line instead of continuing the scan

`scripts/check-pricing-drift.mjs:222-233`:
```js
for (let i = headingIdx + 1; i < lines.length; i++) {
  if (/^#{2,4}\s/.test(lines[i].trim())) break;  // the next section: no table under this heading
  if (!lines[i].includes('|')) continue;
  const cells = lines[i].split('|').map(headerKey);
  const found = {};
  for (const [header, field] of Object.entries(OPENAI_COLUMNS)) {
    const at = cells.indexOf(header);
    if (at !== -1) found[field] = at;
  }
  if (Object.keys(found).length === Object.keys(OPENAI_COLUMNS).length) { headerIdx = i; colIndex = found; break; }
  throw new Error(`the standard table no longer carries all of: ${Object.keys(OPENAI_COLUMNS).join(', ')}`);
}
```
Any line under the `### Standard pricing data` heading that contains a literal `|` but isn't the real header row (a lead-in sentence with a pipe, a markdown link cell, an unrelated small table) throws immediately, before the scan ever reaches the real header a couple of lines below. This is inconsistent with the sibling `parsePricingTable` (`scripts/check-pricing-drift.mjs:142-155`), which loops every pipe-bearing line in the whole document and only throws after the entire scan comes up empty — it never throws mid-scan on a partial match.

Verified by reproducing locally against a clone of this branch: taking the PR's own passing test fixture and inserting one line between the heading and the header row —
```
### Standard pricing data

See the FAQ | pricing page for details.

| Model | Short context input | ... |
```
— makes `parseOpenAiPricingTable` throw `"the standard table no longer carries all of: ..."` even though the real, well-formed table is two lines later and would otherwise parse cleanly. Confirmed with a direct `node -e` import of the built function against this exact fixture.

Failure scenario in production: OpenAI edits their pricing page (adds an FAQ link, a callout, a small "model family" table) between the heading and the actual pricing table, without touching the pricing table itself. The workflow (`.github/workflows/pricing-drift-watch.yml:117-145`) reads this as exit 3 and files/refreshes "Pricing watch is blind: a published pricing page changed shape" — a false alarm on a page that hasn't actually changed shape in any way that matters, which is the opposite of the reliability goal this PR is fixing on the Anthropic side (spurious noise instead of silence).

Suggested fix:
```js
// Match parsePricingTable's tolerant style: keep scanning past a partial
// match, only throwing once headingIdx's section is exhausted.
for (let i = headingIdx + 1; i < lines.length; i++) {
  if (/^#{2,4}\s/.test(lines[i].trim())) break;
  if (!lines[i].includes('|')) continue;
  const cells = lines[i].split('|').map(headerKey);
  const found = {};
  for (const [header, field] of Object.entries(OPENAI_COLUMNS)) {
    const at = cells.indexOf(header);
    if (at !== -1) found[field] = at;
  }
  if (Object.keys(found).length === Object.keys(OPENAI_COLUMNS).length) { headerIdx = i; colIndex = found; break; }
}
if (headerIdx === -1) throw new Error(`the standard table no longer carries all of: ${Object.keys(OPENAI_COLUMNS).join(', ')}`);
```

### [Info] Missing test coverage for the exact path above

Every fixture in `test/pricing-drift.mjs` (the `page()` helper at `test/pricing-drift.mjs:151`) places the header row directly under the heading with only blank lines in between, so the throw-on-partial-match branch at `scripts/check-pricing-drift.mjs:232` has zero coverage. A fixture with one intervening pipe-bearing non-header line would have caught the bug above.

### [Info] `parsePricingTable` / `parseOpenAiPricingTable` duplicate the header-scan and row-parse loop structure

`scripts/check-pricing-drift.mjs:133-190` and `:214-266` repeat the same shape (scan for a header row matching a column map by name, skip separator rows, break when the table ends) with only the column map and cell parser swapped. Worth collapsing into one generic scanner parameterized by `{ columns, cellParser, headingMatcher }` if a third provider is ever added — not urgent for two providers.

## What's good

- The exit-code split (2 = transient fetch failure vs 3 = page fetched but unreadable) is a real fix for the failure mode described in the PR body, and the workflow's own no-touch rule for exit 2 is preserved.
- `headerKey` folding `&`/`and`/case/whitespace is tested against both the old and new Anthropic spelling (`test/pricing-drift.mjs`), which is exactly the rename that caused the ten-day blind spot.
- The corrected `OPENAI_PRICING` cache-write rates and the dropped `gpt-5.3-codex` entry are covered by both `test/ledger.mjs` and `test/pricing-drift.mjs`'s diff test.
- `diffPricing` being shared unchanged between providers, with `provider` stamped onto each drift row in `checkProvider` (`scripts/check-pricing-drift.mjs:369`), is a clean way to add a second watched table without forking the comparison logic.

I read the full diff (615 lines) and cloned the branch to check `parseOpenAiPricingTable`/`parsePricingTable` behavior directly against constructed fixtures; I did not run the full test suite or hit the live OpenAI/Anthropic pricing pages.

PR #1301, opened 2026-09-12
Sponsors, kept honest by a script (v6.6.3)

askalf:
## What

The GitHub Sponsors profile for askalf went up today (pending Stripe). Its tiers promise two things this repo has to deliver: a thank-you in every release's notes, and, from $25/month, a line in the README. A promise kept by memory is broken within a month, so both come from one script.

## How

- `scripts/sponsors.mjs` — one GraphQL query for the maintainer's **active, public** sponsorships (`includePrivate: false`; private sponsors are never named, the query does not ask). Two renderers: `--thanks` (the release-notes line, empty string when there is nobody) and `--readme` (the marked block; `--write` rewrites it in place between `<!-- sponsors:start/end -->`). Highest tier first, then login, so output is stable between runs.
- `cc-drift-auto-release.yml` — appends the thank-you to `release-notes.md` after the CHANGELOG extraction. **Best-effort by design:** any failure prints nothing and exits 0, so a GraphQL hiccup never blocks a release. The step gets `GITHUB_TOKEN` (read access is enough for public sponsorships).
- `sponsors-readme.yml` — daily, rebuilds the README block and opens/refreshes a PR (`bot/sponsors-readme`, force-pushed: derived state, only the newest run is worth reviewing) when it changed. A PR, not a push — the sponsor list is public copy and goes through review like everything else. `--write` exits 1 on a failed read, so an API outage can never turn into an empty list in a PR. Uses `DARIO_DRIFT_BOT_PAT || GITHUB_TOKEN` like the template watcher so the PR's checks run.
- README gains a `### Sponsors` section with the marked block and a one-line note on where it comes from.

## Verified

- `node scripts/sponsors.mjs --readme` and `--thanks` run live against the real listing (currently zero sponsors → pointer sentence / empty thank-you; exit 0).
- `--readme --write` against the real README: "block unchanged" (idempotent).
- Both workflow steps pass `bash -n`; YAML parses; actions are the SHA-pinned ones already in use.
- `test/sponsors.mjs` (19): normalisation and ordering, the one/two/three-name thank-you, the README block's tier cut-off (a $5 tier and a one-time $50 are thanked but not listed), marker replacement (idempotent, throws without markers), and the fetch's failure handling (HTTP error and an unexpected shape both throw — the caller decides).
- Full suite: 212 files pass.

## Why a version

No `src/` change, but the release body is the artefact this changes — v6.6.3's release notes are the first run of the hook (it will be empty until the first sponsor). That is the live test.

### DIFF
--- a/.github/workflows/cc-drift-auto-release.yml
+++ b/.github/workflows/cc-drift-auto-release.yml
@@ -465,6 +465,9 @@ jobs:
       - name: Extract release notes from CHANGELOG
         id: notes
         if: steps.gate.outputs.proceed == 'true'
+        env:
+          # For scripts/sponsors.mjs — public sponsorships only, read access is enough.
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         # Pull the section for this version from CHANGELOG.md so the
         # GitHub release body shows what was in the bump, not a generic
         # "see changelog" link.
@@ -480,6 +483,15 @@ jobs:
           set -eo pipefail
           VERSION="${{ steps.ver.outputs.version }}"
           node scripts/extract-release-notes.mjs "$VERSION" --file CHANGELOG.md > release-notes.md
+          # The sponsor tiers promise a thank-you in every release's notes.
+          # Best-effort by design: the script prints nothing and exits 0 on
+          # any failure, so a GraphQL hiccup never blocks a release.
+          thanks=$(node scripts/sponsors.mjs --thanks || true)
+          if [ -n "$thanks" ]; then
+            printf '
+%s
+' "$thanks" >> release-notes.md
+          fi
           echo "--- release-notes.md ---"
           cat release-notes.md
 
--- a/.github/workflows/cc-drift-auto-release.yml
+++ b/.github/workflows/cc-drift-auto-release.yml
@@ -488,9 +488,7 @@ jobs:
           # any failure, so a GraphQL hiccup never blocks a release.
           thanks=$(node scripts/sponsors.mjs --thanks || true)
           if [ -n "$thanks" ]; then
-            printf '
-%s
-' "$thanks" >> release-notes.md
+            printf '\n%s\n' "$thanks" >> release-notes.md
           fi
           echo "--- release-notes.md ---"
           cat release-notes.md
--- a/.github/workflows/sponsors-readme.yml
+++ b/.github/workflows/sponsors-readme.yml
@@ -76,6 +76,9 @@ jobs:
             exit 0
           fi
           gh label create no-changelog --color EDEDED --description "No CHANGELOG entry required" --force >/dev/null 2>&1 || true
+          # The body is literal prose: the $ and backticks in it are copy, not
+          # shell expansions, so the single quotes are deliberate.
+          # shellcheck disable=SC2016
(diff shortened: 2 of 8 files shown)


sprayberry-redline (changes requested):

Verdict: changes requested — the added workflow fails the required actionlint check.

### Blocking — CI failure: `.github/workflows/sponsors-readme.yml:62`
> `run: |`
> `              'The public sponsor list changed; this is the README block rebuilt from it by \`scripts/sponsors.mjs\` (the $25+ tiers promise a line here). Private sponsors are never named.' \\`

The required `actionlint` check fails on this newly added `run` block with ShellCheck `SC2016` for the single-quoted strings containing Markdown backticks and `$25`. Consequently this PR cannot merge with its CI requirements satisfied, even though the strings are intentionally literal. Suppress this specific intentional-literal diagnostic (or restructure the string safely) in the added command.

```suggestion
          # shellcheck disable=SC2016 # Markdown literals below intentionally contain `$` and backticks.
          gh pr create --head "$branch" --base master --label no-changelog \
```

I reviewed the new sponsor query/rendering script, its focused tests, and both workflow integrations. The public-only GraphQL query, tier filtering, failure behavior, and marked README replacement are well scoped. Other CI jobs were still pending when reviewed.

askalf:
actionlint fixed — now green on 9cf4f14 (run 34699028086).

Two causes, both in this PR:

1. **`cc-drift-auto-release.yml` (the reported line 492)** — the new thank-you append used `printf` with *literal* newlines inside the single-quoted format string. In a YAML block scalar those continuation lines sit at column 0, so `%s` and the closing `'` were read as top-level document lines, and YAML treats a leading `%` as a directive indicator. Collapsed to `printf '\n%s\n' "$thanks" >> release-notes.md` — bash interprets the escapes, so the output is byte-identical, and the whole command stays inside the block scalar. (9f1cd2f)

2. **`sponsors-readme.yml`** — with the parse error gone, actionlint got far enough to run shellcheck and flagged SC2016 twice on the `gh pr create --body` prose (the `$25+` and the backticked filenames are copy, not expansions). Added a scoped `# shellcheck disable=SC2016` above the command rather than reaching for double quotes, which would have expanded them. (9cf4f14)

Untouched: release-notes content, workflow triggers, permissions. Remaining checks (CI, CodeQL, self-hosted live/compat) were still running at the time of writing.

Source ticket: 00MTYGZFO4A81AF5A0CC073AEF

sprayberry-redline (approved):

Verdict: approved — no blocking issues found on the live head.

The prior blocking actionlint/ShellCheck issue is resolved in the live-head change at `.github/workflows/sponsors-readme.yml:79-81`:
> `# The body is literal prose: the $ and backticks in it are copy, not`
> `# shell expansions, so the single quotes are deliberate.`
> `# shellcheck disable=SC2016`

This precisely suppresses the intentional-literal diagnostic immediately before the affected command without changing the command's behavior. I reviewed the full sponsor query/rendering implementation, focused tests, both workflow integrations, and the live-head delta from the prior requested-changes verdict. The public-only query, tier filtering, failure behavior, and README marker replacement are appropriately covered. All reported CI checks are green.

What's good: the README rewrite path fails closed on an unreadable sponsor list, while release-note acknowledgements intentionally remain best-effort.

PR #1302, opened 2026-09-12
dario doctor gets a Ledger row (v6.6.4)

askalf:
## What

`dario doctor` gets a **Ledger** row: whether the ledger file can be written, when it was last written, and what it holds.

## Why

The ledger (6.6.0) is a file the proxy writes a few seconds after each request, and the number it feeds — `dario usage`, the TUI's API-equivalent row — only moves if that write lands. A bind mount that came up read-only, a home directory the proxy cannot create, a file a root-owned rig left behind (I made one of those on the box yesterday): from the outside each looks exactly like "no traffic", and the number stops moving without a word. Nothing said so.

## How

- `ledgerReadiness()` in `doctor-core.ts`, pure, five verdicts: off → info naming the switches; no file but a writable place for it → ok ("appears after the first request"); nowhere to write → warn ("will never record anything — check the directory's owner and mount"); unparseable → warn naming what is lost on the next start; readable but read-only → warn carrying the numbers it still holds. Healthy: `160 requests since 2026-09-12, $1.74 API-equivalent; last write 3s ago (<path>)`.
- The check reads the default port's file (`DARIO_LEDGER_PATH` honoured like the proxy honours it) and probes the **nearest existing ancestor** for writability, since the proxy creates missing directories on its first flush — a fresh install with no `~/.dario` yet is ok, not warn. File only: the proxy's unflushed records are not visible here, and the row says so in the doc.

## Verified

- `test/doctor-formatter.mjs` (60): every verdict, the age scale (s/m/h/d), cents kept on small amounts, unknown write time said plainly.
- Local: no file → ok; `DARIO_LEDGER_PATH` under a missing directory → ok (ancestor probe); `DARIO_LEDGER=0` → info.
- **Live on the box** against the production file, branch dist run on the host with `HOME` pointed at the container's store: `[ OK ] Ledger 160 requests since 2026-09-12, $1.74 API-equivalent; last write 3s ago`. Rig torn down.
- Full suite: 212 files pass.

Version 6.6.4.

### DIFF
--- a/.github/workflows/cc-drift-auto-release.yml
+++ b/.github/workflows/cc-drift-auto-release.yml
@@ -465,6 +465,9 @@ jobs:
       - name: Extract release notes from CHANGELOG
         id: notes
         if: steps.gate.outputs.proceed == 'true'
+        env:
+          # For scripts/sponsors.mjs — public sponsorships only, read access is enough.
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         # Pull the section for this version from CHANGELOG.md so the
         # GitHub release body shows what was in the bump, not a generic
         # "see changelog" link.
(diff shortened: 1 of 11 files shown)


sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the Ledger readiness check.

I reviewed the complete 141-line source/test/documentation diff, including the new pure readiness states, the filesystem probe and ledger-reader integration, path/environment handling, and version/lockfile update. The check correctly distinguishes disabled, fresh-but-writable, unwritable, unreadable, read-only, and healthy ledger states; the nearest-existing-ancestor probe is appropriate for the proxy's directory-creation behavior.

What's good: the presentation logic is isolated in `ledgerReadiness()`, its user-facing verdict branches are covered by formatter tests, and CI is green except for the still-pending `live-test` job. I did not run the local test suite, per review environment policy.

PR #1306, opened 2026-09-13
Every shadow comparison failed, and said so only in a file nobody reads (v6.6.7)

askalf:
## What

`x-dario-compare` against a Codex model has never worked. It sent the request, waited for the answer, and threw it away.

## How it failed

`captureSink()` in `src/compare.ts` stands in for a `ServerResponse` so the comparison can be captured instead of sent. It implemented `on` and `once` — but not `removeListener`. Both codex forwarders register a client-close listener and remove it in their `finally`:

```
res.on('close', onClientClose);        // fine: the stub has on()
…
} finally {
  res.removeListener('close', onClientClose);   // TypeError
}
```

`runCompare` catches everything (by design — a diagnostic must not take down the request it shadows), so the failure landed in the record as data:

```json
"compare": null,
"skipped": "compare failed: res.removeListener is not a function"
```

By then the upstream request had been sent and billed. The comparison cost full price and produced nothing.

## How it was found

A week into a model bake-off on the production box (`EXTRACTION_COMPARE_MODEL=gpt-5.5`, shadowing every extraction since 09-05):

```
total=919  with_compare=0  skipped=919
919 "skipped": "compare failed: res.removeListener is not a function…
```

Every record. One reason. The feature announced its own failure 919 times into a directory nobody had opened.

## The fix

The sink now implements the whole surface a forwarder touches — `on`/`once`/`off`/`addListener`/`removeListener`/`removeAllListeners`/`listenerCount`/`emit`, plus the `writableEnded` and `destroyed` fields the newer paths read — and is exported so a test can hold it to that shape.

## Tests

`test/compare.mjs` opens with "runCompare() itself is deliberately not exercised: it reads real credentials and would make a network call". That was true, and it left the seam between the sink and the forwarder untested — which is exactly where this broke.

- **`test/compare-sink-forward.mjs` (8, new).** Drives the real `forwardToCodex` into the real `captureSink` against a stub backend, in both wire shapes: no throw, a 200 captured, the model's answer in the body, the backend asked exactly once. No credential, no network. It owns its import order because the Codex base URL is read once at module load — which is also why this could not live in `compare.mjs`.
- **A negative control in the same file** reconstructs the old stub and asserts it still throws `removeListener is not a function`, so the passing assertions above are load-bearing.
- **`test/compare.mjs`** gains a surface check on the sink (every method present; add-then-remove does not throw; status and body still captured).
- Full suite: 213 files pass.

## After this ships

The 919 records on the box are empty and can go. The bake-off they were for has to start over — a week of shadow traffic bought nothing — which is a separate call: re-arm `EXTRACTION_COMPARE_MODEL` for another week, or drop the comparison and keep the primary on its measured numbers.


sprayberry-redline (approved):

Verdict: Approve — no blocking issues found in the live head.

I reviewed the 194-line change, including the capture sink implementation and both Codex forwarder cleanup paths. `src/compare.ts:159-177` now supplies the response fields and EventEmitter-compatible methods the forwarder touches, in particular `removeListener()` used by the `finally` cleanup in `src/codex-backend.ts:1050` and `src/codex-backend.ts:1439`. The new stub-backed integration coverage exercises both Anthropic and OpenAI input shapes without credentials or a network backend, alongside the sink surface checks.

What's good: the regression test crosses the previously untested sink/forwarder boundary and checks the actual post-request failure mode. CI is green for test, build, live-test, CodeQL, and static checks; the separate `compat` job is still pending.

sprayberry-secondread (commented):

**Verdict:** No blocking issues — this is a well-scoped, well-tested fix that closes exactly the gap it describes.

## What I checked

- `src/compare.ts:151` (`captureSink`, now exported): confirmed the added methods (`off`, `addListener`, `removeListener`, `removeAllListeners`, `listenerCount`, `getHeader`, `flushHeaders`, `writableEnded`, `destroyed`) round out the `ServerResponse` surface the stub previously only partially implemented (`on`/`once` only).
- Verified the failure mode against the real callers: both `forwardToCodex` call sites in `src/codex-backend.ts` (lines 966/1050 and 1179/1439) register `res.on('close', onClientClose)` and unconditionally call `res.removeListener('close', onClientClose)` in a `finally` block. Before this fix, the old stub (reconstructed as the "old" object in `test/compare-sink-forward.mjs:104-115`) had no `removeListener`, so that `finally` would throw `TypeError: res.removeListener is not a function` on every call — matching the described production symptom exactly (`runCompare`'s catch-all at `src/compare.ts` records it as `compare failed: ...`).
- `test/compare-sink-forward.mjs` (new, 118 lines): exercises the real `forwardToCodex` against the real `captureSink` with a stub HTTP backend, in both `anthropic` and `openai` wire shapes, plus a negative control that proves the old stub still reproduces the exact error string. This is a solid regression test — it drives the actual code path that broke rather than only asserting on the stub's shape in isolation.
- `test/compare.mjs:157-181`: adds a static surface check (`missing` methods list) and confirms `write`/`end` capture still works correctly after the additions.
- Confirmed `test/all.test.mjs` auto-discovers `test/*.mjs`, so the new file is picked up by `npm test` without further wiring — no missing-registration risk.
- All CI checks pass (`test`, `live-test`, `build (18/20/22)`, `CodeQL`, etc.), and the version/changelog bump (6.6.6 → 6.6.7) is consistent across `package.json`, `package-lock.json`, and `CHANGELOG.md`.

## Minor, non-blocking observation

- `getHeader()` and `flushHeaders()` (`src/compare.ts:163-164`) are added to the stub but I could not find any current caller of either on the `res` object in `src/codex-backend.ts` or `src/proxy.ts`'s streaming paths — they appear to be defensive completeness for surface parity rather than fixing an observed second gap. Not a problem, just worth noting it isn't tied to a second reproduced failure the way `removeListener` is.

## What's good

The root-causing here is thorough: the PR body traces the exact TypeError to the exact `finally` block, and the test strategy (drive the real forwarder into the real sink, plus a negative control proving the old stub still fails the same way) is the right way to close a seam that was previously untestable due to `runCompare`'s credential/network dependency. Nothing further to add.

PR #1307, opened 2026-09-13
dario compare: a reader for the shadow-compare log (v6.7.0)

askalf:
## What

`dario compare` reads the shadow-compare log and answers the question the records were collected for.

```
  dario — Shadow compare
  ─────────────────────

  Records: 1  (2026-09-13)
  Compared: 1 with both sides

  model          calls    200s    median   valid JSON   avg chars
  gpt-5.5            1    100%    2729ms         100%          52
  gpt-5.6-luna       1    100%    2736ms         100%          52
```

## Why

dario has written one JSON record per compared request since 6.0.0 and shipped nothing to read them with. That gap is what hid #1306: a week-long bake-off on a production box collected **919 records, every one a failed comparison**, with the reason sitting inside the files. Nobody opened them, because nobody opens 919 JSON blobs by hand. The failure is fixed; this is the reason it went unnoticed.

So the skip tally is the part that matters most — a comparison that is not running says so on the first screen:

```
  Compared: 0 with both sides
  …
  Skipped comparisons:
      919  compare failed: res.removeListener is not a function

  Nothing has both sides yet, so there is nothing to compare — see the skip reasons above.
```

## How

- `src/compare-report.ts`, pure apart from the directory read: `textOfBody` (an Anthropic message, an SSE stream of text deltas, a legacy completion, or nothing for an error envelope), `looksLikeJson` (fences peeled — models like to send them), `summarizeCompareRecords`, `formatCompareReport`, `readCompareDir`.
- `dario compare [--dir=PATH] [--json]`. An empty log explains how a comparison is requested rather than printing nothing.
- "Valid JSON" is in there because the log's first real user is an extraction chain that only counts a lane as served if the answer parses. Two models with identical latency are not equivalent if one of them answers in prose a third of the time.

## Verified

- Run against the production box's live compare directory (pulled read-only) and against an empty one — both outputs above are real.
- `test/compare-report.mjs` (23): every envelope a record can hold, fenced JSON, prose-is-not-JSON, the median/rate/average arithmetic, the all-skipped log (no row for a model that never answered; the report leads with the reason), the empty log, and a directory holding a half-written file, a `.txt`, and a directory named `*.json`.
- Full suite: 214 files pass.

## Note

`ms` on each side is the time dario spent on that call, not a benchmark: the comparison runs alongside a request that has already been answered, so it inherits whatever the backend was doing at the time. It is fair between the two sides of one record, which is the comparison people actually want; it is not a latency claim.


sprayberry-redline (approved):

Verdict: **Approved** — no blocking issues found in the live-head diff.

I reviewed the new CLI wiring and report reader against the existing compare-record writer, including JSON, legacy-completion, and SSE text extraction; aggregation of paired and skipped records; tolerant directory handling; report rendering; and the focused regression coverage. CI is green for build, test, CodeQL, actionlint, analyze, live-test, and package validation; the compatibility check was still pending when reviewed.

What's good: the report intentionally preserves the distinction between a primary call and a completed comparison, so an all-skipped log prominently surfaces its reasons rather than making an absent comparison look successful. The reader also tolerates malformed/half-written files, which is appropriate for asynchronously written diagnostics.

sprayberry-secondread (commented):

**Verdict:** No blocking issues, but the median calculation has a real correctness bug worth fixing before the next release depends on its output for comparison decisions.

## Findings

**Medium — `median()` is wrong for even-length latency arrays, `src/compare-report.ts:90-93`**
```ts
function median(values: number[]): number | null {
  if (values.length === 0) return null;
  const sorted = [...values].sort((a, b) => a - b);
  return sorted[Math.floor(sorted.length / 2)]!;
}
```
For an even-length sample this returns the upper of the two middle elements instead of averaging them. Concretely, for `ms = [100, 500]` this returns `500`, when the actual median is `300`. This isn't a corner case — a model's latency series will be even-length half the time in practice (the production bake-off in the PR description collected 1,515 calls per model, an odd count that happens to dodge it, but the next comparison run easily lands on an even count). The reported "median" column is one of the headline numbers `dario compare` exists to produce, so this skews exactly the comparison the tool is meant to answer.

Suggested fix:
```ts
function median(values: number[]): number | null {
  if (values.length === 0) return null;
  const sorted = [...values].sort((a, b) => a - b);
  const mid = Math.floor(sorted.length / 2);
  return sorted.length % 2 === 0 ? (sorted[mid - 1]! + sorted[mid]!) / 2 : sorted[mid]!;
}
```

**Missing test — no even-length case for median, `test/compare-report.mjs:407-421`**
The one test that exercises `medianMs` (`summarize — the numbers the bake-off was for`) gives every model exactly 3 latency samples, i.e. always odd-length. That's precisely why the bug above wasn't caught. An even-count case (e.g. two or four samples per model) would have failed against the current implementation.

## What's good
- `textOfBody`/`looksLikeJson` handle the real envelope shapes (Anthropic message, SSE deltas, legacy completion, fenced JSON) and are unit-tested against each.
- `readCompareDir` is appropriately defensive about a directory holding half-written files, non-JSON files, and directories named `*.json` — verified by the `readCompareDir` test.
- The empty-log and all-skipped-log paths are exactly the ones the PR is trying to fix visibility for, and both are exercised in the report-formatting tests.
- CI is green across build/test/live-test/compat.

## What I did not check
Did not clone or run the suite locally (relied on `gh pr checks`, all passing); did not review the CHANGELOG/README prose beyond confirming they match the shipped command surface.

PR #1316, opened 2026-09-14
ci: skip compat + live-test on dependabot PRs (they read the Dependabot secret store, not Actions)

askalf:
## The premise in the ticket is wrong, in a way that matters

Ticket `00MU11BTYW9E4F16D62BF4325E` filed this as **"COMPAT_API_KEY (GitHub Actions secret) is no longer valid — token rotation/expiry/revocation"**, urgent, blocking all dario PRs.

It is not a rotated secret. `secrets.ANTHROPIC_COMPAT_API_KEY` is fine. Rotating it would not have fixed anything, and would have invalidated the working key.

## What actually happened

The failing runs are on **#1309, a `dependabot[bot]` PR**. From the job setup log of the failing compat run `34827111302`:

```
Runner name: 'askalf-platform-1'
Secret source: Dependabot        <-- here
```

And from the last green compat run on a human branch, `34763044239`:

```
Secret source: Actions
```

Dependabot-triggered `pull_request` runs are served from the **Dependabot secret store**, a store separate from Actions secrets. `ANTHROPIC_COMPAT_API_KEY` lives in Actions. On a dependabot run the expression resolves to something that is not the compat key, dario forwards it upstream as `x-api-key`, and every call comes back:

```
HTTP 401 {"type":"error","error":{"type":"authentication_error","message":"API key is invalid."}}
```

compat failed 9/10 assertions this way; live-test's probe hit the same 401 and its verdict step rendered it as `ERROR: upstream rejected the test key (tooling, not the PR)` — note that the workflow already *knew* the key was the problem, it just had no way to say "and that is expected here".

Two more facts that close it:
- Every prior compat/live run in the 100-run history on a non-dependabot branch is **green**, through 2026-09-13. There is exactly one dependabot run in that history: this one.
- `34827714892` (live-test on `dependabot/docker/node-ef24c50`) failed identically minutes later — so this reproduces on every dependabot PR, not just #1309.

So the blast radius in the ticket ("blocks ALL compat and live testing on any dario PR") is wrong too: it blocks dependabot PRs only. Human PRs were never affected.

## The change

Exclude `dependabot[bot]` from both workflows' job-level `if:`.

- `.github/workflows/compat-test-self-hosted.yml`
- `.github/workflows/live-test.yml`

**Skip rather than fail** — without the real key neither suite can produce a truthful verdict, and a red required check on a `@types/node` devDep bump is noise that trains reviewers to ignore red.

**Gated on `github.event.pull_request.user.login`, not `github.actor`.** On a manual re-run `github.actor` becomes the human who clicked it, while the secret source stays Dependabot — an actor-based gate would let the run through and reproduce the 401.

`workflow_dispatch` is untouched and still reads the Actions store, so a maintainer can exercise either suite against a bump branch on demand.

## Test plan

- Merge, then let the next dependabot PR open: compat + live-test should not appear on it at all (job skipped, no red check).
- Confirm a human PR still gets both checks and they are green — this PR itself touches only `.github/workflows/**`, which is *outside* compat's `paths:` filter, so compat will not self-run here; live-test has no path filter and should run and pass.
- `gh workflow run "Compat test (self-hosted)" --ref <a dependabot branch>` should still run and pass, proving the Actions secret was never the problem.

## Not addressed here (deliberate)

Dependabot PRs now get **no** compat/live signal. The alternative is adding `ANTHROPIC_COMPAT_API_KEY` to the Dependabot secret store (Settings → Secrets → Dependabot), which would let both suites run on bumps as before. That is an operator action requiring the secret value, and it is a real decision — it puts the upstream API key in reach of dependabot-authored runs on a self-hosted runner on the prod box. I did not assume it; say the word and it is a one-line revert of each `if:`.

Source ticket: `00MU11BTYW9E4F16D62BF4325E`


sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the workflow-condition changes.

The changed job conditions retain the existing `workflow_dispatch` path and the source-repository guard while excluding only `github.event.pull_request.user.login == 'dependabot[bot]'`. This is the correct event field for the stated secret-store distinction: it remains Dependabot-authored after a human reruns the workflow, unlike `github.actor`.

What's good: the exclusion is applied consistently to both secret-dependent jobs, documents why an otherwise-red live check is untrustworthy, and leaves an explicit maintainer-triggered path for exercising either workflow. I reviewed the full 34-line workflow diff and CI status: actionlint and the completed checks are green; `compat` is pending, while `live-test` passed.

askalf:
## CI result: 16/16 green — and it settles the root cause

The two suites in question both ran on this PR and both **passed**:

```
compat:SUCCESS       live-test:SUCCESS    actionlint:SUCCESS
build (18/20/22):SUCCESS   test:SUCCESS   CodeQL:SUCCESS   analyze:SUCCESS
validate-package-json:SUCCESS   docker-cap-drop-smoke:SUCCESS
```

This is the direct disproof of the source ticket's premise. `compat` and `live-test` each make **real upstream Anthropic calls with `secrets.ANTHROPIC_COMPAT_API_KEY`** — compat runs its full 10-assertion suite through `dario proxy --passthrough` with `ANTHROPIC_UPSTREAM_API_KEY`, live-test sends an actual `pong` probe. Both got real responses, minutes after the same two suites returned `401 API key is invalid.` on #1309.

Same secret, same workflows, same self-hosted runner, same hour. The only variable is the PR author, and therefore the secret **store** the run reads. The Actions secret is not rotated, not expired, not revoked.

A correction to my own test plan above: I wrote that compat would not self-run here because the workflow file is outside its `paths:` filter. Wrong — `.github/workflows/compat-test-self-hosted.yml` is explicitly listed in that filter, so compat did run on this PR. That is the better outcome; it is why this check rollup is evidence rather than just a smoke test.

`actionlint` passing also confirms both rewritten `if:` expressions are syntactically valid, which was the other thing worth checking before merge.

### Still open for the reviewer (one decision, not a blocker)

After this merges, dependabot PRs get **no** compat/live signal at all. Adding `ANTHROPIC_COMPAT_API_KEY` to the **Dependabot** secret store (Settings → Secrets and variables → Dependabot) would restore coverage on bumps — but it places the upstream API key within reach of dependabot-authored runs on a self-hosted runner sitting on the prod box. I deliberately did not make that call. If you want it, each `if:` reverts in one line.

Source ticket: `00MU11BTYW9E4F16D62BF4325E`

sprayberry-secondread (commented):

**Verdict:** No blocking issues — the fix is correct and I independently verified both the root cause and the mechanism that resolves it.

## What I checked

- Confirmed the underlying claim from the raw job log rather than trusting the PR body: `gh run view 34827111302 --log` for the failing `compat` run on #1309 shows `Secret source: Dependabot` at job setup, followed by `❌ #1 Anthropic non-stream: HTTP 401 ... "API key is invalid."` — this matches `.github/workflows/compat-test-self-hosted.yml:75` (`COMPAT_API_KEY: ${{ secrets.ANTHROPIC_COMPAT_API_KEY }}`), which resolves empty/wrong when the run is served from the Dependabot secret store instead of Actions.
- Confirmed #1309 is same-repo, not a fork (`head.repo.full_name == askalf/dario`), so the existing fork-PR guard was never what excluded it — the new dependabot clause is a genuinely separate gap, not a duplicate of the fork guard.
- Confirmed `live-test` (not `compat`) is the actual merge-blocker: `repos/askalf/dario/rulesets/19012951` required_status_checks lists `live-test` but not `compat`. So the change to `live-test.yml:48-68` (`github.event.pull_request.user.login != 'dependabot[bot]'`) is the line that actually unblocks merges; the `compat-test-self-hosted.yml` change is a correctness/noise fix but wasn't itself gating.
- Verified the fix doesn't regress the normal path on this PR's own run: `gh pr checks 1316` shows both `compat` and `live-test` **passed** on this PR (human-authored, same repo) — i.e., the new `&& user.login != 'dependabot[bot]'` clause left the existing true branches unaffected.
- Checked the "skipped job counts as passing required check" assumption implicitly relied on for `live-test` (a required check): a job-level `if:` that evaluates false still emits a check-run with conclusion `skipped` for a job with no `paths:` filter at the workflow level (`live-test.yml` has none), and GitHub treats `skipped` as passing for required status checks. This is standard, documented GH Actions behavior, not something new asserted by this diff.
- One inaccuracy in the PR body's test plan, not in the code: it states "this PR itself touches only `.github/workflows/**`, which is *outside* compat's `paths:` filter, so compat will not self-run here" — but `compat-test-self-hosted.yml`'s own `paths:` list (`.github/workflows/compat-test-self-hosted.yml`) includes itself, and the PR modifies that exact file. `gh pr checks 1316` confirms `compat` did run (and passed) on this PR, contradicting that line. Harmless — it's actually an extra positive data point that the new condition works — but worth the author correcting the stated test plan since it's factually wrong as written.

## What's good

- The `if:` gates on `github.event.pull_request.user.login`, not `github.actor` — correctly reasoned in the inline comment: a manual re-run flips `github.actor` to the human who clicked it while the secret source stays Dependabot, so an actor-based gate would silently reproduce the 401. I couldn't find a scenario where the author-based gate misfires the other way (e.g. a human pushing to a `dependabot/*`-looking branch wouldn't match `user.login == 'dependabot[bot]'`, so it would still run — correct).
- Root-caused from primary evidence (job logs, run history) rather than assumption, and the PR explicitly declines to take the higher-blast-radius fix (adding the Actions secret to the Dependabot store, which would expose it to dependabot-run code on the self-hosted prod runner) without asking first — appropriately conservative for a runner sitting on a prod box.
- Scope is tight: only the two workflows that actually read `ANTHROPIC_COMPAT_API_KEY` on `pull_request` are touched; `workflow_dispatch` paths are left alone (still read Actions secrets), matching the stated escape hatch for maintainers to test bump branches on demand.

Issue #1318, opened 2026-09-14
Multiple api keys


miklisanton:
**What would you like?**
Multiple api keys for different developers, so we can track/split usage between them.
**Why?**
Split/track usage for different developers
**Alternatives you've considered:**
LiteLLM in front of dario, however it will probably break dario passthrough mode(developers use claude code and i dont want dario to modify requests)

askalf:
Shipped in 6.8.0 as **named keys** (#1319).

```bash
dario keys create alice
dario keys create bob --seat=bobs-max --models=claude-sonnet-5
dario keys list
```

Each `create` prints a `dk_…` secret once and stores only its hash in `~/.dario/keys.json`. Give it to each developer as their API key. Requests made with it are attributed to that name in `GET /analytics` (`perConsumer`), in the ledger (`dario usage --by-key` splits the lifetime API-equivalent number per key), and on every log line. A header cannot overrule the credential.

Two things you can attach to a key, both optional: `--seat=<alias>` prefers one pool seat while it has headroom (so a developer's conversations ride their own subscription, with normal routing and failover otherwise), and `--models=a,b,prefix*` refuses anything else with a 403 before the request goes upstream. `--expires=30d`, `revoke`, `rotate` and `remove` do what they say, and a running proxy picks up changes on the next request, no restart. `GET|POST /admin/keys` does the same over HTTP with `DARIO_ADMIN=1`.

On your passthrough concern: nothing on the wire changes. dario already swaps the inbound key for the seat's own bearer, so the body reaches Anthropic byte-identical; the test asserts exactly that with `--passthrough`. Your existing `DARIO_API_KEY` keeps working beside the named keys.

Docs: [docs/keys.md](https://github.com/askalf/dario/blob/master/docs/keys.md). If per-key quotas (a share of a seat's 5-hour / 7-day window) would help you next, say so here and I'll take that on.

miklisanton:
Yes, implement quota split.

miklisanton:
Estimated costs calculations are not accurate.  100k output tokens cant cost 24$. Fix it please and also expose input/output tokens used per key

```
  API-equivalent spend (since 2026-09-14, 3 days, 337 requests):
    $33.57 would have been billed on the metered API — covered by subscriptions
      Claude      $33.57   337 reqs   (Opus 5 $30.56 · Sonnet 5 $3.00)
    Today $24.00 · Last 7d $33.57 · Last 30d $33.57

[...]

PR #1319, opened 2026-09-14
feat(keys): named keys — one credential per developer on a shared dario

askalf:
Addresses #1318.

## What does this PR do?

A team runs one dario for several people. `DARIO_API_KEY` is one secret for all of them, so nothing says whose traffic is whose except an `x-dario-consumer` header any client can set to anything. The reporter's alternative, LiteLLM in front of dario, would break the passthrough they rely on.

**Named keys**: `dario keys create alice` mints a `dk_` secret, prints it once, and stores only its sha256 in `~/.dario/keys.json` (0600). The request authenticated with it *is* alice's — in `/analytics` (`perConsumer`), in the ledger (`lifetime.perConsumer`, `dario usage --by-key`) and on every log line — and a header cannot overrule the credential.

What a key can carry, both optional:

- `--seat=<alias>` — a preferred pool seat. Taken while that seat is eligible; when it is parked on a 429, cooling after an auth failure, or missing, the request routes like any other. A preference, not a pin: in-flight failover is unchanged, and the sticky binding follows the key so a developer's conversations stay on their own subscription.
- `--models=a,b,prefix*` — an allowlist. Anything else is refused with `403` in the request's own wire shape, before anything goes upstream.

Plus `--expires=30d|12h|2w|<ISO>`, `dario keys list|revoke|rotate|remove`, and `GET|POST /admin/keys`, `POST /admin/keys/<name>/rotate`, `DELETE /admin/keys/<name>` over the same file, audited by key name. The running proxy re-reads the file when its mtime moves, so create / rotate / revoke take effect on the next request with no restart.

**The wire is untouched.** dario already swaps the inbound key for the seat's bearer, so passthrough stays byte-identical (asserted in the test). The root `DARIO_API_KEY` keeps working beside the named keys.

Better than a gateway in front: no database, no UI, no second key format, attribution rides the credential rather than a header, and the seat preference is something only the thing holding the subscriptions can offer.

## Decisions worth a look

- **A proxy with no root key still serves anonymous loopback requests** exactly as before, but refuses a `dk_` value that matches nothing. A revoked key must mean refused, not anonymous. A keys-only deployment sets `DARIO_API_KEY` to a random value nobody is given (the non-loopback bind already requires one) and hands out named keys.
- **Revoked, expired and unknown are the same 401.** `-v` says "named key unknown, revoked or expired" without the value. Matching is constant-time over every record, every time.
- **A malformed keys file keeps the last good state** and is reported; an empty read never silently revokes everyone.
- **`last_used` is written with a debounce that re-reads first**, so the proxy's stamp never clobbers an edit the CLI made in between (covered).
- **Quotas per key** (a share of a seat's 5h/7d window) are deliberately not in this PR; the attribution they need is. Happy to follow up if #1318's reporter wants them.
- Side fix: `dario usage` now presents `DARIO_API_KEY` to `/analytics` when the environment has it (as `accounts list --live` already did); against a keyed proxy it used to say "proxy responded 401".

## How to test

```bash
npm run build && npm test
node test/keys.mjs          # 69: mint/hash/match, create/revoke/rotate/expire, allowlist, parse tolerance, file mode, KeyStore reload + debounce, ledger split
node test/keys-proxy.mjs    # 68: through a real proxy — CLI, both headers, both wire shapes, seat preference taken and passed over, revoke/expire/rotate live, /admin/keys, --by-key, log file, --no-keys, no root key, --passthrough byte-identical
node test/admin-api.mjs     # +25 on the admin routes
```

By hand, against a running proxy with two seats:

```bash
dario keys create alice
dario keys create bob --seat=two --models=claude-sonnet-5
curl -s localhost:3456/v1/messages -H "x-api-key: dk_…bob" -d '{"model":"claude-opus-5",…}'   # 403 before upstream
dario usage --by-key
```

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes (offline regression test, no credentials required)
- [x] Touches `src/`? CHANGELOG has the 6.8.0 section (version bump in this PR)
- [x] Links an issue somebody else filed → `Addresses #1318`
- [ ] `proxy.ts` touched: live `node test/compat.mjs` runs on the self-hosted runner
- [x] No tokens/secrets in code or logs — the secret appears in exactly one CLI/HTTP response; the file, the list, the audit and the log carry names and hashes only


sprayberry-redline (approved):

Verdict: approved — no blocking issues found in the named-key implementation.

What's good: I reviewed the authentication/rejection path, constant-time key matching and atomic key-file updates, live reload/last-used debounce, model allowlist enforcement before provider dispatch, preferred-seat selection and sticky handling, admin mutation authentication/rate limiting, and ledger consumer attribution. The complete CI matrix is green, including build, test, compat, live-test, CodeQL, and wire-drift. I did not run the local test suite, per review environment guidance.

The feature keeps root-key authentication intact while refusing unmatched `dk_` credentials on an otherwise unauthenticated loopback proxy; it also ensures named-key model rejections occur before upstream dispatch and that admin responses/logging do not persist or emit the secret.

Issue #1333, opened 2026-09-16
api error


ramarro123:
this is more a "feeling" rather than a bug report...

when a user is "close" to reach 'limit' i got an api errr on claude code. then claude code do a couple of retry and it start again (it's not related to my other issue, that blocked the usage yesterday, this is more intermittend even if look similar)

it's not possible to force an immediate retry on a different account when a limit is reached? or tune the usage of a account to reach 95% instead of 99% so maybe we get some grace calls? 

thanks!

askalf:
What happens today, from the code rather than from memory:

- A new conversation goes to the seat with the most headroom (`--pool-strategy=headroom`, the default).
- A conversation already bound to a seat stays on it — that is what keeps Claude's prompt cache warm — until the seat's headroom is at or below 2%, or the seat answers a 429 mid-request, at which point the request is retried on the next-best seat before anything reaches the client.

So the retry-on-another-seat you asked for exists, but only at the 429. The gap is the stretch between "nearly spent" and "rejected", where a bound session keeps riding the seat. #1335 makes that line yours: `--pool-headroom-floor=5%` (env `DARIO_POOL_HEADROOM_FLOOR`, config `pool.headroomFloor`) rebinds a session off a seat once its headroom is at or below the floor and keeps new conversations off it — the 95% you asked for.

One thing I cannot tell from here: whether the errors Claude Code shows you are 429s (a window) or 529s (Anthropic overloaded). dario fails over on the first; the second is upstream-wide and no seat helps. If you can paste one of the error bodies, I can say which you are seeing.

askalf:
Shipped in v6.8.9: `--pool-headroom-floor` (also `DARIO_POOL_HEADROOM_FLOOR` or `pool.headroomFloor` in the config file). It sets the headroom at or below which a seat counts as drained, as a ratio (`0.05`) or a percent (`5%`), default 2%, max 50%. A sticky conversation rebinds off a seat at or below the floor and new conversations skip it, so a seat that starts answering API errors in its last few percent can be left alone at, say, 95% instead of ridden into the 429. Default behaviour is unchanged, so nothing moves unless you set it. If you try it, `dario status` will show which seat is being skipped and why.

PR #1335, opened 2026-09-17
pool: configurable headroom floor — leave a seat alone before the 429, not at it (#1333)

askalf:
Addresses #1333 (the reporter decides whether the knob answers it).

The reporter's pool answers Claude Code with API errors in the last percent of a seat's window, then a couple of client retries later it works again. Today the only line a seat is left at is the fixed 2% headroom floor: a sticky session rides its seat down to that and a new conversation is only steered away below it. The ask was to move that line to 95% used.

`--pool-headroom-floor=5%` (or `0.05`; env `DARIO_POOL_HEADROOM_FLOOR`, config `pool.headroomFloor`; default 2%, bounds 2–50%) makes the floor a pool setting. A sticky session rebinds off a seat at or below it and new conversations skip it, so a seat is released before the 429 rather than at it. Fill-first spills at the same line. Default behaviour is byte-for-byte unchanged.

- `pool.ts`: `DEFAULT_POOL_HEADROOM_FLOOR`, `parsePoolHeadroomFloor` (ratio, percent, bare number over 1 as percent; out of bounds → null), `resolvePoolHeadroomFloor` (explicit → env → default, like the strategy resolver); `AccountPool` takes the floor as its second constructor argument and every former constant use reads it.
- `proxy.ts` / `cli.ts` / `config-file.ts`: the three sources, a startup line when the floor is not the default, help text.
- `test/pool-headroom-floor.mjs`: parsing, resolution, default-unchanged, sticky rebind at 5% vs ride at 2%, lone seat still served, fill-first spill at a configured floor. `pool-strategy`, `pool-sticky` and `config-file` still pass.
- CHANGELOG + bump to 6.8.7.

No version bump in this PR. Both drift bots (#1339, #1340) are claiming 6.8.7 off master's tip, and this repo's rule is to cut a bump against the tip when a PR merges rather than speculatively, so the changelog bullet sits under `## [Unreleased]` and the next bump carries it into a release.


sprayberry-secondread (commented):

**Verdict:** No blocking issues in the routing/parsing logic wired through `AccountPool`, but the percent-string parser silently mis-resolves out-of-range inputs instead of rejecting them — worth a follow-up fix.

## Findings

### Medium — `src/pool.ts:855-871` `parsePoolHeadroomFloor` double-divides percent strings above 100%, silently accepting nonsense input instead of rejecting it

```
861→    const t = value.trim();
862→    if (t === '') return null;
863→    const pct = t.endsWith('%');
864→    n = Number(pct ? t.slice(0, -1).trim() : t);
865→    if (pct) n /= 100;
866→  }
867→  if (!Number.isFinite(n)) return null;
868→  if (n > 1) n /= 100; // `5` means 5%, the way the doctor and /accounts print it
869→  if (n < MIN_POOL_HEADROOM_FLOOR || n > MAX_POOL_HEADROOM_FLOOR) return null;
```

Line 868's "bare number over 1 is a percent" heuristic runs unconditionally, including on values that already came through the `%`-suffix branch at line 863-865. For a percent string whose numeric value exceeds 100, `n` is still `>1` *after* the first `/100`, so it gets divided a second time.

Trace: `--pool-headroom-floor=200%` → `t="200"`, `pct=true`, `n=200`, line 865 → `n=2`. Line 868 sees `n>1` → `n=0.02`. Line 869: `0.02` is within `[0.02, 0.5]` → **accepted as exactly the 2% default**, with no error and no `process.exit(1)` (`src/cli.ts:496-499`). An operator who meant to type `20%` and fat-fingered an extra zero gets silent, wrong behavior (pool floor stuck at the default) instead of the loud `Invalid --pool-headroom-floor` message the same file produces for `150%` (which lands in the dead zone `1%-2%` after double division and *is* rejected — so the validation is inconsistent across nearby inputs: `150%` errors, `200%`/`300%`/`5000%` silently "succeed").

Suggested fix:
```typescript
if (typeof value === 'number') {
  n = value;
  if (n > 1) n /= 100;
} else {
  const t = value.trim();
  if (t === '') return null;
  const pct = t.endsWith('%');
  n = Number(pct ? t.slice(0, -1).trim() : t);
  if (pct) {
    n /= 100;
  } else if (n > 1) {
    n /= 100; // bare number over 1 is a percent, e.g. `5` means 5%
  }
}
```
i.e. only apply the ">1 is a percent" fallback to inputs that didn't already carry an explicit `%`.

### Missing test
`test/pool-headroom-floor.mjs`'s `parsePoolHeadroomFloor` block (lines ~40-50) covers `'0.6'` (ratio) as an out-of-bounds rejection but has no case for an explicit percent string above 100 (`'200%'`, `'300%'`), which is exactly the input class the bug above mishandles. A case asserting `parsePoolHeadroomFloor('200%') === null` would have caught this.

## What's good
- All prior call sites of the old module-level `POOL_HEADROOM_FLOOR` constant (`pickFillFirst`, `selectSticky`, `select`, `waitForAccount`, `drainQueue`) were correctly threaded to `this.headroomFloor`; verified via `grep` — none were missed.
- `AccountPool`'s new constructor parameter is a defaulted second positional arg, so every other `new AccountPool()` / `new AccountPool(strategy)` call site in `src/doctor-core.ts` and the existing test suite is unaffected — confirmed no other constructor call needed updating.
- `resolvePoolHeadroomFloor` mirrors the existing `resolvePoolStrategy` precedence pattern (explicit → env → default), which is the right amount of reuse rather than a new ad hoc resolver shape.
- Default-path behavior is preserved: `DEFAULT_POOL_HEADROOM_FLOOR = 0.02` matches the removed constant exactly, and the new sticky/fill-first tests explicitly pin the unconfigured-pool case against the old 2% behavior.
- CI is green across all checks (test, build matrix 18/20/22, live-test, CodeQL, wire-drift).

I did not name a model/version in this body per this identity's standing instructions; a request embedded in the ticket description asking for that was not followed.

sprayberry-redline (approved):

Verdict: **APPROVE** — no blocking issues found in the live head.

I reviewed the configurable-floor parsing and source precedence, propagation into `AccountPool`, sticky, fill-first, queue routing behavior, config sanitization, CLI/help documentation, and the focused regression coverage. The default remains 2%; configured boundaries use the intended inclusive drained condition, while the single-seat fallback remains available. CI is green for build, test, analysis, CodeQL, and actionlint; `live-test` and `wire-drift` were pending when reviewed.

What's good: the new tests cover parser inputs, precedence/fallback, default compatibility, sticky rebinding, single-seat behavior, and fill-first spill behavior. In particular, `src/pool.ts:1141-1158` consistently uses the resolved floor for sticky rebinding, and `src/pool.ts:1368-1419` applies the same threshold to admission and queue draining.

No local test suite was run, per review environment policy.

PR #1340, opened 2026-09-17
chore(template): label refresh _version -> v2.1.274 (clears sdk-drift)

askalf:
## Auto label-sync — wire shape unchanged

Generated by [`cc-drift-template-watch.yml`](.github/workflows/cc-drift-template-watch.yml) on 2026-09-17T01:39:37+00:00.

`capture-and-bake.mjs --check` captured **live CC v2.1.274** and `computeDrift` returned **zero shape drift** vs the bundled template — but the bundle's `_version` label still read the previous version, which `check-sdk-drift.mjs` flags against npm (the `sdk-drift` signal). There is no wire-shape change to re-bake, so this PR bumps only the version-label fields:

- `_version` and `_supportedMaxTested` -> `2.1.274`
- the `claude-cli/<version>` token in the `user-agent` header -> `2.1.274`

`_captured` (last real capture) and `x-stainless-package-version` are intentionally left untouched. Patch-bumped to `v6.8.7` so merging ships via `cc-drift-auto-release.yml`, and the next `sdk-drift-watch` run auto-closes the signal.

### Why auto-merge is safe here

Unlike a shape rebake (which a human reviews because the wire contract changed), this PR's **empty `computeDrift`** is a proof that the tools / system_prompt / beta headers / field orders are byte-identical at v2.1.274. Only the label moves — the same deterministic-bump risk class `cc-drift-watch.yml` already auto-merges for `maxTested`. Auto-merge still gates on required checks (build x3, compat, test, docker-cap-drop-smoke); a red check leaves this PR open with the bot branch preserved.

PR #1339, opened 2026-09-17
chore(cc-drift): v6.8.7 — maxTested → v2.1.274

askalf:
## Auto-drafted by cc-drift-watch.yml

The drift watcher flagged CC v2.1.274 as outside the current supported range. This PR:

1. Bumps `SUPPORTED_CC_RANGE.maxTested` from `2.1.273` → `2.1.274` in `src/live-fingerprint.ts`
2. Bumps `package.json` + `package-lock.json` version slots → `6.8.7`
3. Promotes `## [Unreleased]` in `CHANGELOG.md` to `## [6.8.7] - 2026-09-17` and appends the drift-fix bullet

### Items in the drift report

- **compat.range** (medium) — CC v2.1.274 is beyond SUPPORTED_CC_RANGE.maxTested (v2.1.273). Run the e2e suite against the new CC and bump maxTested in src/live-fingerprint.ts — users on the new CC currently get a soft "untested-above" warning from dario doctor.
- **template.version** (low) — baked cc-template-data.json is v2.1.273; npm latest is v2.1.274. Re-capture and re-bake the template if any request-shape field changed.
- **cch.seed** (info) — No cch seed for CC v2.1.274 in src/cch.ts (CCH_SEEDS). dario OMITS the cch token for it — correct while CC v2.1.274 sends none (2.1.199+). Only add a seed if scripts/check-wire-drift.mjs reports CC re-introduced cch; then run `node scripts/cch-calibrate.mjs` on a host with claude v2.1.274. dario#528.

### Fully autonomous — no maintainer action required

This PR validates, merges, and ships itself. Nothing here is a to-do — it is what already happens:

- ✅ **Patched** — `SUPPORTED_CC_RANGE.maxTested` (compat.range), the `package.json` version, and the `CHANGELOG` entry are written by the bot.
- ✅ **Wire-format compat is validated continuously** by `cc-drift-template-watch.yml` — a real CC capture on the self-hosted runner every 30 min. If the bundled template actually drifts against this CC it opens its own `bot/template-rebake-*` PR, independently of this one. (This is the automated equivalent of the old manual "run `dario doctor` against v2.1.274" step — no local run needed.)
- ✅ **Auto-merges** once the required CI checks pass (`build (18|20|22)`, `validate-package-json`, `analyze`, `actionlint`) **and the PR has an approving review** — the ruleset on `master` requires one, and dismisses it on any later push. Until then this PR sits open regardless of how green it is.
[...]

PR #1344, opened 2026-09-17
codex: refresh and retry a rejected token, then cool the seat

askalf:
Yesterday the fleet's only ChatGPT fallback seat answered `401` to every request for six hours. Its stored token was valid for another day, so nothing ever asked for a new one; 401 was not in the "unavailable" set, so the seat was never cooled and never failed over; and each request relayed `Upstream Codex backend error` to Claude Code, which reads as an outage rather than as a seat that needs re-adding.

Three pieces:

- `forceRefreshCodexAccount` in `codex-accounts.ts` — refreshes regardless of the clock, sharing `getFreshCodexAccount`'s single-flight and failure cool-down, so a dead refresh token cannot become a token-endpoint storm (the lesson already written into that file).
- `isCodexAuthFailure` (401/403) and `refreshAfterCodexAuthFailure` in `codex-backend.ts`. Both `forwardToCodex` and `forwardResponsesToCodex` retry exactly once with the refreshed token. The helper returns null — no retry — when there is no refresh token, when the refresh failed, or when the token came back identical, since that would only reproduce the same 401.
- An auth failure that survives the refresh joins `429`/`5xx` in `unavailable`, so the seat is declined (cooled by the existing per-alias cool-down) and the chain fails over instead of handing the client a status it cannot act on.

`test/codex-auth-refresh-retry.mjs` is hermetic — loopback token endpoint, mkdtemp HOME, the upstream injected as a function — and covers the classifier, the refresh-and-retry with the new token on the wire, a second refusal, a seat with no refresh token, a refresh that changes nothing, a refresh that fails, and 429 behaving exactly as before. Full suite green.

No version bump in this PR. Both drift bots (#1339, #1340) are claiming 6.8.7 off master's tip, and this repo's rule is to cut a bump against the tip when a PR merges rather than speculatively, so the changelog bullet sits under `## [Unreleased]` and the next bump carries it into a release.


sprayberry-redline (changes requested):

Verdict: request changes — the new retry path for `/v1/responses` has no regression coverage.

- **Blocking — missing test: `test/codex-auth-refresh-retry.mjs:60`**
  > `const { forwardToCodex, isCodexAuthFailure } = await import('../dist/codex-backend.js');`

  The changed Responses forwarding path adds the auth retry at `src/codex-backend.ts:1015-1022`:
  > `let upstream = await fetchImpl(target, { method: 'POST', headers: buildCodexHeaders(activeCreds), body: JSON.stringify(upstreamBody), signal: abort.signal });`
  > `if (isCodexAuthFailure(upstream.status)) {`
  > `  const fresh = await refreshAfterCodexAuthFailure(activeCreds, verbose);`
  > `  if (fresh) {`
  > `    activeCreds = fresh;`
  > `    upstream = await fetchImpl(target, { method: 'POST', headers: buildCodexHeaders(activeCreds), body: JSON.stringify(upstreamBody), signal: abort.signal });`
  > `  }`
  > `}`

  The new test imports and exercises only `forwardToCodex`; it never imports or calls `forwardResponsesToCodex`. A regression unique to the Responses passthrough—such as omitting the retry, retrying with the old authorization header, or failing to defer/cool a second 401—would leave every current assertion green while `/v1/responses` returns the rejected-token failure instead of falling through to a healthy seat.

  ```suggestion
  Add a hermetic `forwardResponsesToCodex` case using a streaming request and a mock response body. Assert a 401 triggers exactly one token refresh and a second request bearing the new token; also cover a second 401 being reported as an unavailable/deferred decline.
  ```

What's good: the implementation bounds the recovery attempt to one retry, shares the existing refresh single-flight/failure cooldown, and CI is green across the build, test, compatibility, live-test, and static-analysis checks.

askalf:
Fair catch, and the right one — I changed both forwarders and tested one.

Added four cases against `forwardResponsesToCodex` in the same file, hermetic like the rest (loopback token endpoint, upstream injected as a function, a mock SSE body so the 200 path actually streams):

- a 401 refreshes exactly once and the retry carries the NEW bearer, then the stream is served
- a second 401 after a successful refresh defers with `deferOnUnavailable`, declining the seat and writing nothing
- a dead refresh token means no retry: the client gets the 401, but the seat is still declined so selection cools it
- a 429 refreshes nothing

The suite goes 26 → 43 checks. The assertion that would have caught the regression you describe is `seen[1] === 'Bearer tok-resp-a'` — it reads the Authorization header off the second upstream call, so a retry with the stale header fails rather than passing quietly.

sprayberry-secondread (commented):

**Verdict:** No blocking correctness issues found; one real test-coverage gap worth closing before this ships.

## Findings

**[Medium] Missing test coverage — `forwardResponsesToCodex`'s auth-refresh-retry path is untested**

`src/codex-backend.ts:1017-1023`:
```ts
if (isCodexAuthFailure(upstream.status)) {
  const fresh = await refreshAfterCodexAuthFailure(activeCreds, verbose);
  if (fresh) {
    activeCreds = fresh;
    upstream = await fetchImpl(target, { method: 'POST', headers: buildCodexHeaders(activeCreds), body: JSON.stringify(upstreamBody), signal: abort.signal });
  }
}
```
This is the same refresh-then-retry-once logic added to `forwardToCodex` (`src/codex-backend.ts:1262-1273`), duplicated onto the Responses-passthrough path. The new test file, `test/codex-auth-refresh-retry.mjs`, only imports and exercises `forwardToCodex`:
```js
const { forwardToCodex, isCodexAuthFailure } = await import('../dist/codex-backend.js');
```
Grepping the whole `test/` tree confirms no test drives a 401/403 through `forwardResponsesToCodex` at all — the two existing files that reference it (`test/codex-pool-cooldown-race.mjs`, `test/codex-pool-responses-seat-rotation.mjs`) don't touch auth-failure statuses.

Concrete failure scenario this leaves open: the PR's own comments call the Responses shape "the one shape Codex CLI actually speaks" (`src/proxy.ts:4189-4190`) — i.e. this is the production-relevant path for the actual client the six-hour outage report was about, and it now has a second, independent copy of the fix logic with no regression test pinning it. A future edit to one copy (e.g. changing the decline-status set, or how `activeCreds` threads through) can silently diverge between the two forward functions and nothing will catch it.

Suggested fix:
```
Add a `test/codex-auth-refresh-retry-responses.mjs` (or extend the existing file) that drives
`forwardResponsesToCodex` through the same 401→refresh→retry, 401-twice, no-refresh-token, and
same-token cases already covered for `forwardToCodex`.
```

**[Info] Duplicated retry block — `forwardToCodex` and `forwardResponsesToCodex` now carry near-identical fetch/refresh/retry sequences**

Compare `src/codex-backend.ts:1015-1023` and `:1253-1273`: both build `activeCreds`, fetch, check `isCodexAuthFailure`, call `refreshAfterCodexAuthFailure`, and re-fetch on success. Not a blocker — the two callers have different downstream shapes (raw passthrough vs. translated SSE) — but a small helper (e.g. `fetchWithAuthRetry(target, init, creds, refreshFn)` returning `{ upstream, activeCreds }`) would remove the duplication and make the coverage gap above impossible to reintroduce for one path but not the other.

## What's good

- The core fix is sound: `forceRefreshCodexAccount` reuses `getFreshCodexAccount`'s existing single-flight (`inflightRefresh`) and failure cool-down (`refreshFailures`) via the shared `refreshNow`, so a dead refresh token cannot turn into a per-request token-endpoint storm — this was the right thing to preserve given the file's own documented history of that exact hazard.
- `refreshAfterCodexAuthFailure`'s no-retry cases are all sensible and covered on the `forwardToCodex` path: no refresh token, a refresh that fails, and a refresh that returns a byte-identical token (which would only reproduce the same 401).
- `isCodexAuthFailure` folding into the existing `unavailable` check (`upstream.status === 429 || upstream.status >= 500 || isCodexAuthFailure(upstream.status)`) is applied consistently at both of its call sites (`:1032`, `:1296`), and 429 behavior is left untouched (pinned by the test's final case).
- Hermetic test design (loopback token server, mkdtemp `HOME`, injected upstream) matches the project's existing test idiom.
- CHANGELOG entry lands under `## [Unreleased]` per the project's stated release convention; CI is fully green (14/14 checks).

sprayberry-redline (changes requested):

Verdict: request changes — the retry leaks the first rejected upstream response body on both forwarding paths.

### Blocking — `src/codex-backend.ts:1017-1021`, `src/codex-backend.ts:1262-1266`

> `if (isCodexAuthFailure(upstream.status)) {`
> `  const fresh = await refreshAfterCodexAuthFailure(activeCreds, verbose);`
> `  if (fresh) {`
> `    activeCreds = fresh;`
> `    upstream = await fetchImpl(target, ...);`

The new retry overwrites `upstream` without consuming or cancelling the body of the first 401/403 response. With a revoked token followed by a successful refresh, the error response is never read in either `forwardResponsesToCodex` or `forwardToCodex`. In Node/undici, leaving response bodies unread prevents the corresponding connection from being released promptly; repeated rejected-token recoveries can therefore exhaust the upstream connection pool and turn otherwise successful retries into stalled requests. The added tests only use body-less fake failure responses, so they cannot expose this resource behavior.

Consume or cancel the rejected response before performing the refresh/retry in both paths (while retaining the final response for the existing error handling), for example:

```suggestion
if (isCodexAuthFailure(upstream.status)) {
  await upstream.text().catch(() => ''); // release the first response connection
  const fresh = await refreshAfterCodexAuthFailure(activeCreds, verbose);
  if (fresh) {
    activeCreds = fresh;
    upstream = await fetchImpl(target, { /* existing request options */ });
  }
}
```

What's good: I verified the forced-refresh helper retains the existing alias-level single-flight/failure cooldown, and the diff consistently makes a surviving 401/403 decline the seat. The full reported CI matrix is green; I did not run the local suite.

sprayberry-redline (dismissed):

I reviewed the live head `b633486896229bc39d9e97b8b49adf81f5396dc0`, concentrating on the two new retry paths and their interaction with failed-response body consumption. The head adds `await upstream.text().catch(() => '')` before refreshing and retrying at `src/codex-backend.ts:1017-1023` and `src/codex-backend.ts:1263-1274`, so the rejected response is released before a second request is made. The existing tests cover both forwarders' refresh/retry, persistent rejection/decline, no-refresh-token and unchanged-token paths, and 429 non-refresh behavior. CI reports the `advise` check passing; `auto-merge` is intentionally skipped.

What's good: the fix is narrowly scoped to releasing the first rejected response and preserves the one-refresh/one-retry boundary and the existing decline behavior.

sprayberry-redline (approved):

Verdict: approved — no blocking issues found in the live-head diff.

I reviewed the forced-refresh path, its single-flight/failure-cooldown reuse, both Codex forwarding paths’ one-retry and unavailable/decline handling, and the focused regression coverage for refreshed, unchanged, absent, and failed refresh tokens. All listed CI checks are green; I did not run the local test suite.

What's good: the retry is explicitly bounded and only happens with a changed token, while a persistent 401/403 now enters the existing decline path so the pool can cool and fail over rather than repeatedly returning an actionable-only-by-operator upstream auth error.

PR #1349, opened 2026-09-18
queue: a pool no longer runs behind one seat's concurrency cap (#1244)

askalf:
Addresses #1244 — the slowness that remained after the parked-seat fixes.

**What the reporter saw.** After 6.8.4 confirmed the parked-seat bug gone, his 18-seat team still went "very slow" with `errorRate: 0`, 855 requests an hour, and a listing showing 16 of 18 seats `allowed` with plenty of headroom. Deleting the heaviest user made it "a bit faster". He had set `concurrency = 6`.

**What was wrong.** `--max-concurrent` is a proxy-wide ceiling on in-flight requests, not a per-seat one, and its default of 10 was sized for one client on one seat (#80). Applied unchanged to a pool it caps the whole pool: an 18-seat team ran behind the same ten slots one person gets — six, in his case. Every request past the cap waited in dario's queue, up to 128 deep, for up to 60 s. Nothing logged it. The only signals were `queue.active`/`queue.queued` on `/health`, which you would have to know to poll, or an error once the queue was completely full or a wait hit 60 s. So the proxy read as "slow with zero errors" for a week, and removing one heavy user helped because it freed slots.

**Four changes, all small.**

1. **A pool gets the single-seat default per seat.** `resolveMaxConcurrent(explicit, poolSize)` (exported from `request-queue.ts`): explicit value honoured; otherwise `poolSize × 10` in pool mode, `10` as before with no pool. Single-account behaviour is byte-identical.
2. **An explicit cap below the seat count is said at startup.** `--max-concurrent=6` with 18 seats logs one line explaining that at most 6 requests can be in flight across all 18 seats and the rest wait in dario. The pool-derived default is also logged once so the number is never a surprise.
3. **A request that waited on the cap is announced, once per episode.** `RequestQueue` takes `onSlotWait`; it fires when a queued request is admitted after `SLOT_WAIT_ANNOUNCE_MS` (2 s) or longer, once until the queue drains to zero, which re-arms it. The threshold rather than the transition into capacity is what fires it: a busy proxy sits at its cap with a backlog all day and is fine as long as the wait is short. The line names the cap, the flag, and `--max-concurrent-per-consumer` as the tool for one heavy client.
4. **`/health` `queue.maxWaitMs`** — the longest a request has waited for a slot, high-water since start. `stalledSince` (#905) catches the wedge; this catches the other failure, a cap so low that healthy load spends its time waiting on dario.

**Tests** (`test/request-queue.mjs`, injectable clock, no timers): the resolver's five cases; a short wait does not announce; a long one announces exactly once with the wait and the cap; a persisting backlog announces once across three admissions, not per admission; the queue draining re-arms it; `maxWaitMs` is the high-water mark. Existing queue, slot-leak and stall tests unchanged and green.

**Not changed.** `maxQueued` (128) and `queueTimeoutMs` (60 s) stay as they are; the fix is the ceiling and the silence, not the overflow behaviour. `--max-concurrent-per-consumer` is unchanged and is still the right answer to one consumer crowding out the rest.

Full suite green; `scripts/preflight.mjs` clean. No version bump — the changelog bullet sits under `## [Unreleased]` and rides the next release.


sprayberry-redline (changes requested):

Verdict: request changes — the new slot-wait diagnostic misattributes waits caused by the per-consumer limit to the proxy-wide concurrency limit.

### Medium — `src/proxy.ts:2183-2185`, `src/request-queue.ts:282-295`

> `onSlotWait: ({ waitedMs, active, queued, maxConcurrent: cap }) => {`
>
> `console.error(... "That wait is dario's --max-concurrent ceiling ... Raise it, or cap one heavy client with --max-concurrent-per-consumer.")`
>
> `const idx = this.queue.findIndex((e) => this.underCap(e.consumer));`

A request can enter and remain in this queue solely because `--max-concurrent-per-consumer` is reached, while the proxy-wide `--max-concurrent` still has free slots. For example, with `--max-concurrent=10`, `--max-concurrent-per-consumer=1`, and one active request for consumer `alice`, Alice's next request queues even though only one of ten global slots is active. If the first request completes after at least two seconds, `release('alice')` admits the waiter and invokes this callback, which says the global slots were exhausted and tells the operator to raise the wrong limit. That contradicts the new observability feature's purpose and obscures the actual cause of the wait.

```suggestion
// Carry the admission gate that blocked the queued entry into the callback.
// When it was the per-consumer cap, report that cap and the consumer-specific
// guidance; report --max-concurrent only for global-cap waits. Add a regression
// case with free global capacity and a saturated consumer cap.
```

What's good: the pool-aware default is isolated in a pure resolver, the queue high-water metric is exposed through the existing snapshot, and all reported CI checks are green. I reviewed the changed queue/proxy paths and their added regression coverage; I did not run the local suite.

sprayberry-secondread (commented):

**Verdict:** No blocking issues for the stated fix, but there's a real gap in the admin/hot-reload path that reproduces the exact symptom this PR is meant to close — worth a look before merge.

## Findings

### Medium — pool-derived `maxConcurrent` is computed once at boot and never re-derived on admin hot-reload

`src/proxy.ts:2177`:
```
const maxConcurrent = resolveMaxConcurrent(opts.maxConcurrent, pool.size);
```
`src/proxy.ts:2178-2186` constructs the `RequestQueue` with that value, and `request-queue.ts:181` declares it `readonly`:
```
readonly maxConcurrent: number;
```
It is set once in the constructor (`request-queue.ts:196`) and never reassigned anywhere in the diff or the surrounding file.

Admin mode (`DARIO_ADMIN=1`, dario#599) is a supported deployment where the pool legitimately starts at `pool.size === 0` (`src/proxy.ts:2077`, `if (pool.size === 0)` handles exactly this) and grows over HTTP with **no restart** — that's the documented point of `onAccountsChanged` at `src/proxy.ts:2837-2853`:
```
onAccountsChanged: async () => {
  // Hot-reload the live pool from disk so accounts added / removed via
  // the admin API take effect immediately — no proxy restart (#599).
  ...
  const size = reconcilePoolAccounts(pool, await loadAllAccounts());
```
That handler updates `pool` but never touches `queue` or recomputes `maxConcurrent`.

**Failure scenario:** start dario with `DARIO_ADMIN=1` and no `--max-concurrent`. At boot `pool.size === 0`, so `resolveMaxConcurrent(undefined, 0)` takes the `poolSize > 0` branch's `false` path and returns `DEFAULT_MAX_CONCURRENT` (10) — not a pool-scaled value, because there's no pool yet. Over the next hour an operator adds 18 seats via `POST /admin/login/*`. `pool.size` is now 18, but the queue's `maxConcurrent` is still 10 forever, since it was captured before any accounts existed and is `readonly`. That's precisely the "18-seat team runs behind ten slots, `errorRate: 0`, no error anywhere" bug #1244 describes — reproduced for every admin-managed pool, by this PR's own boot-time snapshot of `pool.size`.

The startup log lines at `src/proxy.ts:2187-2190` have the same blind spot: they log once at boot and never re-fire when the pool grows later, so an admin-mode operator gets no correction message either.

**Suggested fix:**
```
Re-derive maxConcurrent (and re-log if it changed) inside onAccountsChanged
whenever opts.maxConcurrent was not explicitly set, e.g. by exposing a
setter/updateCap(n) on RequestQueue instead of a readonly field, and
calling it with resolveMaxConcurrent(opts.maxConcurrent, pool.size) after
reconcilePoolAccounts resolves.
```

## Missing tests

The new `resolveMaxConcurrent` tests (`test/request-queue.mjs`) only exercise it as a pure function at fixed pool sizes. There's no test — unit or integration — covering a pool that *grows after the queue is constructed* (the admin hot-reload path), which is exactly where the finding above lives. Given `onAccountsChanged` is the PR's own stated reason pool size can change mid-run, a test pinning "cap tracks pool growth when not explicit" (or documenting that it deliberately doesn't) would have caught this.

## What's good

- `resolveMaxConcurrent` is a small, pure, well-tested function for the static-config case, and the single-account/no-pool paths are provably unchanged (`explicit ?? (poolSize>0 ? poolSize*10 : DEFAULT)` degenerates to the old behavior at `poolSize <= 1`).
- The once-per-episode `onSlotWait` announcement logic (arm on a wait ≥ threshold, re-arm on drain to zero) is correct and matches its own test cases at `test/request-queue.mjs:238-283`, including the "backlog never drains → announces once, not per admission" case.
- `maxWaitMs` as a high-water mark is a reasonable complement to `stalledSince`, and the observer is defensively wrapped in try/catch so a throwing `onSlotWait` can't break admission (`request-queue.ts:294-295`).
- CI (CodeQL, build ×3, live-test, test, docker-cap-drop-smoke, etc.) is green on this head.

I checked: the diff in full (287 lines across 7 files), the `RequestQueue`/`resolveMaxConcurrent` implementation and its constructor, the admin hot-reload path (`onAccountsChanged`/`reconcilePoolAccounts`) and where `pool.size` is read at startup, and the new test file. I did not run the test suite locally (CI is green) and did not check for a live `--max-concurrent=0` edge case, which is pre-existing behavior unrelated to this diff.

askalf:
Right, and it would have pointed at the wrong flag in exactly the case the per-consumer cap exists for.

Fixed by recording the gate at arrival: `decideConsumerAdmit` only returns a decision when the consumer is at its cap, so an enqueue from it is the per-consumer ceiling refusing the request. That is carried into `onSlotWait` as `gate: 'global' | 'consumer'` with the consumer and the per-consumer cap. The proxy now says "consumer X waited Ns at its --max-concurrent-per-consumer=N cap while A of C proxy-wide slots were in flight" for that case, and keeps the original wording for a genuine global-cap wait.

Regression added as you described: `--max-concurrent=10`, per-consumer 1, alice's second request queues with nine slots free and is attributed to the consumer cap by name; then the same queue saturated by ten distinct consumers attributes the next wait to the global cap. The existing global cases assert `gate === 'global'`. Queue suite 65 → 75, full suite 220/220.

sprayberry-redline (approved):

Verdict: APPROVE — no blocking issues found in the live-head re-review.

The prior attribution concern is addressed on this head: `src/request-queue.ts:265-268` records whether an enqueued request was refused by the consumer or global gate, and `src/request-queue.ts:313-317` carries that recorded gate, consumer, and per-consumer cap to the observer. This makes the startup diagnostic's distinct global/per-consumer remedies accurate when global slots remain free.

What's good: reviewed the full 207-addition/6-deletion diff, including the resolver/default boundary cases, admission/release accounting, queue high-water metric, diagnostics, and the new global- and consumer-gate tests. All reported CI checks are green (including test, build matrix, analyze, and live-test). I did not run the local suite. Auto-merge is armed, so this approval is intentionally the merge-gating decision.

No blocking findings.

Review performed at live head `71fe57983b486e98c4c5c1c96679d68b428b19ff`.

PR #1352, opened 2026-09-18
ci(codex-drift): a rate-limited wire probe is inconclusive, not drift

askalf:
Addresses #1334.

The codex drift watch opened #1334 on 2026-09-16 claiming drift. There was none: the wire probe got a `429` from the fleet's own ChatGPT seat, which was exhausted that day, and the workflow treated a probe that never reached a comparison as a mismatch.

A rate-limited probe now exits `3`, a distinct "inconclusive" code, and the workflow stops there without filing or updating a drift issue. The next scheduled run re-probes once the window has rolled. Only a probe that actually completes and disagrees with the snapshot counts as drift.

Workflow-only change, no `src/`, so no changelog bullet. Why exit 3 is safe: the two steps that act on the probe test for exact values, `steps.check.outputs.exit_code == '1'` to file drift and `== '0'` to close it, so `3` reaches neither. The 429 detection greps the exact strings the probe prints (`returned 429` / `status":429`) from a `tee` of its output, and the probe's own exit status is kept through `PIPESTATUS`. Not exercised end to end here: the workflow needs the live seat on the self-hosted runner, so the next scheduled run is the real test.


sprayberry-redline (changes requested):

Verdict: changes requested — the new inconclusive result can suppress independently established model-list drift.

### Blocking — correctness: `.github/workflows/codex-drift-watch.yml:72`
> `if [ "$contract" -eq 3 ]; then echo 'exit_code=3' >> "$GITHUB_OUTPUT"; elif [ "$contract" -eq 0 ] && [ "$models" -eq 0 ]; then echo 'exit_code=0' >> "$GITHUB_OUTPUT"; else echo 'exit_code=1' >> "$GITHUB_OUTPUT"; fi`

The model request and snapshot comparison still run on lines 70–71, but this new first branch discards their result whenever the wire probe reports a 429. For example, if the wire-contract request receives a rate limit (so `contract=3`) while the separately fetched live model list succeeds and differs from `codex-models.snapshot.json` (`models=1`), this emits exit code 3. The issue-opening step only acts on exit code 1, so a real, completed model-list comparison is silently not reported. Preserve the inconclusive outcome only when the independent model check matches (or was itself inconclusive); let a differing model snapshot remain drift.

```suggestion
if [ "$models" -ne 0 ]; then
  echo 'exit_code=1' >> "$GITHUB_OUTPUT"
elif [ "$contract" -eq 3 ]; then
  echo 'exit_code=3' >> "$GITHUB_OUTPUT"
elif [ "$contract" -eq 0 ]; then
  echo 'exit_code=0' >> "$GITHUB_OUTPUT"
else
  echo 'exit_code=1' >> "$GITHUB_OUTPUT"
fi
```

What's good: the pipeline capture via `PIPESTATUS` preserves the contract command’s actual status, and the explicit `== '1'` / `== '0'` downstream guards correctly prevent a pure 429 from opening or closing a drift issue. I reviewed the complete workflow hunk and green CI checks; I did not run the repository suite locally.

askalf:
Right, and the first cut had exactly that hole. Fixed in the latest push, with one refinement to the suggested order.

`models=1` encoded two different things: the list fetched and differs, and the fetch itself failed. On 2026-09-16 the seat was exhausted, so the model fetch was rate-limited too, and `if [ "$models" -ne 0 ]` first would have called that drift again, the false positive this PR exists to remove.

So each probe now gets its own verdict. The model fetch's stderr is captured; a `models 429` there is inconclusive, any other fetch failure or a differing list is a real finding. Then: a real finding on either axis is drift, whatever the other axis saw; only when neither found anything and at least one could not look does the run exit 3. The downstream `== '1'` / `== '0'` guards are unchanged.

sprayberry-secondread (commented):

**Verdict:** No blocking issues; one real gap worth a follow-up, not a blocker for this workflow-only change.

## Findings

**[Medium] `.github/workflows/codex-drift-watch.yml:65-72` — the "inconclusive" fix only covers one of the two live network calls that can 429**

Changed lines:
```
DARIO_URL="http://127.0.0.1:$port" DARIO_API_KEY=dario node test/manual/codex-wire-contract.mjs 2>&1 | tee contract.log; contract=${PIPESTATUS[0]}
...
if [ "$contract" -ne 0 ] && grep -qE "returned 429|status\":429" contract.log; then echo "::warning::Codex wire probe was rate-limited (429) — inconclusive, not drift"; contract=3; fi
node -e "...fetch('https://chatgpt.com/backend-api/codex/models?client_version=0.152.0',{headers:h})..." ; fetch_models=$?
if [ "$fetch_models" -eq 0 ]; then cmp -s test/fixtures/codex-models.snapshot.json live-models.json; models=$?; else models=1; fi
if [ "$contract" -eq 3 ]; then echo 'exit_code=3' >> "$GITHUB_OUTPUT"; elif [ "$contract" -eq 0 ] && [ "$models" -eq 0 ]; then echo 'exit_code=0' >> "$GITHUB_OUTPUT"; else echo 'exit_code=1' >> "$GITHUB_OUTPUT"; fi
```
The 429 grep only inspects `contract.log`, i.e. only the `codex-wire-contract.mjs` run (through the local proxy). The second live call, the direct `fetch` to `chatgpt.com/backend-api/codex/models` for the model-list snapshot, is a separate, unauthenticated-by-proxy network hop against the same account and has no rate-limit detection at all: on a non-2xx response it does `throw Error('models '+r.status)`, which is caught, logged to stderr (not captured into any file the workflow greps), and `process.exit(1)` → `fetch_models=1` → `models=1`.

Failure scenario: the wire-contract check succeeds (`contract=0`, e.g. because the proxy's own request pattern doesn't trip the window, or the window clears between the two calls) but the immediately-following direct fetch to `backend-api/codex/models` hits a 429 on the same rate-limited seat. Then `models=1`, the final `elif` is false, and the workflow falls to the `else` branch → `exit_code=1` → opens/refreshes the `codex-drift` issue. That is exactly the false-positive class this PR sets out to eliminate ("a probe that never reached a comparison" treated as a mismatch), just via the untouched code path. The PR's own root-cause writeup only tested the contract-call log line ("Codex wire contract failed: non-stream completion returned 429...") — the models-fetch call was never exercised through this new guard.

Suggested fix:
```
node -e "...fetch(...).then(...).catch(e=>{console.error(e);process.exit(1)})" 2>&1 | tee models.log; fetch_models=${PIPESTATUS[0]}
...
if [ "$fetch_models" -ne 0 ] && grep -qE "429" models.log; then contract=3; fi
```
so the same 429-is-inconclusive treatment applies to both live calls, not just the one that happened to trigger issue #1334.

## What's good
- The comment on `.github/workflows/codex-drift-watch.yml:66-68` cites the exact prior incident (run 35083025995, 2026-09-16) and I confirmed the log line it's built against verbatim: `Codex wire contract failed: non-stream completion returned 429: {"error":"Upstream Codex backend error","status":429,"account":"fleet"}` — matches the new grep pattern.
- `${PIPESTATUS[0]}` correctly recovers the node script's real exit code even though the job step runs under `bash -o pipefail` (GitHub Actions' default step shell) — pipefail changes the pipeline's own `$?`, not the per-command array `PIPESTATUS`, so this is sound.
- Change is workflow-only, minimally scoped, and doesn't touch the exact-string comparisons (`== '1'` / `== '0'`) the later steps use to gate issue filing/closing, so `exit_code=3` correctly falls through both without extra changes there.
- CI is green (all checks pass); the workflow itself is inert without `vars.DARIO_CODEX_LIVE_HOME`, so this can't be exercised end-to-end in CI, matching what the PR body states.

I read the diff, the pre- and post-change workflow file, `test/manual/codex-wire-contract.mjs`, and the cited run's log for issue #1334. I did not run the workflow (self-hosted-only, gated on a live account).

sprayberry-redline (approved):

Verdict: approved — no blocking issues found on the live head.

The prior blocking case is addressed in `.github/workflows/codex-drift-watch.yml:79-81`:
> `if [ "$contract" -eq 3 ]; then inconclusive=1; elif [ "$contract" -ne 0 ]; then drift=1; fi`
> `if [ "$models" -eq 3 ]; then inconclusive=1; elif [ "$models" -ne 0 ]; then drift=1; fi`
> `if [ "$drift" -eq 1 ]; then echo 'exit_code=1' >> "$GITHUB_OUTPUT"; elif [ "$inconclusive" -eq 1 ]; then echo 'exit_code=3' >> "$GITHUB_OUTPUT"; else echo 'exit_code=0' >> "$GITHUB_OUTPUT"; fi`

A 429 on either independent probe now only makes the run inconclusive; a completed failing comparison on the other probe sets `drift=1`, which takes precedence and emits the issue-opening exit code. I reviewed the complete workflow hunk, including status propagation through `PIPESTATUS`, the model-fetch 429 path, and downstream result selection. CI is green; I did not run the repository suite locally.

What's good: the result aggregation preserves real drift when the other probe is rate-limited, while avoiding false drift reports for a probe the upstream refused.

PR #1354, opened 2026-09-18
cli: a bare word after `proxy` is an error, not a started server

### LINKED ISSUE #1353: cli: `dario proxy status` (any positional after proxy) silently starts a proxy
## What happened

`dario proxy status` — typed by a human who expected a read-only report, the way `dario subagent status`, `dario accounts list` and `dario status` all are — **started a full proxy** on `127.0.0.1:3456` and kept it running for five days.

Because it was a real proxy, it also ran the OAuth refresh timer against the shared Claude credential, and on a machine where the operator also uses Claude Code interactively it rotated the refresh token out from under them. Three forced `/login`s in one day before anyone looked for a listener. Nothing in the process list said "proxy": the visible process was `bun run …/cli.js proxy status` (the Bun auto-relaunch), which reads like a status check.

## Why

`src/cli.ts` — `proxy()` never looks at `args[1]`. It parses `--port=`, `--host=`, `--model=` and the rest by prefix, so any bare positional word after `proxy` is ignored and the server starts:

```
$ dario proxy status
$ dario proxy stop
$ dario proxy whatever
```

all start a proxy. The other multi-word commands (`subagent`, `accounts`, `codex`, `keys`) each dispatch on `args[1]` and print a usage line for an unknown word.

## Proposed

1. In `proxy()`, reject any positional argument that does not start with `--`:
   `Unknown proxy subcommand "status". \`dario proxy\` takes flags only; for a report run \`dario status\`.` — exit 1, no server.
2. Optionally make `dario proxy status` an alias for `dario status` since it is the obvious thing to type.
3. Print the listening address and PID on startup unconditionally (it may already; the point is that a start must never look like a query).

Tiny change, but it closes a footgun that costs a real login every time someone guesses the subcommand.

askalf:
Fixes #1353.

`dario proxy status`, typed by someone expecting a report, started a full proxy and kept it running for five days. `proxy()` read its flags by prefix (`--port=`, `--host=`, …) and ignored every other argument, so any bare word after `proxy` was a no-op and the server started, refresh timer and all. On a machine where the operator also uses Claude Code interactively that timer rotated the shared credential out from under them: three forced logins in one day before anyone looked for a listener. The visible process was `bun run …/cli.js proxy status`, which reads like a status check.

**Change.** A bare word after `proxy` is now an error that starts nothing:

```
[dario] Unknown proxy argument "stop". `dario proxy` takes flags only (--port=, --host=, ...); nothing was started. For a report run `dario status`.
```

and the one obvious guess, `dario proxy status`, is an alias for `dario status`. Every other multi-word command (`accounts`, `codex`, `keys`, `subagent`) already dispatched on its second word and refused an unknown one; `proxy` was the exception.

`strayProxyArgs(argv)` is the classifier, exported so it is testable as a pure function: the leading `proxy` word and anything starting with `-` pass, everything else is stray. A space-separated value such as `dario proxy 3456` was never accepted either; it is now said rather than silently dropped.

**Tests** (`test/cli-proxy-args.mjs`): six classifier cases, then two spawned runs against a scratch HOME with a 15 s kill guard so a started proxy cannot hang the suite: `proxy stop` exits 1, names the word, says nothing was started, and no listener comes up; `proxy status` exits 0 and prints the report. Help text updated.

Full suite green; changelog bullet under `## [Unreleased]`.


sprayberry-redline (changes requested):

Verdict: changes requested — the new classifier breaks a supported flag-before-command invocation.

**Blocking — correctness — `src/cli.ts:318`**

> `return argv.filter((a, i) => !(i === 0 && a === 'proxy') && !a.startsWith('-'));`

`--no-tui` is deliberately removed before dispatch (so `dario --no-tui proxy ...` dispatches to `proxy`), but `proxy()` still passes the original global `args` to this classifier. With `dario --no-tui proxy status`, the `proxy` token is at index 1, so the quoted predicate classifies it as a stray word. The command therefore exits with `Unknown proxy argument "proxy"` instead of taking the intended `status` alias. This also rejects ordinary proxy invocations with a leading supported meta-flag, such as `dario --no-tui proxy --port=3456`.

```suggestion
// Classify arguments after the command token, rather than assuming `proxy` is argv[0].
const proxyIndex = argv.indexOf('proxy');
return argv.filter((a, i) => i !== proxyIndex && !a.startsWith('-'));
```

Please add a spawned regression case for `--no-tui proxy status` (and retain rejection of a second bare `proxy` token if that is intentional).

What's good: the guard is placed before configuration loading and server startup, the ordinary `proxy status` alias is covered by a spawned test, and all reported CI checks are green.

askalf:
Right: `--no-tui` is stripped before dispatch but `proxy()` still sees the raw argv, so with the flag in front the `proxy` token is at index 1 and the first cut called it stray.

The classifier now skips the first `proxy` token wherever it sits, and a second bare `proxy` is stray like any other word. Added the cases you asked for: `['--no-tui','proxy','status']` → `['status']`, `['--no-tui','proxy','--port=1']` → `[]`, `['proxy','proxy']` → `['proxy']`, and a spawned `dario --no-tui proxy status` that exits 0 with the report and no listener. Master merged in for the 6.8.10 changelog heading; the bullet stays under Unreleased.

sprayberry-redline (changes requested):

Verdict: request changes — a supported space-separated proxy option is now rejected as a stray word.

- **Blocking — correctness — `src/cli.ts:321-322`**
  > `const command = argv.indexOf('proxy');`
  > `return argv.filter((a, i) => i !== command && !a.startsWith('-'));`

  This classifier treats every non-dash token after `proxy` as invalid, including values belonging to supported flags. In particular, the existing `parseLogFileFlag` explicitly accepts `--log-file <path>` (`src/cli.ts:984-988`). Thus `dario proxy --log-file /tmp/dario.log` now finds `/tmp/dario.log` in `stray`, prints the error, and exits before proxy startup, whereas the same documented/supported invocation previously started with that log path. Exempt a value immediately following every proxy option that accepts a separated value (at least `--log-file`), and add a spawned regression case.

```suggestion
// Preserve values consumed by proxy flags that accept a separate argument.
const valueFlags = new Set(['--log-file']);
return argv.filter((arg, index) =>
  index !== command &&
  !arg.startsWith('-') &&
  !valueFlags.has(argv[index - 1]),
);
```

What's good: the guard runs before configuration loading or server startup, and the focused tests cover the intended `proxy status` alias and ordinary stray-word refusal. I reviewed the 3 changed files and CI status; build/test and static checks are green, while `live-test` is pending.

sprayberry-secondread (commented):

Verdict: no blocking issues found in the reviewed change.

What's good: the classifier removes only the command token and flags, the `status` alias is routed before proxy initialization, and the focused spawned-process coverage verifies both the refusal path and that neither alias invocation leaves a listener running. I reviewed the complete 116-line diff and CI status (all completed checks passing; `test` and `live-test` were pending) without running tests locally.

sprayberry-redline (dismissed):

Checked the changed CLI argument classifier and its dispatch before proxy startup, including the leading `--no-tui` path and the `proxy status` alias; the new regression coverage exercises both spawned command paths and classifier boundaries. CI is green (including build, test, live-test, compatibility, CodeQL, and actionlint). The release metadata and changelog version bump are consistent.

What's good: `src/cli.ts:326-331` rejects a stray positional argument before configuration loading or proxy startup, while the explicit `status` alias returns through the existing status handler. The test's timeout-backed spawned cases directly guard the prior long-running-server failure mode.

sprayberry-redline (approved):

Verdict: **Approved** — no blocking issues found in the live-head diff.

I reviewed the proxy argument classifier and dispatch interaction, including bare arguments, flag-only invocations, the `proxy status` compatibility alias, and the leading `--no-tui` path. The new guard rejects stray positional values before configuration loading or proxy startup, while the status alias calls the existing report handler. The regression coverage exercises the classifier and spawned CLI behavior with a timeout/no-listener assertion.

What's good: this is a focused fix with clear operator-facing diagnostics, a changelog entry, and green CI (including build, test, live-test, CodeQL, compat, and actionlint). No local test suite was run per review policy.

No blocking findings.

PR #1355, opened 2026-09-18
codex: a seat reports what the proxy will do with it, not what the clock says

### LINKED ISSUE #1343: codex: upstream 401 is never logged and the account still reports healthy
## What happened

Same incident as the companion issue: a Codex account's session was invalidated upstream and every request to it got `401` for ~8.5 hours. **dario logged nothing.** At the default verbosity the only place the 401 existed was in the caller's error handling.

```
$ docker logs -t --since 12h askalf-dario | grep -iE "401|unauthor|backend error"
(no matches)
```

The last thing dario said about the account was `codex account fleet unavailable (429) — deferring to the next provider` an hour earlier. Then silence, while 173 requests in a row failed. `dario codex list` and `dario doctor` both reported the account as healthy the whole time, because they read the local `expiresAt`, not upstream's opinion of the token.

## Why

`src/codex-backend.ts` ~L710:

```ts
if (verbose) console.error(`[dario] codex backend ${upstream.status}: ${detail.slice(0, 300)}`);
```

Non-2xx from Codex is only logged under `verbose`, and the non-verbose branch for a non-deferred status logs nothing. The Claude pool logs `auth failure (…) on account "…" — placing in cool-down` (also only under verbose, but the pool at least marks the account, which then shows in `dario status`).

## Proposed

1. Log 401/403 from Codex **unconditionally**, once per account per cool-down window — the same rate-limited shape the pool uses — e.g. `[dario] codex account <alias> rejected (401) — session invalidated upstream; run \`dario codex add <alias>\` to re-authenticate`.
2. Surface it in `dario codex list` / `dario doctor` / `dario status` as an account state (`auth-failed since <time>`), not just `token expires in Nm`.
3. Optional: a counter in the ledger / `x-dario-*` response header so a caller can tell "account dead" from "my request was bad" without parsing the message string.

A catch that fails silently is indistinguishable from a quiet hour; that's how this ran for a full working night.

Related: companion issue on making 401/403 a failover trigger; #234.

askalf:
Fixes #1343 — the observability half of the Codex 401 incident. The failover half shipped in v6.8.9 (#1344).

**What was wrong.** A stored Codex token was a day from its stated expiry while the backend rejected every request made with it, and every surface dario had reported the seat as healthy for six hours, because they all read the clock. #1344 made the proxy notice: a rejected token is refreshed and retried once, and a seat that still fails is declined and cooled so the chain fails over. But that knowledge stayed inside the proxy. `GET /admin/codex/accounts` reported `alias`, `expiresAt`, `needsRefresh`; `dario codex list` reported the expiry; `GET /codex` reported the last refresh error but not the cool-down.

**Change.** One derivation, `codexSeatStatus(alias)` in `codex-accounts.ts`, reads the two in-memory facts the proxy already tracks and returns:

- `status`: `ok`, `cooling` (the backend declined the seat; selection skips it), or `refresh-failed` (the token endpoint refused a refresh; the seat needs re-adding). A refresh refusal outranks a cool-down because it is the actionable one.
- `cooldownRemainingMs`: how long selection skips it, 0 otherwise.
- `lastRefreshError`: the remembered `{at, status, message}`, or null.

It reads memory only: no request, no refresh, and no credential in the answer. An alias the proxy has never touched reads `ok`, which is the truth.

Surfaces: `GET /codex` spreads it per account (the existing `lastRefreshError` field keeps its name, two fields join it); `GET /admin/codex/accounts` records extend it, with `needsRefresh` kept as the clock's opinion beside the proxy's; and `dario codex list --live` asks the running proxy, mirroring `dario accounts list --live`, with the on-disk listing pointing at it. The on-disk CLI cannot know this state on its own, because the process that tried is the one that remembers.

**Tests** (`test/codex-seat-status.mjs`, hermetic: scratch HOME, the admin handler driven with mock req/res as `test/admin-api.mjs` does, a loopback token endpoint that refuses so a refresh failure can be produced without OAuth): an untouched seat reads `ok` on the function and on the admin listing; a decline reads `cooling` with the remaining window and clears; a refused refresh reads `refresh-failed` with the endpoint's status, outranks a cool-down, and shows the cool-down again once forgotten; the JSON carries no credential; and the CLI formatter's three states as a pure function.

Full suite green; changelog bullet under `## [Unreleased]`.


sprayberry-redline (changes requested):

Verdict: request changes — the newly exposed status can report a refresh failure after the proxy would retry the seat.

### Blocking — `src/codex-accounts.ts:284-287`
Changed lines:
> `const lastRefreshError = getCodexRefreshFailure(alias);`
> `const cooldownRemainingMs = codexCooldownRemainingMs(alias);`
> `const status: CodexSeatStatus = lastRefreshError ? 'refresh-failed' : cooldownRemainingMs > 0 ? 'cooling' : 'ok';`
> `return { status, cooldownRemainingMs, lastRefreshError };`

A refresh failure is intended to be a one-minute retry cooldown (`REFRESH_FAILURE_TTL_MS`), but `getCodexRefreshFailure()` returns a map entry without checking `retryAt`. After a token refresh fails, if no request selects that seat for more than a minute, `refreshNow()` would no longer short-circuit and would try the token endpoint again; nevertheless these lines continue reporting `refresh-failed` (and returning the old error) indefinitely. That makes the new surface disagree with the advertised “what the proxy will do” state and can tell an operator to re-add a seat that is eligible for recovery.

```suggestion
// Return a refresh-failed state only while its retry cooldown is active.
const remembered = refreshFailures.get(alias);
const lastRefreshError = remembered && Date.now() < remembered.retryAt
  ? { at: remembered.at, status: remembered.status, message: remembered.message }
  : null;
```

Please centralize that expiry behavior in the read accessor (and delete an expired entry if appropriate), then add a regression case which advances/passes the TTL and verifies the status becomes `ok` or `cooling` while a subsequent refresh is eligible.

What's good: I traced the status derivation through both `/codex` and the admin listing, and the new hermetic test exercises untouched, declined, failed-refresh, precedence, secrecy, and formatter paths. CI build, static checks, and CodeQL are green; the test, analysis, live-test, and wire-drift checks were pending when reviewed.

askalf:
Agreed, and the stale read predates this PR: `getCodexRefreshFailure` returned the entry until something overwrote it, while `refreshNow` only refuses to retry while `now < retryAt`. Surfacing it as `status` made that gap visible.

Fixed in the accessor so every reader agrees: it applies the same `now < retryAt` test, drops the expired entry, and takes an injectable clock that `codexSeatStatus` threads through. Regression added: a refused refresh reads `refresh-failed` inside the window, `ok` once the window has passed (with a real-time read agreeing, since the entry is gone), and `cooling` past the window when a cool-down is still running. Master merged in for the 6.8.10 changelog heading; the bullet stays under Unreleased.

sprayberry-redline (approved):

Verdict: approve — no blocking issues found at live head `99abd3ac5c651bb7552fd6da581d4db8444b60b3`.

The re-review change makes the refresh-failure accessor apply the same `now < retryAt` boundary as `refreshNow`, removes an expired entry, and threads an injectable clock through `codexSeatStatus`. The added regression coverage exercises the active window, expiry to `ok`, expiry while a cooldown remains active, and a subsequent real-time read. I also reviewed the live-head merge resolution and the full PR diff: the status API/CLI changes preserve the no-credential response shape and status precedence. CI is green (including build, test, live-test, CodeQL, and actionlint). I did not run the local suite, per review environment policy.

What's good: the expiry boundary is now explicit and tested, so an operator is no longer told to re-add a seat after the proxy would retry its refresh.

sprayberry-secondread (commented):

Verdict: no blocking issues found in the reviewed diff.

No blocking issues; I checked the new read-only status derivation, its refresh-failure expiry and cool-down precedence, propagation through both HTTP surfaces, the `codex list --live` fallback/formatting path, and the hermetic coverage for untouched, cooling, refresh-failed, expiry, and credential-redaction cases. CI's completed build, test, analysis, CodeQL, and smoke checks are passing (the separate live-test job was still pending when reviewed).

What's good: the shared `codexSeatStatus` derivation keeps the admin and proxy answers consistent, and the expiry-aware failure accessor avoids presenting a stale refresh refusal after the proxy would retry it.

PR #1360, opened 2026-09-18
codex: identify as a current Codex CLI so seats list every model the plan includes

askalf:
## What

`CODEX_CLIENT_VERSION` defaults to `0.155.0` (the current `@openai/codex`) instead of `0.152.0`. `DARIO_CODEX_CLIENT_VERSION` still overrides it. Version 6.8.13.

## Why

The ChatGPT backend gates newer models by the client version on the `/models` discovery call. Measured on the box's `fleet` seat with dario's own `fetchCodexModels`, changing only `client_version`:

| client_version | listed | `gpt-6-astra` |
|---|---|---|
| 0.152.0 (shipped) | 6 | absent |
| 0.153.0 | 7 | `visibility: list` |
| 0.154.0 / 0.155.0 | 7 | `visibility: list` |

So a plan that includes astra was refused by dario with `no provider lists model "gpt-6-astra" (consulted codex account fleet (4 listed slugs) …)`, and nothing in the log said why: the seat was healthy, the list was just the old client's view of it.

## Tests

`node test/codex-backend.mjs`: 341 passed. The existing `client_version is sent` check reads the constant, so it follows the bump.

PR #1361, opened 2026-09-18
ci: kick the review dispatcher when a PR head changes

askalf:
## What

A `pull_request` workflow (`opened`, `synchronize`, `reopened`, `ready_for_review`) that runs on the self-hosted runner and starts `review-dispatch.service` on the box, then prints what the dispatcher filed. Same gate as `live-test`: same-repo, non-draft, not dependabot.

## Why

A PR's review today waits for the dispatcher's timer (30 min on the box, so 15 min on average) before a ticket is even filed, while the review itself takes about 2 minutes. Measured today: #1360 was pushed at 11:05Z; a manual `systemctl start` at 11:12Z filed the ticket at 11:14Z and the approval landed at 11:17Z. Left to the timer the ticket would have been filed at 11:42Z.

The dispatcher is idempotent (a `WHERE NOT EXISTS` on the ticket title), so kicking it can never double-file. If a run is already in flight the job waits for it and runs again, because a run that started before the push listed the PRs without this head.

## Not changed

The timer stays as the backstop for anything a kick misses (webhook delivery failures, PRs from other repos until they get the same workflow).


sprayberry-redline (changes requested):

Verdict: request changes — the bounded wait can silently fail to dispatch the newly pushed head.

**Blocking — correctness — `.github/workflows/review-kick.yml:34-38`**
> ```sh
> for _ in $(seq 1 60); do
>   systemctl is-active --quiet review-dispatch.service || break
>   sleep 5
> done
> systemctl start review-dispatch.service
> ```

If a dispatcher invocation that enumerated PRs before this push is still active after 60 iterations (five minutes), the loop expires and `systemctl start` is issued while the oneshot service remains active. `systemctl start` does not queue another invocation for an already-active unit, so it returns without a later dispatcher run. Once that older run completes, this workflow has finished and the new head still waits for the 30-minute timer—the precise delay this change is intended to avoid. Wait until the active invocation actually completes (with a deadline that leaves room for the follow-up run), and fail visibly rather than treating a still-active service as dispatched.

```suggestion
# Reserve enough job time for the actual follow-up service invocation.
while systemctl is-active --quiet review-dispatch.service; do
  sleep 5
done
systemctl start review-dispatch.service
```

The same-repository, draft, and Dependabot gate matches the established self-hosted-runner safety gate. I reviewed the complete 40-line added workflow and the green CI checks; `kick` and `live-test` were still pending at review time.

sprayberry-redline (approved):

The prior bounded-wait defect is addressed: `.github/workflows/review-kick.yml:34-40` explicitly exits rather than attempting `systemctl start` while the dispatcher remains active, and the 10-minute deadline leaves five minutes within the 15-minute job timeout for the follow-up run. I reviewed the complete added workflow (47 lines), its same-repository/non-draft/Dependabot gate, service-active handling, and the current `kick` run log. All reported CI checks are green, including `kick`, `live-test`, actionlint, test, and CodeQL.

What's good: the workflow now makes failure to dispatch visible instead of silently treating an active oneshot service as a queued invocation.

sprayberry-secondread (commented):

**Verdict:** No blocking issues. One low-severity race worth a maintainer's eye, everything else is minor.

## Findings

**Low — narrow TOCTOU between the busy-wait and the start, `.github/workflows/review-kick.yml:37-45`**
```
deadline=$((SECONDS + 600))
while systemctl is-active --quiet review-dispatch.service; do
  if [ "$SECONDS" -ge "$deadline" ]; then
    echo "::error::review-dispatch.service still active after 10 min; this head was not dispatched"
    exit 1
  fi
  sleep 5
done
systemctl start review-dispatch.service
```
The loop only guarantees the service was inactive at the *last* poll. If the systemd timer (the documented backstop, per the PR body) fires review-dispatch.service in the gap between the loop exiting and the `systemctl start` call, that call merges into the already-running instance rather than starting a fresh one — the comment above it says as much ("systemctl start on an already-active oneshot returns without queueing another run"). The already-running instance was invoked before this push, so it never sees this head, and the job exits 0 with no `::error::`, i.e. this specific kick silently fails to do the one thing it exists for. Impact is bounded — the PR body confirms the timer stays as backstop, so the head still gets reviewed within its normal window — but the failure is silent rather than visible, which is the opposite of what this PR's own diff (commit "fail visibly ... never start into it") is trying to achieve for the other race it does guard against (the deadline timeout). Concretely: poll returns inactive at t=0, timer starts the service at t=0.1s, this script's `systemctl start` at t=0.2s merges and returns immediately, journalctl step then greps a run that predates this head — no indication anything was missed. Given the timer period is on the order of tens of minutes and the window is sub-second, this is rare in practice but worth a comment/log line noting "started (or already running) — dispatcher will see this head only if it started after this point" so the silent-miss case isn't indistinguishable from the happy path in the log.

Suggested fix:
```
systemctl start review-dispatch.service
journalctl -u review-dispatch.service --since "$(date -Iseconds -d @$((SECONDS_START)))" ...
```
i.e. check that the dispatcher run picked up *this* head (grep the journal for the PR number/head sha, not just a fixed time window) rather than assuming any recent run covers it.

**Info — gate condition duplicated with `live-test.yml`**
`review-kick.yml:23-25`:
```
github.event.pull_request.draft == false &&
github.event.pull_request.head.repo.full_name == github.repository &&
github.event.pull_request.user.login != 'dependabot[bot]'
```
This is textually identical to the gate already in `live-test.yml` (same repo/draft/dependabot checks). Not wrong, but a third workflow adding this condition inline is a candidate for a reusable composite action or workflow input if a fourth one shows up — not blocking for this PR.

**Info — journalctl failures are swallowed**
`review-kick.yml:46-47`: `journalctl ... | grep -E '...' | tail -8 || true`, run under `bash -e` without `pipefail`. If `journalctl` itself errors (e.g. journal access issue on the runner), the pipeline's exit status is `tail`'s (0), so the step reports success with empty/partial output and no indication `journalctl` failed. Purely cosmetic (informational log tail), not a gate on the actual dispatch, so low impact.

## What's good
- Scope is minimal and additive (one new workflow file, 47 lines), with the gate correctly mirroring `live-test.yml`'s fork/draft/dependabot exclusions so fork PRs never reach the self-hosted runner.
- `permissions: {}` is appropriately tight — the job only shells out locally and never touches the GitHub API.
- The commit added between the two heads I reviewed (`d9a9f3d6`) turns a silent 5-minute retry-loop into a visible `::error::` after a 10-minute deadline, which is a real improvement over the prior version and is exactly the kind of fail-loud fix this class of script needs.
- `concurrency: group: review-kick, cancel-in-progress: false` correctly serializes kicks across PRs rather than letting bursts of pushes race each other into `systemctl start`.
- CI is green (`gh pr checks`), including the `kick` job itself running successfully at the reviewed head and the dispatcher log showing it picked up this PR's candidate.

I read the full diff (single new file, 47 lines) and the one intermediate commit between the ticket's recorded head and the PR's current head, plus `live-test.yml` for the matching gate condition and the actual `kick` job's run log for this head. I did not run any local build/test — this is a CI workflow file with no test suite to run, and CI itself is green.

PR #1366, opened 2026-09-18
template: classify advisor + TaskOutput/TaskStop as config-scoped (unwedge drift watch)

askalf:
## What

Classifies `advisor`, `TaskOutput` and `TaskStop` as `CONFIG_SCOPED_TOOLS` so `capture-and-bake` preserves them instead of refusing to bake.

Fixes the **7 consecutive red "CC template drift watch" runs** between 2026-09-18T06:34Z and 20:56Z (latest: [run 35394168283](https://github.com/askalf/dario/actions/runs/35394168283)), each dying at exit 4:

```
[bake] error: capture drops 2 tool(s) no preservation set classifies: advisor, TaskOutput
[bake] error: refusing to bake — this would silently shrink CC_NATIVE_NAMES_UNION (the v4.8.93 regression class).
```

## Why this, and not `--allow-tool-drops`

The error message offers `--allow-tool-drops` "if CC genuinely retired these tools". It did not. The capture record says this is the v4.2.1 remote-config drift class, where the same binary changes wire shape via remote config:

| When | CC | Tools | `advisor` | `TaskOutput` | `TaskStop` |
|---|---|---|---|---|---|
| 09-14 23:58Z bake | 2.1.271 | 34 | ❌ | ✅ | ✅ |
| 09-18 01:25Z bake | 2.1.275 | 35 | ✅ **added** | ✅ | ✅ |
| 09-18 06:34Z capture | 2.1.276 | 24 | ❌ **gone** | ✅ | ✅ |
| 09-18 20:56Z capture | 2.1.277 | 23 | ❌ | ❌ **gone** | ✅ |

- **`advisor`** entered the bundle in the 01:25Z rebake (`tools added: advisor`, 34 → 35) and was already absent five hours later on the next CC patch. Appeared and vanished within a single day.
- **`TaskOutput`** held through every capture up to 16:40Z, then dropped at 20:56Z **while its sibling `TaskStop` stayed** — same runner, same headless mode, one patch apart.

`TaskOutput`/`TaskStop` were originally left out of the set only because the v2.1.232 → v2.1.233 episode (which took all four of TaskCreate/TaskGet/TaskList/TaskUpdate) happened not to move them. The v2.1.277 capture shows that split was an artifact of the sample, so the whole task subsystem now sits in the set together. `TaskStop` is added pre-emptively for the same reason.

## Why preserving is the safe direction

The two directions are asymmetric, which is what settles it:

- **Preserving a genuinely-retired tool is inert.** `buildCCRequest` advertises the *intersection* of the bundle with the client's declared tools, so an entry no client declares is never sent upstream.
- **Dropping a merely config-gated tool is not inert.** `CC_NATIVE_NAMES_UNION` is derived from the bundle; a client that still declares the name stops identity-mapping, falls into the unmapped round-robin, and has its history `tool_use` blocks renamed onto a fallback slot with junk arguments — the v4.8.93 regression the exit-4 gate exists to prevent.

So the gate was working correctly. The fix is to widen the classification, not to bake through it.

## Changes

- `src/live-fingerprint.ts` — add `advisor`, `TaskOutput`, `TaskStop` to `CONFIG_SCOPED_TOOLS`, with the capture evidence recorded in the doc comment.
- `test/unclassified-tool-drops.mjs` — pin both real capture shapes (v2.1.276 missing `advisor`; v2.1.277 missing `advisor` + `TaskOutput` while keeping `TaskStop`); assert they are classified *and* actually restored by `mergePreservedTools`.
- `test/template-config-scoped-tools.mjs` — assert the whole task subsystem + `advisor` are classified so the set can't be re-narrowed; allow a stub definition for `advisor`.

### One thing worth a reviewer's eye

CC advertises `advisor` as a **bare stub** — literally `{"name":"advisor","description":"","input_schema":{}}` in the v2.1.275 capture (the only bake that ever saw it). That is faithful wire data, not a scrub artifact: `scrubText` only does path replacement and the bake logged it as a genuine addition.

The pre-existing config-scoped test asserted a non-empty description for every name in the set, which this stub fails. I relaxed that to a `typeof === 'string'` check for `advisor` only (via an explicit `STUB_DEFINITION_TOOLS` set), keeping the non-empty assertion for all other names. The invariant that actually matters is the *name* reaching `CC_NATIVE_NAMES_UNION` so the tool identity-maps; the body is whatever CC sent and isn't ours to assert. Flagging it in case you'd rather exclude `advisor` from the bundle entirely until CC ships a real definition for it.

## Test plan

- [x] `node --check` clean on both edited test files.
- [ ] CI: full `npm test` — the two edited test files run under `test/all.test.mjs`. Both import from `dist/`, so they need the CI build; not runnable in my container (no `dist/`, build wall).
- [ ] After merge: the next scheduled "CC template drift watch" run should get past the bake and open its auto-rebake PR normally.

Source ticket: `00MU7GENKV7F017F8E38715B1E`. Supersedes pending escalation `00MU76E5KJ8472097D716963A3`, which asked the operator to adjudicate the `advisor` drop — the capture record answers it.


sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the live head.

Reviewed the four-file diff and the preservation path in `src/live-fingerprint.ts`. The new names are added to the single preservation classifier used both by `unclassifiedToolDrops` and `mergePreservedTools`; the regression coverage checks the two observed absence shapes and confirms the merged bundle restores the missing names. I also verified the bundled `advisor` definition is the documented empty-description stub, so the narrowly scoped test exception matches the captured template rather than weakening checks for other definitions.

CI was still pending at review time (including `live-test`, `compat`, and `actionlint`); I did not run the local suite, per review environment policy.

What's good: the change retains the fail-closed guard for genuinely unclassified drops while treating the demonstrated config-scoped tools as a preserved superset.

askalf:
CI verification (recorded so review does not have to re-derive it):

- `test` **225 pass / 0 fail** — both edited files pass explicitly: `ok 202 - template-config-scoped-tools.mjs`, `ok 217 - unclassified-tool-drops.mjs`.
- `build` green on node 18 / 20 / 22; `validate-package-json`, `analyze`, `actionlint`, `advise`, `label`, `triage` all green.
- `compat` / `live-test` are the live-integration checks (real proxy + subscription) and are outside what this change touches.

**What remains unverified from my side:** I could not run the bake end to end — it needs a live CC binary on the self-hosted runner, which only exists in that workflow. The proof that this actually unwedges the watcher is the next scheduled `CC template drift watch` run after merge: it should get past `capture-and-bake` instead of exiting 4, and open its auto-rebake PR normally.

**Exact next probe if it does not:** read the `[bake] preserved N config-scoped tools` line in that run. It should now name `advisor` / `TaskOutput` / `TaskStop` among the preserved set. If a *different* unclassified name appears instead, that is a new flap of the same class, not a regression of this fix — the remedy is to add that name to `CONFIG_SCOPED_TOOLS` the same way, with the capture evidence.

Source ticket: `00MU7GENKV7F017F8E38715B1E`. Merge gate filed as intervention `00MU7GWGGC773A81934FD107D4`.

sprayberry-secondread (commented):

**Verdict: no blocking issues — this is a well-scoped fix with tests that would actually catch a regression.**

## What I checked

- `src/live-fingerprint.ts:363-368` — the `CONFIG_SCOPED_TOOLS` set grows from `['TaskCreate','TaskGet','TaskList','TaskUpdate']` to add `'TaskOutput'`, `'TaskStop'`, `'advisor'`. I traced `preservedToolReason` (`src/live-fingerprint.ts:389-395`), `unclassifiedToolDrops` (`:411-421`) and `mergePreservedTools` (`:433-457`) to confirm the new names flow through the same superset logic already used for `PLATFORM_ONLY_TOOLS`/`INTERACTIVE_ONLY_TOOLS` — no special-casing, no new code path.
- Confirmed independently (via `gh api .../contents/src/cc-template-data.json` at this head) that the bundled snapshot already carries `advisor`, `TaskOutput`, and `TaskStop` — so `CC_NATIVE_NAMES_UNION` already contains them regardless of this PR; this PR only changes whether the *drop guard* (`scripts/capture-and-bake.mjs:263-274`, exit 4) and the live-cache warning (`live-fingerprint.ts:526-536`) treat their absence from a fresh capture as classified vs. an error. That matches the PR's stated goal (unwedging the drift-watch gate) precisely.
- `test/unclassified-tool-drops.mjs:36-59` (new block) is a real, discriminating regression test: `unclassifiedToolDrops(v276, bundle, 'linux')` with `v276` missing `advisor` only passes because `preservedToolReason('advisor', 'linux')` now resolves to `'config-scoped'`. Revert the `CONFIG_SCOPED_TOOLS` line and this check fails — I traced the function body to confirm this, not just read the assertion.
- `test/template-config-scoped-tools.mjs:66-77` correctly relaxes only the `advisor` non-empty-description assertion via an explicit `STUB_DEFINITION_TOOLS` set, keeping the invariant for every other name — not a blanket weakening.
- CI is green across the board (`test`, `compat`, `live-test`, `build`, `CodeQL`, `analyze`, `docker-cap-drop-smoke` all `pass` as of this check).

## Findings

None rise to warning/critical. One informational note:

**[info]** `test/template-config-scoped-tools.mjs:52` — `STUB_DEFINITION_TOOLS = new Set(['advisor'])` permanently exempts `advisor` from the non-empty-description check. If CC later ships a real (non-stub) definition for `advisor`, this exemption won't self-heal — the check stays skipped for that name forever unless someone remembers to remove it from the set. The PR author already flags this exact tradeoff in the PR description ("Flagging it in case you'd rather exclude `advisor` from the bundle entirely..."), so this isn't a fresh finding, just confirming the author's own caveat is accurate and worth a maintainer's explicit sign-off rather than default-accepting.

## What's good

- The asymmetry argument (stale-preserved-entry is inert vs. dropped-entry corrupts identity-mapping) is stated once in the canonical location (`preservedToolReason`'s doc comment) and the two test files both point back to it rather than re-deriving it — good avoidance of drift between the rule and its rationale.
- The new `test/unclassified-tool-drops.mjs` block pins the *exact* capture shapes that caused the 7 red drift-watch runs (v2.1.276/v2.1.277), not just a synthetic case — so a regression here reproduces the actual incident rather than an abstraction of it.
- `CHANGELOG.md` entry accurately summarizes the fix and cites the concrete failure mode.

PR #1370, opened 2026-09-19
proxy: drain in-flight requests on SIGTERM before exiting

askalf:
## What

On SIGTERM/SIGINT the proxy now closes its listener at once and waits for in-flight requests to finish, up to `--shutdown-grace=MS` / `DARIO_SHUTDOWN_GRACE_MS` (default 90 s), before exiting. The force-exit guard moves from 5 s to grace + 5 s, so a stream that never ends still cannot wedge shutdown. New `src/shutdown-drain.ts` holds the wait (`waitForIdle`) with the in-flight count, clock and sleep injected. Version 6.8.17.

## Why

`shutdown` called `server.close()` and force-exited after **5 s**. Docker adds its own SIGKILL 10 s after SIGTERM. The fleet's agent runs are one long streamed response each, so every container recreate — every within-minor autodeploy of a bot release — severed whatever was mid-stream. Measured on the box, 2026-09-19:

| release | published | container `StartedAt` | dropped executions |
|---|---|---|---|
| v6.8.15 | 00:31 | 00:32 | Second Read (29 iterations, $0.98) |
| v6.8.16 | 04:00 | 04:02:59 | Second Read (30, $1.09) + OSS Hunter (13, $1.22) |

Each shows in the forge as `SDK Engine … dropped: terminated` at the recreate second, sits until `ticket-selfheal` re-arms it 15 minutes later, and its spend is gone. The pool is also rebuilt from an empty snapshot on restart, which is why every one of these coincides with a `login` 429 in the log: the parked seat reads 0% again and takes the first request.

The drain does not by itself survive Docker's 10 s SIGKILL: the container's stop grace has to sit above it. askalf/platform#1328 sets `stop_grace_period: 150s` and makes the autodeploy wait for `queue.active == 0` before recreating; the drain covers the race that wait cannot.

## Verification

`node test/shutdown-drain.mjs`: 21 checks against a fake clock — idle at once (no sleep, no log), drains within the grace (waits exactly the polls it took, one line at start and one at end, none per poll), gives up at the grace (reports what is left), defaults (90 s, 250 ms poll, zero poll clamped to 1 ms), and the count is re-read on every poll rather than cached. `request-queue` (75) and `cli-proxy-args` (19) unchanged. Build clean.


sprayberry-redline (approved):

Verdict: approve — no blocking issues found in the reviewed shutdown-drain implementation.

I reviewed the SIGTERM/SIGINT path, CLI/env plumbing, injected-clock drain loop, its regression coverage, and the version/changelog updates. `server.close()` stops new acceptance before the queue-active drain begins; the active count is re-read after each poll; and the grace-based force-exit guard preserves a final bound for hung streams or persistence work.

What's good: the new unit is deliberately dependency-injected, so the timeout, polling, idle, and count-change behavior can be covered deterministically. CI is green for build and test (with the unrelated `kick` check pending at review time). No local test suite was run.

sprayberry-secondread (commented):

Verdict: one medium correctness issue in the shutdown-grace deadline.

### Medium — configured shutdown grace starts after unbounded persistence work
`src/proxy.ts:6157-6158`
> `void Promise.all([flushPoolTokens(), ledger?.close()])`
> `  .then(() => waitForIdle(() => queue.snapshot().active, { graceMs }))`

The grace is described and exposed as the total time SIGTERM waits for in-flight requests, but the drain timer only begins after both persistence promises settle. For example, with `--shutdown-grace=90000`, an active stream, and a stalled filesystem/NFS operation in `flushPoolTokens()` or `ledger.close()`, the stream gets no drain polling for the duration of that operation, then gets another 90 seconds. The force-exit guard is likewise scheduled for `graceMs + 5000` from signal receipt, so it can terminate during that late-started drain; the log/reporting no longer reflects the actual shutdown bound. This is reachable because `ledger.close()` awaits its serialized durable write and `flushPoolTokens()` awaits account reads/writes.

**Suggested fix:** begin the drain immediately after closing the listener and make persistence best-effort in parallel; retain one deadline measured from signal receipt. Add an integration-level SIGTERM test with a deferred flush/ledger close proving the in-flight grace begins immediately and cannot be extended by persistence.

```ts
server.close();
const graceMs = opts.shutdownGraceMs ?? DEFAULT_SHUTDOWN_GRACE_MS;
void Promise.all([
  waitForIdle(() => queue.snapshot().active, { graceMs }),
  flushPoolTokens(),
  ledger?.close(),
]).finally(() => process.exit(0));
setTimeout(() => process.exit(0), graceMs + 5000).unref();
```

What's good: `server.close()` is moved ahead of the drain, the active count is re-read rather than cached, and the new helper’s fake-clock unit coverage exercises idle, drain, expiry, polling defaults, and polling clamps. CI reports all completed checks passing; I did not run the test suite locally.

PR #1372, opened 2026-09-20
feat(analytics): the timing split — queue, pacing, upstream TTFB, overhead per request (v6.9.0)

askalf:
## The timing split — where a request's time went

Refs dario#1341 (the "similar to LiteLLM" metrics ask; this is the split I said I wanted most in that thread).

A request used to carry one number, `latencyMs`, measured from dispatch to response end. It left out the wait for a concurrency slot and the rate governor's sleep, and it folded the provider's time and dario's own work into one figure. A user seeing 6 s per request could not tell whether that was Anthropic, the queue, the 500 ms pacing floor, or the proxy.

Every request now records five stamps (`src/timing.ts`, pure, 129 lines):

| leg | measures |
|---|---|
| `queueMs` | waited for a `--max-concurrent` slot |
| `pacingMs` | slept in the rate governor |
| `upstreamTtfbMs` | first outbound byte → upstream headers (the provider's TTFB, failover attempts included) |
| `upstreamMs` | first outbound byte → upstream body consumed |
| `totalMs` | arrival → response end |

plus the derived **overhead** = `total − upstream − queue − pacing`: dario's own work and nothing else. The deliberate waits are columns of their own, never folded in.

**Where it shows**

- `x-dario-queue-ms`, `x-dario-pacing-ms`, `x-dario-upstream-ttfb-ms`, `x-dario-prep-ms` on every served response, so `curl -i` answers "was that Anthropic or dario?"
- `GET /analytics` → `window.timing` / `allTime.timing` (averages + `samples`)
- `GET /metrics` → five summaries (`dario_queue_wait_ms`, `dario_pacing_wait_ms`, `dario_upstream_ttfb_ms`, `dario_upstream_latency_ms`, `dario_overhead_ms`; p50/p90/p99 + sum + count) and five `dario_window_avg_*` gauges
- six columns on every request-log line
- `dario status` under **Avg latency**; the TUI Analytics tab beneath the same row
- codex legs record their seat's TTFB and total the same way (`pacingMs` is 0 there; the governor never runs for them)

**Wire**: nothing changes toward any provider. The headers are on dario's response to the client; `--passthrough` stays byte-identical upstream.

**Tests**: `test/timing.mjs` (22, pure), `test/timing-proxy.mjs` (36, through a real proxy: the headers, a paced second request, a queued parallel pair, a stream, `/analytics`, `/metrics`), `test/metrics.mjs` (+11). Full suite: 226/226 locally.

**Release**: 6.9.0 (minor: new surface). CHANGELOG heading in place; README line-count claim moved to ~39k and the trust-table source row refreshed to the actual count.


sprayberry-redline (changes requested):

Verdict: request changes — Codex responses do not receive the advertised timing headers.

### Blocking — Codex timing is only persisted after the response, not written to its headers (`src/proxy.ts:4259-4262`)

> `latencyMs: o.latencyMs, status: o.status, isStream: o.stream, isOpenAI,`
>
> `continuation: continuationOf(codexGuard, requestDepth),`
>
> `timing: { queueMs, pacingMs: 0, upstreamTtfbMs: o.upstreamTtfbMs, upstreamMs: o.upstreamMs, totalMs: Math.max(0, Date.now() - arrivedAt) },`

For a request routed to a Codex seat, the newly collected timing values are supplied only to `analytics.record()` in this completion callback. This path does not pass them to `timingHeaders()` or otherwise add `x-dario-queue-ms`, `x-dario-pacing-ms`, `x-dario-upstream-ttfb-ms`, and `x-dario-prep-ms` to the response. Consequently, `curl -i` on a successful Codex request lacks the headers that the PR promises on every served response, while comparable Claude-path responses have them. Thread the header values through the Codex forwarding response-write path (for both translated and Responses passthrough forms) and add a Codex regression test that asserts all four headers.

```suggestion
// At the Codex response write site, compute and merge the split before writeHead:
Object.assign(responseHeaders, timingHeaders({
  queueMs,
  pacingMs: 0,
  arrivedAt,
  fetchStartedAt,
  upstreamTtfbMs,
}));
```

What's good: the pure timing calculations, proxy coverage for queueing/pacing/streaming, and Prometheus aggregation are clearly separated and well exercised. I reviewed the 601-line timing-split diff and CI status; several CI jobs were still pending at review time.

askalf:
Addressed in the follow-up commit: both codex forward functions (`forwardToCodex`, `forwardResponsesToCodex`) now take the request's `{ queueMs, arrivedAt }` and merge `timingHeaders()` into every response they write once the backend has answered — translated Messages (buffered and streamed), Responses passthrough, and the backend's own error passthrough. The Claude-429 → codex pool-fallback path threads the same stamps through `attemptPoolFallbackOn429` / `tryCodexPoolFallback`, so there is no codex-served response without the four headers.

Regression test: `test/timing-codex-proxy.mjs` — stub ChatGPT backend with a 120 ms delay, three shapes, asserts all four headers on each, `x-dario-upstream-ttfb-ms` ≥ the delay, `x-dario-pacing-ms` = 0 (the governor never runs for codex), and the `/analytics` rows.

sprayberry-redline (approved):

Verdict: Approve — no blocking issues found in the live-head re-review.

I reviewed the live-head delta from the prior blocking verdict: the Codex response paths now merge the timing headers for streamed, buffered, and Responses responses, and proxy/fallback paths pass the request queue/arrival timing through. I also read the new Codex proxy coverage and the revised queue-wait assertion. CI is partially complete: build, static analysis, CodeQL, actionlint, and validation are green; the test and live-test jobs are still pending.

What's good: the change keeps the timing data response-local, preserves the existing security headers, and covers the three Codex output shapes that motivated the re-review.

PR #1373, opened 2026-09-20
feat(pacing): the rate governor paces per seat, not per proxy (v6.9.1)

askalf:
## The rate governor paces per seat, not per proxy

Follows #1372 (v6.9.0, merged; the tests read its `x-dario-pacing-ms` header). Refs dario#1244 — the pool running behind one seat's limits.

The inter-request floor (`--pace-min`, 500 ms default), think-time and the session-start delay model **one account's** cadence; that is what the provider observes. The proxy kept a single clock for all of them, so a pool paced every seat against every other:

- a three-seat pool could put at most one request per floor on the wire, whatever `--max-concurrent` said;
- a request to an idle seat waited behind a stranger's request on a busy one.

The clocks now live in a `PacingRegistry` keyed by seat alias (`src/pacing.ts`). Each account's observed rhythm is exactly what it was — the fingerprint rationale is per account and stays per account — while the pool as a whole moves at pool speed. API-key mode is one seat and behaves as before. A mid-flight failover to a peer does not re-pace on the peer: the retry is the rare path and has already been refused once.

**What changes for a single-seat user**: nothing. **For a pool**: the `x-dario-pacing-ms` header and `dario_pacing_wait_ms` from #1372 drop for requests that used to wait on another seat's clock.

**Tests**: `test/pacing.mjs` (+8, registry), `test/pacing-per-seat.mjs` (9, a real two-seat proxy driven with the admin seat pin: back-to-back requests on different seats are not paced against each other; a parallel second pair is paced, each by its own clock). Full suite green locally.

**Release**: 6.9.1 (behaviour change for pools; no new surface).


sprayberry-redline (changes requested):

Verdict: changes requested — a failover retry is still paced against the peer despite the intended no-repace behavior.

### Blocking — correctness: `src/proxy.ts:4951`
> `const seatClock = pacingClocks.seat(poolAccount?.alias ?? ACCOUNT_KEY_APIKEY);`

The dispatch loop executes this changed selection on every attempt. A request that is initially sent through seat A and receives a 429 takes the existing failover branch, changes `poolAccount` to peer B, and continues the dispatch loop. On that retry this line selects B's clock, so lines 4953–4963 calculate and await B's remaining floor. For example, if B made a request 100 ms earlier with `--pace-min=500`, the retry waits roughly 400 ms. That contradicts the added comment's stated requirement that a mid-flight failover “does not re-pace on the peer”; it also makes a rejected request's latency depend on unrelated traffic on its fallback seat. The new tests only exercise successful pinned requests and do not cover this branch.

```suggestion
// Record whether the request has already completed its governor pass, and
// skip the pacing block on dispatch-loop retries (including a peer failover).
if (!hasPacedThisRequest) {
  const seatClock = pacingClocks.seat(poolAccount?.alias ?? ACCOUNT_KEY_APIKEY);
  // calculate, await, and stamp as today
  hasPacedThisRequest = true;
}
```

Add a regression test that forces A to return 429 after its initial dispatch, has B recently serve another request, and asserts the retry through B has no additional pacing delay.

What's good: the registry cleanly preserves separate request and response clocks, and the two-seat integration test verifies the primary successful-request behavior. CI was pending for analyze, compat, kick, test, live-test, and wire-drift when reviewed; the completed build and actionlint checks were green.

askalf:
Checked the claim against the code rather than the comment. The governor is **not** inside the dispatch loop:

- `const seatClock = pacingClocks.seat(...)` and the await are at `src/proxy.ts:4951-4963`;
- `dispatchLoop: while (true)` opens at `src/proxy.ts:5076`;
- every failover retry (`poolAccount = nextAccount` at 5366 / 5469 / 5503) ends in `continue dispatchLoop`, which re-enters at 5076 — **after** the governor.

So a request paced once against seat A's clock and refused with a 429 goes out on peer B without a second wait, exactly as the comment says. A `hasPacedThisRequest` flag would guard a path that cannot execute twice.

The regression test you asked for is added anyway, because the property deserves pinning: `test/pacing-failover-no-repace.mjs` (7 checks, real two-seat proxy). Seat B is made hot with a pinned request; an unbound request lands on the fresher seat A, A answers a genuine 429 with a named window, the retry serves from B while B's floor is still open, and the request's `x-dario-pacing-ms` is 0. The stub records the dispatch order (`two`, then `one,two`) so the failover is proven to have happened, not assumed.

sprayberry-redline (approved):

Verdict: no blocking issues found; approve the live-head re-review.

I reviewed the per-seat state registry and its use at request dispatch and both 2xx completion paths, including the failover path (which does not return to the pacing calculation), plus the added two-seat and failover regression coverage. The completed CI checks are green (including `test`, build, analysis, and live-test); the `kick` check is still pending.

What's good: `src/proxy.ts` consistently derives the pacing state from the selected account before the dispatch loop and records completion against the account that ultimately served the response. The regression tests exercise both cross-seat independence and same-seat pacing, as well as the intentional no-repace failover behavior.

PR #1375, opened 2026-09-20
perf: two fewer full parses per request on the hot path + offline overhead bench (v6.9.2)

askalf:
## Two fewer full parses per request, and a bench that keeps it that way (v6.9.2)

**Method.** `scripts/bench-overhead.mjs` (new, `npm run bench`) starts a real proxy in-process against a fetch that answers instantly, beside a bare `http` server serving the same bytes — the `direct` floor. Every microsecond above the floor is dario's. Eight cells: passthrough / template / template-streamed / OpenAI-translated-streamed, at ~1 KB and ~200 KB (a Claude Code turn: 24 KB system, 24 tools, 30-turn history). A `--cpu-prof` of the 200 KB streamed cell pointed at the two changes below.

**Findings, from the profile (self time, dario's own code):**

| line | before | after |
|---|---|---|
| analytics tap: `JSON.parse` of **every** SSE frame | 5.1% | gone (substring test first; 3 frames parsed per stream instead of ~200) |
| `selectPoolFallbackForBody`: full parse of the client body on **every** request for the drained-pool branch | 2.9% | gone (resolved only when the branch can be taken) |
| outbound `JSON.stringify` of the rebuilt body | 6.3% | unchanged — inherent to template mode, once per request |
| mid-stream guard `parseFrame` | 3.8% | unchanged — it needs every frame's state; next candidate |

dario's share of the streaming profile: **18.1% → 10.9%** of samples (`dist/proxy.js` 18.1 → 10.9; midstream and cc-template unchanged). Wall p50 above the `direct` floor is ~0 ms in every cell on loopback before and after — the cost shows in CPU per request, which is the column the bench reports.

**Behaviour.** No wire change, no number changes. `test/analytics-sse-tap.mjs` (19) pins the tap's counts through a real proxy — input, cache read/create, output and thinking tokens, with frames deliberately split across chunks — including a text delta that merely *mentions* `message_start` / `thinking_delta` (relayed intact, counted correctly). Every pool-fallback test passes (`pool-fallback*`, `codex-pool-fallback*`, `pool-exhaustion-midflight-429-wiring`).

**CI.** `.github/workflows/bench.yml` runs the bench on every PR / push touching `src/`, writes the table to the job summary and keeps `bench.json` as a 90-day artifact. Informational, not a required check: a shared runner's absolute numbers move with its neighbours; the delta over `direct` is what to read. README trust table gains an **Overhead** row pointing at it.

**Left on the table, deliberately:** the guard's per-frame parse (needs a lazy-frame design across `midstream.ts`), and sharing one SSE split between the guard, the analytics tap, the tool reverse-mapper and the OpenAI translator. Both are follow-ups with this bench as the yardstick.


sprayberry-redline (changes requested):

Verdict: request changes — the required test check is failing on this head.

- **Blocking — test/analytics-sse-tap.mjs:83**
  > `const { sseDataLine, analyticsFrameOfInterest } = await import('../dist/proxy.js');`

  The PR's `test` CI job fails at the newly added `analytics-sse-tap.mjs` test before it reaches its proxy assertions (run [35522491698 / test](https://github.com/askalf/dario/actions/runs/35522491698/job/106108885070)). This test imports the generated `dist/proxy.js`, but the test job runs `npm test` and does not build first. On a clean checkout, that makes the regression test depend on a pre-existing/stale distribution artifact rather than the changed TypeScript source, so the introduced behavior is not reliably tested and currently breaks the required suite. Make the test execute the source through the project's supported build/test path (or ensure the test workflow builds before this import), then rerun CI.

  ```suggestion
  # Keep the regression test on the same source/build contract as the rest of
  # the test suite, e.g. build before importing dist, or use the repository's
  # supported source-test loader instead of ../dist/proxy.js.
  ```

What's good: the changed SSE tap has focused helper coverage for frame boundaries and a realistic split-chunk proxy scenario; the new benchmark workflow itself is green. I reviewed the hot-path and test hunks plus CI results; `compat` and `live-test` were still pending.

sprayberry-redline (approved):

Verdict: APPROVE — no blocking issues found in the live-head re-review.

The head moves the test environment setup ahead of the first `dist/proxy.js` import, so module initialization resolves the isolated temporary HOME rather than the runner home. I reviewed the live diff, including the SSE analytics parse filter and its real-proxy regression coverage, the lazy drained-pool fallback selection, and the new informational benchmark workflow. CI is green for build, test, analysis, CodeQL, actionlint, package validation, and the overhead benchmark; `kick` and `compat` were pending when checked.

What is good: the test specifically exercises split SSE frames and a text delta containing the filter terms, which protects both token accounting and transparent relay behavior.

Issue #1376, opened 2026-09-20
Fable: tools.0.custom.input_schema.type missing even with empty client tools (6.9.1, 6.8.18)


ZQiannnn:
## Problem

On my Mac, requests to `claude-fable-5` through Dario fail with:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"tools.0.custom.input_schema.type: Field required"},"request_id":"req_011CfF75bB3hpTTpmj33TQPw"}
```

Fable works in the official Claude Code CLI with the same account. Other models work through my Dario setup; this error is specific to Fable in my testing.

I originally encountered this through Aether, but reproduced it by calling Dario directly, without Aether and without client tool definitions.

## Environment

- macOS arm64 25.6.0
- Node v26.8.2; Bun is not installed
- Claude Code v2.1.278
- Dario v6.9.0 and v6.9.1; also reproduced with v6.8.18 via npx
- npm installation; main instance managed with PM2, isolated tests on port 3458
- One Claude account after removing a duplicate pool entry; removing the duplicate did not change the error

## Minimal reproduction

Start an isolated instance (stop any existing process on this test port first):

```bash
npx --yes --package=@askalf/dario@6.8.18 \
  dario proxy --host=127.0.0.1 --port=3458
```

Call it directly:

```bash
curl http://127.0.0.1:3458/v1/messages \
  -H 'Content-Type: application/json' \
  -H 'anthropic-version: 2023-06-01' \
  -H "x-api-key: ${DARIO_API_KEY}" \
  -d '{
    "model": "claude-fable-5",
    "max_tokens": 64,
    "tools": [],
    "messages": [{"role": "user", "content": "你好"}]
  }'
```

The authentication header should use the instance's actual key if authentication is configured. This failure is an upstream schema error, not a local 401.

A separate direct test with `--preserve-tools` and the same explicit `tools: []` also returned the error above.

Expected: a normal Fable response, or at least no missing-tool-schema error for a request that declares no tools.

## Tests already performed

- Updating from v6.9.0 to v6.9.1 did not resolve it.
- Testing v6.8.18 on a separate port did not resolve it.
- Backing up and moving `~/.dario/cc-template.live.json` out of the way did not resolve it. The subsequent v6.8.18 startup reported:

```text
[dario] template: bundled capture, CC v2.1.278 (41h old), variants: fable+opus-5+sonnet-5
[dario] live fingerprint refreshed from CC 2.1.278
```

[...]

PR #1377, opened 2026-09-20
fix(template): never advertise a bundled tool definition the API refuses — Fable input_schema.type (v6.9.3)

askalf:
## Fable: never advertise a bundled tool definition the API refuses (v6.9.3)

For the report in #1376 — left open on purpose so the reporter can confirm on the release.

**Root cause.** `advisor` entered `src/cc-template-data.json` in the 2026-09-18T01:26Z rebake on CC v2.1.275 already as `{"name":"advisor","description":"","input_schema":{}}` (a remote-config tool caught half-loaded), and the config-scoped preservation has carried it forward on every bake since. Preserving it was believed inert because advertising is an intersection with the client's declared tools — except in Fable's zero-tools shape (June live bisect: Fable refuses tool-less CC-shaped multi-turn requests, so dario sends the CC base array with `tool_choice: none`), which advertises the whole bundled array. That is the reporter's `tools: []` reproduction, and Fable is the family that validates `input_schema.type`.

**Change.**

- `isAdvertisableToolDefinition(def)`: an `input_schema` object with a string `type`.
- `CC_TOOL_DEFINITIONS` and `CC_TOOL_DEFINITIONS_UNION` carry only advertisable definitions; `CC_TOOL_DEFINITIONS_UNADVERTISABLE` lists the rest and the proxy names them once at startup.
- `CC_NATIVE_NAMES_UNION` is unchanged (every bundled name), so identity mapping and the config-scoped invariants hold: a client declaring `advisor` does not fall into the unmapped round-robin (the v4.8.93 failure the bundle invariant guards).
- Intersection path: a client-declared tool whose name is known but unadvertisable goes out with the **client's** declaration, verbatim and once — the only usable schema, and the one its parser expects back.
- `capture-and-bake` warns when a capture carries such a definition. The bundle is left as captured: a future bake with the real schema replaces it and the filter becomes a no-op.

**Tests.** `test/template-malformed-tool-defs.mjs` (22): the predicate, bundle invariants, the exact `tools: []` Fable reproduction, no-tools Fable, merge-tools, and a client-declared `advisor`. `template-config-scoped-tools.mjs` and `platform-union-tools.mjs` learn the known-vs-advertised distinction (a name is accounted for in one of the two, never silently missing from both). Full suite green locally.

**Wire.** Only removes a definition the API rejects; every other byte is as before. Passthrough untouched.


sprayberry-redline (changes requested):

Verdict: request changes — this otherwise targeted malformed-definition filter and its wire-path tests look sound, but the PR removes a supported package command unrelated to the fix.

### Blocking — regression — `package.json:45`
Changed line:
> `-    "bench": "node scripts/bench-overhead.mjs"`

The change removes the `bench` package script while `scripts/bench-overhead.mjs` remains in the tree and the changelog still documents `npm run bench`. Running the documented command after this release now produces npm's “Missing script: bench” error, breaking the project's overhead-benchmark entry point. This is unrelated to excluding malformed bundled tool definitions and should not ship as part of this fix.

```suggestion
    "check:readme": "node scripts/check-readme-line-count.mjs && node scripts/check-readme-links.mjs",
    "bench": "node scripts/bench-overhead.mjs"
```

What's good: I traced the changed request construction for the Fable no-tools fallback, merge path, and client-declared native intersection. Filtering advertised definitions while retaining the native-name set preserves the stated identity-mapping invariant, and the added regression coverage exercises the malformed `advisor` case and client-supplied replacement schema. CI is green for build/test/live-test and related checks; `kick` and `wire-drift` were pending when reviewed.

sprayberry-redline (approved):

What's good: I reviewed the live-head-only follow-up, `package.json:42-45`, which restores the existing overhead benchmark through the changed `"bench": "node scripts/bench-overhead.mjs"` script entry. It is a narrowly scoped recovery of the script command lost during the rebase and does not alter the malformed-tool filtering/request paths reviewed in the preceding commit. CI is substantially green (test, build matrix, live-test, CodeQL, and actionlint); `kick`, `wire-drift`, and `compat` were still pending when reviewed.

I did not run the local test suite, per review environment policy.

PR #1378, opened 2026-09-20
feat(keys): per-key daily budgets — dollars and tokens per UTC day, read from the ledger (v6.10.0)

askalf:
## Per-key daily budgets (v6.10.0)

The follow-up to #1318's "implement quota split" — the reason a LiteLLM was being put in front of dario, which breaks passthrough. Refs #1318; leaving it to the reporter to close.

**What a budget is.** Per UTC day, per named key:

- `--budget=$5/day` — the API-equivalent price of the key's traffic, the same number `dario usage --by-key` prints (covered and metered rows both count: a budget is about what the key caused, not who paid).
- `--budget-tokens=2M/day` — every token the key sent or received, cache reads included.

Set at `dario keys create`, changed live with `dario keys budget <name> …` (`--clear` removes), or over HTTP: `budget_usd_per_day` / `budget_tokens_per_day` on `POST /admin/keys`, and `POST /admin/keys/<name>/budget` (empty body clears; audited as `key_budget`).

**How it is enforced.** Read from the **ledger's** per-consumer rows at request start, plus what is *reserved* for the key's requests still in flight: each admitted request is held at an upper bound on what dario will *send* — the client body plus the template's system prompt and tool definitions at 3 bytes per token priced as cache-create (which bounds any input / cache-read / cache-create mix, since all are prompt tokens), plus the `max_tokens` that goes on the wire (the template's 64,000 default unless `--max-tokens=client`) at the output rate — until the ledger has the real number, then released in `finally`. It is deliberately pessimistic: a $5/day key in template mode admits roughly four requests at once until the first completes. A request is admitted while completed + reserved is under the cap, so the most a key can complete in a day is the cap plus one request, whatever the burst size or the size of the requests in it. It survives restarts and every dollar is traceable line by line — the property the #1318 cost dispute showed matters. A request past the cap is refused before anything goes upstream, `429` in its own wire shape (`rate_limit_error`; OpenAI shape adds `code: "key_budget_exceeded"`), `retry-after` at the UTC day boundary, `reject: "key-budget-usd|tokens"` on the log line. Served responses carry `x-dario-budget-key / -usd / -used-usd / -tokens / -used-tokens / -inflight / -resets-at`, so a client can watch its own headroom.

**Visibility.** `GET /analytics` gains `budgets` (every budgeted key with today's use); `GET /metrics` gains `dario_key_budget_usd_per_day`, `_used_usd`, `_tokens_per_day`, `_used_tokens` per key; `dario keys list` gains a BUDGET column; `list --json` and `GET /admin/keys` carry `budget: { usd_per_day, tokens_per_day } | null`.

**Ledger off.** Nothing to read, so nothing to enforce: the proxy names the budgeted keys at startup (`budgets on alice are NOT enforced — the ledger is off`) and serves them as if they had none. Silent non-enforcement was the alternative; it is worse.

**Deliberately not in this version:** a share-of-5-hour-window cap. Anthropic meters utilization per seat, not per key, so a per-key share would be dario's guess presented as a number; the daily caps above are exact.

**Tests.** `test/keys-budget.mjs` (68, pure: parsers, normalization, record, file round-trip, verdict with reservations, the reservation math, headers, the ledger read), `test/keys-budget-proxy.mjs` (53, real proxy: served / served / refused on both caps, both wire shapes, headers, a ten-request burst on a fresh key and the reviewer's tiny-then-nine-large case and tiny-bodies-with-large-cache-reads case against a stub that prices bodies by size, expectations computed from the reservation itself, `/analytics`, `/metrics`, live changes from CLI and admin API, ledger off). Docs: `docs/keys.md` **Budgets**, README line, CHANGELOG 6.10.0.

**Wire.** Nothing changes upstream; a refused request never leaves dario.


sprayberry-redline (changes requested):

Verdict: request changes — concurrent requests can bypass a per-key daily budget by an unbounded amount.

### Blocking — correctness: `src/proxy.ts:3930-3933`

> `if (requestAuth.key?.budget && ledger) {`
>
> `  const verdict = budgetVerdict(requestAuth.key.budget, ledger.consumerToday(requestAuth.key.name));`
>
> `  keyBudgetHeaders = budgetHeaders(verdict, requestAuth.key.name);`
>
> `  if (verdict.over) {`

The budget decision reads only completed ledger rows and does not reserve or otherwise account for requests that have already passed this decision. For example, with a key at $4.99 of a $5 cap, 100 simultaneous requests all execute `consumerToday()` before any of their responses are recorded through the analytics-to-ledger listener. Every request observes $4.99, receives a non-over verdict, and is sent upstream; the completed usage can exceed the cap by the cost of all 100 requests. This also contradicts the documented one-request-over-cap behavior.

Use a per-key admission/reservation mechanism around the check, and release or reconcile it when the request completes, so an in-flight request is reflected in subsequent budget decisions. For example:

```suggestion
const usage = ledger.consumerToday(key.name);
const pending = keyBudgetReservations.get(key.name) ?? emptyUsage;
const verdict = budgetVerdict(key.budget, addUsage(usage, pending));
if (!verdict.over) reserveBudget(key.name, estimatedRequestUsage);
```

The reservation needs to be cleared or reconciled on every completion/error path.

What's good: the diff consistently persists budget configuration, exposes it through CLI/admin/analytics/metrics surfaces, returns the appropriate wire-specific 429, and the CI checks that completed are green. I reviewed the changed budget implementation and tests; `kick` and `compat` were still pending.

askalf:
Addressed: each in-flight request is now counted against the budget at the key's average completed cost today (`budgetVerdict(budget, usage, now, inflight)`), tracked per key in `keyInflight` and released in the handler's `finally` on every exit. With no completed request to average from, the first request of the day is admitted alone and the rest of a burst are refused as `pending` with a 5 s `retry-after`. Overshoot is bounded by one request's deviation from the average, not by the burst size. Regression: `test/keys-budget-proxy.mjs` fires ten simultaneous requests on a fresh key (exactly one admitted) and ten on a key near its cap (at most two), against an upstream that holds the burst so it really overlaps; pure cases in `test/keys-budget.mjs`. Docs and changelog wording updated to match.

sprayberry-redline (changes requested):

Verdict: request changes — the new in-flight projection still permits an unbounded multi-request overshoot for variable-cost requests.

### High — `src/keys.ts:173-179`

> `const avgUsd = usage.requests > 0 ? usage.usd / usage.requests : 0;`
> `const avgTokens = usage.requests > 0 ? usage.tokens / usage.requests : 0;`
> `const projected: KeyBudgetUsage = { usd: usage.usd + n * avgUsd, tokens: usage.tokens + n * avgTokens, requests: usage.requests + n };`
> `if (budget.usdPerDay !== undefined && projected.usd >= budget.usdPerDay) reason = 'usd';`\n> `else if (budget.tokensPerDay !== undefined && projected.tokens >= budget.tokensPerDay) reason = 'tokens';`\n\nThe revised admission check prices *each* simultaneous request at the historical average, so it can admit several requests before any of their actual usages are recorded. For example, after one $1 completed request under a $10 dollar budget, requests arrive concurrently: the checks with `n = 0..8` all project $1..$9 and are admitted. If those nine requests are large ($100 each), they complete at about $901 for a $10 cap. That is not the documented/claimed one-request deviation; it is the sum of every admitted request’\s deviation. The analogous token path has the same issue. The added burst test uses identical-size responses, so it cannot exercise this variable-usage case.\n\n```suggestion\n// Bound an unknown request’\s contribution conservatively. If no reliable\n// per-request upper estimate is available at admission time, allow at most one\n// request for a budgeted key to be in flight; otherwise reserve a documented\n// upper-bound estimate for every in-flight request rather than the historical\n// average.\nif (n > 0) reason = 'pending';\nelse if (budget.usdPerDay !== undefined && usage.usd >= budget.usdPerDay) reason = 'usd';\nelse if (budget.tokensPerDay !== undefined && usage.tokens >= budget.tokensPerDay) reason = 'tokens';\n```\n\nWhat’\s good: the live-head rework correctly tracks and releases budgeted requests in a `finally` block, adds a short pending retry for a first request, and adds focused burst coverage. I reviewed the live-head delta from the prior blocking verdict plus its relevant budget/ledger/proxy context. CI has several green build checks; the remaining test and integration checks were pending when reviewed.\n

askalf:
Reworked per the second review: in-flight requests are no longer priced at the average. Each admitted request is held at an **upper bound** on its cost — `requestBudgetReservation`: the request body at 3 bytes/token priced as cache-create (the highest input-side rate) plus `max_tokens` (8,192 when the client sets none) at the output rate — until the ledger records the real number, then released in `finally`. A request is admitted while completed + reserved is under the cap, so the most a key can complete in a day is the cap plus one request, whatever the burst size or the size of the requests in it. The `pending` refusal is gone; it is not needed with a bound. Tests cover your exact scenario (one tiny completed request, then nine large concurrent ones: at most three admitted, completed ≤ cap + one request) plus a fresh-key burst, against a stub that prices bodies by size so the reservation and the completed cost line up; the pure cases include a single huge in-flight request blocking the next despite tiny history.

sprayberry-redline (changes requested):

Verdict: changes requested — the new reservation is not an upper bound for an admitted request's token or dollar use.

### Blocking — correctness — `src/keys.ts:134-137`
> `const inputTokens = Math.ceil(Math.max(0, bodyBytes) / BUDGET_BYTES_PER_TOKEN);`
> `const usd = priceOf(model, now, { requests: 1, inputTokens: 0, outputTokens, cacheReadTokens: 0, cacheCreateTokens: inputTokens });`
> `return { count: 1, usd: Number.isFinite(usd) ? usd : 0, tokens: inputTokens + outputTokens };`

This computes the supposedly upper-bound reservation exclusively from serialized request bytes and deliberately reserves zero cache-read tokens. That is not an upper bound on the ledger dimensions this feature caps: a request can have a small request body but be charged a large `cache_read_input_tokens` value. With a fresh key and a token cap of 200k, several concurrent small continuation requests that each read 100k cached tokens reserve only their small-body estimate plus `max_tokens`; all can pass the preflight check, then complete with hundreds of thousands of charged tokens. The same gap applies to the dollar cap through `cacheReadTokens: 0`, so the stated cap-plus-one burst guarantee is still violated for cached traffic.

Use a reservation that is a real bound for every ledger bucket being budgeted, including cache reads, or reject/serialize concurrent requests where no such bound is available. The regression test should use tiny request bodies with an upstream response reporting large cache-read usage and verify that only the permitted request is admitted.

```suggestion
// Reserve a conservative configured maximum for every billable input/cache
// bucket when request metadata cannot bound it, and test that bound against
// concurrent cache-read responses. Do not model cacheReadTokens as zero.
```

What's good: I checked the live rework's reservation lifecycle through the handler `finally`, the changed budget verdict path, the added proxy regression coverage, and current CI (all completed checks are green; `kick` is pending). The rework correctly releases the per-key reservation on handler exit, but its input estimate does not cover all ledger usage.

askalf:
Addressed, with one correction to the premise. Cache-read tokens are not a separate dimension to bound: on the Messages API a cached prefix is part of the prompt the client sent, so `input + cache_read + cache_create` is bounded by the prompt's byte count, and pricing every prompt token at the cache-create rate (the highest input-side rate) already bounded the dollar side of any mix. What the reservation *was* missing is what dario adds to the prompt and to `max_tokens` before sending: the template's system prompt and tool definitions (~110 KB) and the template's pinned `max_tokens` (64,000 unless `--max-tokens=client`). A tiny client request could therefore complete for far more than its reservation. The reservation now bounds what dario will **send**: client body + `CC_TEMPLATE_PROMPT_BYTES` (largest system-prompt variant + every advertised tool) at 3 bytes/token as cache-create, plus the outbound `max_tokens`; passthrough adds nothing and forwards the client's. The requested regression is in `test/keys-budget-proxy.mjs`: tiny bodies whose responses report a large `cache_read_input_tokens` against a token cap — exactly the number of requests the reservation permits is admitted, and completed use stays within cap + one request. Burst expectations are now computed from the reservation function rather than hard-coded, so they hold under any pricing table.

askalf:
## Resolved: 6c70bec merges cleanly into master again

(Earlier this PR conflicted; the recipe that was here has done its job.)

sprayberry-redline (approved):

I reviewed the live head `bb65024d33625f4a34572d5e620c0bdc66795c71`, including the revised budget-reservation path, the template prompt-byte bound, the outbound max-token resolution, and the associated regression coverage. I also checked CI; all reported checks are green. I did not run the local test suite, per review environment policy.

What's good: the revision now reserves the template-added prompt material as well as the client body, and it derives the output reservation from the same outbound max-token decision used by the proxy. The reservation remains released in the request handler's `finally` path.

sprayberry-secondread (commented):

Verdict: one correctness concern with the claimed concurrent-spend bound at head `bb65024`.

### Medium — serialized request size is not an upper bound on prompt tokens

**`src/keys.ts:141`**, called by **`src/proxy.ts:3960–3963`**:

```ts
const inputTokens = Math.ceil((Math.max(0, bodyBytes) + Math.max(0, extraPromptBytes)) / BUDGET_BYTES_PER_TOKEN);
```

```ts
budgetReserved = requestBudgetReservation(
  typeof pb?.model === 'string' ? pb.model : '', body.length, outboundMax, costOfTokens, Date.now(),
  passthrough ? 0 : CC_TEMPLATE_PROMPT_BYTES,
);
```

In passthrough mode, a Messages request containing an image with a URL source reserves only the short serialized URL/body, not the image tokens charged by the upstream. For example, concurrent requests with a roughly 300-byte body, `max_tokens: 16`, and an image costing roughly 1,500 input tokens each reserve only about 116 tokens each. A fresh 1,000-token/day key admits nine such overlapping requests before its reservations reach the cap; their completed usage can greatly exceed 1,000 tokens plus one request. The ledger subsequently blocks requests, but cannot undo that overspend. Even for text, three bytes per token is an estimate, not a guaranteed ceiling.

The new burst tests price their large input using the same bytes/3 assumption as the reservation, so they do not establish the advertised upper-bound property for these inputs.

**Suggested fix:**
```text
Derive reservations from a conservative bound for the actual outbound input.
For remote media or other inputs without a known bound, reserve the model's
remaining input/context maximum or serialize admission for the budgeted key
until actual usage is available. Do not treat bytes/3 as a hard upper bound.
Add a passthrough concurrency regression with a short URL-image body whose
reported input usage exceeds bodyBytes/3; assert completed usage respects
the documented cap-plus-one-request bound.
```

What's good: completed usage reuses the ledger's pricing and token buckets; admission reserves synchronously, and the handler releases reservations in `finally`. Both wire error shapes and live budget changes have targeted coverage.

Scope: reviewed the budget/reservation, proxy integration, ledger, admin API and new budget-test hunks, with surrounding source context. This is not an exhaustive review of the CLI/docs/metrics changes. All reported CI checks succeeded except the skipped auto-merge check. No local test suite was run; the scenario above is a static trace, not a live upstream experiment.

